PR #5, opened 2014-09-27
Rds refactor

willthames:
Using rds2 allows tags and the control over whether or not DBs are
publicly accessible.
Move RDS towards a pair of interfaces implementing the details of rds
and rds2
Added tests to ensure that all operations work correctly as well as
requirements files that allow virtualenvs to test either boto.rds or
boto.rds2

Replaces the modules component of ansible/ansible#8834

Tests will continue to be in ansible/ansible#8834

### DIFF
--- a/cloud/amazon/rds.py
+++ b/cloud/amazon/rds.py
@@ -302,6 +302,19 @@ except ImportError:
     has_rds2 = False
 
 
+class RDSException(Exception):
+    def __init__(self, exc):
+        if hasattr(exc, 'error_message') and exc.error_message:
+            self.message = exc.error_message
+            self.code = exc.error_code
+        elif hasattr(exc, 'body') and 'Error' in exc.body:
+            self.message = exc.body['Error']['Message']
+            self.code = exc.body['Error']['Code']
+        else:
+            self.message = str(exc)
+            self.code = 'Unknown Error'
+
+
 class RDSConnection:
     def __init__(self, module, region, **aws_connect_params):
         try:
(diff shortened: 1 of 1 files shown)


bcoca:
# Rebase Request

Hi!

which are very active:

willthames:
@bcoca - rebased.

danvaida:
Really looking forward to this... https://github.com/ansible/ansible-modules-core/issues/428 would be also fixed by rds2

mtotheikle:
+1 for this, can create RDS instance that is public ATM

ebridges:
+1 need the ability to create publicly accessible RDS instances.
ebridges/rds-photo-supreme-catalog#1

follower:
I'm a little confused why people are wanting this merged in order to _get_ a publicly accessible RDS instance because I'm following this issue because I want to set an RDS instance to _not_ be publicly accessible.

When I use Ansible to set up an RDS instance I get the setting: "Publicly Accessible:   Yes" in the RDS monitor view. What are other people seeing?

willthames:
@follower, I believe it depends upon whether or not you install the RDS instance in a VPC or not.

http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_VPC.html#USER_VPC.Hiding

Furthermore, the option to allow you to set publicly_accessible is only available in boto.rds2, and not boto.rds - and this PR is what moves the ansible rds module to boto.rds2

http://boto.readthedocs.org/en/latest/ref/rds2.html#boto.rds2.layer1.RDSConnection.create_db_instance vs http://boto.readthedocs.org/en/latest/ref/rds.html#boto.rds.RDSConnection.create_dbinstance

Yet further, this PR actually provides the publicly_accessible flag as an option to the rds module.

cloud/amazon/rds.py
```diff
@@ -274,376 +298,715 @@
...
-        if not module.params.get(v):
-            module.fail_json(msg = str("Parameter %s required for %s command" % (v, command)))
-            
-    for v in invalid_vars:
-        if module.params.get(v):
-            module.fail_json(msg = str("Parameter %s invalid for %s command" % (v, command)))
+    def promote_read_replica(self, instance_name, **params):
+        try:
+            result = self.connection.promote_read_replica(instance_name, **params)
+            return RDSInstance(result)
```
bcoca, line 395: Not sure wher RDSInstance is defined, does this intend to be RDSDBInstance?

bcoca:
something i noticed is that the examples use "_" for instance names but aws complains that it is not a valid name, switching to "-" seems to have fixed that (not added in this PR, but making note of it here to correct in the future).

bcoca:
several cases of:

```
module.fail_json(msg=e.message)
```

It would be nicer to add our own message prefix as (i found out the hard way) boto can return null/no message on several errors and the module will just return 'failed=True'.

Also to note that 'size' input should be checked to see if it is a integer, not requried for this PR, but that, the lack of boto response, as mentioned above, and having a clueless user do size: 10Gb (yes, me) can lead to very frustrating time.

cloud/amazon/rds.py
willthames, line 395: Looks like it.

PR #7, opened 2014-09-28
synchronize: use a single -F instead of -FF

Ernest0x:
This small change corrects behavior when one uses an .rsync-filter file to exclude some paths from both being transferred and being deleted, so that these excluded paths can be handled separately with different tasks (e.g. in order to deploy the excluded paths independently from the rest paths and notify handlers appropriately). The problem with the double -FF option is that it excludes the .rsync-filter file from being transferred to the receiver. However, deletions are done on the side of the receiver, so it is absolutely necessary the .rsync-filter file to be transferred to the receiver, so that the receiver knows what files to delete and what not to delete.

### DIFF
--- a/files/synchronize.py
+++ b/files/synchronize.py
@@ -252,7 +252,7 @@ def main():
     group = module.params['group']
     rsync_opts = module.params['rsync_opts']
 
-    cmd = '%s --delay-updates -FF' % rsync
+    cmd = '%s --delay-updates -F' % rsync
     if compress:
         cmd = cmd + ' --compress'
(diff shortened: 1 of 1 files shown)


bcoca:
@tima any comments on this?

Ernest0x:
Hello,

Could you please push this a little bit? So many months for such a small change to be reviewed and merged is somewhat disappointing. It had been at least 3 months in the pool before the core modules repository restructure happened, and was one of the first PRs to be resubmitted at the new repo, listening to your call and in the hope that this would help in merging faster.

PR #10, opened 2014-09-29
files.stat: Expose path in returned result

techtonik:
This is needed to apply subsequent operation on tested path


techtonik:
Cool. It which version will it be?

bcoca:
1.9

techtonik:
Awesome. Thanks!

PR #16, opened 2014-09-29
Allow specification of no external ip, boot disk size, and use existing disk.

rmarsagoogle:
This is the same patch I previously submitted to ansible before the modules were split.

### DIFF
--- a/cloud/amazon/ec2_vpc.py
+++ b/cloud/amazon/ec2_vpc.py
@@ -506,6 +506,15 @@ def create_vpc(module, vpc_conn):
             'id': sn.id,
         })
 
+    # Sort subnets by the order they were listed in the play
+    order = {}
+    for idx, val in enumerate(subnets):
+        order[val['cidr']] = idx
+
+    # Number of subnets in the play
+    subnets_in_play = len(subnets)
+    returned_subnets.sort(key=lambda x: order.get(x['cidr'], subnets_in_play))
+
     return (vpc_dict, created_vpc_id, returned_subnets, changed)
 
 def terminate_vpc(module, vpc_conn, vpc_id=None, cidr=None):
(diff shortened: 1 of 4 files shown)


bcoca:
One thing is that when adding new features they should get a 'versions added' header in the docs, so people know what is available 'when'. 

For the options it might be nicer to allow ssd or standard, leaving the pd- prefix to be added internally.

The PR also needs a remerge.

rmarsagoogle:
I made these changes.  Do I need to submit a new pull request?

bcoca:
no, this one will be fine, but version added should be 1.9

if we were to publish a 1.8.3 release it would only have bugfixes, not new features.

rmarsagoogle:
Version_added changed to 1.9.

cloud/google/gce.py
```diff
@@ -231,7 +260,7 @@ def get_instance_info(inst):
         'name': inst.name,
         'network': netname,
         'private_ip': inst.private_ips[0],
-        'public_ip': inst.public_ips[0],
+        'public_ip': (len(inst.public_ips)>0) and inst.public_ips[0] or None,
         'status': ('status' in inst.extra) and inst.extra['status'] or None,
         'tags': ('tags' in inst.extra) and inst.extra['tags'] or [],
         'zone': ('zone' in inst.extra) and inst.extra['zone'].name or None,
```
bcoca, line 293: avoid using ternary statements

bcoca:
it seems like a bunch of unintended commits were added, please clean up branch

rmarsagoogle:
Sorry, this was my first pull request and I didn't know what I was doing, so it was done from the trunk and not a branch.  I have issued a new pull request (1071) from a branch which should have everything cleaned up.

PR #21, opened 2014-09-29
Add a module to [un]install postgresql extensions

dschep:
https://github.com/ansible/ansible/pull/7874 updated for reorganized module repos.


bcoca:
module looks good, except the 'state' option does not seem be documented

dschep:
@bcoca, I've added that documentaiton. Upon looking at upstream it seems that DB modules got reorganized. Should I move this module to reflect that too?

bcoca:
yep, i missed that, moving it to the postgrseql directory would be appropriate.

dschep:
Done. @bcoca, do task playbooks now need to refer to postgresql.postgresql_db or is that not necessary?

bcoca:
modules are searched in up to 3 dirs, so no need to add path at the current level.

dschep:
Awesome :+1:

bcoca:
one last thing, can you reopen this PR in the extras repo?

abadger:
Added to the ansible-modules-axtras repo: https://github.com/ansible/ansible-modules-extras/commit/4b574ba29445a29f31adc4396ad8ff710ccd78a5

Thanks for the new module!

PR #25, opened 2014-09-29
extend ec2_elb module to view LB health.

tedder:
works with test-module, and works on my machine(tm).

pre-ansible-migration issue: https://github.com/ansible/ansible/pull/7577

was labeled: cloud, feature_pull_request, module, p3


bcoca:
instead of adding a 'view' option, does it make more sense to have a new module 'ec2_elb_facts' ?

tedder:
sure, but there's usually a bias against adding modules for similar actions.

I can re-merge/rebase if requested.

bcoca:
after some thought, keep the functionality, but in the ec2_elb_lb module. Remove the view state, always return the instance health in state=present

tedder:
The problem with that is "present"/"absent" isn't the load balancer state, it's about an _instance_ on the load balancer. That mean viewing the state of a load balancer effectively requires an action.

The merge was a mess. I'll start over, but we may as well continue the thread here (right?).

PR #28, opened 2014-09-29
apt: support version wildcards.

tedder:
not supporting wildcards is a bug, not a lack of a feature. it's especially problematic on packages that are updated frequently (bash had five versions within a few days from bashpocalypse).

putting this on dev, but if the next version isn't shipping soon let me know and I'll patch the current version for a dot release.

code is by @bryanlarsen. manually recommitted since ansible was split into a few repositories before it was merged. links:
- https://github.com/bryanlarsen/ansible/commit/18ee3ebffc21b01648a03d2fee8b64ca7788b312#diff-2eb0581dea24b78e623924caf21e33de
- https://github.com/ansible/ansible/issues/6885

tags: bug_report, module, p3


hwmrocker:
awesome, i just stumbled upon this bug and there is allready a fix for this!

hwmrocker:
I just applied the patch locally and its working. THX!!!

abadger:
Also, pull request from @ethervoid:
https://github.com/ansible/ansible-modules-core/pull/64

And the one he originally submitted to ansible/ansible which contains new integration tests: https://github.com/ansible/ansible/issues/8893

And @bryanlarsen 's pull request which also includes a documentation update: https://github.com/ansible/ansible/pull/7184

abadger:
Just to be sure I understand the logic we want...

We want a wildcard to be able to be used in either the name or in the version but not both at the same time?

abadger:
Although it took a few tries to get it right, it looks like wildcards in both the packagename and the version are allowed:

```
badger@ubuntu14:~$ sudo apt-get install w3+m$=0.5.*
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'w3m' for regex 'w3+m$'
Selected version '0.5.3-15' (Ubuntu:14.04/trusty [amd64]) for 'w3m'
The following packages were automatically installed and are no longer required:
  linux-headers-3.13.0-32 linux-headers-3.13.0-32-generic
  linux-headers-3.13.0-36 linux-headers-3.13.0-36-generic
  linux-image-3.13.0-32-generic linux-image-3.13.0-36-generic
  linux-image-extra-3.13.0-32-generic linux-image-extra-3.13.0-36-generic
Use 'apt-get autoremove' to remove them.
Suggested packages:
  w3m-img w3m-el cmigemo
The following NEW packages will be installed:
  w3m
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/877 kB of archives.
After this operation, 2,376 kB of additional disk space will be used.
Selecting previously unselected package w3m.
(Reading database ... 189457 files and directories currently installed.)
Preparing to unpack .../w3m_0.5.3-15_amd64.deb ...
Unpacking w3m (0.5.3-15) ...
Processing triggers for mime-support (3.54ubuntu1) ...
Processing triggers for man-db (2.6.7.1-1ubuntu1) ...
Setting up w3m (0.5.3-15) ...
```

jlaska:
@abadger my understanding is that the `apt-get install` command supports the wildcard in either the name or version (if supplied).  For example, I could run any of the following commands:
- `apt-get install 'vsftp'`
- `apt-get install 'vs.*d$'`
- `apt-get install  vsftpd=3.0.2-1ubuntu2.14.04.1`
- `apt-get install  vsftpd=3.0.2-*`
- `apt-get install '^vs.*d$=3.0.2-*'`

More detail in the `apt-get` man page ...

```
   install
           install is followed by one or more packages desired for installation
           or upgrading. Each package is a package name, not a fully qualified
           filename (for instance, in a Debian system, apt-utils would be the
           argument provided, not apt-utils_1.0.1ubuntu2_amd64.deb). All
           packages required by the package(s) specified for installation will
           also be retrieved and installed. The /etc/apt/sources.list file is
           used to locate the desired packages. If a hyphen is appended to the
           package name (with no intervening space), the identified package
           will be removed if it is installed. Similarly a plus sign can be
           used to designate a package to install. These latter features may be
           used to override decisions made by apt-get's conflict resolution
           system.

           A specific version of a package can be selected for installation by
           following the package name with an equals and the version of the
           package to select. This will cause that version to be located and
           selected for install. Alternatively a specific distribution can be
           selected by following the package name with a slash and the version
           of the distribution or the Archive name (stable, testing, unstable).

           ...

           If no package matches the given expression and the expression
           contains one of '.', '?' or '*' then it is assumed to be a POSIX
           regular expression, and it is applied to all package names in the
           database. Any matches are then installed (or removed). Note that
           matching is done by substring so 'lo.*' matches 'how-lo' and
           'lowest'. If this is undesired, anchor the regular expression with a
           '^' or '$' character, or create a more specific regular expression.
```

abadger:
Hey, @mvo5 since you contributed the wildcard code in ansible's apt module in the first place, do you know why it prohibits packagename wildcard and version? https://github.com/ansible/ansible-modules-core/commit/2503c36c6a91b759aff97c67967930b81793d4f3

abadger:
I think that this change would break things like package_status()'s functionality.  If I'm reading everything correctly, we'll end up asking package_status something like whether "foo" "1.0_"  is currently installed.  package_status will find it has a "foo" package.  Then it will check the version and find it has a "1.0-1" installed.  Since "1.0_" == "1.0-1" is false, it will incorrectly tell us the package is not currently installed.  I think that, in turn, would break things like check_mode telling us whether the system will be changed by running the playbook.

That's fixable but would need some additional code.

mvo5:
On Wed, Oct 29, 2014 at 09:02:19AM -0700, Toshio Kuratomi wrote:

> Hey, @mvo5 since you contributed the wildcard code in ansible's apt module in the first place, do you know why it prohibits packagename wildcard and version? https://github.com/ansible/ansible-modules-core/commit/2503c36c6a91b759aff97c67967930b81793d4f3

So from my (limited) understanding of the context it seems like we 
would need to modify the code:
"""
for pkgname_or_fnmatch_pattern in pkgspec:
  # note that any of these chars is not allowed in a (debian) pkgname
  if [c for c in pkgname_or_fnmatch_pattern if c in "*?[]!"]:
"""

to something like:
"""
for thing in pkgspec:
  pkgname_or_fnmatch_pattern = thing.split("=")[0]
  # note that any of these chars is not allowed in a (debian) pkgname
  if [c for c in pkgname_or_fnmatch_pattern if c in "*?[]!"]:
"""

Though I have to say that at this point having a unittest for this is
probably a good idea.

I was thinking that having something like:
"foo_=1.0"
does not make much sense, but a string like
"foo=1.0_"
of course does make sense.

Cheers,
 Michael

mvo5:
So I had a deeper look at this now, The branch should work fine and allows "foo_=1.0_" which maybe useful for some people (however I personally think that you shouldn't mix the two as foo\* may expand to very different packages with very different versions which is why I have not allowed it initially and why I think something like my outlined approach in https://github.com/mvo5/ansible-modules-core/commit/cb0acd3c4ce3bd6e971c19a2ec796e8c349e67fa is preferable). 

But it should work fine because the check_version() is not doing a string compare (in which case it would indeed break) but instead its doing a fnmatch() too in its package_status() code (see line 213).

mvo5:
I pushed a branch based on what I outlined above. I don't really mind if "foo_=1.0_" is legal or not, the important part IMO is that this needs a unittest like https://github.com/mvo5/ansible-modules-core/commit/1fa88ace0643a0182230d6ee101d5767fdd0e885 - probably done in a more elegant way, I lost track of ansible a bit, maybe instead of the (ugly) environment we can simply put `if __name__ == "__main__": main()` in apt.py - but I have to admit I don't know how modules are called these days. Having that would make most of the uglyness go away.

abadger:
<nod> unittests would be very good for functions like this that just parse a string and return output -- disussing with the other people here how we'd want to do this.  This strategy doesn't work for the general case of any function in a module but I think implementing a framework that handles that is far enough into the future that adopting something like this now would be good.

abadger:
Yep, we can do `if __name__ == "__main__": main()` so that part should clean up okay.

Can one of the people reporting this bug tell me what they want to happen in this circumstance:

```
ansible -m apt -a 'package=w3m=5.0* state=present'
```

The system has w3m-5.0-1 installed.  The only package present in the repository is w3m-5.0-3.

Do you want ansible to report no changes and leave w3m-5.0-1 on the system or do you want ansible to update the package to the newer one that's available in the repositories?

jlaska:
> Do you want ansible to report no changes and leave w3m-5.0-1 on the system or do you want ansible to update the package to the newer one that's available in the repositories?

I'd think ansible's behavior should match that of `apt-get` with the same scenario.  In such a scenario, would apt-get install `w3m-5.0-3`, or would it leave the existing `w3m-5.0-1` installed?

abadger:
Commandline apt-get will update:

(Started with apache2=2.4.7-1ubuntu4; apache2=2.4.7-1ubuntu4.1 is available in the repository)

```
badger@ubuntu14:~$ sudo apt-get install 'apache2=2.4.7-1ubuntu4*'
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Selected version '2.4.7-1ubuntu4.1' (Ubuntu:14.04/trusty-updates [amd64]) for 'apache2'
The following packages were automatically installed and are no longer required:
  linux-headers-3.13.0-32 linux-headers-3.13.0-32-generic
  linux-headers-3.13.0-36 linux-headers-3.13.0-36-generic
  linux-image-3.13.0-32-generic linux-image-3.13.0-36-generic
  linux-image-extra-3.13.0-32-generic linux-image-extra-3.13.0-36-generic
Use 'apt-get autoremove' to remove them.
The following extra packages will be installed:
  apache2-bin apache2-data
Suggested packages:
  www-browser apache2-doc apache2-suexec-pristine apache2-suexec-custom
  apache2-utils
The following packages will be upgraded:
  apache2 apache2-bin apache2-data
3 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
Need to get 1,086 kB of archives.
After this operation, 5,120 B of additional disk space will be used.
Do you want to continue? [Y/n] y
```

ethervoid:
I think it should install the new package, thats what the wildcard is for IMHO

abadger:
k.  I think we will have to make changes to check_version() then.  We have to directly compare the versions in the repository against he version installed in this case, not just compare each of those against the string that ansible was given.

abadger:
If anyone would care to review my code, I just pushed it in a branch.  I'll probably merge it to devel tomorrow if no one objects.

https://github.com/ansible/ansible-modules-core/issues/323

abadger:
Merged that change and added @mvo5's unittests to the main ansible repo (with the change to allow wildcard in both package name and version).  Unittests pass.  Added documentation update from bryanlarsen's earlier PR.  Added ethervoid's integration tests from an earlier PR and fixed a bug they exposed :-).  I believe this is now fixed.

Thanks to @mvo5, @tedder, @bryanlarsen, @ethervoid, @jlaska, and everyone else who participated in diagnosing and fixing this bug.

# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #30, opened 2014-09-29
Route53 fix - forcing zone_in, record_in to lower case

phillipholmes:
It turns out the Route53 API cares if the zone and record specified in the playbook are lower case or not when deleting a record. If you use a variable to name your servers and care about case, using that same proper case name will cause Route53 DNS delete requests to fail.

The change requested adds .lower() to the module.params.get for both zone and record when used in the underlying code.

Both zone and record are mandatory variables, and as such a more complicated implementation is not needed, as they must always be specified when using this module see lines 169 and 170 for the required state).

If you use lowercase names (or don't use a name variable and share it between a tag and DNS entries) then you will never see this issue.

Tested/Confirmed as an issue in Ansible 1.6.6 and above.


bcoca:
# Rebase Request

Hi!

which are very active:

phillipholmes:
Rebased as requested.

Issue #40, opened 2014-09-29
tests: how do we submit these in concert with module code?


billwanjohi:
Like many others, I have a recently closed Ansible pull request that I'd like to resubmit here (https://github.com/ansible/ansible/pull/6303).

My original pull request has an integration test for the added functionality, and I'd like to submit that. Unfortunately, the module tests didn't come along with the module code in the repository split.

Given this, what's the best way to submit tests in one repository for functionality that's yet to be merged in another repository?

mpdehaan:
Hi,

This looks like a ticket and not a pull request.  Can you please resubmit this as a pull request?

Thanks very much!

billwanjohi:
Hi, sorry for not being clear. I submitted the user module patch as pull request #41. I intended for this issue to discuss the general confusion around module tests, since I figured many people would be confused about where to put them.

PR #41, opened 2014-09-29
user: add expired state

billwanjohi:
(Ported from https://github.com/ansible/ansible/pull/6303)

It's very useful and routine to disable a *nix user.
I implemented expired instead of locked because this prevents any use of
the account, safer than just preventing password-based authentication.

I have tests [1], but since none of the suite came along with the core
modules, I'm unsure how to submit them. Discussed in #40.

[1] https://github.com/billwanjohi/ansible/blob/add_locked_state/test/integration/roles/test_user/tasks/main.yml

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -84,10 +84,12 @@ options:
     state:
         required: false
         default: "present"
-        choices: [ present, absent ]
+        choices: [ present, absent, expired ]
         description:
-            - Whether the account should exist.  When C(absent), removes
-              the user account.
+            - Whether the account should exist, and whether it is expired.
+              When C(absent), removes the user account.
+              When C(expired), the user will not be able to login through any means.
+              Expired state is only implemented for Linux.
     createhome:
         required: false
         default: "yes"
@@ -318,6 +320,10 @@ class User(object):
             cmd.append('-s')
             cmd.append(self.shell)
 
+        if self.state == 'expired':
+            cmd.append('--expiredate')
+            cmd.append('1')
+
         if self.password is not None:
             cmd.append('-p')
             cmd.append(self.password)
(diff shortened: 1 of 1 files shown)


bcoca:
I'm thinking a more flexible implementation should be expires="date string" (default None == never), which enables the creation of users with expiration dates (common when you have a contractor/consultant), expires=now would then be the particular case that you wanted to add.

making note that  'password_expires' and 'locked' are options that we should implement.

billwanjohi:
That sounds reasonable. @bcoca are you speaking for the collective? i.e. if I reimplemented this, would it be considered for inclusion in some reasonable time period? You may have seen, I originally submitted this last March, made changes based on feedback, only to see it languish.

bcoca:
I can only ask for patience, there are a lot of PRs for new features (good problem to have for an OSS project) and we tend to prioritize bugs. I know its frustrating, (up to a few months ago I was on the other side of this) but we have increased the resources and are making a big effort to go through the backlog.

ahmgithubahm:
Hi all.  Looks like this feature has been implemented, in 1.9, but I am having trouble 'un-expiring' a user.  If I miss out 'expires:', Ansible makes no change to /etc/shadow, so the user stays expired.  If I include 'expires=', I get an error:
  File "/root/.ansible/tmp/ansible-tmp-1441284323.66-108325559784408/user", line 3735, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1441284323.66-108325559784408/user", line 2035, in main
    supports_check_mode=True
  File "/root/.ansible/tmp/ansible-tmp-1441284323.66-108325559784408/user", line 2512, in **init**
    self._check_argument_types()
  File "/root/.ansible/tmp/ansible-tmp-1441284323.66-108325559784408/user", line 3189, in _check_argument_types
    self.params[k] = float(value)
ValueError: empty string for float()

Am I missing something?

ansible-1.9.2-1.el6.noarch on CentOS 6 from EPEL, target server also CentOS 6.

PR #42, opened 2014-09-29
Added 'clear_keys' parameter to the authorized_key module

kustodian:
Added a new `clear_keys` parameter to the `authorized_keys` module which if it is set to yes (default: no), clears all keys from the authorized_keys file before adding a new key.

This parameter could be useful if you would like to be sure that the user's authorized_keys file only has the key(s) you would like it to have and no other. Because the `authorized_keys` module appends new keys and only deletes those whose state is put as `absent`, so it was very hard to delete other keys which were not set by Ansible. `clear_keys=yes` could be an easy and fast solution for clearing all unneeded keys.

This pull request is dependent on the following pull request in core https://github.com/ansible/ansible/pull/9206

### DIFF
--- a/system/authorized_key.py
+++ b/system/authorized_key.py
@@ -70,6 +70,13 @@ options:
     required: false
     default: null
     version_added: "1.4"
+  clear_keys:
+    description:
+      - Whether this module should clear all keys in the authorized_keys file before adding a new one.
+    required: false
+    choices: [ "yes", "no" ]
+    default: "no"
+    version_added: "1.8"
 description:
     - "Adds or removes authorized keys for particular user accounts"
 author: Brad Olson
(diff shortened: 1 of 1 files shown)


system/authorized_key.py
```diff
@@ -332,9 +340,14 @@ def enforce_state(module, params):
...
     state       = params.get("state", "present")
     key_options = params.get("key_options", None)
+    clear_keys  = params.get("clear_keys")
 
-    # extract indivial keys into an array, skipping blank lines and comments
-    key = [s for s in key.splitlines() if s and not s.startswith('#')]
+    # an array of passed SSH keys
+    keys = []
+    # 'key' is a string of all keys separated by a comma
+    for split_key in key.split(','):
```
fabriceb, line 348: If key contains commas, either in the options or in the comments, this will break

kustodian, line 348: That is true. To correct this, changes in the core would need to be made to return a list of keys and not a string separated by commas.

bcoca:
the new exclusive parameter takes care of this, closing this issue

PR #44, opened 2014-09-29
adds error message if socket does not exist

Jmainguy:


Jmainguy:
Recreation of https://github.com/ansible/ansible/pull/8542 which was meant to fix https://github.com/ansible/ansible/issues/6860

PR #49, opened 2014-09-30
Fixes bind mounts

Jmainguy:
This PR is to fix bind mounts. This is a bugfix

### DIFF
--- a/system/mount.py
+++ b/system/mount.py
@@ -320,6 +320,17 @@ def main():
             if os.path.ismount(name):
                 if changed:
                     res,msg = mount(module, **args)
+            elif "bind" in args['opts']:
+                changed = True
+                cmd = 'mount -l'
+                rc, out, err = module.run_command(cmd)
+                allmounts = out.split('\n')
+                for mounts in allmounts[:-1]:
+                    arguments = mounts.split()
+                    if arguments[0] == args['src'] and arguments[2] == args['name'] and arguments[4] == args['fstype']:
+                        changed = False
+                if changed:
+                    res,msg = mount(module, **args)
             else:
                 changed = True
                 res,msg = mount(module, **args)


Jmainguy:
Recreation of https://github.com/ansible/ansible/pull/8524

rmetzler:
Very interested in seeing #86 closed, but this PR seems to be ill-indented. Seems like a mixed use of tabs and spaces.

Jmainguy:
Yeah good call, I will try and clean it up tonight

Jmainguy:
@rmetzler Fixed it I think, look it over and let me know if its nice or ugly.

bcoca:
# Rebase Request

Hi!

which are very active:

Jmainguy:
@bcoca Rebased

aleemb:
I just noticed today I have about half a dozen mounts for the same thing and ended up on this bug.

Can this PR be merged if possible?

piotrplenik:
After investigation, I think problem is with Python method ``os.path.ismount()`. 
On Debian I discovered that this method does not return correct response. Test case: 

```
$ mkdir /tmp/a
$ mkdir /tmp/b
$ sudo mount --bind /tmp/a /tmp/b
$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.ismount('/tmp/a')
False
>>> os.path.ismount('/tmp/b')
False
```

Because, this method is used in mount.py in 5 times, I propose create new function, like `is_mount(path)`, that always return correct response, for bind and non bind mountpoints.

What do you think?

AvtechScientific:
@bcoca, @Jmainguy,

I think parsing mount output doesn't always catch bind-mounted directories.

Here is my case: I'm on Debian9 (with btrfs), Ansible 2.2.1
Here is how I use the mount module:

```
- name: update /etc/fstab and mount /home/{{ user.username }}
  mount:
    src: /home/{{ user.username }}
    # ToDo: change name to path on v2.3
    name: /chroots/{{ user.username }}/home/{{ user.username }}
    fstype: bind
    opts: defaults,bind
    state: mounted
```

Here is how the mount excerpt ouptut looks like:

```
...
/dev/sda3 on /chroots/testuser3/home/testuser3 type btrfs (rw,relatime,space_cache,subvolid=258,subvol=/root_subvolume/home/testuser3)
/dev/sda3 on /chroots/testuser2/home/testuser2 type btrfs (rw,relatime,space_cache,subvolid=258,subvol=/root_subvolume/home/testuser2)
/dev/sda3 on /chroots/testuser1/home/testuser1 type btrfs (rw,relatime,space_cache,subvolid=258,subvol=/root_subvolume/home/testuser1)
/dev/sda2 on /boot type ext3 (rw,relatime,data=ordered)
tmpfs on /run/user/1000 type tmpfs (rw,nosuid,nodev,relatime,size=3281060k,mode=700,uid=1000,gid=1000)
```

It looks like `arguments[0] == args['src']` is false and so directories get mounted on each run...
I think the bug should be reopened...
Thank you!

PR #51, opened 2014-09-30
Support specifying cidr_ip as a list

cewood:
Support specifying ec2_group cidr_ip as a list

**Example:**

``` yaml
ec2_group:
  name: ec2_group list example
  rules:
    - proto: tcp
      from_port: 3306
      to_port: 3306
      cidr_ip: [
        192.168.0.0/24,
        172.16.0.0/16,
        10.0.0.0/8,
      ]
```

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -288,19 +288,24 @@ def main():
                     rule['from_port'] = None
                     rule['to_port'] = None
 
-                # If rule already exists, don't later delete it
-                ruleId = make_rule_key('in', rule, group_id, ip)
-                if ruleId in groupRules:
-                    del groupRules[ruleId]
-                # Otherwise, add new rule
-                else:
-                    grantGroup = None
-                    if group_id:
-                        grantGroup = groups[group_id]
+                # Convert ip to list we can iterate over
+                if not isinstance(ip, list):
+                    ip = [ip]
 
-                    if not module.check_mode:
-                        group.authorize(rule['proto'], rule['from_port'], rule['to_port'], ip, grantGroup)
-                    changed = True
+                # If rule already exists, don't later delete it
+                for thisip in ip:
+                    ruleId = make_rule_key('in', rule, group_id, thisip)
+                    if ruleId in groupRules:
+                        del groupRules[ruleId]
+                    # Otherwise, add new rule
+                    else:
+                        grantGroup = None
+                        if group_id:
+                            grantGroup = groups[group_id]
+
+                        if not module.check_mode:
+                            group.authorize(rule['proto'], rule['from_port'], rule['to_port'], thisip, grantGroup)
+                        changed = True
 
         # Finally, remove anything left in the groupRules -- these will be defunct rules
         if purge_rules:
@@ -328,25 +333,30 @@ def main():
                     rule['from_port'] = None
                     rule['to_port'] = None
 
-                # If rule already exists, don't later delete it
-                ruleId = make_rule_key('out', rule, group_id, ip)
-                if ruleId in groupRules:
-                    del groupRules[ruleId]
-                # Otherwise, add new rule
-                else:
-                    grantGroup = None
-                    if group_id:
-                        grantGroup = groups[group_id].id
+                # Convert ip to list we can iterate over
+                if not isinstance(ip, list):
+                    ip = [ip]
 
-                    if not module.check_mode:
-                        ec2.authorize_security_group_egress(
-                                group_id=group.id,
-                                ip_protocol=rule['proto'],
-                                from_port=rule['from_port'],
-                                to_port=rule['to_port'],
-                                src_group_id=grantGroup,
-                                cidr_ip=ip)
-                    changed = True
+                # If rule already exists, don't later delete it
+                for thisip in ip:
+                    ruleId = make_rule_key('out', rule, group_id, thisip)
+                    if ruleId in groupRules:
+                        del groupRules[ruleId]
+                    # Otherwise, add new rule
+                    else:
+                        grantGroup = None
+                        if group_id:
+                            grantGroup = groups[group_id].id
+
+                        if not module.check_mode:
+                            ec2.authorize_security_group_egress(
+                                    group_id=group.id,
+                                    ip_protocol=rule['proto'],
+                                    from_port=rule['from_port'],
+                                    to_port=rule['to_port'],
+                                    src_group_id=grantGroup,
+                                    cidr_ip=thisip)
+                        changed = True
         elif vpc_id and not module.check_mode:
             # when using a vpc, but no egress rules are specified,
             # we add in a default allow all out rule, which was the


juliedavila:
@cewood could you please rebase this? Also, this _might_ not be an issue anymore.

defionscode:
Recommending we close this PR.

cewood:
@defionscode definitely still an issue. I've re-based and it's ready to go now, apologies I couldn't respond sooner.

juliedavila:
Cool. Before I test though, why are there modifications to `files/copy.py` and `packaging/os/apt.py`?

Could you please clean that up and then squash/fixup your commits?

cewood:
Woops, sorry about that. Looks better now,

juliedavila:
Looks good. Testing now.

juliedavila:
@cewood 
Unfortunately your pull request breaks backwards compatibility to the way the module worked originally.

In detail:
This playbook (taken from your example) works.

``` yaml
---
- name: Test AWS Stuffs
  hosts: localhost
  connection: local
  tasks:
    - ec2_group:
        name: PR-Test
        description: testing pull-request
        region: us-east-1
        rules:
          - proto: tcp
            from_port: 3306
            to_port: 3306
            cidr_ip: [
              192.168.0.0/24,
              172.16.0.0/16,
              10.0.0.0/8,
            ]
```

Going back to the original way of specifying a single CID/R block like below does not work and returns a boto exception

``` yaml
---
- name: Test AWS Stuffs
  hosts: localhost
  connection: local
  tasks:
    - ec2_group:
        name: PR-Test
        description: testing pull-request
        region: us-east-1
        rules:
          - proto: tcp
            from_port: 3306
            to_port: 3306
            cidr_ip: 192.168.0.0/24

```

Gives the following error:

``` shell
~/Tech/debug ansible-playbook -i host aws.yml

ok: [localhost]

TASK: [ec2_group ] ************************************************************
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427305927.34-264006003752680/ec2_group", line 2197, in <module>
    main()
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427305927.34-264006003752680/ec2_group", line 307, in main
    group.authorize(rule['proto'], rule['from_port'], rule['to_port'], thisip, grantGroup)
  File "/usr/local/lib/python2.7/site-packages/boto/ec2/securitygroup.py", line 203, in authorize
    dry_run=dry_run)
  File "/usr/local/lib/python2.7/site-packages/boto/ec2/connection.py", line 3191, in authorize_security_group
    params, verb='POST')
  File "/usr/local/lib/python2.7/site-packages/boto/connection.py", line 1226, in get_status
<Response><Errors><Error><Code>InvalidParameterValue</Code><Message>invalid CIDR subnet specification: 0</Message></Error></Errors><RequestID>9f733eca-2cbc-4296-a6d5-4cc6abed6821</RequestID></Response>


```

Going back to `devel` allows the above to go through without error.

At your convenience please patch your code and let me know when it's ready for a second look.

mspiegle:
I believe the issue is related to the isinstance(ip, unicode) check.  On my Python 2.x system, this check fails every time, and should probably be "basestring" as unicode and str are both derived from it.  If we include support for Python 3.x, then it looks like they did away with all of that and just have "str".  I'm using a hack from here: http://stackoverflow.com/questions/11301138/how-to-check-if-variable-is-string-with-python-2-and-3-compatibility

```
try:
  basestring
except NameError:
  basestring = str
```

EDIT: I suppose you could also say the following.  This pattern is used in the boto library:

```
if not isinstance(ip, list):
  ip = [ip]
```

cewood:
@mspiegle thanks for the comment and suggestions.

This being a python3 issue makes a lot of sense, as I'm an Arch Linux user and may have forgotten to use a pyenv/virtualenv when I did this work (python3 is the default in Arch).

I think I'll go with the latter approach, as it feels slightly simpler/more correct.

@defionscode I've pushed an additional commit up that should resolve this issue. Please let me know if you'd prefer that commit were squashed/amended into the previous commit or similar.

mspiegle:
I may have found a hole in this approach.  It seems to break the use-case where you have a single rule that specifies a group_name.  Example:

```
  - name:        default-test
    description: testing default sg
    <<: *us-east-1-defaults
    rules:
      - proto:      all
        group_name: default-test
    rules_egress: []
```

The reason it breaks is because no cidr_ip was defined, and thus the entire "for thisip in ip" block never runs.

iainlbc:
@Michael, I'm afk so can't help test, but in our experience we had to
specify a group_desc as well as group_name when using this module

mspiegle:
@iainlbc You appear to be correct, however the behavior is either not well documented, or faulty.  Look at the example in the docs: http://docs.ansible.com/ec2_group_module.html.  You can see that a rule can reference the containing group without a group_desc, though in the example, I think the containing group already existed.  In my example, the containing group hadn't existed yet, but it seems silly to supply a description because you're essentially describing the same group twice.

EDIT: And of course now I'm having trouble reproducing my problem.  I should have posted the error message when I had it.

juliedavila:
@cewood apologies for the delay, I've been wrapped up recently. Yes please squash the commits and I'll do a retest right after. I'll try to replicate @mspiegle's issue as well.

mspiegle:
@defionscode I wouldn't spend too much time trying to reproduce my problem.  It hasn't happened since I reported it, and I have a somewhat complex set of rules.  It's also possible that it was implicitly fixed by some other PR that I merged into my local library/ec2_group.py module.  For reference, here's the code I'm using:

https://gist.github.com/mspiegle/9eb300fa79f688672a11

I'm managing about 60 different security groups, each one having ~5 rules on average.

juliedavila:
Sounds good, I'll wait on the squash, do normal testing then, all else equal, will recommend a merge.

cewood:
@defionscode I've squashed the commits and pushed, bring on the merge :smiley:

gregdek:
@defionscode Ready to merge?

cewood:
@defionscode any update?

juliedavila:
Apologies for the delay, was out of town for a good bit. Testing now...

juliedavila:
+1 from me. good to go.

abadger:
This PR drops the usage of `make_rule_key()`.  I don't see a reason for that, should it be restored?

For the potential issue in https://github.com/ansible/ansible-modules-core/pull/51#issuecomment-89042666 where no ip address is specified, I think that this leads to ip being set to None.  That means that the loop becomes:

``` python
for thisip in [None]:
```

Which means that we get a single rule key in that example: "in-all-None-$GROUPID-None"

So I don't think there's an issue there.

gregdek:
So wait @abadger -- are you recommending that we put make_rule_key() back, or are you saying that it's not an issue?

mspiegle:
@gregdek I agree with @abadger that make_rule_key() needs to stay.  The logic to make rule keys is in several places and it makes lots of sense to put it behind a function.

juliedavila:
@cewood friendly ping

cewood:
Sorry, have been on a bit of hiatus... somewhat like this PR I guess ;)

@defionscode I've updated the PR to use `make_rule_key()` again as requested, so we should be good to merge this now ?

juliedavila:
+1

abadger:
All looks good now!  Thanks everyone for the new code and the reviews!

# Merged

Hi!

issue is resolved.

Thank you!

PR #53, opened 2014-09-30
dpkg does not take a --force-yes option

hfinucane:
If you try to install a .deb (uploaded or downloaded outside of the usual `apt` channels), and you set `force=yes`, you'll get an error because `dkpg` (unlike `apt-get` or `aptitude`) will not take `--force-yes`.

Substituting a mostly similar command, which will make sure you will install the package even if it's probably a bad idea.

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -358,7 +358,7 @@ def install_deb(m, debs, cache, force, install_recommends, dpkg_options):
         if m.check_mode:
             options += " --simulate"
         if force:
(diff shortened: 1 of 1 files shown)


skmoen:
This impacts our occasional need to force a reinstall. Would be helpful.

hfinucane:
(it's worth noting that this combination of options has never worked)

hfinucane:
Also, there are tests for this here: https://github.com/ansible/ansible/pull/9957

PR #55, opened 2014-09-30
Add track submodule branches parameter

bibliotechy:
Was https://github.com/ansible/ansible/pull/7877
Allows user to decide if git submodule should track branches/tags or track commit hashes defined in the superproject.

Add track_branches parameter to the git module.

Defaults to track branches behavior.

Original Issue was https://github.com/ansible/ansible/issues/7030
##### Issue Type:

“Bug Report” /  “Feature Idea”
##### Ansible Version:

ansible 1.6 (devel a5e7492c4f) last updated 2014/04/15 21:12:13 (GMT -400)
##### Environment:

Ubuntu 12.04
##### Summary:

Git submodule update is hard coded to track branches (always adds the `--remote` flag, assuming the git version is late enough to support tracking branches.

https://github.com/ansible/ansible/blob/1c9950678a8a226020c6a1a32d6a60f942495072/library/source_control/git#L423-424

I have a use case where we definitely want to track the sha1 recorded in the superproject, not a branch or tag of the submodule. I'm willing to put in a PR on this, but wonder about the best way to add in a choice on this. A submodule_remote parameter? Something else?
##### Steps To Reproduce:

In a git repo with submodules that are tracked by commit hash stored in the superproject rather than branch or tag, run 

`git: repo=reponame dest=place/to/put/it`
##### Expected Results:

I'd expect all of my submodule repos to be checkout out on the commit stored in the the superproject.
##### Actual Results:

All of my submodule repos are be checkout out to the commit at the tip of the master branch.

If the submodules also do not use the label "master" for the master branch, it throws an error

`fatal: Needed a single revision`

`Unable to find current origin/master revision in submodule path`

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -128,6 +128,16 @@ options:
         description:
             - if C(no), repository will be cloned without the --recursive
               option, skipping sub-modules.
+
+    track_branches:
+        required: false
+        default: "yes"
+        choices: ["yes", "no"]
+        version_added: "1.7"
+        description:
+            - if C(no), repository will be cloned without the --recursive
+              option, allowing submodules to be tracked by commit hash
+              instead of branch name.
 notes:
     - "If the task seems to be hanging, first verify remote host is in C(known_hosts).
       SSH will prompt user to authorize the first contact with a remote host.  To avoid this prompt, 
@@ -399,7 +409,7 @@ def get_head_branch(git_path, module, dest, remote, bare=False):
     f.close()
     return branch
 
-def fetch(git_path, module, repo, dest, version, remote, bare):
+def fetch(git_path, module, repo, dest, version, remote, bare, track_branches):
     ''' updates repo from remote sources '''
     (rc, out0, err0) = module.run_command([git_path, 'remote', 'set-url', remote, repo], cwd=dest)
     if rc != 0:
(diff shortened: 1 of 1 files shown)


pztrick:
:+1: My project tracks commit hashes and I'm currently checking out this P/R's 37ed9b1 just so I can deploy at all.

josef-spak:
I'd like to rebase/squash this and change the default behavior to match git behavior if that's OK?

mtotheikle:
+1

We need this fix / option added as well since our projects must use the parents project recorded SHA-1. I'd say that without this fix applied the Git module is broken as most projects I've worked on require tracking the recored SHA-1 or face issues with upstream changes breaking projects.

I'm for making the default behavior of this the same as git as that would cause less confusion.

After trying to figure out where things were breaking in my project I found this commit that introduced the option, https://github.com/ansible/ansible/commit/11895fa43734ce246a7bf64746df618540251c1f. Maybe @addshore could explain the reasoning for adding it as the default?

bcavagnolo:
+1 on merging this. Just ran into this issue today.

addshore:
+1 to this too

abadger:
@bibliotechy not trying to derail getting this feature in, but I have a question -- at the end of your report your note that submodules without a master branch get a fatal error.  From reading up about git submodules --remote, it looks like you should add a `branch=no-master-so-specify-branch-name` to the .gitmodules file for the project.  Is there a reason that should not be done?

@addshore Since you wrote the code that adds --remote, the question was asked why this was made the default.  Is there a reason or was that just an oversight?

addshore:
@abadger oversight on my part

abadger:
Excellent.   I think we should reverse the default to use commit hashes (like git itself does), then.

The parameter name and description also seem a little confusing.  How about something like:

```
track_submodules:
    required: false
    default: "no"
    choices: ["yes", "no"]
    version_added: "1.8"
    description:
         - if C(yes), submodules will track the latest commit on their master branch (or other branch specified in .gitmodules).  If C(no), submodules will be kept at the revision specified by the main project. This is equivalent to specifying the --remote flag to git submodule update.
```

abadger:
Thanks for the work on this @bibliotechy !

# Merged

Hi!

issue is resolved.

Thank you!

PR #56, opened 2014-09-30
Find python via /usr/bin/env for the GCE's inventory plugin.

fcuny:


erjohnso:
Hi @franckcuny, going over (very) old PRs, and I'm thinking we should drop the request to use `/usr/bin/env` since that does not seem to be standard practice looking over other modules.

But it would be great to keep the `target_tags` portion. But I'm not sure this would apply cleanly any longer since cloud modules have been moved under sub-directories.

bcoca:
For modules we need /usr/bin/python for the ansible_python_interpreter facility to work, the python in env is not always the one you want (see Archlinux, Gentoo, etc).

erjohnso:
Great, thanks for clarifying @bcoca. We'll see what Franck wants to do with the other part and getting it moved over to the new dir.

gregdek:
@franckcuny we'd love to have this PR, if you can rebase and remove the /usr/bin/env patch.

fcuny:
@gregdek OK, will do it before the end of the weekend.

Thanks.

fcuny:
@gregdek this option is already present in the code, added by https://github.com/ansible/ansible-modules-core/commit/ea1122d49f84ab4f43637d3770bfe61dce7eb6f9. Closing this PR.

PR #57, opened 2014-09-30
LinuxService: systemd related improvements

ciupicri:
- check if systemd is really in running, not just installed
- instead of listing all the unit files, verify only the current unit
- simplify/shorten code

Basically do things by the book in a way similar with what systemd does.


system/service.py
```diff
@@ -394,48 +394,41 @@ def get_service_tools(self):
...
 
         def check_systemd(name):
-            # verify service is managed by systemd
-            if not location.get('systemctl', None):
+            """Verify if unit (service) is managed by systemd"""
+
+            # sd_booted(3) reimplemented in pure Python
+            # systemd.daemon.booted() could also be used
+            try:
+                st = os.lstat("/run/systemd/system/")
```
abadger, line 402: Will different distros put this in different places?  Do we need to check both /run/systemd and /var/run/systemd, for instance?

ciupicri, line 402: No, unless they fork systemd and modify its paths.
For example Debian has migrated to `/run` according to their [release notes](https://www.debian.org/releases/stable/i386/release-notes/ch-whats-new.en.html#tmpfs-filesystems).

bcoca:
closing in favor of #343

ciupicri:
This patch was proposed quite a while ago and tries to do things _by the book_, so that bugs like #161 don't appear, yet a new patch is suddenly accepted. I understand that it takes some time for patches to be analyzed and merged, but still. I feel like all my work has been in vain.

bcoca:
@ciupicri, I am sorry, it was not my intent, I actually found your commits very useful. There are still parts  that I want to integrate as they are much cleaner and seem more correct than the current code (a0fb50eefbdbd7c01e4fc89f5b891249ac1dbe4a and 6c009630c82fa0e4062648b448e449cc235a22ae).

For expediency I opted for reworking it myself (#343, #346) as the existing code had gotten pretty messy and lead to confusion and bugs, part of it was having the whole flow make sense to me.  Also the detection from existing patches did not work with all implementations/distributions, but I believe the new patch does (let me know if otherwise). 

If it is OK with you, I will cherry pick some of your commits (the ones linked above), otherwise feel free to resubmit the code, I should have a easier time now merging those in.

ciupicri:
Thank you for the feedback.

Speaking of the current mess, I would have proposed splitting the current Linux class into multiple classes, one per init system e.g. `Systemd`, `Upstart`, etc., but knowing that you guys have a lot of pull requests, I preferred limiting myself to this smaller change.

Feel free to cherry pick some of the code if you think it's easier that way. Anyway I've rebased this pull request (branch) against the current devel branch, just in case.

PR #58, opened 2014-09-30
ec2_lc cleanup

ferrants:
- Ensures that we can set the default and necessary value for `instance_profile_name`. Setting `false` or a blank string will cause an error and does not work.
- Clean a spacing issue
- Add information to the successful module return

This PR was previously tagged with `bugfix_pull_request`


nicocesar:
This is just what I need!! anyone could review it and merge it? Thanks,

ferrants:
Rebased for a clean merge

PR #63, opened 2014-09-30
Use MySQLdbs read_default_file

muffl0n:
Removed custom config parsing and some unused code. Let's you specify a config_file that should be read instead of ~/.my.cnf

These changes are quite extensive, it will break config-parsing where non-mysql-compliant formats where used.

PS: this is a re-apply of this PR: https://github.com/ansible/ansible/pull/6295 I Hope nothing did brake while cherry picking the commits.

### DIFF
--- a/database/mysql/mysql_user.py
+++ b/database/mysql/mysql_user.py
@@ -183,11 +183,7 @@ class InvalidPrivsError(Exception):
 # MySQL module specific support methods.
 #
 
-def connect(module, login_user, login_password, config_file):
-    default_file = '~/.my.cnf'
-    if config_file is not None:
-        default_file = config_file
-    
+def connect(module, login_user, login_password, config_file='~/.my.cnf'):
     config = {
         'host': module.params['login_host'],
         'db': 'mysql'
@@ -198,8 +194,8 @@ def connect(module, login_user, login_password, config_file):
     else:
         config['port'] = module.params['login_port']
 
-    if os.path.exists(default_file):
-        config['read_default_file'] = default_file
+    if os.path.exists(config_file):
+        config['read_default_file'] = config_file
     else:
         config['user'] = login_user
         config['passwd'] = login_password
--- a/database/mysql/mysql_user.py
+++ b/database/mysql/mysql_user.py
@@ -100,7 +100,12 @@ options:
     version_added: "2.0"
     description:
       - C(always) will update passwords if they differ.  C(on_create) will only set the password for newly created users.
-
+  config_file:
+    description:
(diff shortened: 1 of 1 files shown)


muffl0n:
The method "connect" should probably be extracted so that the other mysql-modules can use it, too.

muffl0n:
Silly me, I already extracted the connect-method in https://github.com/ansible/ansible/pull/6295. Would make the same change here if this is going to be merged.

bcoca:
This requires a rebase

muffl0n:
Great. Now I totally fucked up this branch. Give me some time to clean up the mess. :(

muffl0n:
When comparing changes between [ansible:devel and muffl0n:mysql_read_default_file](https://github.com/ansible/ansible-modules-core/compare/devel...muffl0n:mysql_read_default_file) it does not look that fucked up at all. 
Any hint what's wrong here would be greatly appreciated. :(

bcoca:
the issue is that your tree has 21 commits and 11 files changed, most of which are unrelated to the change you want. This looks like a bad rebase.

muffl0n:
Thanks for the hint, I just reverted the rebase and will give it another try.

database/mysql_user.py
```diff
@@ -155,6 +160,30 @@
 # MySQL module specific support methods.
 #
 
+def connect(module, login_user, login_password, config_file):
+    default_file = '~/.my.cnf'
+    if config_file is not None:
+        default_file = config_file
```
bcoca: This might be much simpler if you just use:
def connect(module, login_user, login_password, config_file='~/.my.cnf'):
and then just use config_file

bcoca:
The rebase issue is probably due to the file moving.

Aside from that I do like this approach better by removing configparser and relying on the built in mysql client config reading. But I'm concerned on backwards compatibility and keeping all mysql modules using the same methods. 

I'm thinking of moving this into the shared module_utils/mysql.py and have all mysql modules use it.

muffl0n:
Thanks for pointing out the thing with the parameter default. I'll will change that as soon as I get this rebase to work.
And I will also catch up moving connect() out of mysql_user.py as I already did in https://github.com/ansible/ansible/pull/6295.

muffl0n:
I got the rebase to work and added the default value for the parameter config_file. 
Unfortunately I'm not quite sure how to handle the refactoring of connect(). I do have to create a seperate PR for that in ansible/ansible, right?

bcoca:
yes, it would be in ansible/ansible and then the other modules would also need to be updated.

PR #65, opened 2014-09-30
digital ocean: Remove unsupported states

zupo:
The code in the `core` method only supports `present` and `absent` states. If a user chooses `active` or `deleted`, no code is run in the `core` method and no error is raised, leaving the user baffled about what is going on and why nothing happens.

### DIFF
--- a/cloud/digital_ocean_domain.py
+++ b/cloud/digital_ocean_domain.py
@@ -27,7 +27,7 @@ options:
     description:
      - Indicate desired state of the target.
     default: present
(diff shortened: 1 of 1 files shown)


skottler:
Hey @zupo, thanks for the pull request! The format of the repo has changed a little bit so `digital_ocean_domain.py` is now in `cloud/digital_ocean/`. Can you update to reflect that change?

skottler:
Fixed this up in #336, going to work on it over there.

PR #71, opened 2014-09-30
Added support for specifying a folder on vm creation

stewrutledge:
This change allows for folder to be given as a vm_extra_config parameter, defaulting to the Virtual Machines (or whatever the default folder is set to) if none is given.


bcoca:
closing as this seems to have been included in #72 alrady

cmarzullo:
Doesn't look like the folder option is in that commit.

PR #72, opened 2014-09-30
Added support for deploying virtual machine from a virtual template

stewrutledge:
Added support for deploying guests from an existing template, including a bit of documentation on how it works.

### DIFF
--- a/cloud/vmware/vsphere_guest.py
+++ b/cloud/vmware/vsphere_guest.py
@@ -67,7 +67,16 @@ options:
     description:
      - Indicate desired state of the vm.
     default: present
-    choices: ['present', 'powered_on', 'absent', 'powered_off', 'restarted', 'reconfigured']
+    choices: ['present', 'powered_on', 'absent', 'powered_on', 'restarted', 'reconfigured']
+  from_template:
+    description:
+     - Specifies if the VM should be deployed from a template (cannot be ran with state)
+     default: no
+     choices: ['yes', 'no']
+  template_src:
+    description:
+     - Name of the source template to deploy from
+     default: None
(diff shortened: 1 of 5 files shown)


stewrutledge:
I'll get this fixed, it comes from an older version of the module that had the issue in it (this has come up in a few pull requests) but had been fixed since my original pull requests.

bcoca:
also you'll need to rebase before I can merge

stewrutledge:
Ah crap, my rebase didn't go so well apparently. I can kill this and make a new pull request.

bcoca:
you can update the branch in github with --force, no need for a new PR

## 

Brian Coca

stewrutledge:
I think the branch itself is updated, at least for the commit on this file, it's just that a lot of old commits seem to have come along instead of just my new commit.

I'm really new at the forking thing, so if I'm doing something wrong here I'd be more than happy to do it right!

gnunchucks:
Is it necessary to force the use of a destination cluster when cloning from a template? It seems that the root resource pool always exists, but there are plenty of scenarios where someone would want to clone from template but not be utilizing a cluster.

stewrutledge:
I would guess not no, I assume it can be done in a similar manner to the "create_vm" method (if resource_pool:).

I'll take a look at this next week when I have access to a vsphere instance.

gnunchucks:
I just took a look at that section, that is exactly what I had in mind. Thanks for taking it into consideration.

sebastienbonami:
@gnunchuks @stewrutledge I have a standalone ESXi host and I can't use the feature, because of the cluster parameter (that should be optional). Any updates?

PR #75, opened 2014-09-30
Add module for Linode DNS management

vanclevstik:
Module for creating and deleting dns records on Linode Public Cloud.

### DIFF
--- a/cloud/linode_domain
+++ b/cloud/linode_domain
@@ -1,0 +1,282 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


gregdek:
@ansible/core Better for Extras inclusion at this point?

bcoca:
yes, also we might want cloud/misc or cloud/linode as a directory

gregdek:
@ferewuz Thanks so much for this PR! Since you submitted it, though, we changed our policy: new modules now go into the ansible-modules-extras repository here:

https://github.com/ansible/ansible-modules-extras

Please re-file this PR against that repo, with @bcoca's suggestions, and we'll put it into community review as soon as we can. Thanks!

PR #76, opened 2014-09-30
ec2_snapshot idempotency feature + some other changes

hkariti:
Added a snapshot_max_age parameter, that allows you to check for an existing snapshot to the volume and only create a new one if it's older than a given age.

"Side" changes:
- Added snapshot_max_age, wait and wait_timeout to the docs (included a correct `version_added` of 1.6, when these parameters were first added)
- Made the default value of wait to be false. This restores the behavior to what it was before ansible/ansible#6697.

Example for snapshot_max_age:

```
- ec2_snapshot: volume_id={{volume_id}} region={{region}} snapshot_max_age=60
  register: snapshot
```

If there's no snapshot for this volume, or if the latest snapshot is older that 60 minutes, a new one will be created (snapshot.changed is true).

If there's a snapshot that's newer than 60 minutes (even if it's in progress) the module will not create a snapshot (snapshot.changed is false), and the info returned to snapshot will be the info of the found snapshot.

If the newest snapshot is currently in progress and wait is set to true, the module will wait for it to complete before returning.

### DIFF
--- a/cloud/amazon/ec2_snapshot.py
+++ b/cloud/amazon/ec2_snapshot.py
@@ -136,7 +136,7 @@ def main():
             instance_id = dict(),
             snapshot_id = dict(),
             device_name = dict(),
-            wait = dict(type='bool', default='false'),
+            wait = dict(type='bool', default=True),
             wait_timeout = dict(type='int', default=0),
             snapshot_max_age = dict(type='int', default=0),
             snapshot_tags = dict(type='dict', default=dict()),
--- a/cloud/amazon/ec2_snapshot.py
+++ b/cloud/amazon/ec2_snapshot.py
@@ -74,6 +74,11 @@ options:
       - snapshot id to remove
     required: false
     version_added: "1.9"
+  snapshot_max_age:
+    description:
+      - If the volume's most recent snapshot has started less than `snapshot_max_age' minutes ago, a new snapshot will not be created.
+    required: false
+    default: 0
 
 author: Will Thames
 extends_documentation_fragment: aws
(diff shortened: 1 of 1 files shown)


juliedavila:
@hkariti can you rebase this when you have a chance?

hkariti:
@defionscode done.
Also, I'm thinking of putting the default value of 'wait' back to true, as it's been like that for so long. thoughts?

juliedavila:
Yeah, the default should be wait. Held off on looking deeper/testing until post-rebase, will do a more thorough look/test now.

juliedavila:
@hkariti 

So this works as long as `snapshot_max_age` is specified, otherwise it doesn't work properly and hangs at the ec2_snapshot task for a significantly longer amount of time than what it used to. I'll note that it does make the snapshot, it does just stay hanging for a good amount of time. 

With your PR total time = `9:51.65`
Before PR total time = `2:37.78`

When you get some time can you figure out where the hanging is occurring and patch it accordingly?

hkariti:
Can you post a playbook? I can't seem to figure that out

juliedavila:
Sure, it's just a very trivial playbook:

``` yaml
---
- name: Test AWS Stuffs
  hosts: localhost
  connection: local
  tasks:
      - name: test snapshot PR
        ec2_snapshot: volume_id="{{ my_volume}}" region=us-east-1 snapshot_max_age=60
```

It is an 8GB standard volume.

The way I tested was by simply appending `time` to the playbook execution

`time ansible-playbook aws.yml`

robjailall:
Updated this pull here in a pycon sprint:

https://github.com/ansible/ansible-modules-core/pull/1143

hkariti:
Closed this in favor of #1143

PR #77, opened 2014-09-30
Updating SELinux example doc in lininefile module

major:
Encouraging users to use this Ansible module to _enable_ SELinux seems
like a better idea. It also warms Dan Walsh's heart.

Signed-off-by: Major Hayden major@mhtx.net

### DIFF
--- a/files/lineinfile.py
+++ b/files/lineinfile.py
@@ -127,7 +127,7 @@ options:
 """
 
 EXAMPLES = r"""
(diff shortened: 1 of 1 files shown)


mpdehaan:
I agree.  That was set there many years ago to see if someone like you or Dan Walsh would find it.

You may now claim your prize.

Will appear in next docs update.

--Michael

PR #87, opened 2014-09-30
Tweak docs to mention `replace`

offby1:
I only found out about the "preview" module by asking on IRC, so I figured it'd be good to point to it from the module that I already knew about.


mpdehaan:
Definitely all in favor of suggesting alternatives - thanks for this, it will appear in the next docs build.

PR #95, opened 2014-10-01
Use insserv where available

grahamrhay:
Use insserv for enabling/disabling services on Debian 7.


bcoca:
@grahamrhay can you rebase this request?

grahamrhay:
done

PR #100, opened 2014-10-01
Unarchive no longer transfers archive if creates= check succeeds

rdolejsi:
Unarchive no longer transfers the archive in case the creates= parameter yields a successful match, thus preventing lenghty operations over the network (especially with large archive distributions).

### DIFF
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -185,6 +185,7 @@ def main():
             dest              = dict(required=True),
             copy              = dict(default=True, type='bool'),
             creates           = dict(required=False),
+            precheck          = dict(default=False, type='bool'),
         ),
         add_file_common_args=True,
     )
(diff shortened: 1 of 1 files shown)


rdolejsi:
Please note: the unpack module on the client is actually called twice to achieve this - with extra signalling using precheck=True/False parameter. Please suggest other form of signalling if this is not the correct way of doing it.

rdolejsi:
This is a replacement request for the closed pullreq #9139 (no changes, update to current HEAD).

This is a half of the patch, complementary to core Ansible pullreq http://github.com/ansible/ansible/pull/9232. Both pullreqs are required.

abadger:
Added a comment on https://github.com/ansible/ansible/pull/9232 -- I think we merged a different pull request in November that achieved the same goals.

abadger:
Commit that implemented a fix for this: https://github.com/ansible/ansible/commit/e79acf3f882a2f9d46a44df500ce9cd9955ca227

# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #102, opened 2014-10-01
template, copy: archive, (m/a)time support added, file: (m/a)time support added

rdolejsi:
template, copy: added support for archive parameter to retain permissions, ownership and timestamps of the source(s) on the destination(s).

template, copy, file: added support for mtime, atime - setting access/mod datetime of the destination(s) in unix touch format HHMMDDhhmi[.ss].

Please note: mtime, atime suport addition is a side-effect of archive and cannot be split in separate pull request.

### DIFF
--- a/files/copy.py
+++ b/files/copy.py
@@ -148,7 +148,10 @@ def main():
             backup            = dict(default=False, type='bool'),
             force             = dict(default=True, aliases=['thirsty'], type='bool'),
             validate          = dict(required=False, type='str'),
-            directory_mode    = dict(required=False)
+            directory_mode    = dict(required=False),
+            atime             = dict(required=False),
+            mtime             = dict(required=False),
+            archive           = dict(type='bool',required=False), # Dissolved into other parameters before delegation to remote host
         ),
         add_file_common_args=True,
         supports_check_mode=True,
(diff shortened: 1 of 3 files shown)


rdolejsi:
The template archive option is a crucial one when iterating and templating directory structure which includes executable scripts, configurations with diverse access rights, etc.

rdolejsi:
This is a replacement request for the closed pullreq #9143 (no changes, update to current HEAD).

This is a half of the patch, complementary to Ansible pullreq http://github.com/ansible/ansible/pull/9233. Both pullreqs are required.

mpdehaan:
I'm almost thinking retaining the original times might be a _bad_ thing as that doesn't indicate things have changed.  Though if it's possible already via user accounts, perhaps this is not our problem.

rdolejsi:
Good point.

Changing the times is fully supported on user accounts (using touch) and is useful during copy command.
On the other hand, having correct permissions while updating the time might be valid scenario as well - i.e. updated properties on Ansible master will change files on hosts when templating (as administrators would probably like to see the time of change).

When thinking about it, it might be good to have multiple 'archive' modes - i.e. retain all, retain permissions only on content change.. What do you think?

gregdek:
@rdolejsi, thanks for submitting this pull request. Please accept our apologies for taking so long to review it. 

gregdek:
@rdolejsi A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

gregdek:
Closed due to contributor inactivity. Note to future contributors: if you would like to pick up this contribution, please do so by opening a new pull request rather than reopening this one. Thanks!

PR #104, opened 2014-10-01
Add documentation that add_host is a host list bypass plugin

sivel:
Ref: https://github.com/ansible/ansible/pull/7499

Over the past few weeks there has been an increase in the number of people being confused and submitting issues to the fact that add_host only runs on the first host in a playbook.

This pull requests adds a few notes items to the add_host module docs to hopefully clarify this some place in the docs.

### DIFF
--- a/inventory/add_host.py
+++ b/inventory/add_host.py
@@ -19,6 +19,15 @@ options:
     description:
     - The groups to add the hostname to, comma separated.
     required: false
+notes:
+  - This module is most commonly used in conjunction with a C(with_) style loop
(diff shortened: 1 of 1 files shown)


mpdehaan:
Thanks -I still find this documentation a bit confusing but leaving this open so we can figure out what we may want to write here.

An example or comment int he examples block might be easier.

bendavies:
Let's get this progressed. I've just wasted a lot of time misunderstanding this.

sivel:
Closing due to #782

PR #105, opened 2014-10-01
ec2_lc does not create valid launch configuration for EC2-Classic

luisguilherme:
With the introduction of `assign_public_ip`, `ec2_lc` always sets a value for this parameter, which is invalid for EC2-Classic requests (being accepted only on EC2-VPC). 

Solution: when `assign_public_ip` is omitted, will not pass this parameter to boto.


luisguilherme:
This fixes #273

forty:
Hello,

I still have this issue https://github.com/ansible/ansible-modules-core/issues/273 on 1.9.0.1
I believe this PR was supposed to fix it, but it does not appear to be the case here

forty:
My bad, actually the launch configuration I was working on was incorectly created with the previous version and the new version did not fix it (which might be ok), but when I deleted the LC, I was able to recreate it properly with 1.9.

PR #108, opened 2014-10-01
Fix ec2 asg json serialization error with Tag object

filipenf:


cloud/ec2_asg.py
```diff
@@ -245,6 +245,7 @@ def get_properties(autoscaling_group):
                 properties['pending_instances'] += 1
         properties['instance_facts'] = instance_facts
     properties['load_balancers'] = autoscaling_group.load_balancers
+    properties['tags'] = [{ 'key': t.key, 'value' : t.value} for t in tags ]
```
filipenf: The issue was here. Without this line we would get:

TypeError: Tag(chaordic:product=mail) is not JSON serializable

jimi-c:
Hi @filipenf, I took a slightly different tact in fixing this bug, which is in the commit above. Thanks for your contribution though, and if you have any further questions about this please let us know!

PR #110, opened 2014-10-01
Adding Google DNS module

erjohnso:
New module to add support for Google DNS [https://cloud.google.com/dns/]. This is modeled very closely to the route53 module.

As with the other Google modules, this is implemented using libcloud and a safety import-check is included.


cloud/gc_dns.py
```diff
@@ -0,0 +1,349 @@
+#!/usr/bin/python
```
fcuny, line 1: could you use `/usr/bin/env python` instead ? That way it also works nicely with virtualenv

cloud/gc_dns.py
```diff
@@ -0,0 +1,349 @@
...
+            value = dict(required=False),
+            pem_file = dict(),
+            service_account_email = dict(),
+            project_id = dict(),
+        )
+    )
+    conn = gce_connect(module, Provider.GOOGLE)
+    # TODO: figure out why this is needed to prevent SSL-required error
+    conn.connection.secure = True
+    conn.connection.port = 443
```
fcuny, line 201: This was due to a bug in libcloud. I've submitted a fix over there.

fcuny:
LGTM!

erjohnso:
Per agreement with @mpdehaan, starting this out in extras.

PR #117, opened 2014-10-02
Add word boundary in apache2_module regexp

direvus:
Add a word boundary \b to the regexp for checking the output of a2{en,dis}mod,
to avoid a false positive for a module that ends with the same text as the
module we're working on.

For example, the previous regexp r'.*spam already enabled' would also match
against 'eggs_spam already enabled'.

Also, get rid of the redundant '.*' from the end of the regexp.


direvus:
Note: this is a resubmit of ansible/ansible#8662.

direvus:
Ugh, reopen after accidental close.

PR #123, opened 2014-10-03
Restart Docker container when volumes, ports, or env vars have been changed

aaronfeng:
The current implementation only restarts the running container if the image:tag combo is different for named containers.

This PR will also restart the container if volumes, ports, and env vars do not match current state of Ansible.  This is desirable for development for ease of restarting the containers while tweaking the parameters.  It is also beneficial for production since those parameters can evolve over time.

I know this PR is also missing other potential important parameters such as links.  I have no problem adding further support for other fields, but I want to make sure this will be accepted by upstream.  With that said, I do think this PR provides value as it and other parameters can be part of another PR.


matthughes:
+1 This directly addresses the concerns raised here: https://groups.google.com/forum/#!topic/ansible-project/Du7gC9tiKH4

abevoelker:
This would be really nice to have! :+1:

lapluviosilla:
This would be great. :+1: Trying to figure out a workaround in this very moment.

aaronfeng:
@lapluviosilla @abevoelker @matthughes mind testing it out?  It might get merged faster this way :) cc/ @mpdehaan

dkerwin:
:+1:  Any updates?

bcoca:
@aaronfeng can you rebase, we reorganized the modules and this does not merge cleanly anymore

aaronfeng:
@bcoca OK, I will check it out soon.

aaronfeng:
@bcoca I was able to merge it cleanly, but it seems like the docker syntax have changed.  Is the example section of docker.py outdated?

`ERROR: docker is not a legal parameter in an Ansible task or handler`

I assume this is caused by the directory structure change.  I tried `docker/docker` in my yml, but no luck.

bcoca:
this is probably caused by either submodules update or the subcategories one, try updating the main ansible repo to latest devel and the same with submodules.

aaronfeng:
@bcoca oh right, totally forgot about that.  It's working now.

aaronfeng:
Just a quick status check on this.  Let me know if anything is needed before merging.

PikachuEXE:
Anything prevents this from merging?

P.S. Also how do I get module update if this is merged?

abadger:
latest comments on this PR: https://github.com/ansible/ansible-modules-core/pull/502 have started to discuss whether we should control this via a new state or parameter.

aaronfeng:
@abadger I see the other PR as orthogonal to this one since they don't accomplish the same thing.  This PR will detect if the container configuration has been changed.  If so, restart it.

What's the timeline for this PR to be merged?

jimi-c:
@aaronfeng testing this out today, and I've corrected one or two minor bugs I've found, but also ran into a bit of a problem. In your `key_val_changed` and `ports_changed` methods, you compare the lengths of the values to those specified in the module execution, however in my testing case I have an image that sets some values outside of the module execution. For example (this is a tomcat container), the Env looks like this:

<pre>        "Env": [
            "CATALINA_OPTS=-Dconfig.zookeepers=zookeeper1,zookeeper2",
            "HOME=/",
            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/tomcat/bin",
            "TOMCAT_VERSION=8.0.9",
            "DEPLOY_DIR=/maven",
            "CATALINA_HOME=/opt/tomcat",
            "DEBIAN_FRONTEND=noninteractive",
            "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/lib/x86_64-linux-gnu/"
        ],
</pre>

But the only env options specified in the playbook were:

<pre>
      env:
         CATALINA_OPTS: "-Dconfig.zookeepers=zookeeper1,zookeeper2"
</pre>

Similarly, the ports were not specified in the playbook but are already exposed in the Dockerfile. This leads to the result that my container is always restarted, despite nothing changing in the playbook.

However, removing these length comparisions from your code also present a problem, in which you could then add environment settings/ports but could not later remove them. This may not be a huge concern, as we could document that users should use state=restarted when removing options, but that is less than optimal.

Any thoughts on how to address this?

jimi-c:
Just thinking about this a bit more, the values for the container you're inspecting should be pulled out before and after any changes are made, and then those states should be compared instead of the current state being compared against the module params.

As a result of this, we're going to go ahead and close this issue and let you rework it as specified above. A full list of other minor fixes I've made to this patch as I've tested it can be found here:

https://gist.github.com/jimi-c/b489a515e0a7b3c64f3c

Which include:
1) Change the name of the `restart_named_container` to `should_restart_named_container`, which more accurately reflects what the method is doing.
2) Make sure ports are strings before trying to use them as such.
3) Make sure splits have a limit, to avoid tuple results with more than 2 values inside a dict() constructor.
4) The `remote_port` value may be None, so test that before trying to use it as a string

Thanks!

aaronfeng:
@jimi-c Well, I think the assumption is that you will be fully managing Docker from your playbook.  Any changes outside of the playbook should be considered "dirty" and the playbook will act accordingly (in this case restart the container) since it's the authority of the configuration.  I think once outside intervention has been introduced there will be no way for Ansible to know or manage properly.

Any changes outside of the playbook, I consider configuration drift and I want to normalize it.

jimi-c:
@aaronfeng be that as it may, users are not going to know about every EXPOSE or ENV option in the Dockerfile, especially since those are not visible when pulling an image. The better solution is to compare the state of the container before and after ansible makes changes (if any) and to then take action accordingly. 

I agree this is a very useful feature, however we don't want people to suddenly have containers restarting after every ansible run because of this change, which would almost certainly be the case if we merged this in as-is.

abadger:
@aaronfeng  I think jimi-c's usage is more inline with other modules.  We want to ensure the state of the parameters that are specified rather than the state of the container as a whole.  I can't quite nail down why this is but the example that springs to my mind is that you can mange the docker container in ansible via multiple plays in the playbook.

Say you want to say all docker containers running in group internal need to have a certain port open and in a different part say that all docker containers running a web client need to have a specific http_proxy env var set.  As long as we're adding and subtracting these options, the user can manage that in two tasks with those logical groups.  If we're mandating the entire state, then the user needs a third play where we've had to calculate the overlap between the two groups.

abadger:
@aaronfeng, maybe an additional parameter to specify whether the specified values should be applied to the existing set of values or to a clean slate would be appropriate here.  Or maybe we should discuss this on the mailing list (github isn't a good place to have contentious discussions because (1) the issue tracker is just a single threaded forum which doesn't lend itself to branching discussions and (2) only few of the stakeholders are participating.  Probably should use https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

abadger:
Okay, as part of #502 and its followup #912, I think we've merged something that does pretty much the same thing as here while addressing @jimi-c's bug report so that users don't have to specify all of the settings that the container itself makes and the containers don't end up restarting every single time.  We dicsussed this at some length and decided that we could add a parameter as talked about in https://github.com/ansible/ansible-modules-core/pull/123#issuecomment-76231289 to do additive/subtractive instead of replacement at some time in the future if it becomes an issue; fixing the configured-by-container vs configured-by-ansible problem was sufficient to make this not too problematic for end users.

@aaronfeng does using state=reloaded with the present code in the devel or stable-1.9 branch fit your needs?

abadger:
Closing this as it should be implemented by #502 

# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #126, opened 2014-10-03
Gentoo/OpenRC+systemd: check for systemd if falling back to service command

Tatsh:
Gentoo with OpenRC fix
See https://bugs.gentoo.org/show_bug.cgi?id=504116

This is a re-do of https://github.com/ansible/ansible/pull/8041


ghost:
i think this PR might be obsolete as ansible as moved on towards using /proc/1/comm to see if systemd is not just installed, but also enabled: https://github.com/ansible/ansible-modules-core/blob/devel/system/service.py#L415  This should appear in the next ansible major release.

ghost:
you can probably test it by simply replacing with service.py with the one from master.

Tatsh:
I like that approach better. I will give it a try.

Tatsh:
Works for me. Closing this in favour of the `/proc/1/comm` way of detection.

PR #128, opened 2014-10-03
apt_key: add keyserver example to docs

lorin:
For the `apt_key` module, add keyserver example to docs.


jimi-c:
Merged, thanks!

PR #136, opened 2014-10-05
adding json fail when state == absent and name does not exist

mschurenko:
IMO there should be a failure if the asg you are trying to delete does not exist. Currently it exits with changed=False.


mschurenko:
The more I thought about it the more I agree that the existence of an AWS resource should be idempotent.  Going to close.

PR #137, opened 2014-10-05
chaging type of vpc_zone_identifier from str to list

mschurenko:
I've changed the type for vpc_zone_identifier from 'str' to 'list' as this makes more sense and is more consistent. I've also changed the example.

Before:

```
vpc_zone_identifier: 'subnet-abcd1234,subnet-1a2b3c4d'
```

Now:

```
vpc_zone_identifier:
      - subnet-abcd1234
      - subnet-1a2b3c4d
```

This change may not be backwards compatible; however IMO making ansible syntax more consistent is a win for everyone.


juliedavila:
@mschurenko can you rebase when you have a moment?

juliedavila:
@mschurenko It's been 2 weeks, I'm recommending this PR be closed.

mschurenko:
It looks like this was updated already so I'll close this.

PR #141, opened 2014-10-06
get_url: Allow additional hashing algorithms

muffl0n:
Allow all hashing algorithms that hashlib provides (e.g. 'md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512'). Introduces a new parameter "checksum" with format: `<algorithm>:<checksum>`, e.g. `checksum="md5:66dffb5228a211e61d6d7ef4a86f5758"`
Deprecates parameter `sha256sum`. Uses hashlib directly instead of methods in module_utils/basic.py.

Also gives helpful error messages:
`msg: Could not hash with algorithm 'md6'. Available algorithms: ('md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512')`
`msg: The checksum parameter has to be in format <algorithm>:<checksum>`

Moved from https://github.com/ansible/ansible/pull/7639. I'm aware of the fact that my original PR was already denied. But I'm still convinced that this is the most future-proof solution as the possitibilities grow while hashlib grows. So I leave this here, open for discussion.

### DIFF
--- a/network/basics/get_url.py
+++ b/network/basics/get_url.py
@@ -72,7 +72,20 @@ options:
       - If a SHA-256 checksum is passed to this parameter, the digest of the
         destination file will be calculated after it is downloaded to ensure
         its integrity and verify that the transfer completed successfully.
+        This option is deprecated. Use 'checksum'.
     version_added: "1.3"
+    required: false
+    default: null
+  checksum:
+    description:
+      - If a checksum is passed to this parameter, the digest of the
+        destination file will be calculated after it is downloaded to ensure
+        its integrity and verify that the transfer completed successfully.
+        Format: <algorithm>:<checksum>, e.g.: checksum="sha256:d98291acbedd510e3dbd36dbfdd83cbca8415220af43b327c0a0c574b6dc7b97"
+        If you worry about portability, only the sha1 algorithm is available 
+        on all platforms and python versions.  The third party hashlib 
+        library can be installed for access to additional algorithms.
+    version_added: "2.0"
     required: false
     default: null
   use_proxy:
@@ -144,15 +157,13 @@ EXAMPLES='''
 
 - name: download file with custom HTTP headers
   get_url: url=http://example.com/path/file.conf dest=/etc/foo.conf headers: 'key:value,key:value'
+
+- name: download file with check
+  get_url: url=http://example.com/path/file.conf dest=/etc/foo.conf checksum=sha256:b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c
+  get_url: url=http://example.com/path/file.conf dest=/etc/foo.conf checksum=md5:66dffb5228a211e61d6d7ef4a86f5758
 '''
 
 import urlparse
-
-try:
-    import hashlib
-    HAS_HASHLIB=True
-except ImportError:
-    HAS_HASHLIB=False
 
 # ==============================================================
 # url handling
@@ -209,6 +220,28 @@ def extract_filename_from_headers(headers):
 
     return res
 
+def get_available_hash_algorithms():
+    available_hash_algorithms = dict()
+    try:
+        import hashlib
+
+        # python 2.7.9+ and 2.7.0+
+        for attribute in ('available_algorithms', 'algorithms'):
+            algorithms = getattr(hashlib, attribute, None)
+            if algorithms:
+                break
+        if algorithms is None:
+            # python 2.5+
+            algorithms = ('md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512')
+        for algorithm in algorithms:
+            available_hash_algorithms[algorithm] = getattr(hashlib, algorithm)
+    except ImportError:
+        import md5
+        import sha
+        available_hash_algorithms = {'md5': md5.md5,
+                                     'sha1': sha.sha}
+    return available_hash_algorithms
+
 # ==============================================================
 # main
 
@@ -219,6 +252,7 @@ def main():
         url = dict(required=True),
         dest = dict(required=True),
         sha256sum = dict(default=''),
+        checksum = dict(default=''),
         timeout = dict(required=False, type='int', default=10),
         headers = dict(required=False, default=None),
     )
@@ -233,6 +267,7 @@ def main():
     dest = os.path.expanduser(module.params['dest'])
     force = module.params['force']
     sha256sum = module.params['sha256sum']
+    checksum = module.params['checksum']
     use_proxy = module.params['use_proxy']
     timeout = module.params['timeout']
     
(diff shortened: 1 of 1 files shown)


muffl0n:
As predicted in https://github.com/ansible/ansible/pull/7639#issuecomment-51659891 @mgcrea is [asking for sha1-support](https://github.com/ansible/ansible/issues/10216#issuecomment-86570038). So there is the next reason to merge this PR as this already covers this requirement.

mgcrea:
:+1:, I like the future-proof single arg.

gregdek:
Clearly people like this PR. Thanks @muffl0n for submitting it. Please accept our apologies for taking so long to review it.

abadger:
- I like the format.  Making this automatically expandable to the set of available hashes is good
- Remember to update version_added in documentation to 2.0
- In documentation of checksum, add that only sha1 is currently available on all known platforms.  State that the hashlib library from pypi can be used for additional hashes on python<2.5
- If hashlib is not available, we should still try the stdlib's sha1 and md5 libraries
- I'd like to avoid regex if we can.  Maybe we should take the approach of validating the checksum instead of trying to correct it?  So something like:

``` python
try:
    method, checksum = checksum.rsplit(':', 1)
    # Ensure the checksum portion is a hexdigest
    int(checksum, 16)
except ValueError:
    module.fail_json(msg="The checksum parameter has to be in format <algorithm>:<checksum>")
```
- hashlib.algorithms isn't available until 2.7.0.  Prior to that I think we have to rely on the guaranteed hashes (SHA-224, SHA-256, SHA-384, and SHA-512 .)
- hashlib.algorithms_available is better than hashlib.algorithms but it is only available from 2.7.9 onwards.  Taking all that into account, figuring out the valid algorithms should looks something like this:

``` python
hasher = None
if not HAS_HASHLIB:
    if method == 'md5':
        import md5
        hasher = md5.md5()
    elif method == 'sha1':
        import sha
        hasher = sha.sha()
else:
    try:
        hasher = hashlib.new(method)
    except ValueError:
        pass
if hasher is None:
    if HAS_HASHLIB:
        # python 2.7.9+ and 2.7.0+
        for attribute in ('available_algorithms', 'algorithms'):
            algo = getattr(hashlib, attribute, None)
        if algo is None:
            # python 2.5+
            algo = ('sha224', 'sha256', 'sha384', 'sha512')
    else:
        algo = ('md5', 'sha1')
    os.remove(dest)
    module.fail_json(msg="Could not hash with algorithm '%s'. Available algorithms: %s " % (method, algo))
```
- instead of pushing all of the file's contents into the hasher's contructor, probably a good idea to read a chunk of the file at a time and then pass the chunk to the hasher's .update() method.  That will prevent running short of memory with extremely large files or low memory computers.
  
  I'm pretty tired right now so forgive me if I've missed anything!

muffl0n:
@abadger: Sounds reasonable! Especially the backward compatibility stuff. :)

As I'm still feeling not that confident with python I would be glad if you could do a new PR with your suggestions added. It would need me a pretty long time to understand everything and get the PR right. And I don't want to risk this feature not getting merged now it has a bit of attention. :D

gregdek:
My guess is @abadger will be busy getting 2.0 ready for release. Maybe @mgcrea could help? Also still needs to be rebased.

gregdek:
Not sure how to proceed here -- looks like @muffl0n is not confident in his ability to make the necessary improvements, and @abadger is busy elsewhere currently. And I don't want to leave it indefinitely or it will bitrot.

I think we've got three choices:
1. @muffl0n give it a try anyway despite your reservations. :) If that's the choice, rework the code, rebase to latest, and we'll review it again.
2. Have a new contributor like @mgcrea (or someone else!) pick it up and run with it; if that's the case, please pick up the branch and submit it in a new PR, and close this one; be sure to credit @muffl0n in the new PR.
3. Set it aside with regrets for now and close the PR.

muffl0n:
Despite the fact that my abilities are not good enough yet I'm quite eager to learn and get this done. :) I just rebased the PR onto the current devel and will have a look at the features/changes proposed by @abadger asap.

muffl0n:
I gave it a try and it doesn't look _that_ bad. :) Perhaps get_available_hash_algorithms() should be moved to module_utils so that other modules don't have to reimplement this.

Unfortunately I could only test on Python 3.x, 2.7 and 2.6. I could not get my 2.5 test setup running because of missing SSL-capabilities:

``````
fatal: [default]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "BECOME-SUCCESS-wnebiluqbsihjdkmjkgbgjasuywhkgid\r\nTraceback (most recent call last):\r\n  File \"/home/vagrant/.ansible/tmp/ansible-tmp-1437745549.42-23029274097793/get_url\", line 2165, in <module>\r\n    PROTOCOL = ssl.PROTOCOL_TLSv1\r\nNameError: name 'ssl' is not defined\r\n", 
    "parsed": false
}```
``````

File: network/basics/get_url.py [hunk 1]
```diff
@@ -212,6 +220,26 @@ def extract_filename_from_headers(headers):
...
     return res
 
+def get_available_hash_algorithms():
+    available_hash_algorithms = dict()
+    try:
+        import hashlib
+
+        # python 2.7.9+ and 2.7.0+
+        for attribute in ('available_algorithms', 'algorithms'):
+            algorithms = getattr(hashlib, attribute, None)
```
abadger: Oops, the sample code I posted was missing a break here.  Should be:

``` python
for attribute in ('available_algorithms', 'algorithms'):
    algorithms = getattr(hashlib, attribute, None)
    if algorithms:
        break
```

network/basics/get_url.py
```diff
@@ -212,6 +220,26 @@ def extract_filename_from_headers(headers):
...
+            algorithms = ('md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512')
+        for algorithm in algorithms:
+            available_hash_algorithms[algorithm] = getattr(hashlib, algorithm)
+    except ImportError:
+        import md5
+        import sha
+        available_hash_algorithms = {'md5': md5.md5,
+                                     'sha1': sha.sha}
+    return available_hash_algorithms
+
```
abadger, line 220: nice.  I like this.  You're right we might want to move this into the module_utils/\* code at some point in the future.

network/basics/get_url.py
```diff
@@ -75,9 +75,19 @@
...
+        This option is deprecated. Use 'checksum'.
     version_added: "1.3"
     required: false
     default: null
+  checksum:
+    description:
+      - If a checksum is passed to this parameter, the digest of the
+        destination file will be calculated after it is downloaded to ensure
+        its integrity and verify that the transfer completed successfully.
+        Format: <algorithm>:<checksum>, e.g.: checksum="sha256:d98291acbedd510e3dbd36dbfdd83cbca8415220af43b327c0a0c574b6dc7b97"
```
abadger, line 84: Add something like this in here:

```
If you worry about portability, only the sha1 akgorithm is available on all platforms and python versions.  The third party hashlib library can be installed for access to additional algorithms.
```

abadger:
Okay, just a few notes to address.  Otherwise looks good!

I also saw that you did a merge instead of a rebase so it pulled in an unrelated commit.  If you're familiar with git you can try to get rid of that.  If not, I can try to once the final changes are made and it's time to merge.

muffl0n:
@abadger: Thank you very much for you helpful and motivating comments! :) I already tried to get rid of the merge-commit. I somewhat succeeded, but there are still some commits that are quite senseless. But I'm eager to get this cleaned up: learning git the hard way. :)

muffl0n:
Is sha1 really the only algorithm that is available in all versions of python? If that's correct, then there might be some versions where this will fail because of the missing md5 module:

```
    except ImportError:
        import md5
        import sha
```

bcoca:
fips enabled systems will not have md5, so you might want to split each import into diff exception blocks

muffl0n:
- catch ImportError when importing md5
- removed example with parameter "sha256sum" <- I'm not sure about this one. This parameter should only be kept for backward compatibility, but maybe a deprecation-note at the example line would be more sane?
- As we return checksum_src and checksum_dst, this should be useless, right?

```
    # Backwards compat only.  We'll return None on FIPS enabled systems
    try:
        md5sum = module.md5(dest)
    except ValueError:
        md5sum = None
```

lukehoersten:
This functionality is useful for any module that downloads a file. For example, unarchive now supports downloads. Is there a way this code could be shared with other modules so code doesn't have to be replicated. It wasn't an issue when it was just sha256 because of the simplicity. See my associated PR: https://github.com/ansible/ansible-modules-core/pull/1840

muffl0n:
@LukeHoersten: Indeed! :) 

I added a PR for this change in ansible/ansible#11840 and changed the code for get_url in this PR accordingly. Feel free to have a look and see if that does work for you.

lukehoersten:
@muffl0n care to update the `unarchive` module as well as a part of this PR? Otherwise I'll just have to rewrite the PR I made for it.

muffl0n:
I would prefer not to mix the changes for `get_url` and `unarchive`. For now, let's just hope that ansible/ansible#11840 get's merged so that we can both use the same functionality. :)

lukehoersten:
Sounds good.

abadger:
Okay, squashed the commits into a single one and merged!  github doesn't seem to automatically detect the merge when I do that, but it's here: https://github.com/ansible/ansible-modules-core/commit/4f0cf6d2ca0b46d289f631496af373fac02adf1c

abadger:
Thanks @muffl0n !

# Merged

Hi!

issue is resolved.

Thank you!

muffl0n:
Yay! :) Thanks for merging!

PR #144, opened 2014-10-06
Add `cache_subnet_group_name` to elasticache module

quodlibetor:
According to the [AWS docs on elasticache and the VPC](http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/CacheSubnetGroups.html) subnet groups are required inside VPCs.

### DIFF
--- a/cloud/amazon/elasticache.py
+++ b/cloud/amazon/elasticache.py
@@ -485,6 +485,7 @@ class ElastiCacheManager(object):
 
 def main():
     argument_spec = ec2_argument_spec()
+    default = object()
     argument_spec.update(dict(
             state={'required': True, 'choices': ['present', 'absent', 'rebooted']},
             name={'required': True},
@@ -494,7 +495,7 @@ def main():
             num_nodes={'required': False, 'default': None, 'type': 'int'},
             cache_port={'required': False, 'default': 11211, 'type': 'int'},
             cache_subnet_group={'required': False, 'default': None},
-            cache_security_groups={'required': False, 'default': ['default'],
+            cache_security_groups={'required': False, 'default': [default],
                                    'type': 'list'},
             security_group_ids={'required': False, 'default': [],
                                    'type': 'list'},
@@ -523,6 +524,14 @@ def main():
     zone = module.params['zone']
     wait = module.params['wait']
     hard_modify = module.params['hard_modify']
+
+    if cache_subnet_group and cache_security_groups == [default]:
+        cache_security_groups = []
+    if cache_subnet_group and cache_security_groups:
+        module.fail_json(msg="Can't specify both cache_subnet_group and cache_security_groups")
+
+    if cache_security_groups == [default]:
+        cache_security_groups = ['default']
 
     if state == 'present' and not num_nodes:
         module.fail_json(msg="'num_nodes' is a required parameter. Please specify num_nodes > 0")
--- a/cloud/amazon/elasticache.py
+++ b/cloud/amazon/elasticache.py
@@ -58,6 +58,12 @@ options:
       - The port number on which each of the cache nodes will accept connections
     required: false
     default: 11211
+  cache_subnet_group:
+    description:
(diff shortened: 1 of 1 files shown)


rrana:
Any progress on this?

craigmarvelley:
I'd really appreciate this being merged in once the conflicts are resolved. Happy to do that and resubmit if @quodlibetor isn't able to?

quodlibetor:
I'm extremely willing to do any work to help get this merged :)

quodlibetor:
Also, FWIW the way that I'm working around not having this merged upstream is just having a copy of the fixed version in my `library/cloud/` dir, which the default `ansible.cfg` has as a higher priority than the global one.

bcoca:
@quodlibetor can you rebase, this PR currently does not merge cleanly

quodlibetor:
Rebased, no conflicts when rebasing. :8ball:

quodlibetor:
Any other problems?

InfiniteLee:
Bump for merging!

rouge8:
I don't think this actually works, because `cache_security_groups` will always be set to `['default']`, which is invalid within a VPC.

rouge8:
Nevermind, you can set `cache_security_groups: []`.

barraponto:
and is there a way to create a cache subnet group using ansible?

juliedavila:
@quodlibetor this looks good, can you change the parameter to `cache_subnet_group` instead of being `cache_subnet_group_name`
@rouge8 Hmm, seems hacky, I'll refactor this later so it can detect if it's going to go into a vpc and trigger a `[]` default.
@barraponto not atm, but will look into adding that into the module elasticache after the PR is merged.

quodlibetor:
Updated to use `cache_subnet_group` and rebased.

quodlibetor:
I've added a change that defaults `cache_security_groups` to an empty list if nothing was provided.

I used a sentinel value to set the default since it's plausible that people are explicitly setting the list to `default`, in which case they would get some unexpected behavior if we just overwrote it. If the argument spec is ever serialized to json this would cause a ValueError traceback, I believe. I don't know how likely that is. 

I'm willing to change it to use the less-robust option if that's what you'd like.

juliedavila:
@quodlibetor 

Unfortunately that doesn't seem to work. With the sentinel value it gives the following error:

``` shell
ok: [localhost]

TASK: [Elasticache] ***********************************************************
msg: argument cache_security_groups is of invalid type: <type 'object'>, required: list

```

And for the sake of trying I attempted to cast default as a list and it returned

```

ok: [localhost]

TASK: [Elasticache] ***********************************************************
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427391056.94-38111842336176/elasticache", line 2365, in <module>
    main()
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427391056.94-38111842336176/elasticache", line 498, in main
    cache_security_groups={'required': False, 'default': list(default),
TypeError: 'object' object is not iterable


```

The less robust option might be the answer here, though I'm open to a different work around.

quodlibetor:
Your specific issue comes from the fact that `list()` actually creates a list by iterating over the contents of its argument, using `[default]` just wraps `default` in a list, which does work and is in the lastest version.

The newest version was tested with the following playbook:

```
---
- hosts: localhost
  tasks:
    - local_action:
        module: elasticache
        name: test-bwm
        state: present
        cache_subnet_group: doesnotexist
        num_nodes: 1
        region: us-east-1
```

And resulted in:

```
$ ansible-playbook -i ec2.py test.yml

PLAY [localhost] **************************************************************

ok: [localhost]

TASK: [elasticache] ***********************************************************
failed: [localhost -> 127.0.0.1] => {"failed": true}
msg: Cache Subnet Group doesnotexist does not exist.
```

Which is a correct error that comes from the call boto, so it passes all of the ansible verification now.

quodlibetor:
Rebased.

juliedavila:
Tested. Good to go. Thanks.

quodlibetor:
:sunrise:

PR #150, opened 2014-10-07
Add recursive support for ACL module

### LINKED ISSUE #5053: ACL module should support defaults
Using ansible 1.4

I'm trying to set a default acl on a directory.
With the command line tool one would do it this way:

```
setfacl -m d:group:mygroup:rwx directory/
```

so I tried to translate that in an ansible format:

```

---
- name: Setup ACL
  acl: name=/var/www entry="d:group:dev:rwx" state=present
```

But I get this error:

```
Invalid entry: 'd:group:dev:rwx', it requires 3 sections divided by ':'
```

The expected result is: no error should be thrown.

This might not be a bug, per se; according to @bcoca:

```
The module currently doesn't support the 'default' option as some systems seem to have problems with it and I didn't get enough feedback to enable detection (POSIXLY_CORRECT is only usable in systems that actually allow for the default option but disable it). 

so:
- name: Setup ACL
  acl: name=/var/www entry="group:dev:rwx" state=present

will work for the directory but not set the default for all files under it. 
```

### LINKED ISSUE #7276: acl recursive
Hi,

Could you add recursive option for acl module?

http://docs.ansible.com/acl_module.html

Thanks.

astorije:
This PR enables the support for recursive ACLs. Some refactoring was needed to make this work. It is the new repo structure augmented clone of ansible/ansible#8380.

It fixes ansible/ansible#5053, ansible/ansible#5550 and fix ansible/ansible#7276) as well as a few things along the way:
- PEP8 compliance(-ish)
- Fix erroneous entry param `etype` to `type`
- Param checking messages use RFC 2119 wording)
- Add checks for incompatible states

I did not specify the version in the documentation bit. I will add it when this is merged to have consistent docs.
## Advantages of this PR
- Permissions do not need to be normalized anymore using the `--test` checking system
- `--test` does not check using `getfacl`, so the code is less verbose and less likely to introduce a parsing error (since there is actually - almost - no parsing)
- `--test` makes `--recursive` possible at no performance cost, and checking the `changed` status is neither overly complicated or slow
- `default` can now be used on acl queries as well
- [`X` (`capital-x` / _Special execute_)](https://en.wikipedia.org/wiki/Chmod#Symbolic_modes) is now supported in the permissions with no extra code
## Inconvenients of this PR
- `--test` does not seem to be supported by BSD's `setfacl`. However, this can be solved in the `acl_changed` function only without affecting the rest of the file. Suggestions below.
- `recursive=yes` and `state=query` are currently incompatible because that needs further discussion to ensure backward compatibility. This should be done in another issue/PR though. I will be more than happy to work on it, but one step at a time :)
## General notes (and motivations) to help for the review
- `normalize_permissions` did not support `X` (`capital-x`) and its support would cause problem with the former checking system (since `X` itself is not stateless)
- `_run_acl` was making the usage of `--test` impossible because for the last line returned when using `--test` is relevant and not empty. It now makes sure that the last line is empty before stripping it.
- 2 new builder helpers `build_entry` and `build_command` now factorize some recurring code
- `acl_changed` is a function meant to test if a command changes ACLs. The checking can be extended there rather than in the `main` function
- I removed the `get_acls`, `set_acl` and `rm_acl` wrappers for the following reasons:
  - They were used only once each
  - They only contained one call to `build_command` and one call to `run_acl` in the end
  - The command is already built when it is `--test`ed before so I reuse the command rather than rebuilding it
- `_run_acl` is now `run_acl` because it is not a hidden/private function anymore, it is used in the `main` function multiple times
## Suggestions for Linux / BSD support

My PR breaks support for BSD because of its lack of `--test`, but adds support for `--recursive` on Linux. I suggest that when `--test` is not available, `changed` is **always** true when using `recursive=yes`. When not using `recursive=yes`, we can either do the same or replicate the old code for checking state. What I described earlier (notably the `X` bit) will still be a problem. I'll talk with @bcoca on IRC about that.

The documentation should then mention the support differences if there are  any (for example no support for `recursive` on BSD).

### DIFF
--- a/files/acl.py
+++ b/files/acl.py
@@ -21,6 +21,8 @@ version_added: "1.4"
 short_description: Sets and retrieves file ACL information.
 description:
      - Sets and retrieves file ACL information.
+notes:
+    - As of Ansible 2.0, this module only supports Linux distributions.
 options:
   name:
     required: true
@@ -80,7 +82,7 @@ options:
       - DEPRECATED. The acl to set or remove.  This must always be quoted in the form of '<etype>:<qualifier>:<perms>'.  The qualifier may be empty for some types, but the type and perms are always requried. '-' can be used as placeholder when you do not care about permissions. This is now superseded by entity, type and permissions fields.
 
   recursive:
-    version_added: "@@@"
+    version_added: "2.0"
     required: false
     default: no
     choices: [ 'yes', 'no' ]
@@ -150,7 +152,10 @@ def split_entry(entry):
 
 def build_entry(etype, entity, permissions=None):
     '''Builds and returns an entry string. Does not include the permissions bit if they are not provided.'''
-    return etype + ':' + entity + (':' + permissions if permissions else '')
+    if permissions:
+        return etype + ':' + entity + ':' + permissions
+    else:
+        return etype + ':' + entity
 
 
 def build_command(module, mode, path, follow, default, recursive, entry=''):
@@ -163,6 +168,7 @@ def build_command(module, mode, path, follow, default, recursive, entry=''):
         cmd.append('-x "%s"' % entry)
     else:  # mode == 'get'
         cmd = [module.get_bin_path('getfacl', True)]
+        # prevents absolute path warnings and removes headers
         cmd.append('--omit-header')
         cmd.append('--absolute-names')
 
@@ -187,7 +193,11 @@ def acl_changed(module, cmd):
     cmd = cmd[:]  # lists are mutables so cmd would be overriden without this
     cmd.insert(1, '--test')
     lines = run_acl(module, cmd)
-    return not all(line.endswith('*,*') for line in lines)
+
+    for line in lines:
+        if not line.endswith('*,*'):
+            return False
+    return True
 
 
 def run_acl(module, cmd, check_rc=True):
@@ -206,6 +216,9 @@ def run_acl(module, cmd, check_rc=True):
 
 
 def main():
+    if get_platform().lower() != 'linux':
+        module.fail_json(msg="The acl module is only available for Linux distributions.")
+
     module = AnsibleModule(
         argument_spec=dict(
             name=dict(required=True, aliases=['path'], type='str'),
@@ -295,7 +308,7 @@ def main():
             run_acl(module, command, False)
         msg = "%s is absent" % entry
 
-    else:
+    elif state == 'query':
         msg = "current acl"
 
     acl = run_acl(
--- a/files/acl.py
+++ b/files/acl.py
@@ -87,7 +87,7 @@ options:
     default: no
     choices: [ 'yes', 'no' ]
     description:
-      - Recursively sets the specified ACL (added in Ansible @@@). Incompatible with C(state=query).
+      - Recursively sets the specified ACL (added in Ansible 2.0). Incompatible with C(state=query).
 
 author: "Brian Coca (@bcoca)"
 notes:
--- a/files/acl.py
+++ b/files/acl.py
@@ -79,6 +79,14 @@ options:
     description:
       - DEPRECATED. The acl to set or remove.  This must always be quoted in the form of '<etype>:<qualifier>:<perms>'.  The qualifier may be empty for some types, but the type and perms are always requried. '-' can be used as placeholder when you do not care about permissions. This is now superseded by entity, type and permissions fields.
 
+  recursive:
+    version_added: "@@@"
+    required: false
+    default: no
+    choices: [ 'yes', 'no' ]
+    description:
+      - Recursively sets the specified ACL (added in Ansible @@@). Incompatible with C(state=query).
+
 author: "Brian Coca (@bcoca)"
 notes:
     - The "acl" module requires that acls are enabled on the target filesystem and that the setfacl and getfacl binaries are installed.
@@ -110,20 +118,6 @@ acl:
     sample: [ "user::rwx", "group::rwx", "other::rwx" ]
 '''
 
-def normalize_permissions(p):
-    perms = ['-','-','-']
-    for char in p:
-        if char == 'r':
-            perms[0] = 'r'
-        if char == 'w':
-            perms[1] = 'w'
-        if char == 'x':
-            perms[2] = 'x'
-        if char == 'X':
-            if perms[2] != 'x':  # 'x' is more permissive
-              perms[2] = 'X'
-    return ''.join(perms)
-
 def split_entry(entry):
     ''' splits entry and ensures normalized return'''
 
@@ -132,9 +126,9 @@ def split_entry(entry):
     if len(a) == 3:
         a.append(False)
     try:
-        p,e,t,d = a
+        p, e, t, d = a
     except ValueError, e:
-        print "wtf?? %s => %s" % (entry,a)
+        print "wtf?? %s => %s" % (entry, a)
         raise e
 
     if d:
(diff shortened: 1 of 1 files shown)


tristanbes:
acl module still not supports recursive option ? :'( too bad. Any move on this PR ?

astorije:
Still waiting for feedback from the Ansible guys, sorry :/

issei-m:
:+1:

tjanez:
This is a must have for my scenario, where I want to recursively set ACLs for a dedicated backup user to `rX` on all directories that are to be backed up.

@astorije, I suggest rebasing the PR to fix the merging conflict and then to send an email requesting attention to ansible-devel mailing list.

abadger:
We also have weekly public meetings to work on reviewing and merging the pull requests of people who stop by to discuss them.  If you're available, the meetings are currently every Tuesday from 17:00 UTC to 19:00 UTC.  We send an email to the mailing lists and message to the #ansible and #ansible-devel IRC channels just before the meeting starts with google hangout information (and you can chat on irc during the meeting time too if you can't do a google hangout.)

gregdek:
@astorije Thanks for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

astorije:
Oops yes sorry I dropped the ball, @tjanez, @abadger and @gregdek!
I'm on it for the rebase.

astorije:
Voilà, @gregdek, rebasing is done :-)
Happy reviewing!

gregdek:
Thx @astorije :)

This module is critical to Ansible's functionality, so it's actually owned by the Core team, which is why it's taken a long time to review. Thanks for your patience here.

astorije:
No worries! I'll be happy to contribute more when you guys can review and come back with some feedback.

cbrunnkvist:
Gah, just ran into this issue. 

Gonna have to work around it with `command:
 setfacl -R -m u:appuser:rwx /mypath` in the meantime.

astorije:
@gregdek, I just checked and no rebasing is necessary at the moment. I'll try to stop by maybe once a month to make sure this branch is up-to-date in case I win the review lottery :-)

@cbrunnkvist I'm afraid that's your only choice :-/
If the status of the execution matters to you, I advise you to take a look at the `--test` option.

files/acl.py
```diff
@@ -151,69 +145,87 @@ def split_entry(entry):
...
 
-    p = normalize_permissions(p)
+    return [d, t, e, p]
 
-    return [d,t,e,p]
 
-def get_acls(module,path,follow):
+def build_entry(etype, entity, permissions=None):
+    '''Builds and returns an entry string. Does not include the permissions bit if they are not provided.'''
+    return etype + ':' + entity + (':' + permissions if permissions else '')
```
puiterwijk: You cannot use ternary if because that's not in python 2.4

files/acl.py
```diff
@@ -151,69 +145,87 @@ def split_entry(entry):
...
+    '''Builds and returns agetfacl/setfacl command.'''
+    if mode == 'set':
+        cmd = [module.get_bin_path('setfacl', True)]
+        cmd.append('-m "%s"' % entry)
+    elif mode == 'rm':
+        cmd = [module.get_bin_path('setfacl', True)]
+        cmd.append('-x "%s"' % entry)
+    else:  # mode == 'get'
+        cmd = [module.get_bin_path('getfacl', True)]
+        cmd.append('--omit-header')
```
puiterwijk, line 172: You might want to re-add the warning: # prevents absolute path warnings and removes headers

files/acl.py
```diff
@@ -151,69 +145,87 @@ def split_entry(entry):
...
-        cmd.append('-k')
-    entry = entry[0:entry.rfind(':')]
-    cmd.append('-x "%s"' % entry)
-    cmd.append(path)
+def acl_changed(module, cmd):
+    '''Returns true if the provided command affects the existing ACLs, false otherwise.'''
+    cmd = cmd[:]  # lists are mutables so cmd would be overriden without this
+    cmd.insert(1, '--test')
+    lines = run_acl(module, cmd)
+    return not all(line.endswith('*,*') for line in lines)
```
puiterwijk: I'm having a hard time parsing this, can you refactor this?

files/acl.py
```diff
@@ -226,79 +238,72 @@ def main():
...
+        )
+        changed = acl_changed(module, command)
+
         if changed and not module.check_mode:
-            rm_acl(module,path,':'.join([etype, entity, '---']),follow,default)
-        msg="%s is absent" % ':'.join([etype, entity, '---'])
+            run_acl(module, command, False)
+        msg = "%s is absent" % entry
+
     else:
```
puiterwijk: Maybe make this "elif state == 'query' "

files/acl.py
```diff
@@ -226,79 +238,72 @@ def main():
     state = module.params.get('state')
     follow = module.params.get('follow')
     default = module.params.get('default')
-
```
puiterwijk, line 229: Could you add a check if this is linux and do a module.fail_json if not?

abadger:
Cool.  Talked with @puiterwijk as he was reviewing this.  Seems like once his review comments are addressed, this can be merged.

files/acl.py
```diff
@@ -79,6 +79,14 @@
...
       - DEPRECATED. The acl to set or remove.  This must always be quoted in the form of '<etype>:<qualifier>:<perms>'.  The qualifier may be empty for some types, but the type and perms are always requried. '-' can be used as placeholder when you do not care about permissions. This is now superseded by entity, type and permissions fields.
 
+  recursive:
+    version_added: "@@@"
+    required: false
+    default: no
+    choices: [ 'yes', 'no' ]
+    description:
+      - Recursively sets the specified ACL (added in Ansible @@@). Incompatible with C(state=query).
+
```
abadger, line 91: If the other review comments get addressed pretty soon, this module should be in 2.0 so you can go ahead and fill that version number in here too.

astorije:
Thanks @puiterwijk and @abadger, I'll try to address these ASAP.

files/acl.py
astorije: Good call, I did not know that.

Addressed in f8027a7. Does that seem good/Python enough to you? (Keep in mind that this PR is 1-year old and I haven't done any Python since then :) )

files/acl.py
astorije: Hi @puiterwijk, as I was building my defense for this syntax, I noticed that `all()` was Python 2.5+ only, so I replaced it with an ugly `for`-loop in f8027a7. Hope this looks better to you.

About the original syntax though, I do believe (and I checked with a very experienced Python dev colleague of mine) that this was the easiest way to put/read this.
In fact, this makes much more sense if you read it out loud, as this translates in English to:

```
if not all lines end with '*,*' then acls have changed, otherwise acls haven't changed
```

On top of that, I was using a generator so it can cleanly exit prematurely, for efficiency. Not exactly sure what difference that makes here, but assuming one is running the recursive option on a folder of hundred thousands sub-folders and files, on old hardware (after all, we do support Python 2.4 :-) ), it's always nice to terminate early if we can. My `for`-loop keeps this behavior.

files/acl.py
astorije: Addressed in f8027a7. Thanks for catching this.

files/acl.py
astorije, line 229: Done in f8027a7.

On one hand, I am very happy we now support recursive ACLs, on the other hand I am very sad we are dropping support for BSD because of this.

Note however that there are some artifacts left: the `-h`option (now on [line 179](https://github.com/ansible/ansible-modules-core/pull/150/files#diff-9c3c3b66bc72c68c9eafd9cefa084c71R179)) is BSD-only while the equivalent must be set with `-L`/`-P` for Linux. See the [Linux man](http://linux.die.net/man/1/setfacl) and the [FreeBSD man](http://www.freebsd.org/cgi/man.cgi?format=html&query=setfacl%281%29) for reference.

I'd be more than happy to give BSD its `acl` module back and to fix that issue described above once this PR makes it upstream. I'm afraid it will be forever delayed if making them part of this PR.

What do you think of my suggestions in the "Suggestions for Linux / BSD support" section of this PR description?

files/acl.py
astorije, line 91: Done in f8027a7.

astorije:
@puiterwijk and @abadger, thanks again for your review. I have now addressed all your comments and, as I figured you'd prefer few commits, I stashed all the fixes into one single commit: f8027a7.

Let me know if you have any further comments and if these are minor, I might be able to address them in a similar timeframe.

files/acl.py
```diff
@@ -21,6 +21,8 @@
 short_description: Sets and retrieves file ACL information.
 description:
      - Sets and retrieves file ACL information.
+notes:
+    - As of Ansible 2.0, this module only supports Linux distributions.
```
bcoca, line 25: is it possible to just make the recursive option ~~BSD~~ Linux only?

astorije, line 25: You mean Linux-only, right?

bcoca, line 25: yep, that is what i wrote :-)

astorije, line 25: Ah ah!

Not really, as everything this refactoring does is based off the `--test` option.

Once this PR makes it upstream, I will definitely work right away on a new PR to fix this. This implies a lot of code though, and this PR already does too much (add/fix capital-`X` support, other fixes, ...).
It would be very sad to see this PR delayed furthermore (initially opened more than a year ago). I promise I'll take care of BSD support right after that. I think I have shown good reactivity (if not always < 24h) every time I was asked, haven't I? :)

astorije, line 25: Plus we are talking about a new major version for Ansible (therefore making non-BC changes if following semantic versioning).
When is Ansible 2.0 due? Chances are I'll send a further PR before it gets out.

bcoca, line 25: OK, will evaluate this patch as is and work for the 'rebsdization' after it

astorije, line 25: Great, thanks a lot! Plus you just find the name of my future PR, thanks :)

astorije:
Note that I don't have any test playbooks with me so I did not run the last version of this code...

astorije:
Youhouuuuu! Thanks @bcoca!

Okay, to re-BSD-ize this module now, I see a few possible strategies. None of them are perfect, so I'll evaluate them and provide a comparison in a new PR. I remember I did this comparison a year ago, but my memory isn't fresh anymore so I'll work back on this before saying anything wrong.

drzraf:
worth noting ? issue  ansible/ansible#12514

jimi-c:
@drzraf closed that issue, as it should be opened in this code repo. Thanks!

drzraf:
acl: permissions=rX always marked as changed : #2151

astorije:
@drzraf, you didn't specify your environment in either ansible/ansible#12514, #2151 or here. I believe this PR fixed this issue along the way (for Linux, support for BSD being still in progress) but without more details it is hard to say.

drzraf:
Ansible 2.0.1.0

```
- name: operators permissions                                                                                                                                                             
  acl:                                                                                                                                                                                    
    recursive: yes                                                                                                                                                                        
    name: /srv/www/                                                                                                                                                                       
    etype: group                                                                                                                                                                          
    entity: operator                                                                                                                                                                      
    permissions: rwX                                                                                                                                                                      
    default: yes                                                                                                                                               
    state: present        
```

is always set as "changed"

drzraf:
Side note : `default=yes` should be prohibited for regular files

drzraf:
Side question: is there a reason `entry` would not support multiple comma-separated entries?

PR #153, opened 2014-10-07
Refactor force basic auth, now all modules which use fetch_url() can use force_basic_auth

woltage:
Check out PR: https://github.com/ansible/ansible/pull/9281 which is related to this PR.

### DIFF
--- a/network/basics/get_url.py
+++ b/network/basics/get_url.py
@@ -110,6 +110,15 @@ options:
         parameter is not specified, the C(url_password) parameter will not be used.
     required: false
     version_added: '1.6'
+  force_basic_auth:
+    description:
+      - httplib2, the library used by the uri module only sends authentication information when a webservice
+        responds to an initial request with a 401 status. Since some basic auth services do not properly
+        send a 401, logins will fail. This option forces the sending of the Basic authentication header
+        upon initial request.
+    required: false
+    choices: [ "yes", "no" ]
+    default: "no"
   others:
     description:
       - all arguments accepted by the M(file) module also work here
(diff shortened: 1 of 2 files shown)


gregdek:
@woltage thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

woltage:
Fixed, but I need to fix ansible/ansible#9281 too.

gregdek:
@jpmens can you review? If you give this a +1 we will tag "P1_reviewed" for quicker inclusion.

jpmens:
+1

network/basics/get_url.py
```diff
@@ -108,6 +108,15 @@
         parameter is not specified, the C(url_password) parameter will not be used.
     required: false
     version_added: '1.6'
+  force_basic_auth:
```
bcoca: requires version added

network/basics/uri.py
```diff
@@ -369,7 +368,6 @@ def main():
             body_format = dict(required=False, default='raw', choices=['raw', 'json']),
             method = dict(required=False, default='GET', choices=['GET', 'POST', 'PUT', 'HEAD', 'DELETE', 'OPTIONS', 'PATCH']),
             return_content = dict(required=False, default='no', type='bool'),
```
bcoca, line 371: patch seems to remove feature, not add it

jpmens, line 371: I thought so too until I looked at the linked issue which seems to add it there.

bcoca:
ah, this one does remove it, but the local version and the other request makes it global.

bcoca:
waiting for rebase in the other ticket, won't remove the feature till it is added there

woltage:
@jpmens @bcoca: My mistake look's like uri -module is not using module_utils/urls.py from Ansible core. I think it would be nice rewrite uri -module to use urls.py. (let's see if will have some time on my vacation)

woltage:
@bcoca: It would be great if you could merge this too.  You already merged ansible/ansible/pull/9281

PR #158, opened 2014-10-07
[git] Find the actual commit annotated tags refer to instead of the tag object.

ajsalminen:
There is a bug in the git module handling of git tags. When the git module version parameter is an annotated tag (all github tags are annotated), the git module incorrectly gets the tag object id instead of the commit it points to in get_remote_head(). This bug results in check mode always reporting a task as changed and unnecessary work in real runs.

The proposed change fixes the bug by always finding the dereferenced tag from ls-remote for annotated tags.

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -287,6 +287,7 @@ def reset(git_path, module, dest):
 def get_remote_head(git_path, module, dest, version, remote, bare):
     cloning = False
     cwd = None
+    tag = False
     if remote == module.params['repo']:
         cloning = True
     else:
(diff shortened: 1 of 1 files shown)


ajsalminen:
Sorry for the close-reopen, I thought I had made a mistake with the repo I was opening the pull request to now that the repos have been split. It seems I was mistaken and it is fine after all.

abadger:
Thanks!  It took a bit of googling to figure out what was going on here but it makes sense.

# Merged

Hi!

issue is resolved.

Thank you!

PR #163, opened 2014-10-08
Parse command output

RickyCook:
Allow parsing of stdout/stderr in the command module, so that things like `failed_when`, `changed_when`, and `register`/`when` conditionals are much easier, and more robust.

eg (I know that this would be easier with the exit code, but it's just for show; there are plenty of other, more complex uses):

```
- name: validate downloaded dmg
  command: hdiutil verify ~/Downloads/command_line_tools_for_os_x_10.9_for_xcode_6.dmg -plist
           parse_stdout=plist
  register: dmg_verify
  failed_when: "dmg_verify.parsed_stdout.calculated['Checksum Value'] != dmg_verify.parsed_stdout.expected['Checksum Value']"
```

Based on fixes in #162

### DIFF
--- a/commands/command.py
+++ b/commands/command.py
@@ -100,6 +100,7 @@ EXAMPLES = '''
     creates: /path/to/database
 '''
 
+# Dict of options and their defaults
 OPTIONS = {'chdir': None,
            'creates': None,
            'executable': None,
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hi!

Unfortunately, we're not sure if we want this feature in the program, and I don't want this to seem confrontational.  Our reasons for this are:
- we believe the idea of data format parsing should be left up to Jinja2 filters
- we are not sure data-format parsing should have specifics for the command module when other modules may want this feature

In the future, sometimes starting a discussion on the development list prior to implenting a feature can make getting things included a little easier, but it's not always neccessary.

PR #166, opened 2014-10-09
Allow users to specify the volume type for EC2 volumes (for SSD volumes)

suprememoocow:
Currently the `ec2_vol` will automatically choose between `io1` (PIOPS) and `standard` (magnetic) volumes, depending on whether the `iops` parameter has been set. EBS now supports a third volume type, `gp2` (SSD). 

This pull-request allows the volume type to be explicitly set for a new volume. It it's not specified, the default behaviour is backwards compatible with the current behaviour.


davestern:
:+1:

Jonnymcc:
Duplicate of #3 ?

bcoca:
closing this in favor of #3 which was just merged

PR #168, opened 2014-10-10
Added option 'ubuntu_base' to apt_repository.

jericksanjuan:
This option forces apt_repository to identify distros such as Linux Mint as an Ubuntu target to allow PPAs to be added.

### DIFF
--- a/packaging/apt_repository.py
+++ b/packaging/apt_repository.py
@@ -63,6 +63,13 @@ options:
         required: false
         default: 'yes'
         choices: ['yes', 'no']
+    ubuntu_base:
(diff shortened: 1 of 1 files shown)


mpdehaan:
Hmm,

I'm not sure most people would understand what this option does, with the description "Force identify host distro as Ubuntu-base to allow PPAs to work."

Might there be some other way of phrasing this?

In particular I'm thinking of whether this has bearing for Ubuntu users at all, etc.

jericksanjuan:
Okay please disregard this one. I have found a better approach and will make another pull request later. Thanks.

PR #171, opened 2014-10-10
Don't put spaces around equals sign in sysctl entries

merenbach:
Don't add spaces around the equals sign of sysctl entries when rewriting.

**Background:** FreeBSD doesn't like spaces around the equals sign in sysctl entries: `vfs.usermount=1` is valid syntax there, while `vfs.usermount = 1` is not.

**Expected behavior:** When adding one ore more sysctl entries in a playbook, no spaces exist around equals signs for any of those entries when the playbook is complete.  This works equally on Linux and FreeBSD.

**Current behavior:** When adding a sysctl entry, no spaces are placed around the equals sign of the added entry (good!).  `fix_lines()` rewrites other entries, however, to have the spaces around the equals sign, so if I add three entries in one playbook, two of them have spaces around the equals sign when the playbook is done (bad!).  This renders the behavioral outcome non-idempotent.

**Fix behavior:** Don't add spaces to either side of the equals sign in written/rewritten lines.

(Edited to clarify the "Current behavior" section above.)


gangelop:
This bug also exists against OpenBSD hosts. Tested @merenbach's patch and it works.

merenbach:
Looks as though this was closed by the subsequent #298, which addressed bug #287.  Closing.

PR #89, opened 2014-09-30
Updates to win_feature module

cchurch:
Fixes https://github.com/ansible/ansible/issues/8723 and https://github.com/ansible/ansible-modules-core/issues/88 by returning the content of the  `Microsoft.Windows.ServerManager.Commands.FeatureMessage` object(s):

```
"message": [{"error_code": 0, "message_type": "Info", "text": "[Installation] Succeeded: [Telnet Client] Telnet Client."}]
```

Replaces https://github.com/ansible/ansible/pull/9150; that PR uses `ToString()`, which only returns the string representation of the `message_type`.

Fixes https://github.com/ansible/ansible/issues/8739 by detecting if the `Install-WindowsFeature` cmdlet exists (Server 2012) or not (Server 2008).

Replaces https://github.com/ansible/ansible/pull/9046 by checking for the cmdlet instead of using OS version.

Partially addresses https://github.com/ansible/ansible/issues/8798 by failing with `"Not supported on this version of Windows"` on non-server Windows versions.

PR #180, opened 2014-10-13
Fix win_feature result message (fixes old#8723)

wozz:
Fix win_feature result message (fixes old#8723)

ported change from old repo


wozz:
nevermind it's been fixed here: https://github.com/ansible/ansible-modules-core/pull/89

PR #183, opened 2014-10-14
Added user, tenant and password to module parameters declaration

sb98052:
Without this change, user creation via Keystone fails with the errors "msg: unsupported parameter for module: user/tenant/password"


jimi-c:
Merged, thanks!

PR #187, opened 2014-10-14
ec2_asg: Initialize create_changed and replace_changed

jorgebastida:
Running `ec2_asg` with an already existing `asg` makes ansible crash if there is no value for `replace_all_instances` or `replace_instances`. (i.e wating for `viable_instances > X`)

```
- name: Wait for viable_instances
  local_action:
        module: ec2_asg
        name: asg-name
        region: eu-west-1
  register: asg
  until: asg.viable_instances|int >= 2
  delay: 10
  retries: 120
```

**Expected behavior:** Do not crash and return the status of the `asg`.

**Current behavior:** Crash with the following error `UnboundLocalError: local variable 'replace_changed' referenced before assignment`.

**Fix behavior:** Expected behavior.


jimi-c:
Merged, thanks!

PR #189, opened 2014-10-15
Openbsd rcctl

jarmani:
Work to improve the OpenBsdService service backend.

It uses rcctl on recent OpenBSD to enable / disable services. 
The behavior is not changed on older OpenBSD version.

Thanks to @eest who did a lot of the work and ajacoutot@openbsd.org who helped and improved  rcctl to fit your needs.

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -979,6 +979,9 @@ class OpenBsdService(Service):
 
         rc, stdout, stderr = self.execute_command("%s %s %s" % (self.enable_cmd, 'status', self.name))
 
+        if stderr:
+            self.module.fail_json(msg=stderr)
+
         if self.enable:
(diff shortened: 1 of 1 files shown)


eest:
For whatever my word is worth as the original author of the OpenBSD bits: I am OK with this going in.

gangelop:
Yes, I want this! :+1:

gangelop:
~~If I understand this correctly, this also fixes the status checking for the service module (#200), which will also work on OpenBSD 5.5 (which doesn't have rcctl yet). Is that correct?~~

I'm wrong. Nothing to see here...

eest:
@lathan: I might be missing something, but I am not sure what you think is broken. Please see my comment in #200.

nmeum:
Looks good and seems to work. :+1:

PR #204, opened 2014-10-20
synchzonize: allow use_ssh_args to take from ansible.cfg

FauxFaux:
I have some important things set in `ssh_args`, like `-F custom-ssh-config`, which I would like to apply during `rsync` as well as during `ssh`.  This patch adds an optional argument to `synchronize`, named `use_ssh_args`, which allows you to opt-in to this.

I've left the patch as off-by-default, to maintain compatibility, although I strongly believe that this shouldn't even be an option, and that it should just use `ssh_args` everywhere.

For anyone blocked by this issue: The patch is different for ansible 1.7.2 (stable release at time of writing), but the whole new module (as of this commit) Just Works when backported, so you can just use that.

### DIFF
--- a/files/synchronize.py
+++ b/files/synchronize.py
@@ -138,6 +138,10 @@ options:
       - put user@ for the remote paths. If you have a custom ssh config to define the remote user for a host
         that does not match the inventory user, you should set this parameter to "no".
     default: yes
+  use_ssh_args:
+    description:
+      - Honour ssh_args set in ansible.cfg.
+    default: no
   rsync_opts:
     description:
       - Specify additional rsync options by passing in an array.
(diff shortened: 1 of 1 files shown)


r3nat:
+1

kowalskey:
+1

tima:
I can see the value in this feature, but, without going in to my own extensive testing, I don't think this will work as coded. 

In line 314 you are importing ansible.constants to get the ssh_args. Thing is ansible.constants is probably not available on the remote host and the ansible.cfg is almost certainly not where this code is getting shipped and run. 

In order to get the ssh_args in to the module you need to add that logic to the associated synchronize action module that runs locally and sets up the synchronize module that you modified. The action module has access to ansible.cfg and ansible.constants. You'd fetch the ssh_args in the action module and transparently (to the playbook author) pass those to the module getting run on the remote.

BTW: This seems to be going at the same problem as #1073. I just reviewed that PR and it has issues to be resolved also.

tima:
With #1073 being merged in to HEAD I'm thinking this one can be closed now.

bcoca:
closing this in favor of #1073 and ansible/ansible#10612

PR #206, opened 2014-10-21
Add support for Google Cloud DNS

fcuny:
This CL add support for Google Cloud DNS. The module allows the creation, update and deletion of zones and records.

I'll submit another CL to add documentation to the GCE's guide in the ansible repository later.

This can't be merged yet since it's dependent on a couple of fix for libcloud.

cc @erjohnso


fcuny:
Oh well, I see that there's already a PR [1] for this by @erjohnso :)

[1] https://github.com/ansible/ansible-modules-core/pull/110

erjohnso:
@franckcuny - Thanks! I believe ansible would rather see all new modules go into 'extra' first and then after a few releases, get promoted to 'core'.  With that, I should whack #110 and re-submit.  I quickly glanced at your version and I like it.  But I tried to model mine after the route53 module so it would be easier for ansible users to swap back/forth as they prefer.

WDYT? Should we go with yours or mine?

fcuny:
@erjohnso let's go with yours then. I'm closing this one, and hopefully we can get this module merged somewhere soon :)

Thanks

PR #208, opened 2014-10-21
correct requirement of parameter dest

CptLausebaer:
The parameter "dest:" is required, but it is documented as "required: false".


jimi-c:
Merged, thanks!

PR #217, opened 2014-10-23
Added cidr_ip to ec_group rules_egress example in documentation.

jnsaff:
Fixed really annoying omission in the example.

Rules_egress needs cidr_ip, without it the task would complete without errors or warnings but delete all egress rules.


jimi-c:
Merged, thank you!

PR #219, opened 2014-10-23
Update syntax for cloudformation example

adamchainz:
The spare `>` freaked me out when I saw it (not a yaml line continuation since it was after a word), and the `action:` syntax is deprecated.


jimi-c:
Merged, thank you!

PR #224, opened 2014-10-25
add retry with exponential backoff when we receive throttling error code...

semifocused:
... from cloudformation

Occasionally when polling for completion of the creation or deletion of a CloudFormation stack, AWS will return an exception indicating that a rate limit has been exceeded and throttling is occurring.  It looks like this:

```
boto.exception.BotoServerError: BotoServerError: 400 Bad Request
<ErrorResponse xmlns="http://cloudformation.amazonaws.com/doc/2010-05-15/">
  <Error>
    <Type>Sender</Type>
    <Code>Throttling</Code>
    <Message>Rate exceeded</Message>
  </Error>
  <RequestId>2ab5db0d-5bca-11e4-9592-272cff50ba2d</RequestId>
</ErrorResponse>
```

This can be avoided by retrying the operation repeatedly with an exponential backoff up to a max number of retries.

This PR implements such a retry for all describe_stacks operations within the cloudformation module.

retry is only invoked when the error code received indicates throttling is occurring, otherwise the exception is re-raised to maintain prior functionality.

### DIFF
--- a/cloud/cloudformation.py
+++ b/cloud/cloudformation.py
@@ -159,7 +159,7 @@ def stack_operation(cfn, stack_name, operation):
     operation_complete = False
     while operation_complete == False:
         try:
-            stack = cfn.describe_stacks(stack_name)[0]
+            stack = invoke_with_throttling_retries(cfn.describe_stacks, stack_name)[0]
             existed.append('yes')
         except:
             if 'yes' in existed:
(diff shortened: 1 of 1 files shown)


semifocused:
Not sure what the merge conflict was, but I've created a new PR that applies the same changes atop the current devel branch. https://github.com/ansible/ansible-modules-core/pull/569.

I will close this PR in favor of that one.

PR #231, opened 2014-10-26
Adding running and stopped states to nova compute

tkinz27:
This commit adds the ability to power on and power off a VM
managed by openstack. This capability is very similar to the
ec2 module.


jimi-c:
cc @emonty

PR #232, opened 2014-10-27
Fixes service module for servers where systemctl is installed, but not h...

Jmainguy:
...andeling services.


Jmainguy:
Tested against Centos7 where systemd is handling services and against ubuntu1410 where it is not.

```
#Centos7
---

- name: Install mailman
  yum: name=mailman state=latest

- name: Restart mailman
  service: name=mailman state=restarted enabled=yes
```

```
#Ubuntu 1410
---

- name: Install systemd
  apt: name=systemd state=latest

- name: Install server-redis
  apt: name=redis-server state=latest

- name: Restart redis-server
  service: name=redis-server state=restarted enabled=no
```

ghost:
there's another related PR here https://github.com/ansible/ansible-modules-core/pull/57

claco:
Just for fun, I've hit a similar issue from the opposite direction. Currently running devel from earlier today against a Fedora 20 machine.

```
ansible 1.8 (devel e79acf3f88) last updated 2014/11/13 21:57:22 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 6317d3a988) last updated 2014/11/08 00:12:28 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 5a514ccdda) last updated 2014/11/08 00:12:38 (GMT -400)
  v2/ansible/modules/core: (detached HEAD cb69744bce) last updated 2014/11/08 00:12:49 (GMT -400)
  v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/11/08 00:12:59 (GMT -400)
```

I won't name the guilty package, but in my case, it is [mistakenly imho] installing both a .service file AND a script in init.d. The .service runs the binary directly, and the init.d script runs daemon on the binary. It seems, the service module finds, and uses the script for the service if it finds one in init.d and never tries to start the systemd version...on a systemd system. To make matters worse, if the init.d script returns a 0 exit status, ansible things it's started and takes no action it seems. If I manually delete the script the package is dropping into init.d, ansible does the right thing.

I can reproduce this with an other systemd service, like ntpd, by dropping in an executable script in /etc/init.d/ntpd that does an exit 0.

Now, I'm not defending the package install here at all. But, it seems this is a case for checking the system units for the service name (another PR I think). 

I would also advocate that with all of this inspection and guessing if systemd is in place, and valid, or init.d should be used, etc, etc.. it might just be time to add a "type" parameter to this module so the user can short circuit that inspecting, and just tell the module explicitly what type service  it is interacting with.

ghost:
@claco : is it possible that #57 solves that?

claco:
@jrobeson I'll give it a whirl in the am and see if that fixes my oddball situation.

claco:
@jrobeson Applied patch in the ./lib/ansible/modules/core. No effect. Service still using init.d script over systemd unit.

Jmainguy:
@claco what service is it btw, would love to test it my self

claco:
I haven't shamed them until I can get some context to why they're doing what they're doing. But, you can repro this by just make an executable in init.d for any systemd only service. I did that w/ ntpd.

abadger:
@claco I believe the behaviour you are experiencing is the way systemd is currently designed.  Would need to talk to the systemd folks about why they prefer the init file over the systemd unit file and to the maintainers of the package in question as to why they have two installed files trying to occupy the same name.

abadger:
@Jmainguy Looking at your patch, is it going to be portable to non-Linux systems?  if someone wants to run systemd on their BSD for instance?  Perhaps a compromise would be to fall back on the old behaviour if the /proc file is not detected.

claco:
How would I tell?

If I log into the machine, and issue `systemctl start ntpd`, it indeed starts the ntpd unit .service, and not the ntpd script in init.d. This is what led me to it being an ansible service  module issue.

I'll dig further to see what commands it is actually running.

## 

@abadger , so I dug into the code a bit, and dumped out `self.svc_cmd` in the `service_control` method.

With both files in place, the `svc_cmd` is `/usr/sbin/service`
When I delete the ntpd init.d script, leaving just the .service,  `svc_cmd` is `/usr/bin/systemctl`

Sure not sure where in the code the discovery around setting that goes wrong yet.

## 

I suspect line ~475: It's finding service, but prioritizes that over systemd:

```
        # Locate a tool for runtime service management (start, stop etc.)
        if location.get('service', None) and os.path.exists("/etc/init.d/%s" % self.name):
            # SysV init script
            self.svc_cmd = location['service']
        elif location.get('start', None) and os.path.exists("/etc/init/%s.conf" % self.name):

```

Final thought. There are other places, where it does a secondary `systemctl` type check:

```
        # Locate a tool for enable options
        if location.get('chkconfig', None) and os.path.exists("/etc/init.d/%s" % self.name):
            if check_systemd(self.name):
                # service is managed by systemd
                self.enable_cmd = location['systemctl']
            else:
                # we are using a standard SysV service
                self.enable_cmd = location['chkconfig']

```

Jmainguy:
@abadger Added failback as you recommended.

abadger:
@claco I see a way that ansible could be failing and not systemd :-(  If ansible is using /sbin/service, then it would first try the init script and then try the unit file.  If ansible is using /sbin/systemctl it will first try the unit file and then the init script.  I haven't looked at whether ansible is using /sbin/service yet or not but that's a likely path to the behaviour you're seeing.

Could you open a new bug report for this?   I'd like to make sure I can track both bugs independently.  Thanks!

system/service.py
```diff
@@ -395,9 +395,20 @@ def get_service_tools(self):
...
+            systemd_enabled = 'false'
+            if os.path.exists('/proc/1/cmdline'):
+                try:
+                    f = open('/proc/1/cmdline', 'r')
+                except IOError, err:
+                    self.module.fail_json(msg="%s" % (err))
+                for line in f:
+                    if 'systemd' in line:
+                        systemd_enabled = 'true'
+            elif not location.get('systemctl', None):
```
abadger: Need to reverse this conditional (because systemd_enabled is being set to false as the default above).

We should be able to use boolean values True and False instead of strings.

Is there a reason to only fallback to the old behaviour if the /proc/ file doesn't exist rather than, for instance, also if the file were unreadable?

Jmainguy:
https://github.com/bcoca/ansible-modules-core/tree/service_systemd_fix seems to knock it out of the park, will close this request once that one gets merged.

bcoca:
closing in favor of #343

PR #199, opened 2014-10-19
Add support for new docker restart policies

gottwald:
Docker deprecated the "-r" switch for the docker daemon.
Since docker 1.2 every container can now have it's own restart policy.

Module restart_policy option defaults are the same as the docker command defaults.
Needs docker-py >= 0.5.0.
Settings get ignored when used with a lower docker-py version. Tested it with docker-py 0.3.2.

PR #235, opened 2014-10-29
Add 'restart_policy' to docker module

konradkonrad:
Since version 1.2.0/ApiVersion 1.14 docker supports the flag
`restart_policy`. docker.py supports that flag since version 0.5.0.

This adds the new module parameter, to take advantage of the progress
on dockers side.

I'm not overly happy with the argument_spec for this, since it should be
a choice, but it also needs the numbered parameter for on-failure (see:
https://github.com/docker/docker-py/blob/master/docs/api.md#start ).

I chose to ignore values other than the defined ones ('always', 'on-failure').
The outcome of this is, that docker will assume the "default"
dict(Name='', MaximumRetryCount=0) in such instances.

### DIFF
--- a/cloud/docker.py
+++ b/cloud/docker.py
@@ -204,6 +204,13 @@ options:
     default: ''
     aliases: []
     version_added: "1.8"
+  restart_policy:
+    description:
+      - Set the restart policy for the containers (always, on-failure[:<retrycount>]). Requires docker >= 1.2.
+    required: false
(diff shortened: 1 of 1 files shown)


konradkonrad:
Ooops. I saw #199 only now. I leave this open, but feel free to close in favor of it :)

bcoca:
closing this in favor of dupe

PR #236, opened 2014-10-29
Alternative fix for #28

mvo5:
Based on the discussion in https://github.com/ansible/ansible-modules-core/pull/28 a alternative way of solving the issue. I personally don't mind either way but I think it would be great if you would pull in the unittest so that this kind of code is easier to change in the future.


abadger:
Thanks for the fix!  Merged it along with the additional changes discussed in #28 .

PR #244, opened 2014-10-30
Deprecate old OpenStack modules

emonty:


abadger:
Increasing priority slightly because I think we should decide whether to do this in time for the 1.9 release.

bcoca:
have we merged or do we have PR with new modules?

emonty:
@bcoca This is also now ready to land (I updated it to match the move of the modules into the openstack dir) I'd love to get this merged sooner rather than later to make it clear that these should really only be taking bugfixes

PR #251, opened 2014-10-31
Add working status detection for pf on FreeBSD

davidfritzsche:
The return code of "service pf onestatus" is usually zero on FreeBSD (tested with FreeBSD 10.0), even if pf is not running. So the service module always thinks that pf is running, even when it needs to be started.


bcoca:
I believe this is a bug with the pf init script, the service module has a 'pattern' parameter to match something in the process list to compensate for this.

Since we have the functionality above and IIRC pf does show up in the process table I'm going to close this ticket. We really don't want an exception in code per bad init script out there, the service module is large enough already.

davidfritzsche:
The pf kernel thread always shows up in the process table and can therefore not be used to detect if pf is active (service running) or not (service not running).

PR #254, opened 2014-10-31
use simpler ec2_connect ec2 util call

danielabel:
This allows boto profile use.

Coded with @tomwadley


juliedavila:
Looks good. Thanks. Recommend a merge.

PR #258, opened 2014-10-31
EC2 route tables module

ghost:
Early implementation of idempotent EC2 route tables module `ec2_route_tables` with backward-compatible option for those who are already using `ec2_vpc`.
- [ ] needs support for `absent` state
- [ ] needs tests
- [ ] needs documentation

Feedback would be much appreciated.  I made the module idempotent by checking each table and rebuilding the table _iff_ the configuration and the existing table differ (exhaustive comparison of all known/configurable properties).


ghost:
Closed in favor of #322.

PR #260, opened 2014-11-01
Added better region handling and enabled eu-central-1

willthames:
Make use of improved connect_to_aws that throws an exception
if a region can't be connected to (e.g. eu-central-1 requires
boto 2.34 onwards)

Add eu-central-1 to the two modules that hardcode their regions

This pull request makes use of the changes in ansible/ansible#9419

### DIFF
--- a/cloud/amazon/ec2_ami_search.py
+++ b/cloud/amazon/ec2_ami_search.py
@@ -56,7 +56,8 @@ options:
     required: false
     default: us-east-1
     choices: ["ap-northeast-1", "ap-southeast-1", "ap-southeast-2",
-              "eu-west-1", "sa-east-1", "us-east-1", "us-west-1", "us-west-2", "us-gov-west-1"]
+              "eu-central-1", "eu-west-1", "sa-east-1", "us-east-1",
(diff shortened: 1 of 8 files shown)


jjshoe:
@willthames - You've got merge conflicts :\

simao:
Any chance we get this soon? Thanks!

willthames:
@jjshoe - module conflicts resolved, thanks for the notification.

bcoca:
so 2 things, 1 i would separate the regions from the error handling, 2 I would change the modules that still use their own region list to use the common one in ansible.module_utils.ec2 and do any pertinent modifications there.

willthames:
@bcoca, I state quite clearly why the error handling is needed for the region changing - because if not, anyone using an older version of boto with the new region will have their code fail in a nasty way. 

I think the two remaining modules wouldn't benefit that much from the common code. Either way, I'd recommend separating _that_ change from this one.

willthames:
@bcoca, it's worth taking a look at the related change  ansible/ansible#9419 where I mention the remaining two modules.

Issue #149, opened 2014-10-06
Request for increased speed from git module when ensuring repo is on a hash


addshore:
##### ISSUE TYPE

Bug Report
##### COMPONENT NAME

git module
##### ANSIBLE VERSION

N/A
##### SUMMARY

Same as https://github.com/ansible/ansible/issues/8916

> When the git module is given a hash as the version of the repo to checkout the git module should first do a quick check to see if we are already on the given hash!
> Currently even if the repo is already on the version / hash we want it seems to pull from the remote and update.
> Instead we should make sure the repo exists (pull if it doesn't) and then check the current hash of HEAD and if it matches the hash version we want, do nothing, if it doesn't, pull said hash version.
> 
> This would drastically increase the speed of playbooks that depend on, say 100 git repos and specify a given version hash

josef-spak:
The problem here is that the git module calls 'git ls-remote' a LOT of times, often unnecessarily. I've been looking at the code for the past hour and it's quite ugly. I'll try to come up with a PR but for me personally it seems that going back to a shell script instead of the git module might work just as well right now (it's that bad).

caspark:
See also [ansible/#14501](https://github.com/ansible/ansible/issues/14501).

andreaso:
These superfluous git ls-remote calls (when version=SHA1SUM) aren't only a potential performance issue, they also have a negative impact in regards to robustness. Assuming that the git remote in question is unreachable the entire play will unnecessarily fail, even if it already had the desired explicit version locally.

Yes, one might be able to use update=no as a workaround, except then you have to toggle it back and forth whenever there is a new version you want to use.

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

jctanner:
As of today the git module performance for this test case ...

```
...
Tuesday 16 August 2016  12:07:48 -0400 (0:00:01.059)       0:00:34.490 ******** 
=============================================================================== 
clone ansible repository ----------------------------------------------- 27.10s
pull ansible repository using ansible's git module ---------------------- 5.88s
pull ansible repo using git directly ------------------------------------ 1.06s
ensure target is empty -------------------------------------------------- 0.38s

[jtanner@fedmac AMC-149]$ ansible --version
[...]

Issue #70, opened 2014-09-30
git: get_head_branch incorrect when branch name contains a slash


thpani:
[re-posting module issue from ansible/ansible#6672]
##### Issue Type:

Bug Report
##### Ansible Version:

1.5.3
##### Environment:

OS X 10.9.2
##### Summary:

`get_head_branch()` fails to correctly compute the HEAD branch if the branch name contains a slash. This is caused by splitting on '/' and indexing [-1] [here](https://github.com/ansible/ansible-modules-core/blob/c8e1a2077ed356fb893cd9886f758ac3d284e6b9/source_control/git.py#L398).
##### Steps To Reproduce:

Run this playbook:

```
- name: checkout llvm
  git: repo=https://github.com/thpani/llvm.git
       dest=/home/vagrant/llvm/
```

Checkout fails because branch is computed as
`release_33_nullsucc`
instead of
`patches/release_33_nullsucc`.
##### Expected Results:

`get_head_branch()` computes the correct branch name.

I.e. behave as if the above example were

```
- name: checkout llvm
  git: repo=https://github.com/thpani/llvm.git
       dest=/home/vagrant/llvm/
       version=patches/release_33_nullsucc
```
##### Actual Results:

```
TASK: [checkout llvm] *********************************************************
failed: [default] => {"failed": true}
msg: Failed to checkout branch release_33_nullsucc
```

robynbergeron:
Hi @thpani -- Thanks for submitting this bug report. Apologies for taking so long to triage it.

Going to close this issue -- It looks like #441 fixed this last year -- so hopefully that resolved the problem for you. If not, let us know!

pierrre:
It looks like #441 wasn't merged.
Is it really fixed?

PR #263, opened 2014-11-03
[git] couple of fixes/enhancements

tverlaan:
One of the issues I noticed is that the branch doesn't change if the sha is the same. I think a user expects the branch to change even if the sha is the same. I haven't created a bug report but instead created a PR :)

I also made a fix for #70 and I made an enhancement because of #149 (I'm not sure if it covers all cases, but it should be faster now when you enter a sha as version).


abadger:
Separate Pull requests will likely be easier to review.  That way I'll know which issues are being fixed by which code and be able to test each change individually.  Thanks!

tverlaan:
I'll create separate pull requests once I find the time.

PR #268, opened 2014-11-04
made subcategories for cloud modules for better organization

bcoca:
this takes advantage of the just updated ansible core support


mpdehaan:
Merged, thanks!

PR #269, opened 2014-11-04
use aws connect call in rds_subnet_group that allows boto profile use

tomwadley:
coded with @danielabel


juliedavila:
Good to go. Thanks. Merge recommended.

PR #272, opened 2014-11-04
docker: use env vars if present

### LINKED ISSUE #176: docker and docker_image should respect $DOCKER_HOST environment variable
In cases like remote docker (e.g. in virtual machine like boot2docker for osx) with variable respected by docker itself it's expected behavior that docker modules will use it 'by default' instead of socket.

lorin:
If the DOCKER_HOST environment variable is set, use environment variables
to configure the docker client.

This makes it much simpler to use this module with boot2docker on OS X. 

Fixes #176.

### DIFF
--- a/cloud/docker/_docker_image.py
+++ b/cloud/docker/_docker_image.py
@@ -29,6 +29,7 @@ version_added: "1.5"
 short_description: manage docker images
 description:
      - Create, check and remove docker images
+     - If DOCKER_HOST is set, will get connection info from environment variables
 options:
   path:
     description:
@@ -108,6 +109,7 @@ try:
     import re
     import json
     import docker.client
+    import docker.utils
     from requests.exceptions import *
     from urlparse import urlparse
 except ImportError, e:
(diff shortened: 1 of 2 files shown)


lorin:
Rebased against devel to resolve merge conflict

tcurdt:
cannot get boot2docker to work with ansible without this - please merge!

abadger:
Can we change this so that if you specify these in the playbook the playbook values would override what's in the environment or does that still break this use case?

lorin:
@abadger I like the idea of only using the environment variable if the user does not explicitly specify a `docker_url`, but I'm not sure how to actually implement this.

The docker_url parameter has a default value of `unix://var/run/docker.sock`. 

Consider the following two invocations: 

```
- docker: image=ubuntu
- docker: image=ubuntu docker_url=unix://var/run/docker.sock
```

We would want to use the `DOCKER_HOST` environment variable in the first case but not in the second. Is there a way to differentiate between the following two invocations of the docker module in the code?

deiga:
Can I be of some assistance with this issue? I can't get ansible to run my docker containers in the current state..

abadger:
@lorin -- sorry, I have too many docker tickets and this one got lost in my queue of tickets.  Yes, there is a way to do that.  We'd have to change the default value in the argument_spec to a sentinel value.  Then in the module code we'd need to set the if any of the values we'd have to merge the values from the environment variable, actually specified in the playbook, and defaults if they weren't in the environment variable and the variable from the playbook was the sentinel value.  None is probably the best sentinel value.

abadger:
@lorin, how does this match up with the environment handling from the recent TLS merge?  https://github.com/ansible/ansible-modules-core/pull/926

lorin:
#926 looks good, other than #946. I'm going to close this PR.

lorin:
Looks like the env var logic was originally added in #502.

PR #281, opened 2014-11-06
Fixed postgresql_db failing on Python 2.4 with --check

kustodian:
Fixed a minor bug where `postgresql_db` module fails with `--check`. This happens because on Python 2.4 (Centos 5) `module.exit_json` calls `sys.exit()` which throws an exception which is then caught by the try/except around it, so it reports an error.

This is just a minor change it the code, where I only put if/else instead of the exit_json() after the first if.

### DIFF
--- a/database/postgresql/postgresql_db.py
+++ b/database/postgresql/postgresql_db.py
@@ -281,15 +281,19 @@ def main():
             elif state == "present":
                 changed = not db_matches(cursor, db, owner, template, encoding,
                                          lc_collate, lc_ctype)
-        else:
-            if state == "absent":
-                changed = db_delete(cursor, db)
+            module.exit_json(changed=changed,db=db)
 
-            elif state == "present":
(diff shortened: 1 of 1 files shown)


abadger:
Thanks for finding this.  I had altogether forgotten that SystemExit was a subclass of Exception  in 2.4.  I think to help people remember that in the future we'd be better off changing the exception handling rather than removing the module.fail_json() calls (otherwise someone else will add a module.fail_json() later and break this again.)

So change the exception handler to be something like this:

``` python
except SystemExit:
    # Avoid catching this with Exception on python2.4 
    raise
except Exception, e:
    module.fail_json(msg="Database query failed: %s" % e)
[...]
```

kustodian:
@abadger I agree. I changed it as you suggested.

abadger:
Thank you for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #286, opened 2014-11-08
made hostname work with python 2.4

bcoca:
should fix #9413

### DIFF
--- a/system/hostname.py
+++ b/system/hostname.py
@@ -313,7 +313,7 @@ class OpenRCStrategy(GenericStrategy):
     def set_permanent_hostname(self, name):
         try:
             f = open(self.HOSTNAME_FILE, 'r')
-            lines = [x.strip() for x in f]
+            lines = (x.strip() for x in f)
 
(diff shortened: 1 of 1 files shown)


system/hostname.py
```diff
@@ -298,31 +298,35 @@ class OpenRCStrategy(GenericStrategy):
...
+            f.close()
+
+        return None
 
     def set_permanent_hostname(self, name):
         try:
-            with open(self.HOSTNAME_FILE, 'r') as f:
-                lines = [x.strip() for x in f]
+            f = open(self.HOSTNAME_FILE, 'r')
+            lines = [x.strip() for x in f]
```
abadger: Note: Python2.4 has iterators and generator expressions.  It would be good if we got into the habit of using iterators instead of lists and generator expressions instead of list comprehensions when the list is unnecessary overhead.

This would then look like:

```
lines = (x.strip() for x in f)
```

abadger:
Other than the note about generator expressions above, +1 to this pull request.

PR #288, opened 2014-11-10
set_fact.py: Make it obvious that facts don't survive between runs

techtonik:


bcoca:
Actually the original docs are correct, it survives across plays, not playbooks. 

A playbook is a collection of plays, in ansible itself this is 'the list of plays run in this execution'.  We also refer to the files that contain plays as playbooks. In both cases facts do not survive across playbooks. 

Facts survive across plays in the current playbook (or current ansible execution). If you setup the new facts cache facility, this enables them to survive across playbooks also.

techtonik:
Docs are confusing. Play is a verb that means action of ansible-playbook and an attempt to redefine it to something more academic doesn't work for all users. The problem with current documentation in this context is to understand if fact persists between launches of ansible-playbook executable.

bcoca:
> Play is a verb that means action of ansible-playbook
> I'm sorry, I don't see that anywhere.
> 
> The problem with current documentation in this context is to understand
> if fact persists between launches of ansible-playbook executable.

This I can understand, but will probably be obsoleted by fact caching.

## 

Brian Coca

techtonik:
Whatever. I've made an experiment for myself, and if I can't help other users to get that - it not my problem. I tried.

PR #296, opened 2014-11-10
CommandModule: detect change based on update to some file

quinot:
When generating a file by running an ad hoc command on the target
host, allow detection of change based to a change in the contents
of a named file specified as a parameter to the action
(updates=<filename>).

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -443,6 +443,9 @@ def install_deb(m, debs, cache, force, install_recommends, dpkg_options):
         if force:
             options += " --force-all"
 
+        for (k,v) in APT_ENV_VARS.iteritems():
+            os.environ[k] = v
+
         cmd = "dpkg %s -i %s" % (options, " ".join(pkgs_to_install))
         rc, out, err = m.run_command(cmd)
         if "stdout" in retvals:


commands/command.py
```diff
@@ -188,12 +202,21 @@ def main():
...
         args = shlex.split(args)
     startd = datetime.datetime.now()
 
     rc, out, err = module.run_command(args, executable=executable, use_unsafe_shell=shell)
 
+    changed = True
+    if updates and updates_md5 is not None:
+        if updates_md5 == check_md5(module, updates_file, 'after', warnings):
+            changed = False
+
```
clonedagain: How about a more direct writing? maybe :

```
    if updates and updates_md5 is not None:
        changed = (updates_md5 == check_md5(module, updates_file, 'after', warnings))
    else:
        # We didn't get a "updates" parameter, or we couldn't check the MD5 so by default we'll assume something changed
        changed = True
```

clonedagain:
Apart from the style note I just wrote on the diff, :+1: I'd like to have this feature in Ansible and the patch looks reasonable (though I've not tested it yet).

quinot:
clonedeagain: thanks for your comment, indeed your version is clearer (note: you need to change "==" to "!=", obviously).

quinot:
Hello, do you have any target date for integration of this proposed enhancement?

gregdek:
Hi @quinot. Thanks for the PR.

Because this module is a very central module, it's maintained by the Ansible Core team, and changes to these central modules can be conservative and very slow. Thanks for your patience with us here.

gregdek:
@quinot Thanks for your continued patience here; the core team is trying to get 2.0 released, so we'll push this out until after 2.0.

quinot:
@gregdek thanks for the update.

Issue #35, opened 2014-09-29
ec2_eip module creates new ip on repeated run


arnuschky:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.7.1
##### Environment:

Ubuntu 14.04.1
##### Summary:

If an EC2 instance already has an EIP associated to it, I would expect the ec2_eip module to do nothing/reuse the existing IP. Instead, it creates a new IP and associates this.

Note that I am not asking for the functionality provided by `reuse_existing_ip_allowed`.
##### Steps To Reproduce:
1. Associate an EIP to any instance using the ec2_eip module, for example `- ec2_eip: instance_id=i-1212f003`
2. repeat
##### Expected Results:

An instance without an EIP gets a new one allocated.
An instance with an EIP gets the same one allocated (idempotent),
##### Actual Results:

An instance with an EIP gets the another new EIP allocated (no idempotent, every new run creates and allocated a new EIP)

Here's the playbook that creates an instance and associates an EIP.
(vpc and security group exist already)

```
- name: EIP module bug
  hosts: 127.0.0.1
  connection: local
  gather_facts: no
  tasks:
    - include_vars: "aws/eip-bug/vpc_info"
    - name: Creating instance
      local_action:
        module: ec2
        key_name: "keyname"
        region: us-east-1
        group_id:
          - "{{ sg_default.group_id }}"
        instance_type: t2.micro
        image: "ami-0619bf6e"
        wait: yes
        vpc_subnet_id: "{{ subnet_test_id }}"
        assign_public_ip: yes
        exact_count: 1
        count_tag:
          Environment: "eip-bug"
          Class: eip-bug
        instance_tags:
          Environment: "eip-bug"
          Class: eip-bug
          Name: eip-bug
      register: instance
    - name: Allocating elastic IP to instance
      ec2_eip:
        in_vpc: yes
        reuse_existing_ip_allowed: yes
        state: present
        region: us-east-1
        instance_id: "{{ instance.tagged_instances[0].id }}"
      register: instance_eip
    - debug: var=instance_eip.public_ip
```

The instance task is idempotent thanks to exact_count/count_tag. The EIP task creates and associates a new EIP on every invocation. Accordingly, the number of EIPs increases; after 4 invocations the task fails due to Amazon's restriction to 3 unallocated EIPs.

EIP module output with -vvv:

```
TASK: [Allocating elastic IP to instance] ************************************* 
[...]

PR #300, opened 2014-11-11
Fix EC2_EIP: don't try to associate an already-associated instance.

herbyg-axial:
Also stops it from allocating a new Elastic IP on every run.

### DIFF
--- a/cloud/amazon/ec2_eip.py
+++ b/cloud/amazon/ec2_eip.py
@@ -207,7 +207,7 @@ def allocate_address(ec2, domain, module, reuse_existing_ip_allowed):
         domain_filter = { 'domain' : 'standard' }
       all_addresses = ec2.get_all_addresses(filters=domain_filter)
 
-      unassociated_addresses = filter(lambda a: a.instance_id == "", all_addresses)
+      unassociated_addresses = filter(lambda a: not a.instance_id, all_addresses)
       if unassociated_addresses:
         address = unassociated_addresses[0];
       else:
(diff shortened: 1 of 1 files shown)


herbyg-axial:
This should resolve **ansible/ansible-modules-core#35**.

herbyg-axial:
Additional changes to add a "state='allocated'" option, as well as correct the behavior when looking for unused elastic IPs.

Old behavior of allocating a new elastic IP when specified as just:
action: eip
..has been preserved.

pguillebert:
This one bites me as well.

juliedavila:
Could you elaborate what you mean by "don't try to associate an already-associated instance"

pguillebert:
@nousdefions : The current behaviour of the plugin always allocate a new EIP ("creates" an IP) even if the EC2 instance is already happily associated with an EIP already. In other words, it breaks idempotence we expect from ansible.

lamielle:
Any chance we can get this PR finished up and merged?  I'm running into the problem that this fixes.

daniilyar:
Tested by including updated module as 'library' to my Ansible codebase. Works well, +1 to merge this one

moorereason:
This bug is really annoying, but this PR could be tidied up a bit before merging.
- New `state` string should be an adjective, such as "allocated."
- Why define `allocate_eip()` [inside of main](https://github.com/axialops/ansible-modules-core/blob/c5d94ccac71273cf63d44c4f30698dc0e9b8c3f5/cloud/amazon/ec2_eip.py#L284)?  That function should probably be moved.

Looking forward to this PR being merged!!

juliedavila:
@herbyg-axial 

I agree with @moorereason, the allocate_eip() should be moved out of main. Are you still working on this?

theonlysinjin:
Has this not merged yet? I used this copy, it doesn't create new addresses on run, but it's still not associating with already existing, but free addresses

herbyg-axial:
@moorereason @defionscode @sinjin I'll get this tidied up between today and tomorrow.  Thank you for taking a look :)

juliedavila:
@herbyg-axial are you still working on this?

jml:
I don't know if this is the done thing, but I really would like this fixed, so #1015 is a PR that's a clone of this one but with the recommended changes made.

juliedavila:
@jml that looks good, just need the merge commit fixed.

bendavies:
how are people working around this bug?

moorereason:
@bendavies I've disabled the `ec2_eip` tasks in my playbooks until I know I have an instance that needs a new EIP.  Not much of a workaround.

klj613:
any update on this?

bennojoy:
@klj613  seems like #1015 fixes it , can you give the latest devel a try

bennojoy:
closing, since fixed via #1015

PR #310, opened 2014-11-12
Compare DNS names case insensitively.

feanil:
##### Summary:

If I try to override the value of a DNS name that is in lowercase with one that is uppercase, the Ansible module things they are no the same DNS name.
##### Steps To Reproduce:

Create a dns name in route53 that has mixed case. Then try to change its value with 'overwrite' set to true.
##### Expected Results:

DNS name gets updiated with the new value.
##### Actual Results:

18:12:42 msg: Tried to create resource record set [name='******', type='CNAME'] but it already exists


phillipholmes:
I've not been able to reproduce your 'steps to reproduce' step - if I try creating an upper or mixed case dns entry, Route53 automatically converts it to lower case. I put in a pull request a while ago to have the route53 module treat all dns entries as lower case no matter what the Ansible script has them as due to a similar error. referenced pull request:  https://github.com/ansible/ansible-modules-core/pull/30

bcoca:
Closing, as this seems redundant at this point after previous merges already guarantee the data is lower case before this point.

PR #311, opened 2014-11-12
Apt state fix

bcoca:
now present state completely equates to installed and i removed installed from all error messages.


packaging/os/apt.py
```diff
@@ -519,8 +525,8 @@ def main():
                     p['default_release'], dpkg_options)
 
         if p['deb']:
-            if p['state'] != "installed":
-                module.fail_json(msg="deb only supports state=installed")
+            if p['state'] == 'present':
```
abadger: The condition here changed from != to ==.  I think it's supposed to stay !=.

PR #315, opened 2014-11-13
Add new flavor_name parameter

johnpbatty:
Allow the instance flavor to be specified by name rather than id.

Specifying objects by name is much more human-friendly and portable
between systems than specifying by id.

The nova_compute module already allows images to be specified by name
or id, so this is just providing equivalent function for flavors.

OpenStack does not (unfortunately) enforce that names are unique.
However, any users who have the same name for different flavors are
asking for trouble!  If there are multiple flavors with the same name,
the first one returned by nova.flavors.list() is used.


johnpbatty:
I screwed up and submitted this from the wrong git branch.  I have resubmitted the change as #387 on a fresh branch.

PR #316, opened 2014-11-13
Docker: Handle images with multiple tags

dkerwin:
## Scenario: Tags returned from docker and ansible may not always match

Single docker image has multiple tags associated (latest, prod, 456). Playbook looks like this:

```
tasks:
    - name: stop old containers
      local_action:
        module: docker
        docker_url: tcp://{{ facter_ipaddress_eth0 }}:2375
        image: hub.example.com/base
        state: stopped
    - name: run container
      local_action:
        module: docker
        docker_url: tcp://{{ facter_ipaddress_eth0 }}:2375
        image: hub.example.com/base:prod
        command: sleep 300
        count: 3
        state: running
```

Before the patch you end up with 3 new containers and the old containers would still be running. Docker will return the tag 456 (or one of the other tags) and the match in docker.py will fail. The patch will get all tags associated with the given image and return the correct list of containers.

Cheers :beers: 
Daniel

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -356,7 +356,6 @@ HAS_DOCKER_PY = True
 
 import sys
 from urlparse import urlparse
-from collections import defaultdict
 try:
     import docker.client
     import docker.utils
@@ -683,11 +682,9 @@ class DockerManager(object):
             name = '/' + name
         deployed = []
 
-        all_valid_tags = []
         for img in self.client.images():
             if image in img['RepoTags']:
-                for x in img['RepoTags']:
-                    all_valid_tags.append(x.split(':', 1)[1])
+                all_valid_tags = set(x.rsplit(':', 1)[1] for x in img['RepoTags'])
 
         # if we weren't given a tag with the image, we need to only compare on the image name, as that
         # docker will give us back the full image name including a tag in the container list if one exists.
@@ -703,7 +700,7 @@ class DockerManager(object):
             image_matches = (running_image == image)
 
             # images can and most probably have multiple tags associated
-            tag_matches = (not tag or running_tag == tag or running_tag in all_valid_tags)
+            tag_matches = (not tag or running_tag in all_valid_tags)
 
             # if a container has an entrypoint, `command` will actually equal
             # '{} {}'.format(entrypoint, command)
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -356,6 +356,7 @@ HAS_DOCKER_PY = True
 
 import sys
 from urlparse import urlparse
+from collections import defaultdict
 try:
     import docker.client
     import docker.utils
@@ -686,6 +687,13 @@ class DockerManager(object):
         # docker will give us back the full image name including a tag in the container list if one exists.
         image, tag = get_split_image_tag(image)
 
+        # Get list of all images and create dict image_tag_map[image] = [tag1, tag2]
+        image_tag_map = defaultdict(list)
+        for i in self.client.images():
+            for t in i['RepoTags']:
+                a, b = t.split(':', 1)
+                image_tag_map[a].append(b)
+
         for i in self.client.containers(all=True):
             running_image, running_tag = get_split_image_tag(i['Image'])
             running_command = i['Command'].strip()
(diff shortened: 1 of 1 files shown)


dkerwin:
Sorry for the confusion. The previous commit would return all running instances without reference to the supplied tag. Commit 091e9d4 reduces the runtime and returns only the matching containers.

dkerwin:
Hey @bcoca,

is there anything i can do to get this PR merged before the 1.8 release?

Thanks!
Daniel

bcoca:
@dkerwin we are trying to get through all the P1 and P2 tickets first, we'll merge this as soon as we can after that.

cloud/docker/docker.py
```diff
@@ -329,6 +329,7 @@
 
 import sys
 from urlparse import urlparse
+from collections import defaultdict
```
abadger: modules have to be compatible with python2.4 so unfortunately there's no defaultdict in py2.4.  Fortunately it doesn't look like defaultdict is being used in your current patch?

cloud/docker/docker.py
```diff
@@ -562,6 +563,12 @@ def get_deployed_containers(self):
             name = '/' + name
         deployed = []
 
+        all_valid_tags = []
+        for img in self.client.images():
+            if image in img['RepoTags']:
+                for x in img['RepoTags']:
+                    all_valid_tags.append(x.split(':', 1)[1])
```
abadger: Are the repotags returned from img['RepoTags'] simple enough that splitting on the first ':' doesn't run into any cornercases?  I think that "image" contains the registry information, do the strings we get back from img['RepoTags'] contain that as well?

cloud/docker/docker.py
```diff
@@ -574,7 +581,10 @@ def get_deployed_containers(self):
             if i["Names"]:
                 name_matches = (name and name in i['Names'])
             image_matches = (running_image == image)
-            tag_matches = (not tag or running_tag == tag)
+
+            # images can and most probably have multiple tags associated
+            tag_matches = (not tag or running_tag == tag or running_tag in all_valid_tags)
+
```
abadger, line 706: Is tag one guaranteed to be one of the tags in all_valid_tags?  If so we can  omit the check for running_tag == tag.

cloud/docker/docker.py
```diff
@@ -562,6 +563,12 @@ def get_deployed_containers(self):
             name = '/' + name
         deployed = []
 
+        all_valid_tags = []
```
abadger: Might be slightly better to use a set instead of a list here since we don't care about order and we're doing a "img in all_valid_tags" check later.

abadger:
The concept here looks good.  I made some comments on the implementation that should be addressed.

cloud/docker/docker.py
dkerwin: You are right. It's no longer required and will be removed

cloud/docker/docker.py
dkerwin: Valid point. Will replace split with rsplit

dkerwin:
Hey @abadger. Is there still something to be done or is it ready to be merged?

abadger:
Just me needing to get a chance to go back through my buffer of tasks.  /me checks the changes.

cloud/docker/docker.py
```diff
@@ -682,6 +682,10 @@ def get_deployed_containers(self):
             name = '/' + name
         deployed = []
 
+        for img in self.client.images():
+            if image in img['RepoTags']:
+                all_valid_tags = set(x.rsplit(':', 1)[1] for x in img['RepoTags'])
+
```
abadger, line 690: I noticed that previously, this could potentially save tags from multiple img's.  Was that correct behaviour?  I kinda think it might have been incorrect.  If so, we should break out of the loop once we've set all_valid_tags.

cloud/docker/docker.py
```diff
@@ -682,6 +682,10 @@ def get_deployed_containers(self):
             name = '/' + name
         deployed = []
```
abadger, line 685: Need to set all_valid_tags to an empty set()/frozenset() here just in case we don't detect any images that match.

abadger:
Okay, just a pair of comments around one area of the code.

cloud/docker/docker.py
dkerwin, line 685: Valid point. Had it already fixed locally

cloud/docker/docker.py
dkerwin, line 690: You are right. f115707 should fix that

abadger:
Sorry :-(  I got pulled away from docker onto other issues for a while and then forgot about this one.  I think that when I merged #502 and some associated fixes (like https://github.com/ansible/ansible-modules-core/pull/912 ) that this was fixed.

Can you confirm that alternate tags don't cause issues with devel or the stable-1.9 branch?

abadger:
Closing as this should have been implemented when we merged #502 

# Closing This Ticket

Hi!

issue is resolved.

Thank you!

PR #318, opened 2014-11-13
Allow empty EC2 Security Group rules.

rwe:
Previously, there was no way to specify "no rules" on a security group short of specifying a phony rule. (e.g. you would need to add something like "only allow access to localhost").

This was because the module would check whether any rules were defined, and if not, add an "anything goes" default rule. This was counter to intuition about what `purge_rules` etc. did and causes unintentionally permissive security groups.

This PR allows you to continue using that default behavior, if desired, by omitting either `rules` or `rules_egress`; however you may now specify that there are no rules at all by supplying an empty list.

Personally, I believe the "anything goes" default is a bad one, but I did not address that in this change. This PR should be backwards compatible except in the case that someone explicitly specified an empty list with the expectation that Ansible would create the `0.0.0.0/0` rule in its place.

Now you are able to set up a rule like:

``` yaml
- name: Set up security group for package clients
  local_action:
    module: ec2_group
    name: 'package-client'
    description: Can retrieve packages from package servers
    vpc_id: '{{vpc_res.vpc.id}}'
    region: '{{vpc_res.vpc.region}}'
    rules: []
    rules_egress:
      - proto: tcp
        from_port: 80
        to_port: 80
        group_name: 'package-server'
        group_desc: Build server and apt repository security group
      - proto: tcp
        from_port: 443
        to_port: 443
        group_name: 'package-server'
```

Which will create a security group with _only_ the rules specified.

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -24,11 +24,15 @@ options:
     required: false
   rules:
     description:
-      - List of firewall inbound rules to enforce in this group (see example).
+      - List of firewall inbound rules to enforce in this group (see'''
+''' example). If none are supplied, a default all-out rule is assumed.'''
+''' If an empty list is supplied, no inbound rules will be enabled.
     required: false
   rules_egress:
     description:
-      - List of firewall outbound rules to enforce in this group (see example).
+      - List of firewall outbound rules to enforce in this group (see'''
+''' example). If none are supplied, a default all-out rule is assumed.'''
+''' If an empty list is supplied, no outbound rules will be enabled.
     required: false
     version_added: "1.6"
   region:
@@ -268,7 +272,7 @@ def main():
         addRulesToLookup(group.rules, 'in', groupRules)
 
         # Now, go through all provided rules and ensure they are there.
-        if rules:
+        if rules is not None:
             for rule in rules:
                 group_id, ip, target_group_created = get_target_from_rule(module, ec2, rule, name, group, groups, vpc_id)
                 if target_group_created:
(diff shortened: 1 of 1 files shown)


kalefranz:
Hi @erydo, I have a problem right now that this pull request could easily address (but right now doesn't).  Around line 347 exists the following comment:

```
# when using a vpc, but no egress rules are specified,
# we add in a default allow all out rule, which was the
# default behavior before egress rules were added
```

In some use cases I would like _no_ egress rules, and the module just doesn't allow that right now, even with `purge_egress_rules` set to `true`.

I propose modifying line 351 of your pull request from

```
if default_egress_rule not in groupRules:
```

to

```
if default_egress_rule not in groupRules and not purge_rules_egress:
```

I guess another option to maintain backward compatibility but achieve the same result would be to create an additional argument for the module called something like `include_default_egress_rule` that defaults to `true` but can be flipped to `false`. However, that seems to me like a different PR, and furthermore when first using the `purge_egress_rules` feature, my assumption was that if I included no egress rules, I would end up with _no rules_, not egress that's wide open.

rwe:
Hi @kalefranz, I'm not sure I understand your request. To specify that there should be _no_ egress rules, you set `rules_egress: []` in this pull request: that will remove all (including the default) egress rules.

Is that different from what you are looking for?

kalefranz:
Hey @erydo, sorry for the confusion.  I just tested your PR, and I can confirm it works as expected.

Could I get you to add an additional documentation example that shows the usage?  Maybe something like

```
- name: respond to ping, traceroute, and other ICMP requests
  local_action:
    module: ec2_group
    name: IcmpResponseSG
    description: respond to ICMP ping, traceroute, etc
    purge_rules: yes
    purge_rules_egress: yes
    rules:
      - proto: icmp
        from_port: -1
        to_port: -1
        cidr_ip: 0.0.0.0/0
    rules_egress: []
```

That example would have saved me a _ton_ of time actually.

juliedavila:
@erydo can you add the example from @kalefranz? Everything else looks good.

mspiegle:
I am using this and it appears to work as advertised for my use case.  Most of my instances should NOT have open outbound access.  The ones that do get it via a special group.  IMO, it's dangerous to use a permissive default where security is concerned.  Thanks for your work!

PR #320, opened 2014-11-13
Improved AMI search using EC2 image directory.

rwe:
This improves the `ec2_ami_search` module to search the entire EC2 AMI repository, allowing for things like running Amazon-provided AMIs (NATs, etc.), other third party AMIs (WordPress, etc.) as well as official Ubuntu AMIs as before. There are some additional filtering options that were not present before.

The previous module interface _only_ supported Ubuntu images by scraping an Ubuntu-specific datasource.

The new module interface is similar to the earlier one, but **not entirely compatible** (e.g. there is no standard way to search for `distro` or `stream` in the EC2 AMI directory). Examples are given for both Ubuntu and Amazon AMIs.

### DIFF
--- a/cloud/amazon/_ec2_ami_search.py
+++ b/cloud/amazon/_ec2_ami_search.py
@@ -20,51 +20,78 @@
 DOCUMENTATION = '''
 ---
 module: ec2_ami_search
-short_description: Retrieve AWS AMI information for a given operating system.
+short_description: Searches AWS for matching AMI.
 deprecated: "in favor of the ec2_ami_find module"
 version_added: "1.6"
 description:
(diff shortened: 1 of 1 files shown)


juliedavila:
@erydo plz fix the merge conflicts/rebase when you get a sec

rwe:
@defionscode - This is rebased, though it may have been partially(?) addressed by the new `ec2_ami_find` module in the interim.

bcoca:
@erydo yes, the new module should take over now and ec2_search is deprecated, which is why I'm closing this ticket. 

Feel free to resubmit the improvements to the new module, just be mindful about incompatible changes to the interface, which is the reason we went with a new module, to avoid breaking existing playbooks.

PR #324, opened 2014-11-14
Validate Security Group rule parameter names for ec2_group module.

rwe:
Previously, if you mistyped rule parameter names (e.g. `cdir_ip` instead of `cidr_ip`), rather than failing the module would silently assume those parameters were omitted and do unexpected things like skipping rules.

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -127,7 +127,7 @@ def make_rule_key(prefix, rule, group_id, cidr_ip):
     """Creates a unique key for an individual group rule"""
     if isinstance(rule, dict):
         proto, from_port, to_port = [rule.get(x, None) for x in ('proto', 'from_port', 'to_port')]
-        #fix for 11177 
+        #fix for 11177
         if proto not in ['icmp', 'tcp', 'udp'] and from_port == -1 and to_port == -1:
             from_port = 'none'
             to_port   = 'none'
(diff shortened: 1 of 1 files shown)


rwe:
@bcoca Perhaps this project has different definitions, but I'd suggest this be labeled as a bugfix rather than feature. I believe the previous validation was mistakenly incomplete, not intentionally so; and its behaviour is both unexpected, silent, and dangerous.

juliedavila:
Hi @erydo 

So this is definitely a good thing you caught, much appreciated. The module fails with improper rules params now, however there seems to be an issue...

``` shell
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427478015.88-158714273873022/ec2_group", line 2199, in <module>
    main()
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427478015.88-158714273873022/ec2_group", line 281, in main
    validate_rule(module, rule)
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427478015.88-158714273873022/ec2_group", line 130, in validate_rule
    module.fail_json(msg='Invalid rule parameter \'{0}\''.format(k))
NameError: global name 'k' is not defined


```

rwe:
Thanks @defionscode, that was a bug in the most recent commit ("Cleaner validation check"). That commit has been rolled back and I'm rebasing against current devel now.

rwe:
Rebased.

PR #325, opened 2014-11-14
Minor style cleanups to ec2_eip module.

rwe:
No functional changes; only whitespace and some other minor style cleanups.


bcoca:
we don't take reformat requests as they break attribution for very little gained benefit

PR #326, opened 2014-11-14
Refactor and improve the EC2 Elastic IP module.

rwe:
This includes the style cleanups from #325. I separated them out into separate requests as this is a larger change.

This refactor substantially reduces the number of long boto calls. Additionally, it will more intelligently match existing Elastic IP associations when an `instance_id` is provided and will skip unnecessary changes. (So providing an `instance_id` causes this module to be idempotent).

It also (I believe) obsoletes the need for `wait_timeout` functionality, as the `address` object does not need to be searched for repeatedly.

### DIFF
--- a/cloud/amazon/ec2_eip.py
+++ b/cloud/amazon/ec2_eip.py
@@ -228,7 +228,7 @@ def ensure_present(ec2, domain, address, instance_id,
     return {'changed': changed, 'public_ip': address.public_ip}
 
 
-def ensure_absent(ec2, domain, address, check_mode):
+def ensure_absent(ec2, domain, address, instance_id, check_mode):
     if not address:
         return {'changed': False}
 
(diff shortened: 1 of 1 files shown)


juliedavila:
So this looks good except when specifying `state=absent`

Running this playbook:

``` yaml
---
- name: Test AWS Stuffs
  hosts: localhost
  connection: local
  tasks:

    - name: allocate a new elastic IP and associate it with an instance
      ec2_eip: instance_id=i-9cf6a86c region=us-east-1 state=absent
```

Gives the following error:

``` shell

ok: [localhost]

TASK: [allocate a new elastic IP and associate it with an instance] ***********
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427486745.87-46492369806187/ec2_eip", line 2092, in <module>
    main()
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427486745.87-46492369806187/ec2_eip", line 280, in main
    result = ensure_absent(ec2, domain, address, module.check_mode)
  File "/Users/jdavila/.ansible/tmp/ansible-tmp-1427486745.87-46492369806187/ec2_eip", line 234, in ensure_absent
    if instance_id:
NameError: global name 'instance_id' is not defined


```

rwe:
@defionscode - That error has been fixed and the branch rebased.

PR #334, opened 2014-11-15
Make cloudformation module's template argument not required in case of stack deletion

adamchainz:
The AWS API doesn't even accept a template when deleting a stack, it makes the module confusing to use.


juliedavila:
@adamchainz This is very good finding. Can you please rebase/fix merge conflicts when you have time?

adamchainz:
Done!

PR #338, opened 2014-11-16
apt_repository: fix file mode

pilou-:
'set_mode_if_different' method should be called on destination filename (not temporary filename) and after
'atomic_move' method.

This one allows me to find ansible/ansible#9565.

### DIFF
--- a/packaging/os/apt_repository.py
+++ b/packaging/os/apt_repository.py
@@ -238,10 +238,6 @@ class SourcesList(object):
                 d, fn = os.path.split(filename)
                 fd, tmp_path = tempfile.mkstemp(prefix=".%s-" % fn, dir=d)
 
-                # allow the user to override the default mode
-                this_mode = module.params['mode']
-                module.set_mode_if_different(tmp_path, this_mode, False)
-
                 f = os.fdopen(fd, 'w')
                 for n, valid, enabled, source, comment in sources:
                     chunks = []
@@ -259,6 +255,10 @@ class SourcesList(object):
                     except IOError, err:
                         module.fail_json(msg="Failed to write to file %s: %s" % (tmp_path, unicode(err)))
                 module.atomic_move(tmp_path, filename)
+
+                # allow the user to override the default mode
+                this_mode = module.params['mode']
+                module.set_mode_if_different(filename, this_mode, False)
             else:
                 del self.files[filename]
                 if os.path.exists(filename):


abadger:
The documentation says that the mode applies to "- The octal mode for newly created files in sources.list.d".  it looks like the current code would do that?  (ie: if a file already exists, its mode would be used.  If it's a wholly new file, then the mode parameter is used to set that).

I'm not sure if that was an intentional design choice or if the documentation merely reflects the way the code works, though.  (Bug discussion: https://github.com/ansible/ansible/issues/6908  commit: https://github.com/psa/ansible/commit/7d7a838b71b2665d65735549101cefbbde9f2b07 )  

If the current code doesn't do what's documented I'll be happy to look into this more.  If you'd like to change the way it works and is documented, ask on the mailing list about the change -- if there's no objections, update this ticket and I'll merge this.

pilou-:
The `mode` parameter is not used when a new file is created in `/etc/apt/sources.list.d`.

Tested with the following task:

<pre>
- name: "Add backports repository"
  apt_repository: repo='deb http://ftp.debian.org/debian/ wheezy-backports main' state=present mode=0640
</pre>

If the file `/etc/apt/sources.list.d/ftp_debian_org_debian.list` doesn't exist, it is created with `0644` mode.

gregdek:
Hi @abadger -- what's the path forward with this PR?

abadger:
Sorry @pilou-  this had dropped off of my radar.  Took a bit to look through the code today and saw that atomic_rename takes its permissions from the destination file (if it already exists) that is being replaced rather than from the file at the source... different than what /bin/mv does.  So your change makes sense for new files.  it's not correct for existing files, though.  I'm looking into how I can add to your code to fix that.

abadger:
https://github.com/ansible/ansible-modules-core/commit/50912c9092eb567c5dc61c47eecd2ccc585ae364

Okay, that should take care of it (not modifying the mode of existing sources files) for existing files as well.

I think the parameters are a little strange for this module but I'm not sure if there's a good reason for that.  I'd expect that mode would apply to whichever sources file gets the new (or subtracted) repository rather than to a sources file that's newly created.  Reading the code I can see how it's not as trivial to add this here but I'm not sure if that's a case of the code reflecting what we actually want to happen or what has been documented as expected behaviour reflecting how the code was implemented.  In any case, if someone wants to change that we can discuss and entertain a new pull request that does so.

# Merged

Hi!

issue is resolved.

Thank you!

PR #341, opened 2014-11-17
digital_ocean: Add note about using API v1

axilleas:
This was not referenced in docs, which caused confusion since v2 is now the standard.

Let me know if you want to make any changes rephrasing it.


GR360RY:
+1 for the change. I spend some time figuring out why digital ocean module did not work properly. This should be included in the docs.

PR #342, opened 2014-11-17
fixes user module for rhel5 by using -n instead of -N

Jmainguy:
fixes user module for rhel5 by using -n instead of -N, all other distros will still use -N

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -299,7 +299,15 @@ class User(object):
             # exists with the same name as the user to prevent 
             # errors from useradd trying to create a group when 
             # USERGROUPS_ENAB is set in /etc/login.defs.
-            cmd.append('-N')
+            if os.path.exists('/etc/redhat-release'):
+                dist = platform.dist()
+                major_release = int(dist[1].split('.')[0])
+                if major_release <= 5:
+                    cmd.append('-n')
+                else:
+                    cmd.append('-N')
+            else:
+                cmd.append('-N')
 
         if self.groups is not None and len(self.groups):
             groups = self.get_groups_set()


abadger:
I think that this would solve the error but it does introduce different behaviour.  -N means that the user will not be added to any group.  -g self.name means that the user will be added to the group with that name.  This could be surprising if that group already existed for another purpose.

Jmainguy:
On RedHat5

```
   -g, --gid GROUP
      The group name or number of the user’s initial login group. The group name must exist. A group number must refer to an already existing group.
      /etc/default/useradd.

   -n A group having the same name as the user being added to the system will be created by default. This option will turn off this Red Hat Linux specific behavior.
      When this option is used, users by default will be placed in whatever group is specified in /etc/default/useradd. If no default group is defined, group 1 will be
      used.
```

-n is out of the question in my opinion, do not want to add this user to gid 1. 
that pretty much just leaves -g, adding the user to a group that matches their name by default.

I imagine when people are using ansible, and they do not specify a group, they expect the user to be added to their own group. Where as the current code (If you do not merge my addition) will not add a user to their own group, if that group already exists.

Do you believe the current behavior of not adding a user to their own group, if it exists already and if group= is not defined is the correct behavior?

bcoca:
what about "-n -g self.name", would that qualify as a default?

## 

Brian Coca

abadger:
After some IRC discussion we realized that rhel5's -n is the same as more modern shadow-utils -N.  (IIRC, a RedHat patch to add -n was implemented upstream and changed to a capital N at the same time).

abadger:
Thanks @jmainguy!

# Merged

Hi!

issue is resolved.

Thank you!

madkiss:
This is also broken on SLES 11.

PR #343, opened 2014-11-17
service now detects systemd is actually running, not just installed

### LINKED ISSUE #161: service module broken by installation of systemd on ubuntu
##### Summary:

If you install the `systemd` package on Ubuntu (Utopic / 14.10), the service module stops working:
##### Steps To Reproduce:

```
$ ansible test-foo -m service -a 'name=redis-server state=stopped enabled=no'

test-foo | FAILED >> {
    "failed": true, 
    "msg": "failure 1 running systemctl show for 'redis-server.service': Failed to get D-Bus connection: No connection to service manager.\n"
}
```
##### Expected Results:

It believes that, just because `systemctl` is on the path, and there's a file named `/lib/systemd/system/redis-server.service`, that the `redis-server` package is managed by `systemd`.  However, the system is using `upstart` for service management, and I believe Ubuntu will be doing so for a long time.

Ironically, `systemd` is installed because `upstart` recommends it:

```
$ dpkg -S $(which systemctl)
systemd: /bin/systemctl
$ aptitude why systemd
i   libdbus-1-3    Recommends dbus                    
i A dbus           Recommends libpam-systemd          
i A libpam-systemd Depends    systemd (= 208-8ubuntu7)
$ aptitude why libdbus-1-3
i   upstart-bin Depends libdbus-1-3 (>= 1.2.16)
$ apt-cache show upstart | fgrep -m 1 Prior
Priority: required
```

Perhaps checking `systemctl` actually works could work?

```
$ systemctl 2>/dev/null || echo "not working"
not working
```

Maybe check what `init` is?

```
% grep -o 'init=[^ ]*' /proc/cmdline || readlink -f /sbin/init
/sbin/upstart
```

I don't have any systems with working `systemd` so I can't say what else might work.

bcoca:
should fix #161

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -394,8 +394,24 @@ class LinuxService(Service):
             location[binary] = self.module.get_bin_path(binary)
 
         def check_systemd(name):
-            # verify service is managed by systemd
-            if not location.get('systemctl', None):
+            # verify systemd is installed (by finding systemctl)
+            if not location.get('systemctl', False):
+                return False
+
+            systemd_enabled = False
+            # Check if init is the systemd command, using comm as cmdline could be symlink
+            try:
+                f = open('/proc/1/comm', 'r')
+            except IOError, err:
+                # If comm doesn't exist, old kernel, no systemd
+                return False
+
+            for line in f:
+                if 'systemd' in line:
+                    systemd_enabled = True
+                    break
+
+            if not systemd_enabled:
                 return False
 
             # default to .service if the unit type is not specified


bcoca:
should supersede #57 and #232

system/service.py
```diff
@@ -394,8 +394,24 @@ def get_service_tools(self):
...
+                f = open('/proc/1/comm', 'r')
+            except IOError, err:
+                # If comm doesn't exist, old kernel, no systemd
+                return False
+
+            for line in f:
+                if 'systemd' in line:
+                    systemd_enabled = True
+                    break
+
```
ciupicri, line 413: File `f` should be closed here, after the `for` statement.

PR #346, opened 2014-11-18
service cleanup

### LINKED ISSUE #345: service module fails to start service when both init.d and systemd scripts exist
This is originally reported in
- https://github.com/ansible/ansible-modules-core/pull/232#issuecomment-63016019
- https://github.com/ansible/ansible-modules-core/pull/232#issuecomment-63119092

In situations where there is both a systemd .service unit file and a init.d startup script, the service module will detect and run the init.d script instead of calling systemd. In this case, I'm on Fedora 20, but I suspect this is true on RHEL/Centos7 as well.

A quick synopsis of the comments above:

With both in place, `self.svc_cmd` returns `/usr/bin/service`
When I delete the script in init.d, `self.svc_cmd` returns `/usr/bin/systemctl`

I suspect line ~475: It's finding service, but prioritizes that over systemd:

```
        # Locate a tool for runtime service management (start, stop etc.)
        if location.get('service', None) and os.path.exists("/etc/init.d/%s" % self.name):
            # SysV init script
            self.svc_cmd = location['service']
        elif location.get('start', None) and os.path.exists("/etc/init/%s.conf" % self.name):

```

Final thought. There are other places, where it does a secondary `systemctl` type check:

```
        # Locate a tool for enable options
        if location.get('chkconfig', None) and os.path.exists("/etc/init.d/%s" % self.name):
            if check_systemd(self.name):
                # service is managed by systemd
                self.enable_cmd = location['systemctl']
            else:
                # we are using a standard SysV service
                self.enable_cmd = location['chkconfig']

```

bcoca:
now handles systemd better, uses it in every case when it is the init system in use (fixes #345)

clarified the if/elif mess for service discovery (at least a bit), removed a lot of redundancy.

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -393,7 +393,12 @@ class LinuxService(Service):
         for binary in binaries:
             location[binary] = self.module.get_bin_path(binary)
 
-        def check_systemd(name):
+        for initdir in initpaths:
+            initscript = "%s/%s" % (initdir,self.name)
+            if os.path.isfile(initscript):
+                self.svc_initscript = initscript
+
+        def check_systemd(name, initscript):
             # verify systemd is installed (by finding systemctl)
             if not location.get('systemctl', False):
                 return False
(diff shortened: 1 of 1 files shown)


abadger:
I've taken a look at the code and nothing looks obviously wrong to me.  I like the cleanups in general but am a little nervous about applying this so close to release.  Code that's full of if-elif-else blocks is somewhat dangerous to update as it's very sensitive to the order that the checks are performed.  Cleaning that up is both good in the long term and fraught with peril in the short term ;-)

Reserved +1 to merge from me.

PR #361, opened 2014-11-19
Add option docker_user for docker module.

### LINKED ISSUE #352: docker: parameter `user` is ignored
##### Issue Type:

Bug Report
##### Ansible Version:

```
% ansible --version
ansible 1.7.2
```
##### Environment:

Debian unstable

```
% cat /etc/debian_version
jessie/sid
% uname -v
#1 SMP Debian 3.16.7-2 (2014-11-06)
% docker --version
Docker version 1.3.1, build 4e9bbfa
% pip show docker-py

---
Name: docker-py
Version: 0.5.3
Location: /usr/lib/python2.7/dist-packages
Requires: requests, six, websocket-client
% python --version
Python 2.7.8
```
##### Summary:

The `docker` module does not handle the `user` parameter which
would correspond to `--user` option of `docker run` or `docker
create` correctly.  It isn't applied but no error is thrown.
The parameter `user` is simply ignored.

The parameter should be either supported (which is what i would
like to see) or it should raise an error.
##### Steps To Reproduce:

Launch a simple container and look at the `failed` and `User`
fields using the following command:

```
ansible -m docker -a 'image=busybox command=true user=nobody' localhost | sed '1c{' | jq -c '[.failed, .ansible_facts.docker_containers[0].Config.User]'
```
##### Expected Results:

There are two possible outcomes:
1. The `user` parameter is not supported.  The appropriate error
   message for that would be "unsupported parameter for module:
   user" and the output of the command above should therefore be
   the following:
   
   [true,null]
2. The `user` parameter supported.  Then the process in the
   container should be launched as the user `nobody` and therefore
   the output of the command above should be the following:
   
   [false,"nobody"]
##### Actual Results:

However neither is the case.  The unknown paramter `user` is simply
ingnored and the process of the container is simply launched as
`root`.  The output of the command above is actually the following

```
[false,""]
```

simono:
docker_user can be used to specify the user or UID to use within the
container.

Fixes #352

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -160,6 +160,13 @@ options:
         specified by docker-py.
     default: docker-py default remote API version
     version_added: "1.8"
+  docker_user:
+    description:
+      - Username or UID to use within the container
+    required: false
+    default:
+    aliases: []
+    version_added: "2.0"
   username:
     description:
       - Remote API username.
@@ -1303,6 +1310,7 @@ class DockerManager(object):
                   'stdin_open':   self.module.params.get('stdin_open'),
                   'tty':          self.module.params.get('tty'),
                   'host_config':  self.create_host_config(),
+                  'user':         self.module.params.get('docker_user'),
                   }
 
         def do_create(count, params):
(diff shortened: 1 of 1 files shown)


gregdek:
Hi @simono. Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

simono:
@gregdek Thanks for the heads-up. I rebased, so this should be fine to merge now.

gregdek:
Thanks @simono. 

@cove @joshuaconner @softzilla @smashwilson please review; if any of you give +1 we will tag "P1_reviewed" for potential inclusion.

joshuaconner:
+1, thanks for your patience!

simono:
Thanks

PR #366, opened 2014-11-20
cloudformation: allow template_url parameter

Rob-Johnson:
Cloudformation offers a template_url parameter in the CreateStack API, allowing one to specify a URL to a Cloudfomation template in an S3 bucket to be used, rather than passing the contents of the Cloudformation stack via the template_body parameter.

This PR implements this functionality.

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -50,7 +50,7 @@ options:
     aliases: []
   template:
     description:
-      - The path of the cloudformation template, required if state is "present".
+      - The local path of the cloudformation template. This parameter is mutually exclusive with 'template_url'. Either one of them is required if "state" parameter is "present"
     required: false
     default: null
     aliases: []
@@ -76,6 +76,11 @@ options:
     default: null
     aliases: ['aws_region', 'ec2_region']
     version_added: "1.5"
+  template_url:
+    description:
+      - Location of file containing the template body. The URL must point to a template (max size: 307,200 bytes) located in an S3 bucket in the same region as the stack. This parameter is mutually exclusive with 'template'. Either one of them is required if "state" parameter is "present"
+    required: false
+    version_added: "2.0"
 
 author: James S. Martin
 extends_documentation_fragment: aws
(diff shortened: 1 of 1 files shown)


Rob-Johnson:
rebased against devel. @jimi-c is there anything I can do to get this merged?

geowa4:
:+1: as I'm blocked by this.

alexanderdean:
+1 would be great to get this in!

juliedavila:
@Rob-Johnson, sorry for the delay, this looks great, but just one thing. Can you please add a check to see if both `template` AND `template_url` are specified and subsequently error out? Alternatively, you could add in a `mutually_exclusive` param to the argument spec if you prefer.

Rob-Johnson:
Hi @defionscode,

The current behaviour follows that of [boto](https://boto.readthedocs.org/en/latest/ref/cloudformation.html#module-boto.cloudformation.stack#create_stack):

> Conditional: You must pass TemplateURL or TemplateBody. If both are
> passed, only TemplateBody is used.

Though I agree that your suggestion is probably what _should_ happen, I think it would be surprising if the module deviated from the behaviour of boto.

What do you think?

juliedavila:
@Rob-Johnson, even if boto supports/does it, I don't think that should translate to the module necessarily needing to support it. There can't possibly be a use-case where someone would intentionally seek out to supply both parameters for the purposes of using `TemplateBody` only. 

A more likely scenario would be someone who is refactoring their role/playbook and forgets to remove one or the other which could then lead to unintended results.

Rob-Johnson:
Hi @defionscode,

I've rebased against devel and updated. I've had to make a couple of changes to account for upstream commits to the module, specifically https://github.com/ansible/ansible-modules-core/commit/d27df0cdb4868ddea8956ab03ce398edc534017a, but otherwise I've just added the `mutually_exclusive` option.

cloud/amazon/cloudformation.py
```diff
@@ -76,6 +76,11 @@
     default: null
     aliases: ['aws_region', 'ec2_region']
     version_added: "1.5"
+  template_url:
+    description:
+      - Location of file containing the template body. The URL must point to a template (max size: 307,200 bytes) located in an S3 bucket in the same region as the stack. This parameter is mutually exclusive with 'template'. Either one of them is required if "state" parameter is "present"
+    required: false
+    version_added: "1.8"
```
juliedavila: Could you change this to 2.0?

Rob-Johnson: Done.

juliedavila:
Good to go. Recommending a merge.

Rob-Johnson:
Thanks @defionscode @bcoca

PR #367, opened 2014-11-20
Add module supporting AWS VPC Peering Connections

herbyg-axial:
This module makes it possible to name, create, accept, and remove peering connections between VPCs in AWS.
The module permits VPCs to be referenced by their VPC ID, CIDR block or name tag.

### DIFF
--- a/cloud/amazon/ec2_vpc_peering.py
+++ b/cloud/amazon/ec2_vpc_peering.py
@@ -1,0 +1,593 @@
+#!/usr/bin/env python
+"""
+Ansible module to manage VPC peering connections
+"""
+
+DOCUMENTATION = '''
+---
+module: ec2_vpc_peering
+short_description: manage AWS VPC Peering Connections
+description:
+    - Create, list or reject/delete AWS VPC peering connections.
+    - This module has a dependency on python-boto
+version_added: "1.7.2"
+author: Herby Gillot <herby.gillot@gmail.com>
+notes:
+    - You can specify VPCs via their VPC ID, CIDR block or tagged name.  If you use a CIDR or name and it matches more than one VPC, the module will abort, at which point you'd need to use something more specific (like the VPC's ID).
+    - This module uses the account ID of the current AWS API user to determine whether or not it can or cannot accept a peering connection that is pending acceptance.  The 'force' flag can be used skip this check, and will attempt acceptance of every pending-acceptance connection seen, no matter what the accepting account ID is.
+options:
+    name:
+        description:
+            - Name to tag this connection with
+        required: false
+        default: null
+    source_vpc:
+        description:
+            - VPC ID, CIDR block or name tag of the requesting VPC
+            - Required for all states except "list"
+        required: false
+        default: null
+    peer_vpc:
+        description:
+            - VPC ID, CIDR block or name tag of the accepting VPC
+            - Required for all states except "list"
+        required: false
+        default: null
+        aliases: ['dest_vpc']
+    peer_owner_id:
+        description:
+            - Account ID of the owner of the accepting VPC, if different.
+        required: false
+        default: null
+    state:
+        description:
+            - Sets desired state for this VPC peering connection. If set to 'pending', if the connection does not exist, it will be created in the 'pending-acceptance' state. If set to 'present' or 'active', and both the source and dest VPC are owned by you, this module will create and set the connection to be active.
+        required: false
+        default: 'present'
+        choices: ['absent', 'present', 'active', 'pending', 'list']
+    force_accept:
+        description:
+            - Normally, this module only accepts pending connections that it owns.
+            - Enabling this option attempts to accept all connections seens that are pending acceptance.
+        required: false
+        default: false
+    ignore_rejection:
+        description:
+            - By default, this module does not attempt to re-create a connection that has been rejected.
+            - Enabling this option will create connections even if they have been rejected.
+        required: false
+        default: false
+    always_delete:
+        description:
+            - By default, when removing connections, connections in a pending-acceptance state will be rejected.
+            - Enabling this flag will always delete a connection whether it is pending acceptance or not.
+        required: false
+        default: false
+
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Create a VPC peering connection between ovpc1 and ovpc2
+- local_action: 
+    module: ec2_vpc_peering
+    name: 'ovpc1-to-ovpc2'
+    source_vpc: 'ovpc1'
+    dest_vpc: 'ovpc2'
+    region: 'us-west-1'
+    state: 'present'
+    
+# List all VPC peering connections
+- local_action:
+    module: ec2_vpc_peering
+    region: 'us-east-1'
+    state:  'list'
+  register: peering_list
+
+- debug: var=peering_list
+
+# Create a VPC peering connection from your VPC with CIDR 10.0.0.0/16 
+# to vpc ID vpc-01234567, owned by the account ID 012345678901
+- local_action: 
+    module: ec2_vpc_peering
+    source_vpc: '10.0.0.0/16'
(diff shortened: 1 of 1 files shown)


herbyg-axial:
(Missing 'cloud' label)

jjshoe:
Either the documentation is wrong, or the code is wrong. To stay consistent I vote code. It looks as if you do indeed require state to be specified: https://github.com/axialops/ansible-modules-core/blob/0fb319f17ec01a21fc39a040216ea2900331c090/cloud/amazon/ec2_vpc_peering.py#L543

@bcoca - input on what's best practice here?

jjshoe:
Also, specifying 'peer_vpc' instead of 'dest_vpc' like the docs say should work, doesn't.

bcoca:
plenty of examples of modules going both ways, but it should be consistent, either modify the code to default to present and not be required or modify the docs to reflect that it is required

herbyg-axial:
@jjshoe Thanks for looking at this.  I've pushed a change to fix the `state` parameter, but I'm not able to reproduce any problems with the `peer_vpc` parameter. Can you reconfirm? And if you're still having problems, can you provide at least which version of Ansible you're using? Thanks in advance.

denmat:
This module has worked as shown in examples. When can this be merged? Currently using axialops's branch in their repo and it's working fine.

juliedavila:
This looks good. Recommending it be merged.

tendant:
Is there any plan to merge this module?

juliedavila:
@tendant new module merges are being held off until v2

tendant:
@defionscode Thanks for the quick response. Is there any timeline for the V2 to be available?

juliedavila:
@tendant looks like I was going off old information. Previously, there was an internal discussion about the potential movement of certain common code; it won't be moved (with the exception of lookup/connection/callback plugins). This should be merged relatively soon. Sorry about the confusion.

tendant:
Cool! Thanks for the information. I will watch close on this issue. 

Please let me know if there is anything I can do to help move this forward.

bennojoy:
@herbyg-axial  @tendant  if the commits could be sqashed to a single commit , i think this shoudl be good to merge

tendant:
@herbyg-axial, I can help squash the commits, if you are fine with it.

juliedavila:
@tendant, it's been a cpl weeks, go ahead and do it, keep the commit credit to @herbyg-axial.

gregdek:
Adding "needs_revision" just to track the needed commit squash. Once that's been done, I believe that @defionscode has given enough approval for us to make this "P1_reviewed".

tendant:
I have squashed commits and pushed to my github fork. The sha1 for the commit is: a12d09483907d15de5d15604dc437f64b8738046

Do I need to create a separate pull request?

bcoca:
just 2 commits, really no need to squash, its new module so we won't be applying it as a fix or to other branches.

What we do need is the license header added

cloud/amazon/ec2_vpc_peering.py
```diff
@@ -0,0 +1,596 @@
...
+"""
+
+DOCUMENTATION = '''
+---
+module: ec2_vpc_peering
+short_description: manage AWS VPC Peering Connections
+description:
+    - Create, list or reject/delete AWS VPC peering connections.
+    - This module has a dependency on python-boto
+version_added: "1.7.2"
```
bcoca, line 13: update to 2.0

cloud/amazon/ec2_vpc_peering.py
```diff
@@ -0,0 +1,596 @@
...
+# and 10.10.2.0/24 are not present.
+- local_action: 
+  module: ec2_vpc_peering
+  source_vpc: '10.10.1.0/24'
+  dest_vpc: '10.10.2.0/24'
+  region: 'us-west-1'
+  state: 'absent'
+'''
+
+from boto.exception import EC2ResponseError
```
bcoca, line 109: look at other amazon modules for example, boto should be imported in a exception handler, set a var to return a 'nice' message to user that the requirement is missing if not found.

cloud/amazon/ec2_vpc_peering.py
```diff
@@ -0,0 +1,596 @@
...
+    if desired_state in ('list',):
+        action = lambda: vpcpmod.list_all()
+    elif desired_state in ('present', 'active'):
+        action = lambda: vpcpmod.ensure_present(want_active=True)
+    elif desired_state in ('pending',):
+        action = lambda: vpcpmod.ensure_present()
+    elif desired_state in ('absent',):
+        action = lambda: vpcpmod.ensure_absent()
+
+    try:
```
bcoca, line 580: we shy away from global exception handlers, the underlying boto messages are not normally very helpful, we would prefer returning a messages explaining what we were doing when we failed, then attach the boto message

cloud/amazon/ec2_vpc_peering.py
```diff
@@ -0,0 +1,596 @@
...
+
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(dict(
+            region = dict(required=True),
+            name = dict(required=False),
+            source_vpc = dict(required=False),
+            peer_vpc = dict(required=False, aliases=['dest_vpc']),
+            peer_owner_id = dict(required=False),
+            state = dict(
+                choices=['absent', 'present', 'active', 'pending', 'list'],
```
bcoca, line 546: we would prefer to shy away from 'list' states and move this to a fact module if possible , or if it makes sense just always return the data with the other states.

bcoca:
@herbyg-axial sorry for the late review, if you catch me in the city remind me I owe you a beer.

juliedavila:
@tendant are you still planning to open a PR? If so, do address the comments made by @bcoca

cloud/amazon/ec2_vpc_peering.py
tendant, line 580: Just want to make sure I understand the requirement correctly, you were talking about the try-catch on the action() call (line 581). 

I agree with you in general it is better to explain in detail when exception happens. 

In this case, we don't really know which action is taken, unless we try-catch each action separately. Also we don't really have better knowledge than boto about why we fail, since it is just simple call to boto function.

cloud/amazon/ec2_vpc_peering.py
tendant, line 546: I am willing to make the change. Do you have any example for fact module? I not familiar with ansible module development. Tried to find some examples in source code and devel maillist, but I didn't find anything. Thanks.

bcoca, line 546: ec2_facts might be a good example

cloud/amazon/ec2_vpc_peering.py
bcoca, line 580: there are many steps and exceptions that can happen in some of those function calls, by detailing specifically what was being attempted you might give the user a better clue to fix his data (if that is what caused the error).

tendant, line 580: I have pushed code with my fix here: https://github.com/ansible/ansible-modules-core/compare/devel...tendant:ec2_vpc_peering

You can find the branch here:

https://github.com/tendant/ansible-modules-core/tree/ec2_vpc_peering

The "fact module" is not changed. Since I am not sure about what is the best way to handle it. 

Let me know whether the code is good or not. I can create pull request if it is good to go.

tendant:
Please review the code https://github.com/ansible/ansible-modules-core/compare/devel...tendant:ec2_vpc_peering

If everything is OK, I can create PR.

gregdek:
This has been quite the football, so I'm closing it so it doesn't show up in my reports any more. :)

But a special note to @tendant -- go ahead and create the PR. We can't really review it until it's in PR form anyway. And thank you!

tendant:
Thanks @gregdek. 

An new pull request has been created at #1985

PR #371, opened 2014-11-21
git: add 'refspec' argument

rohanpm:
This argument may be used to fetch additional refs beyond the default refs/heads/\* and refs/tags/*.
For example, useful to check out a GitHub pull request or Gerrit patch set.

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -80,6 +80,17 @@ options:
         default: "origin"
         description:
             - Name of the remote.
+    refspec:
+        required: false
+        default: null
+        version_added: "1.9"
+        description:
+            - Add an additional refspec to be fetched.
+              If version is set to a I(SHA-1) not reachable from any branch
+              or tag, this option may be necessary to specify the ref containing
+              the I(SHA-1).
+              Uses the same syntax as the 'git fetch' command.
+              An example value could be "refs/meta/config".
     force:
         required: false
         default: "yes"
@@ -166,6 +177,9 @@ EXAMPLES = '''
 
 # Example just ensuring the repo checkout exists
 - git: repo=git://foosball.example.org/path/to/repo.git dest=/srv/checkout clone=no update=no
+
+# Example checkout a github repo and use refspec to fetch all pull requests
+- git: repo=https://github.com/ansible/ansible-examples.git dest=/src/ansible-examples refspec=+refs/pull/*:refs/heads/*
 '''
 
 import re
@@ -279,7 +293,7 @@ def get_submodule_versions(git_path, module, dest, version='HEAD'):
     return submodules
 
 def clone(git_path, module, repo, dest, remote, depth, version, bare,
-          reference):
+          reference, refspec):
     ''' makes a new git repo if it does not already exist '''
     dest_dirname = os.path.dirname(dest)
     try:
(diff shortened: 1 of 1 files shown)


rohanpm:
I considered making the option 'refspecs' and take an array rather than 'refspec'.  I thought refspec is better since it's a bit easier to use a string than an array and since, if you intend to check out a certain SHA1, it's redundant to fetch that commit from more than one ref.
I also have an update to integration tests for this argument.

source_control/git.py
```diff
@@ -434,38 +450,36 @@ def get_head_branch(git_path, module, dest, remote, bare=False):
...
-        (rc, out2, err2) = module.run_command("%s fetch --tags %s" % (git_path, remote), cwd=dest)
-    if rc != 0:
-        module.fail_json(msg="Failed to download remote objects and refs")
-    out_acc.append(out2)
-    err_acc.append(err2)
+        commands.append([fetch_str, [git_path, 'fetch', '--tags', remote]])
+
+    if refspec:
+        commands.append([fetch_str, [git_path, 'fetch', remote, refspec]])
+
```
abadger: Showing my ignorance of git's more advanced features here, is there a reason all of these fetch options are separate?  Your restructuring of the code makes it more obvious how similar they are.  I wonder if the code can combine them into fewer commands (merging the refspecs together for instance, or adding the --tags parameter to one of the fetch operations).

source_control/git.py
```diff
@@ -80,6 +80,17 @@
         default: "origin"
         description:
             - Name of the remote.
+    refspec:
+        required: false
+        default: null
+        version_added: "1.8"
```
abadger: Ssorry it took so long to review this PR.  This will either be 1.8.3 or 1.9 -- Not sure whether we'll have a 1.8.3 yet.

source_control/git.py
```diff
@@ -159,6 +170,9 @@
 
 # Example just ensuring the repo checkout exists
 - git: repo=git://foosball.example.org/path/to/repo.git dest=/srv/checkout update=no
+
+# Example checking out a github repo along with additional refs
```
abadger: If I'm reading this correctly, this example will fetch github pull requests?  Might be useful to mention that in the comment.

abadger:
Sorry it took a while to review this.  I like your restructuring of the fetch function.  I think this could be a useful feature.  Added some inline questions to answer and then I think this could be merged.

source_control/git.py
rohanpm: I think you're right.  The updated version reduces the number of commands.  However it still ends up doing two fetches if you specify a custom refspec in non-bare mode, since AFAIK there's no way to append a refspec to git's default behavior in that scenario.

source_control/git.py
rohanpm: Changed to 1.9.

source_control/git.py
rohanpm: Adjusted it to explain the example better.

rohanpm:
https://github.com/ansible/ansible/pull/9860 adds some integration tests.

abadger:
Thanks for the refspec parameter and the refactoring!  Merged.

# Merged

Hi!

issue is resolved.

Thank you!

abadger:
@rohanpm I had to rework a few bits of your changes so that it would work with git-1.7.x (still in use on RHEL6 and Ubuntu12.04)  It passes the integration tests you added but I figured I should mention it here so you could check that I didn't break something that isn't being tested:

https://github.com/ansible/ansible-modules-core/commit/8f6ae92cf88beda287c6c11d8b4127239c3168e0

(Also fixed a general problem earlier in the code but you probably have a fix for that that you just forgot to commt: https://github.com/ansible/ansible-modules-core/commit/8a03af66083da993c47a970cde44ab8fc39744b6 )

PR #382, opened 2014-11-24
adding Amazon RDS snapshot features

sassman:
this basically supports only 2 use cases:
- make sure a given snapshot is absent (delete)
- make sure a given snapshot is present based on another snapshot (copy)

### DIFF
--- a/cloud/amazon/rds_snapshot.py
+++ b/cloud/amazon/rds_snapshot.py
@@ -17,7 +17,7 @@
 DOCUMENTATION = '''
 ---
 module: rds_snapshot
-version_added: "1.5"
+version_added: "1.7.3"
 short_description: manage RDS database snapshots
 description:
      - copy, delete, and describes. This module has a dependency on python-boto >= 2.5.
--- a/cloud/amazon/rds_snapshot.py
+++ b/cloud/amazon/rds_snapshot.py
@@ -20,11 +20,11 @@ module: rds_snapshot
 version_added: "1.7.3"
 short_description: manage RDS database snapshots
 description:
-     - copy, delete, and describes. This module has a dependency on python-boto >= 2.5.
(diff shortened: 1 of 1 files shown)


bcoca:
not sure if we want the new module as the rds one already supports some snapshot functionality

sassman:
do you suggest to add the features (copy and delete snapshot) to rds module?

bcoca:
no, sorry, the more I look at it the more sense it makes to deprecate the snapshotting features in rds

sassman:
I totally agree. 
IMO an own command as an entry point for all the snapshoting features makes the most sense. 
Indeed it is somehow handy to work with a specific rds instance and just make a snapshot of it. But working with a own snapshot command would even feel more naturally after all.

Do you need any further input/help/contribution/refactoring to get this merged?

juliedavila:
@sassman Sorry for the delay on getting back to you on this pull request. A cpl things:
1. Could you abstract away the function-related bits away from main into their own functions. Maybe `copy_snapshot` and `delete_snapshot` functions?
2. Would it be possible for you to add in the functionality to create a snapshot given a db instance name?
   -  Bonus points if you can add in a `max_snapshot_age` parameter which will skip the task should the snapshot be younger than the max age.
   - Also, if you add this, then perhaps add absent, present and _copy_ as state options.

Thoughts?

sassman:
@defionscode no problem, thanks for the feedback.
1. absolutely!
2. that is a very nice idea, makes it also a bit more smart, as ansible used to be

will try a change on those and ping back

gregdek:
Hi @sassman. Just checking in to see how these changes are coming.

gregdek:
@sassman A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

gregdek:
Closing due to inactivity.

PR #384, opened 2014-11-24
Win copy/file/template modules (for ansible-modules-core)

jhawkesworth:
Note these modules also need action plugins for which a separate PR has been created here: https://github.com/ansible/ansible/pull/9611
Both would be required.

### DIFF
--- a/windows/win_copy.ps1
+++ b/windows/win_copy.ps1
@@ -1,0 +1,84 @@
+#!powershell
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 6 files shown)


cchurch:
:clap: Sweet! I'll review/test and get back to you.

jhawkesworth:
Thanks Chris, I'll put some notes on the other PR.

jhawkesworth:
@cchurch Just to say I realised I'd missed switching the actual modules to use SHA1 checksums, so I've pushed a change for that now.

rasmuswlarsen:
I'm not sure if this is the correct place to ask this. But are these windows module being added to Ansible?

jhawkesworth:
Questions generally best asked on the mailing lists, but since you are here.

There are two parts to this PR - some modifications were also needed to the
main Ansible project, specifically action plugins needed to support
templating.

My understanding is the Ansible core team want to refactor the action
plugins to remove some repeated code so this was unmerged until they have
had the chance to do so.

PR #387, opened 2014-11-25
Add new flavor_name parameter

johnpbatty:
[Resubmitting a change that I originally submitted in #315, but where I screwed up by using the wrong git branch...]

Allow the instance flavor to be specified by name rather than id.

Specifying objects by name is much more human-friendly and portable
between systems than specifying by id.

The nova_compute module already allows images to be specified by name
or id, so this is just providing equivalent function for flavors.

OpenStack does not (unfortunately) enforce that names are unique.
However, any users who have the same name for different flavors are
asking for trouble!  If there are multiple flavors with the same name,
the first one returned by nova.flavors.list() is used.

### DIFF
--- a/cloud/openstack/nova_compute.py
+++ b/cloud/openstack/nova_compute.py
@@ -92,6 +92,11 @@ options:
         - The id of the flavor in which the new VM has to be created. Mutually exclusive with flavor_ram
      required: false
      default: 1
+   flavor_name:
+     description:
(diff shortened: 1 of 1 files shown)


gregdek:
Hi @johnpbatty. Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

gregdek:
@johnpbatty A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

gregdek:
Closing due to inactivity.

PR #392, opened 2014-11-26
Added eu-central-1 AWS region

tricoder42:


tricoder42:
Sorry, it's already reported in #260

PR #394, opened 2014-11-27
Added support for immediate block storage provisioning

cdmitri:
**Added support for immediate block storage provisioning**
- Already exists in ec2 module
- Syntax is the same as in ec2 module
  
  volumes:
       - display_name: vm1-vol1
         device_name: /dev/vdb
         volume_size: 10
         delete_on_termination: true
       - display_name: vm1-vol2
         device_name: /dev/vdc
         volume_size: 20
         delete_on_termination: false
- `delete_on_termination` is not an openstack feature
  but will work during deletion

**Changed the authentication to Password() method**
- Reflects most recent documentation
- Necessary to switch between compute and volume
  actions if API URLs for those are different,
  which is true for a lot of installations

### DIFF
--- a/cloud/openstack/nova_compute.py
+++ b/cloud/openstack/nova_compute.py
@@ -23,8 +23,11 @@ import os
 try:
     from novaclient.v1_1 import client as nova_client
     from novaclient.v1_1 import floating_ips 
+    from novaclient.v1_1.client import Client
     from novaclient import exceptions
     from novaclient import utils
+    from keystoneclient.auth.identity import v2
+    from keystoneclient import session
     import time
 except ImportError:
     print("failed=True msg='novaclient is required for this module'")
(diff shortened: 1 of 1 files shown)


imiu:
I've been looking to have this feature for a while! Kudos to you for making it available!

keithchambers:
LGTM!

keithchambers:
Is there a reason this isn't getting merged?

emonty:
Sorry - these modules are deprecated and not accepting new features. I believe this feature is covered in the new modules at ansible/ansible-modules-extras#92 - albeit in two modules, not one. However, if it's not, I'd love to make those better and make sure it serves your needs. Do you have a strong feeling on being able to combine volume creation and server creation as one play? I don't have a strong aversion to doing that in the os_server module as well ...

emonty:
Created issue ansible/ansible-modules-extras#321 to track this need

PR #406, opened 2014-11-29
Really implement doc-promised upgrade if package(s) already installed

evgkrsk:
To accomplish this goal, we check that hashsum of full packagelist on
system has changed (covers both clean upgrade or dist-upgrade with new
package(s) installation) after apt invocation.

This is exact same implementation as in
https://github.com/ansible/ansible/pull/8530 for ansible-modules-core.

### DIFF
--- a/packaging/os/apt_rpm.py
+++ b/packaging/os/apt_rpm.py
@@ -69,6 +69,13 @@ import shlex
 import os
 import sys
 
+# md5.py is deprecated since python2.5
+try:
+    import hashlib as md5
+except ImportError:
+    # but we need to be compatible with >= python2.4
+    import md5 as md5
+
 APT_PATH="/usr/bin/apt-get"
 RPM_PATH="/usr/bin/rpm"
 
@@ -86,6 +93,14 @@ def query_package_provides(module, name):
     # 1 if it is not installed
     rc = os.system("%s -q --provides %s >/dev/null" % (RPM_PATH,name))
     return rc == 0
+
+def query_packages_digest(module):
+    # compute hashsum of all installed package names
+    # returns None on exception
+        rc, out, err = module.run_command("%s -qa" % RPM_PATH)
+        if rc:
+            return None
+        return md5.md5(out).hexdigest()
 
 def update_package_db(module):
     rc = os.system("%s update" % APT_PATH)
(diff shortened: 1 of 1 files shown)


bcoca:
one thing, there is a  digest_from_file common function for modules, which we have been adapting to work in many contexts, could you switch to using that?

evgkrsk:
digest_from_file needs file as argument, so either:

0) I can leave apt_rpm as is for now
1) I can add temporary file creation in apt_rpm and use digest_from_file
2) Someone can add digest_from_string to basic.py

I like first and last variants most. How do you think?

gregdek:
@bcoca can you reply to @evgkrsk question here?

evgkrsk:
Stupid question: can I just write

``` python
from ansible.utils import secure_hash_s
```

in apt_rpm and use secure_hash_s as replacement to md5 (BTW, seems like md5 will not work in FIPS mode)? Or we must use only ansible.module_utils.\* for module's code?

bcoca:
ansible.utils is not available for the module as it executes on the target host which does not require ansible to be installed, so we are limited to things in module_utils but we are happy to add a equivalent function to secure_hash_s there for this use. 

We checked and all current functions there require a file, which you can use right now by creating a temp file.  Either way is fine with us, let me know .

abadger:
@bcoca and I talked and it would also be okay for you to use sha1 directly without waiting for us to wrap the function.  it will be very similar to your md5 code just using the sha1 algorithm from hashlib or falling back to the sha module from the stdlib.

evgkrsk:
Ok, ping back in this PR when secure_hash_s equivalent wrapper will be available in module_utils, so I can update apt_rpm's code.

gregdek:
So I understand: we're leaving this PR open until a change is made in module_utils, so we can do this the right way?

evgkrsk:
Yes, you are right.

gregdek:
Leaving open for now, just touching on the weekly walkthrough.

gregdek:
@evgkrsk can you tell if the supporting code for this PR has been added to module_utils yet?

evgkrsk:
Nope. There is still NO module_utils function to compute digest from string (current devel branch, ae2447df9136353453c9ed48d44b2c7fa70231b0 ).

abadger:
@evgkrsk If you'd like to get this change off your radar, you can modify your code to use sha1 instead of md5 and we'll accept it.  (md5 is bad to use due to not being available on machines running in FIPS compatibility mode).

evgkrsk:
Sorry about that, but I think that I will close this PR and try to make two new: 1) function that makes digest from string in module_utils 2) unify apt_rpm with apt/yum/zypper/dnf: add **name** as alias of **pkg**, change **state** to one of **latest**, **present**, **absent**.

Reason is simple: practical usage shows that it is inconvenient to always unconditionally install latest version without option to keep currently installed.

PR #407, opened 2014-11-29
[service doc] Clarify supported init systems.

axilleas:
It should be stated which init services are supported. Hope I didn't left any out.


bcoca:
missing BSD init

axilleas:
@bcoca updated thanks.

PR #409, opened 2014-11-29
[authorized_key] added state=replace_all, to ensure only the added keys are present

fabriceb:
I want to improve the authorized_key module, as suggested on the ansible-devel mailing list.
Before that I propose the following refactoring of the existing code to improve readability and maintainability.

### DIFF
--- a/system/authorized_key.py
+++ b/system/authorized_key.py
@@ -37,7 +37,8 @@ options:
     aliases: []
   key:
     description:
-      - The SSH public key, as a string
+      - The SSH public key, as a string.
+        Can be multiple SSH public keys, separated by newline characters "\n"
     required: true
     default: null
   path:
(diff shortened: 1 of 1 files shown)


kosssi:
:+1:

system/authorized_key.py
```diff
@@ -272,25 +272,37 @@ def parsekey(module, raw_key):
...
-        key_data = parsekey(module, line)
-        if key_data:
-            # use key as identifier
-            keys[key_data[0]] = key_data
-        else:
-            # for an invalid line, just append the line
-            # to the array so it will be re-output later
-            keys[line] = line
-    f.close()
+    with open(filename) as f:
```
abadger: Modules have to be compatible with python-2.4 so we can't use context managers

fabriceb:
@abadger : in this new version I do not touch the file i/o anymore, so the module should be compatible with python-2.4 again

I have simplified the pull request, and concentrated on just adding the feature "replace_all". There is already a pull request which would not work with keys containing commas (ie any keys with options) https://github.com/ansible/ansible-modules-core/pull/42

Let me know if you see anything wrong, or if you have a better proposal for the "replace_all" state name.

fabriceb:
By the way, I have written more tests for this module, how do I include them since they are in the ansible repository?

resmo:
I like the feature, but I don't like the name **replace_all**. 

This function does not replace all keys, it even does not replace keys, it removes or purges not defined keys: My suggestion is to rename it to **purge**

bcoca:
at this point I would argue it is easier to just use the template module

fabriceb:
@resmo I do not really like replace_all either. "purge" is interesting but it might be confusing to whether the keys are added or not? I was also thinking about "exclusive"? Or "reset"?

@bcoca I agree that the template module is the way to do it for the moment. But there is parsing of key options and finding automatically where to save the keys, which you don't get if you use the template module. And it also means that depending on your strategy "present, absent, exclusive" you have to use a completely different module. This is why I thought it was a nice to enable all that from the current authorized_key module.

bcoca:
overlaps with #42

fabriceb:
Hi @bcoca It functionally overlaps. But this version takes into account the fact that multiple keys can be passed with \n characters between them, whereas #42 expects the core to be rewritten... imho not so cool.

bcoca:
the new exclusive parameter takes care of this, closing this issue

PR #411, opened 2014-11-30
Fix behavior when insert* doesn't match anything

p-tr0:
If insertbefore/insertafter didn't match anything, lineinfile module was doing nothing, instead of adding the line at end of fille as it's supposed to.

### DIFF
--- a/files/lineinfile.py
+++ b/files/lineinfile.py
@@ -256,9 +256,9 @@ def present(module, dest, regexp, line, insertafter, insertbefore, create,
         msg = 'line added'
         changed = True
     # Add it to the end of the file if requested or
-    # if insertafter=/insertbefore didn't match anything
+    # if insertafter/insertbefore didn't match anything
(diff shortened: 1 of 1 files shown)


abadger:
Looks like the behaviour isn't specified in the documentation, just in the comments in the code.  Would you be able to update the documentation to specify the behaviour as well?

p-tr0:
Done. The docs didn't say what happens when insertbefore/insertafter has multiple matches, so I specified that too - line is added before/after last match. Seems ok for insertafter, but what about insertbefore? I think in that case adding line before first match makes more sense, not sure though.

abadger:
Sorry I lost track of this ticket :-(  merged now.  I agree that before the first match might make more sense.  It would be a backwards incompatible change, though, so I think just documenting the current behaviour is probably the safest thing to do.

Thanks for doing this work!

# Merged

Hi!

issue is resolved.

Thank you!

Issue #412, opened 2014-12-01
Check mode for pip does not work in virtualenv


JimPatterson:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8.1
  configured module search path = None
##### Environment:

Fedora 20 and RHEL 7
##### Summary:

The check mode for the pip module when using a virtualenv always returns changed.
##### Steps To Reproduce:

Using the playbook (t.yaml):

```

---
- hosts: 127.0.0.1
  connection: local
  tasks:
    - name: Ensure we have the python docker module installed
      pip: name=docker-py version=0.6.0 state=present
           virtualenv=/tmp/ansible-bug-pip virtualenv_site_packages=yes
```

Run the playbook twice normally, and once in check mode:

```
rm -rf /tmp/ansible-bug-pip
ansible-playbook t.yaml
ansible-playbook t.yaml
ansible-playbook t.yaml --check
```
##### Expected Results:

The first run creates the virtualenv and installs docker-py.  The 2nd run makes no changes.  The check mode run indicates that nothing needs to change.
##### Actual Results:

The first two runs work exactly as expected, but the check mode always reports changes.

```
$ rm -rf /tmp/ansible-bug-pip
$ ansible-playbook t.yaml 

PLAY [127.0.0.1] ************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [127.0.0.1]

TASK: [Ensure we have the python docker module installed] ********************* 
changed: [127.0.0.1]

PLAY RECAP ******************************************************************** 
127.0.0.1                  : ok=2    changed=1    unreachable=0    failed=0   

$ 
$ 
$ 
$ ansible-playbook t.yaml 

PLAY [127.0.0.1] ************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [127.0.0.1]

TASK: [Ensure we have the python docker module installed] ********************* 
ok: [127.0.0.1]

PLAY RECAP ******************************************************************** 
127.0.0.1                  : ok=2    changed=0    unreachable=0    failed=0   

$ 
$ 
$ 
$ ansible-playbook t.yaml --check

PLAY [127.0.0.1] ************************************************************** 

GATHERING FACTS *************************************************************** 
ok: [127.0.0.1]

TASK: [Ensure we have the python docker module installed] ********************* 
changed: [127.0.0.1]

PLAY RECAP ******************************************************************** 
127.0.0.1                  : ok=2    changed=1    unreachable=0    failed=0   

[...]

PR #413, opened 2014-12-01
Correct check mode for pip in virtualenv.

JimPatterson:
##### Issue Type:

Bugfix Pull Request for #412.

d3cb2d38b7d4b9585adfd4f1ceece1896c2cc952 added better support for check mode, but if the target is in a virtualenv, then it always returned `True`.  Since by that point in the code, we have already verified that we have a valid virtualenv, the pip freeze command will work correctly, so there should be no need to exit early.

### DIFF
--- a/packaging/language/pip.py
+++ b/packaging/language/pip.py
@@ -314,7 +314,7 @@ def main():
         this_dir = os.path.join(this_dir, chdir)
 
     if module.check_mode:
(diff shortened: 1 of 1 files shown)


gregdek:
@mattupstate can you review here? If you give +1 we will tag P1_reviewed for faster inclusion.

mattupstate:
LGTM

PR #419, opened 2014-12-01
Add lucid compatibility

hfinucane:
This has been working OK for me for a while on precise & lucid, and I'm firing off this pull request in case others would like it. 

There was a brief discussion on the <a href="https://groups.google.com/forum/#!searchin/ansible-devel/lucid/ansible-devel/FIZyrcb79bQ/Tw-GrINt5AoJ">mailing list</a>, and there was cautious acceptance of it.

I do not believe trusty's copy of `python-apt` changed the API again, although I don't test much there.

In some ways, I think the best solution would be to drop `python-apt`, and vendor the dpkg parsing code (it's gpl2+). The magic 'install python-apt under the hood' thing seems a little scary to me, and the validation code is more strict than I would like (See my other recent apt-related pull requests).

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -327,10 +327,10 @@ def install_deb(m, debs, cache, force, install_recommends, dpkg_options):
     deps_to_install = []
     pkgs_to_install = []
     for deb_file in debs.split(','):
-        pkg = apt.debfile.DebPackage(deb_file)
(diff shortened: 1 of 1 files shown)


hfinucane:
Fixed here: a08165ca2e05808555077023ffaa1cd4a1499e48

PR #420, opened 2014-12-01
route53 module retry_interval needs to be a float before passing to time.sleep

drfickle:
The retry_interval argument currently comes in as a string, causing a TypeError.

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -160,7 +160,7 @@ def commit(changes, retry_interval):
             code = code.split("</Code>")[0]
             if code != 'PriorRequestNotComplete' or retry < 0:
                 raise e
-            time.sleep(retry_interval)
+            time.sleep(float(retry_interval))
 
 def main():
     argument_spec = ec2_argument_spec()


bcoca:
time.sleep will take a float for greater precision, but this is optional.

I tested in python 2.7 on CentOS/gentoo and FreeBSD without a type error, does this happen on a specific version of python and/or OS?

drfickle:
Correct, but a string is being passed into it. There is nothing in the route53 module which converts it from a string to a number. This was occurred on Ubuntu Precise (12.04) with Python 2.7.3.

bcoca:
What about setting this at the parameter definition? 
       retry_interval  = dict(required=False, default=500,type='float')

drfickle:
Unfortunately that doesn't work because of the integer values the module wants to present to users.

I made a stripped down module to test this and go this output from running it with test-module

```
PARSED OUTPUT
{
    "failed": true, 
    "msg": "argument retry_interval is of invalid type: <type 'int'>, required: float"
}
```

Since the float is really just an implementation detail, I don't think changing the module to require one would be the best option.

drfickle:
This makes no sense to me, but might explain why you're not seeing the problem.

```
TASK: [Update Specific Records for XXX] ************************** 
changed: [XXX_YYY-0 -> 127.0.0.1]
changed: [XXX_YYY-2 -> 127.0.0.1]
changed: [XXX_ZZZ-0 -> 127.0.0.1]
changed: [XXX_ZZZ-1 -> 127.0.0.1]
changed: [XXX_YYY-1 -> 127.0.0.1]
changed: [XXX_ZZZ-2 -> 127.0.0.1]
changed: [XXX_WWW-0 -> 127.0.0.1]
failed: [XXX_WWW-1 -> 127.0.0.1] => {"failed": true, "parsed": false}
invalid output was: SUDO-SUCCESS-ysbsrqljjbrlnmyrannvodbmhekfuzrv
  File "/home/sfox/.ansible/tmp/ansible-tmp-1417621452.6-246597109163412/route53", line 1816, in <module>
    main()
  File "/home/sfox/.ansible/tmp/ansible-tmp-1417621452.6-246597109163412/route53", line 269, in main
    result = commit(changes, retry_interval_in)
  File "/home/sfox/.ansible/tmp/ansible-tmp-1417621452.6-246597109163412/route53", line 163, in commit
    time.sleep(retry_interval)
TypeError: a float is required
```

It seems that each host is not receiving the same retry_interval value for the same task statement? The tasks just reads 'retry_interval=30.0' (using a float value as a work-around instead of my fix).

This is with Ansible 1.7.2.

PR #440, opened 2014-12-03
Fix documentation of the docker.state default value.

andi5:


abadger:
It looks like this has been updated now as part of https://github.com/ansible/ansible-modules-core/pull/502, although the state running has been renamed to started (I need to check if it's a simple rename and create an alias to running if so.)

abadger:
PR to enable state=running to continue to work.  https://github.com/ansible/ansible-modules-core/pull/941

PR #442, opened 2014-12-04
Note that `rsync` must be installed on both local & remote machine

follower:
For `synchronize` to work `rsync` must be installed on both local & remote machine (as appropriate for source/destination) when SSH is used as the transport.

In particular, if `rsync` is not installed on the remote machine the following error message will be encountered:

```
"rsync error: remote command not found"
```


jimi-c:
Merged, thanks!

Issue #347, opened 2014-11-18
ec2: Instance creation failed => InvalidParameter: Security group sg-XxXxXxXx and subnet subnet-YyYyYyYy belong to different networks.


ap1kenobi:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8 (devel 9245acbf44)
##### Environment:

Running on OSX 10.9.5
Managing: various
##### Summary:

Given a Region with multiple VPCs that define two or more security groups by the same name (e.g.: default), attempting to launch an EC2 instance and assign such a security group by name results in ansible error: 

```
msg: Instance creation failed => InvalidParameter: Security group sg-XxXxXxXx and subnet subnet-YyYyYyYy belong to different networks.
```

It seems Ansible (or boto) is trying to assign a security group from VPC other than the one the new instance is being launched into.  When trying to do the same thing via AWS Dashboard, security groups selection is limited to the groups associated with the VPC new EC2 instance is being launched to.

_Ansible should distinguish between similarly named security groups based on Subnet ID the new instance is being launched into._
##### Steps To Reproduce:

Create two VPCs (with public subnets only) in the same Region via AWS Dashboard and leave `default` SG automatically created for you.  Note one of Subnet IDs.

Then attempt to launch an EC2 instance via ansible playbook: 

```

---
- name: Create or maintain AWS environment
  hosts: localhost
  connection: local
  vars:
    env_region: us-east-1
    env_keypair: "your-key-pair"
    env_subnet_Web1: sb-yyyyyyyyy

  tasks:
  - name: Launch an instance of Packer
    local_action:
      module: ec2
      region: "{{ env_region }}"
      keypair: "{{ env_keypair }}"
      group: ["default"]
      instance_type: c3.large
      image:
      vpc_subnet_id: "{{ env_subnet_Web1 }}"
```
##### Expected Results:

New EC2 launched with correct security group assigned to it.
##### Actual Results:

```
TASK: [Launch an instance of Packer] ******************************************
failed: [localhost -> 127.0.0.1] => {"failed": true}
msg: Instance creation failed => InvalidParameter: Security group sg-c7a772a2 and subnet subnet-d220d3f9 belong to different networks.

FATAL: all hosts have already failed -- aborting
```

ap1kenobi:
[...]

PR #446, opened 2014-12-04
ec2: make group only match group names in this VPC

bretmartin:
In an AWS EC2/VPC, it is possible to have two security groups with identical names, but in different VPCs. The Ansible ec2 module currently assumes security group names to be unique when converting them to security group IDs. This is described in detail in issue #347 and this PR is my attempt to correct it.


shaunbrady:
We are interested in this fix as well.  We have a work around in place, but this seems like a logical addition to the ec2 module.

bcoca:
@bretmartin could you rebase?

bretmartin:
@bcoca done

PR #450, opened 2014-12-05
use state parameter in examples

dlbewley:
action parameter no longer exists, so example does not work as written.
proper syntax uses state=present or state=absent


mpdehaan:
Merged, thanks!

PR #453, opened 2014-12-05
Update django_manage to add database option for migrate

pitabwire:
Allow passing the database option to the django_manage module for migrations. This is usefull in situations where multiple databases are used by a django application.


mpdehaan:
Merged, thanks!

PR #459, opened 2014-12-07
Add module option to allow alternative password hashes to be used

alexkingnz:


alexkingnz:
This allows the exiting htpasswd module to support password files with different hashes than the ones used by Apache/Nginx, such as the one supported by libpam-pwdfile (https://github.com/tiwe-de/libpam-pwdfile)

mpdehaan:
Hmm, I'm not sure we need to be passing all that in.   Is there an easier way to express this intent in terms of syntax?

alexkingnz:
Well,

Essentially the underlying module works with a list of possible hashes; 
it can "read" a password in the file created by any of the hashes, but 
will write only in the hash specified in crypt_scheme.

So you can't specify an arbitrary list without adding an extra module 
option or overloading crypt_scheme and allowing it to take a list.

My latest PR is less intrusive and only allows one additional hash to be 
passed in via crypt_scheme, so doesn't let you specify an arbitrary 
list, but it still meets my needs.

Of course there could be other approaches.  Perhaps adding a new module 
altogether in system modules, since this change allows functionality 
beyond just htpasswd files for webservers.

I think the change I'm now suggesting is the minimal, least-surprise 
change yet would allow important new flexibility.

Cheers,
Alex

alexkingnz:
Other PR is #469

gregdek:
@alexkingnz Sorry this PR has been idle for so long.

So I understand: #469 obsoletes this PR, correct? So I can safely close this one?

gregdek:
Closing due to inactivity.

Issue #461, opened 2014-12-07
Add gem "conservative" option


nervo:
##### Issue Type:

Feature Idea
##### Ansible Version:

All known stable version
##### Environment:

N/A
##### Summary:

Adding a gem "conservative" option should avoid gem module to be downloaded and installed everytime the module is played.

See gem option help:

```
    --conservative               Don't attempt to upgrade gems already
                                 meeting version requirement
```
##### Steps To Reproduce:

Define a task:

```
- name: packages
  gem: >
    name=sass
    version=>0
    state=present
    user_install=no
  sudo: yes
```
##### Expected Results:

The gem module should be installed only first time the task is runned
##### Actual Results:

The gem module is installed every time the task is runned

(Hello ansibot :))

mpdehaan:
Typically we do this as force=no (default should probably be no).   Maybe we just want to fix this altogether, but upgrade behavior might not be clear.

hyvs:
Any news about this feature ?

bcoca:
I have no info of anyone working on it yet

nervo:
ping @mpdehaan ?

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
This repository has been locked. All new issues and pull requests should be filed in https://github.com/ansible/ansible

Please read through the <a href="http://docs.ansible.com/ansible/dev_guide/repomerge.html#move-issues-and-prs-to-new-repo">repomerge</a> page in the dev guide. The guide contains links to tools which automatically move your issue or pull request to the ansible/ansible repo.

nervo:
This issue was moved to ansible/ansible#18925

PR #462, opened 2014-12-07
Gem: add support for conservative. Refs #461.

skottler:


mpdehaan:
Per discussion on the other ticket we usually want this as 'force', but need to decide on a smart default and upgrade handling.

Can you please stop by ansible-devel to discuss?

Thanks!

PR #469, opened 2014-12-09
Extend the htpasswd so it can use different hashes in addition to the 4 currently supported by Apache.

alexkingnz:
Extend hashes that can be specified by crypt_scheme beyond those understood by Apache/Nginx, without adding any extra module options.


alexkingnz:
This PR is less intrusive and doesn't add an extra option to the module, sacrificing some functionality, copared with my other PR.

abadger:
Previous PR https://github.com/ansible/ansible-modules-core/pull/459

gregdek:
@lorin please review; if you give +1, we will tag P1_reviewed for faster inclusion.

lorin:
+1

PR #496, opened 2014-12-11
add support for stack policies in cloudformation

haraldsk:
Add very much needed support for using stack policies in aws cloudformation


juliedavila:
Tested, works as intended.

PR #497, opened 2014-12-11
ec2_lc - include all launch config properties in the return

robbwagoner:
make all properties available when registering the result
which is useful when wanting to launch a stand-alone instance based upon
an existing Launch Config.


robbwagoner:
Any chance this will make it into the next release? This is a trivial change and won't break any compatibility.

bcoca:
it seems it would be simpler to just return result itself, adding changed= to that dict

bennojoy:
@robbwagoner  could you please change it to return just the result object as bcoca suggested  when you get time and it should be good to merge. thanks

robbwagoner:
Apologies for the long delay @bennojoy and @bcoca . This is done.

Issue #27, opened 2014-09-29
docker module supposed to be able to upgrade?


bfleming-ciena:
If you're familiar with the apt module, you can specify state=latest to get ansible to upgrade.

I want to do the same with docker. I want to run an image, and when that image has been updated I want the old container stopped and the new version of the container started.

Does the docker module support this?  I've tried and it doesn't work as I hoped.

Example playbook which I hoped to re-use over and over to ensure my target is at the latest.

hosts: localhost 
sudo: yes
tasks:

name: ensure redis container is running
docker: image=dockerfile/redis name=redis command=" bash -c 'redis-server /etc/redis/redis.conf'"

name: ensure latest version of the web container is running
docker: image=snazzy/cyweb ports=5000:5000 links=redis:redis name=cyweb

Thanks

vlfig:
Ansible docker module does not ensure correct container state after an image tag update.

When an image gets built and tagged with the same string that was provided to the docker plugin as the `image` param, the container associated with the previous image (no longer tagged) keeps running. No container associated with the given image name is running.

```
- name: run latest gocd-server
  sudo: yes
  docker: "image=organisation/gocd-server:latest state=running ports=8000:8000"
```

Steps to reproduce:
1. build an image and tag it.
2. run the above ansible task
3. rebuild the image (associate same tag with a different image)
4. run the above ansible task again
5. verify container running is based of the original image and no container associated with the now-tagged image is running.

Example (edited for clarity):

```
docker@mylovelydockerhost:~$ sudo docker images
REPOSITORY                         TAG                 IMAGE ID             CREATED       
organisation/gocd-server           latest              1246944e6539         16 hours ago  
coreplatform/gocd-agent            latest              8d2dee2ff4dc         18 hours ago  

docker@mylovelydockerhost:~$ sudo docker ps -a
CONTAINER ID        IMAGE                              COMMAND              CREATED       
2497cb1c0044        organisation/gocd-agent:latest     "/bin/sh -c '/etc/in 20 minutes ago
1566143ac885        organisation/gocd-server:latest    "/bin/bash"          16 hours ago  
```

All is well. Now run `docker build -t organisation/gocd-server:latest` pointing to a different Dockerfile. The tag gets updated.

```
docker@mylovelydockerhost:~$ sudo docker images
REPOSITORY                         TAG                 IMAGE ID             CREATED       
organisation/gocd-server           latest              d28b2c1fc0b7         20 minutes ago
[...]

PR #484, opened 2014-12-10
Add pull_latest and insecure_registry parameters to Docker module

psviderski:
This is a continuation of the work submitted by @abevoelker in pull request #85. The `restart_policy` part was eliminated as this feature is already implemented, tested and ready to be merged (PR #199).
The correctness of `insecure_registry` option handling was also tested with docker-py versions prior to 0.5.0.

PR #85, opened 2014-09-30
Add pull_latest, restart_policy, and insecure_registry parameters to Docker module

abevoelker:
This is a resubmission of @discordianfish's `ansible` repo pull request [#9095](https://github.com/ansible/ansible/pull/9095).  I did not split it into three separate PRs because if they would be accepted/merged separately it would cause merge conflicts.  Let me know if this is a problem and I will modify as needed.
- `insecure_registry` applies to `ansible` repo [#9095](https://github.com/ansible/ansible/pull/9095)
- `restart_policy` applies to `ansible` repo [#8815](https://github.com/ansible/ansible/pull/8815)
- `pull_latest` applies to `ansible` repo [#8805](https://github.com/ansible/ansible/pull/8805)

PR #502, opened 2014-12-12
Docker: state reorganization and pull option

smashwilson:
This adds a `pull` option to the Docker module, which has two possible settings:
- `missing` _(default)_. The named image and tag will only be pulled if missing from the host. This is backwards-compatible with the existing behavior.
- `always`. A pull will always be performed to check for more recent builds first. If an existing container is already running with an older version of that image, it will be stopped and restarted.

The use case that this captures is continuous delivery of a Docker image. You can use a task like this one:

``` yaml
  - name: Run the docker container
    docker: pull=always image=smashwilson/minimal-sinatra name=minimal-sinatra ports=80:4567 state=running
```

... where "smashwilson/minimal-sinatra" is a container that's generated by an automated build on DockerHub, for example. Each time you execute the play, Ansible will ensure that the most recent build of that image is running. Without this, I find myself using ugly workarounds like:

``` yaml
- name: pull the latest docker build
  command: docker pull myimage
  register: myimage
  changed_when: outsideimage.stdout.find('Downloaded newer image') != -1
  sudo: yes

- name: stop the older image
  docker: image=myimage name=mycontainer state=absent
  when: myimage | changed
  sudo: yes

- name: launch the service
  docker: image=myimage name=mycontainer state=running
  sudo: yes
```

But you don't _always_ want to attempt to pull the latest on every deployment. It would be inconvenient if your database container was unconditionally restarted, for example, and upgraded under your feet. For that reason and for backwards compatibility I gated this behind an module option.

This has been requested before, specifically in #27.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -507,12 +507,50 @@ class DockerManager(object):
 
         self.env = self.module.params.get('env', None)
 
-        # connect to docker server
-        docker_url = urlparse(module.params.get('docker_url'))
+        # Connect to the docker server using any configured host and TLS settings.
+
+        env_host = os.getenv('DOCKER_HOST')
+        env_cert_path = os.getenv('DOCKER_CERT_PATH')
+
+        docker_url = module.params.get('docker_url')
+        if not docker_url:
+            if env_host:
+                docker_url = env_host
+            else:
+                docker_url = 'unix://var/run/docker.sock'
+
+        docker_tls_cert = module.params.get('docker_tls_cert')
+        if not docker_tls_cert and env_cert_path:
+            docker_tls_cert = os.path.join(env_cert_path, 'cert.pem')
+
+        docker_tls_key = module.params.get('docker_tls_key')
+        if not docker_tls_key and env_cert_path:
+            docker_tls_key = os.path.join(env_cert_path, 'key.pem')
+
+        docker_tls_cacert = module.params.get('docker_tls_cacert')
+        if not docker_tls_cacert and env_cert_path:
+            docker_tls_cacert = os.path.join(env_cert_path, 'ca.pem')
+
         docker_api_version = module.params.get('docker_api_version')
         if not docker_api_version:
             docker_api_version=docker.client.DEFAULT_DOCKER_API_VERSION
-        self.client = docker.Client(base_url=docker_url.geturl(), version=docker_api_version)
+
+        tls_config = None
+        if docker_tls_cert or docker_tls_key or docker_tls_cacert:
+            # See https://github.com/docker/docker-py/blob/d39da11/docker/utils/utils.py#L279-L296
+            docker_url = docker_url.replace('tcp://', 'https://')
+            verify = docker_tls_cacert is not None
+
+            tls_config = docker.tls.TLSConfig(
+                client_cert=(docker_tls_cert, docker_tls_key),
+                ca_cert=docker_tls_cacert,
+                verify=verify,
+                assert_hostname=False
+            )
+
+        self.client = docker.Client(base_url=docker_url,
+                                    version=docker_api_version,
+                                    tls=tls_config)
 
         self.docker_py_versioninfo = get_docker_py_versioninfo()
 
@@ -1286,7 +1324,10 @@ def main():
             links           = dict(default=None, type='list'),
             memory_limit    = dict(default=0),
             memory_swap     = dict(default=0),
-            docker_url      = dict(default='unix://var/run/docker.sock'),
+            docker_url      = dict(),
+            docker_tls_cert = dict(),
+            docker_tls_key  = dict(),
+            docker_tls_cacert = dict(),
             docker_api_version = dict(),
             username        = dict(default=None),
             password        = dict(),
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -322,6 +322,8 @@ EXAMPLES = '''
 HAS_DOCKER_PY = True
 
 import sys
+import json
+import re
 from urlparse import urlparse
 try:
     import docker.client
@@ -631,6 +633,18 @@ class DockerManager(object):
 
         return False
 
+    def get_inspect_image(self):
+        image, tag = get_split_image_tag(self.module.params.get('image'))
+        if tag is None:
+            tag = 'latest'
+        resource = '%s:%s' % (image, tag)
+
+        matching_image = None
+        for image in self.client.images(name=image):
+            if resource in image.get('RepoTags', []):
+                matching_image = image
+        return matching_image
+
     def get_inspect_containers(self, containers):
         inspect = []
         for i in containers:
@@ -683,6 +697,41 @@ class DockerManager(object):
 
         return running
 
+    def pull_image(self):
+        extra_params = {}
+        if self.module.params.get('insecure_registry'):
+            if self.ensure_capability('insecure_registry', fail=False):
+                extra_params['insecure_registry'] = self.module.params.get('insecure_registry')
+
+        resource = self.module.params.get('image')
+        image, tag = get_split_image_tag(resource)
+        if self.module.params.get('username'):
+            try:
+                self.client.login(
+                    self.module.params.get('username'),
+                    password=self.module.params.get('password'),
+                    email=self.module.params.get('email'),
+                    registry=self.module.params.get('registry')
+                )
+            except:
+                self.module.fail_json(msg="failed to login to the remote registry, check your username/password.")
+        try:
+            last = None
+            for line in self.client.pull(image, tag=tag, stream=True, **extra_params):
+                last = line
+            status = json.loads(last).get('status', '')
+            if status.startswith('Status: Image is up to date for'):
+                # Image is already up to date. Don't increment the counter.
+                pass
+            elif status.startswith('Status: Downloaded newer image for'):
+                # Image was updated. Increment the pull counter.
+                self.increment_counter('pull')
+            else:
+                # Unrecognized status string.
+                self.module.fail_json(msg="Unrecognized status from pull", status=status)
+        except:
+            self.module.fail_json(msg="failed to pull the specified image: %s" % resource)
+
     def create_containers(self, count=1):
         params = {'image':        self.module.params.get('image'),
                   'command':      self.module.params.get('command'),
@@ -704,11 +753,6 @@ class DockerManager(object):
         if params['volumes_from'] is not None:
             self.ensure_capability('volumes_from')
 
-        extra_params = {}
-        if self.module.params.get('insecure_registry'):
-            if self.ensure_capability('insecure_registry', fail=False):
-                extra_params['insecure_registry'] = self.module.params.get('insecure_registry')
-
         def do_create(count, params):
             results = []
             for _ in range(count):
@@ -721,23 +765,7 @@ class DockerManager(object):
         try:
             containers = do_create(count, params)
         except:
-            resource = self.module.params.get('image')
-            image, tag = get_split_image_tag(resource)
-            if self.module.params.get('username'):
-                try:
-                    self.client.login(
-                        self.module.params.get('username'),
-                        password=self.module.params.get('password'),
-                        email=self.module.params.get('email'),
-                        registry=self.module.params.get('registry')
-                    )
-                except:
-                    self.module.fail_json(msg="failed to login to the remote registry, check your username/password.")
-            try:
-                self.client.pull(image, tag=tag, **extra_params)
-            except:
-                self.module.fail_json(msg="failed to pull the specified image: %s" % resource)
-            self.increment_counter('pull')
+            self.pull_image()
             containers = do_create(count, params)
 
         return containers
@@ -803,6 +831,7 @@ def main():
         argument_spec = dict(
             count           = dict(default=1),
             image           = dict(required=True),
+            pull            = dict(required=False, default='missing', choices=['missing', 'always']),
             command         = dict(required=False, default=None),
             expose          = dict(required=False, default=None, type='list'),
             ports           = dict(required=False, default=None, type='list'),
@@ -845,11 +874,24 @@ def main():
         count = int(module.params.get('count'))
         name = module.params.get('name')
         image = module.params.get('image')
+        pull = module.params.get('pull')
 
         if count < 0:
             module.fail_json(msg="Count must be greater than zero")
         if count > 1 and name:
             module.fail_json(msg="Count and name must not be used together")
+
+        # Explicitly pull new container images, if requested.
+        # Do this before noticing running and deployed containers so that the image names will differ
+        # if a newer image has been pulled.
+        if pull == "always":
+            manager.pull_image()
+
+        # Find the ID of the requested image and tag, if available.
+        image_id = None
+        inspected_image = manager.get_inspect_image()
+        if inspected_image:
+            image_id = inspected_image.get('Id')
 
         running_containers = manager.get_running_containers()
         running_count = len(running_containers)
@@ -873,7 +915,7 @@ def main():
 
                 # the named container is running, but with a
                 # different image or tag, so we stop it first
-                if existing_container and existing_container.get('Config', dict()).get('Image') != image:
+                if existing_container and (image_id is None or existing_container.get('Image') != image_id):
                     manager.stop_containers([existing_container])
                     manager.remove_containers([existing_container])
                     running_containers = manager.get_running_containers()
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -164,12 +164,13 @@ options:
     - '"killed" stop and kill all matching containers. "absent" stops and then'
     - removes any matching containers.
     required: false
-    default: present
+    default: started
     choices:
     - present
     - started
     - reloaded
     - restarted
+    - stopped
     - killed
     - absent
   privileged:
@@ -324,6 +325,8 @@ HAS_DOCKER_PY = True
 import sys
 import json
 import re
+import os
+import shlex
 from urlparse import urlparse
 try:
     import docker.client
@@ -356,8 +359,10 @@ def _human_to_bytes(number):
     print "failed=True msg='Could not convert %s to integer'" % (number)
     sys.exit(1)
 
+
 def _ansible_facts(container_list):
     return {"docker_containers": container_list}
+
 
 def _docker_id_quirk(inspect):
     # XXX: some quirk in docker
@@ -384,6 +389,13 @@ def get_split_image_tag(image):
         resource = image
 
     return resource, tag
+
+
+def is_running(container):
+    '''Return True if an inspected container is in a state we consider "running."'''
+
+    return container['State']['Running'] == True and not container['State'].get('Ghost', False)
+
 
 def get_docker_py_versioninfo():
     if hasattr(docker, '__version__'):
@@ -414,6 +426,7 @@ def get_docker_py_versioninfo():
 
     return tuple(version)
 
+
 def check_dependencies(module):
     """
     Ensure `docker-py` >= 0.3.0 is installed, and call module.fail_json with a
@@ -429,8 +442,12 @@ def check_dependencies(module):
 
 class DockerManager(object):
 
-    counters = {'created':0, 'started':0, 'stopped':0, 'killed':0, 'removed':0, 'restarted':0, 'pull':0}
+    counters = dict(
+        created=0, started=0, stopped=0, killed=0, removed=0, restarted=0, pulled=0
+    )
+    reload_reasons = []
     _capabilities = set()
+
     # Map optional parameters to minimum (docker-py version, server APIVersion)
     # docker-py version is a tuple of ints because we have to compare them
     # server APIVersion is passed to a docker-py function that takes strings
@@ -553,7 +570,6 @@ class DockerManager(object):
 
         return processed_links
 
-
     def get_exposed_ports(self, expose_list):
         """
         Parse the ports and protocols (TCP/UDP) to expose in the docker-py `create_container` call from the docker CLI-style syntax.
@@ -571,7 +587,6 @@ class DockerManager(object):
             return exposed
         else:
             return None
-
 
     def get_port_bindings(self, ports):
         """
@@ -615,6 +630,37 @@ class DockerManager(object):
 
         return binds
 
+    def get_summary_message(self):
+        '''
+        Generate a message that briefly describes the actions taken by this
+        task, in English.
+        '''
+
+        parts = []
+        for k, v in self.counters.iteritems():
+            if v == 0:
+                continue
+
+            if v == 1:
+                plural = ""
+            else:
+                plural = "s"
+            parts.append("%s %d container%s" % (k, v, plural))
+
+        if parts:
+            return ", ".join(parts) + "."
+        else:
+            return "No action taken."
+
+    def get_reload_reason_message(self):
+        '''
+        Generate a message describing why any reloaded containers were reloaded.
+        '''
+
+        if self.reload_reasons:
+            return ", ".join(self.reload_reasons)
+        else:
+            return None
 
     def get_summary_counters_msg(self):
         msg = ""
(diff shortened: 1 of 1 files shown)


thejuan:
It would be awesome if this could ensure only a given tag was present.
i.e.
as time goes on you will build up images with different versions, it would be nice to make sure only the tag you have specified is available.
dockerfile/elasticsearch:1.1
dockerfile/elasticsearch:1.2
dockerfile/elasticsearch:1.3

abadger:
@smashwilson I'm wondering is there a time when you'd want the restart and pull functionality to be separate?  I suppose that you probably don't want to pull if you don't want to restart?  Because you end up waiting a long time for the download but then nothing really changes on the running system afterwards?

OTOH, you might download an image outside of ansible but not want to update immediately (so that you don't have your db server rebooting on the nightly ansible run). I think the current code in this PR which checks image_id means that if you downloaded outside of ansible and then ran the playbook, pull=missing would not protect you from the container being restarted.  Not sure if that should just be a bug in the code (add "and pull=='always'" to some of the conditionals) or if we want to separate out pull from restart, though.) 

I also have been looking at several pull_latest Pr's (this one in particular: https://github.com/ansible/ansible-modules-core/pull/484 ) which seem to handle pulling a new image but only in the case of creating a new container.  I'm not sure if we should combine that concept with this as well.

smashwilson:
> @smashwilson I'm wondering is there a time when you'd want the restart and pull functionality to be separate?

Possibly! I think that conflating the two is a consequence of unifying the `docker_image` and `docker` modules. Before, there were two separate tasks, one for image management and another for container management; now both are handled by a single module. As it stands, I'm not sure how to separate them. Maybe something like:

``` yaml
- name: ensure the image is up-to-date
  docker: image=mystuff/service pull=always
  register: serviceimage

- name: restart if the container is old
  docker: name=microservice image=mystuff/service state=restarted
  when: serviceimage | changed
```

The first task's default state of `present` will start the container if it doesn't exist, though! We'd need a separate state that says "don't touch the containers, just manage the image." But, on that note...

> OTOH, you might download an image outside of ansible but not want to update immediately (so that you don't have your db server rebooting on the nightly ansible run).
> 
> I think the current code in this PR which checks image_id means that if you downloaded outside of ansible and then ran the playbook, pull=missing would not protect you from the container being restarted. Not sure if that should just be a bug in the code (add "and pull=='always'" to some of the conditionals) or if we want to separate out pull from restart, though.)

Those are very good points. I admit I have some confusion about how aggressive the docker module should be.

In general, I tend to think of Ansible tasks not as actions to be performed, but rather as states to be asserted. I prefer to grow a playbook as I evolve an existing architecture, adding and editing tasks and re-running as it changes. For example, to me, it's surprising that I can have an existing task like this one:

``` yaml
- name: ensure the service is running
  docker: name=microservice image=mystuff/service state=running
```

Edit it to this:

``` yaml
- name: ensure the service is running
  docker: name=microservice image=mystuff/service env={ "FOO": "BAR" } state=running
```

... and, when I next run the play, have the task take no action and report no changes, even though the running container doesn't have the environment that the task seems to assert! Ditto for the container's volumes, links, and so on.

On the other hand, I absolutely agree it's _also_ surprising and dangerous for the docker task to kick your containers when all that you requested was `state=running`. The `image_id` check that's part of this PR would be a move in the "aggressively restart" direction, as opposed to the "safely lazy" stance the rest of the module currently adopts. So it's probably not a good idea as it's written.

Could we extend the choices for `state`, maybe? Something like `state=uptodate` that says "if any of these parameters no longer matches the running container, restart it to make it true; if they all still match, do nothing." ("uptodate" is a terrible name for that, by the way, I just can't think of anything better right now.) `state=running` could stay the way it is, then, and remain the safe and lazy choice.

> I also have been looking at several pull_latest Pr's (this one in particular: #484 ) which seem to handle pulling a new image but only in the case of creating a new container. I'm not sure if we should combine that concept with this as well.

I think that approach is more in line with the way the rest of the module presently behaves. It just doesn't let me deploy new images by re-running the playbook :wink:

What do you think about adding the new state I described above? Any ideas for what I should call it?

abadger:
I've been thinkng about your post over the weekend and I'm tentatively in favour of a new state.  I think we probably should write down the combinations of states (and still a new pull parameter?) and which use cases they're implementing.  I don't want to have so many states and parameters that it's confusing which combination a person should use but the amount that we have currently is plainly too few to satisfy all the needs that people have.

smashwilson:
> I've been thinkng about your post over the weekend and I'm tentatively in favour of a new state.

Excellent, now we just need to decide on a good name for it. It sits between "running" and "restarted" like so:
- `state=running`: Create the container _only_ if it does not already exist. Do nothing if a container with the same name is already active.
- `state=???`: Create the container if it does not already exist. Restart the container if it's already running, but with different options (including image ID). If it exists and still has the same options, do nothing.
- `state=restarted`: Always restart the container, regardless of initial state.

"latest" maybe, for consistency with other modules... ? That sort of maps to the the concept. I'm not sure.

> I think we probably should write down the combinations of states (and still a new pull parameter?) and which use cases they're implementing. I don't want to have so many states and parameters that it's confusing which combination a person should use but the amount that we have currently is plainly too few to satisfy all the needs that people have.

Agreed. The ansible-doc examples would be a good place to do this. Personally I find the existing states a bit confusing (I always forget what `state=present` does, and usually end up checking the source) and I'd hate to make it worse. That might be creeping out of the scope of this PR, though.

bcoca:
started instead of running, we already use it elsewhere, restarted and
reloaded can work for the other 2.

## 

Brian Coca

gmalouf:
Is this on hold at this point?

smashwilson:
> Is this on hold at this point?

Kind of. I think I know what I want to do with it now, but one of my other projects is broken so I've been spending time elsewhere.

> started instead of running, we already use it elsewhere, restarted and
> reloaded can work for the other 2.

Wait, @bcoca, do you mean:
- `state=started`: Create the container only if it does not already exist. Do nothing if a container with the same name is already active.
- `state=reloaded`: Create the container if it does not already exist. Restart the container if it's already running, but with different options (including image ID). If it exists and still has the same options, do nothing.
- `state=restarted`: Always restart the container, regardless of initial state.

Or do I have those last two flipped?

bcoca:
@smashwilson no, i think you got it right

MrMMorris:
Really excited for this capability! Without this feature, Continuous Deployment with Docker using Ansible is not as nice...

daviddyball:
What were the original reasons for deprecating `docker_image` and consolidating on the `docker` module? IMHO Docker is too big to be in one module, the only way to do it justice would be to try and recreate the sub-command style they use in the docker CLI by specifying something like `action=[inspect|exec|run|pull|kill|stop|start]`.... Which seems a bit over-complicated for a single module.

I'd actually be in favour of multiple modules that corresponded to the equivalent Client API calls e.g:

```
docker_pull
docker_run   (combines create+start calls)
docker_stop (allows force=true to kill vs. stop)
docker_remove
docker_inspect
docker_exec
docker_ps
```

Things like `docker_run` could have states like `state=present` (make sure a container with `name` is defined) or `state=force` (make sure the container def is _exactly_ as the spec or destroy and recreate).

With finer module control it would be a lot easier to ensure your containers were in exactly the state you expected them to be in, rather than crossing your fingers and hoping that the monolithic `docker` module acts on state in the same way you hoped.

Any thoughts?

MrMMorris:
@daviddyball I actually just ran into a situation where the docker module doesn't seem to have an option to make sure a container is just created (not run). For data-only containers, this is needed and has confused me about the different between `state=running` and `state=present`. It seems we need some state that maps to `docker create` as well

smashwilson:
I'm working on this again, by the way! I've been refactoring and cleaning things up while I'm at it. I haven't tested what I just pushed _at all_, though, that'll be tomorrow.

> With finer module control it would be a lot easier to ensure your containers were in exactly the state you expected them to be in, rather than crossing your fingers and hoping that the monolithic docker module acts on state in the same way you hoped.

That's an interesting idea. The `docker_inspect` module might not be worth having - I'd rather see all of the others offer inspect output for register - but the rest seem reasonable to me. My only objection would be that this approach would lead to "task soup" even for simple, common cases. Would any of the Docker folks care to weigh in?

> @daviddyball I actually just ran into a situation where the docker module doesn't seem to have an option to make sure a container is just created (not run). For data-only containers, this is needed and has confused me about the different between state=running and state=present. It seems we need some state that maps to docker create as well.

This actually seems like a reasonable thing for `state=present` to do - create the container if it doesn't exist, but don't start it, and leave any matching containers alone if they already exist but are stopped.

MrMMorris:
> This actually seems like a reasonable thing for state=present to do - create the container if it doesn't exist, but don't start it, and leave any matching containers alone if they already exist but are stopped.

Agreed. That's what I thought it meant, but it seems to be almost the same as `state=running`

Thanks for taking the time to work on this!

MrMMorris:
@smashwilson is this at a point where I can test it out with your fork? :sake:

smashwilson:
@MrMMorris It's getting there! I've revamped and have been testing all of the states except the one I actually care about the most, `reloaded`.

Actually, if you want to take it for a spin and let me know if they behave according to your expectations, that would be appreciated. The documentation is up to date although the examples are not.

costimuraru:
Looking forward for this feature

abadger:
@smashwilson Just letting you know that we're getting the code together to ship out ansible-1.9 in mid-March (just around the corner).  If this seems done and the only worry is that it has only had light testing of some portions it might be good to merge it and then be sure to get testing on it during our release candidates (we're doing release candidates for 1.9, yay!)

On the other hand, if you do know that there's some things that aren't working it would be great to list them on the chance that someone could help out with pushing it the last little way to completion.

And if it's farther away from done than I think, there's always the next version too.  let me know that and I'll see if there's any other PR's that I've been holding off on merging until the framework here was done to see if they're suitable.

smashwilson:
@abadger Thanks for the heads up!

The short version of the status of this is that I have one more state to implement, but it's the trickiest one. I'm hoping it'll take another day or two of my time to get to a mergeable state, modulo feedback of course - barring things like family emergencies I should have the time at the end of this week and next.

If that's too close to the wire and the rest of my work looks acceptable, I could also omit the last state for this release and resubmit that alone as a new PR post-1.9. It would only take a few minutes to touch up the documentation to match.

---

The long version is that I have all of the refactoring done [and tested](https://github.com/smashwilson/docker-module-test) for the following states:
- [x] `present`
- [x] `started`
- [x] `restarted`
- [x] `stopped`
- [x] `killed`
- [x] `absent`

I feel like I have something in place that's [more comprehensible and consistent](https://github.com/smashwilson/ansible-modules-core/blob/docker-pull-option/cloud/docker/docker.py#L1082-L1180) than what was there before, although I'd be curious to have more feedback about that from someone who didn't write it :wink:

I also addressed #657 along the way because I was testing with boot2docker, although #854 looks like a better implementation with more options.

The remaining work is:
- [x] The `reloaded` state, to restart or recreate a container if and only if its configuration is different from the module parameters.
- [x] Bring the examples up to date and ensure they demonstrate coverage and use cases of each state.

There's also a decent amount of soft-coding in the module to support a wide variety of docker-py and docker versions, which I haven't touched but also haven't tested for.

I'm also seeing a lot of open PRs that will conflict with this one as well: #854, #671, #670, #123, #85 to name a few. Any thoughts on this?

MrMMorris:
I actually just got compose going with ansible-pull, but it was kind of an interim solution anyways and I'm waiting for a CircleCI bug fix so I will try this out today :yum:

Stupid Question: How do I test out the module? Just pull it locally or something?

abadger:
I like the splitting of each state into a function.  I like it a lot :-)  I'm not a big fan of nested functions, though.  I'd probably move those outside of the function and pass in the variables (A brief look seems like they just need container, manager, and count).

For #657 If you think #854 is better, I can review that and merge it and you can rebase on top of it.  Looks like there's some superficial work to be done there at least -- the patch has a weird 2 space indentation scheme that will have to be changed.

It would be very nice to get reloaded in but we are getting tight on time.  How about this -- I want to merge on Monday if possible (I can fix up the nested functions and rebasing onto a fixed up #854 if necessary).  If you can finish up reloaded by then we'll get it into the 1.9 release candidates.  And hopefully people will test those enough to identify and fix any bugs that crop up.

If it's looking like reloaded won't be finished by Monday or Tuesday then it would be good to get the reduced feature version in by that time instead.

For all the other PRs... I haven't taken a comprehensive look in about a month but when I looked then there was a real mess.  Some PRs try to do the exact same thing as this one and we can just close them as duplicating the functionality.  Many of them, though implement some slightly different use cases or unrelated functions.  We'll probably have to ping each of those tickets once this is in and ask the authors if they would like to port their additional changes to the new framework you're creating or if they're satisfied with what gets done here.

abadger:
Slight update on timeframes -- looks like we're shooting to release rc1 on Monday.  So we would want to get whatever portions of the code we can into that.  takes one day away from the time to get the code for reloaded done.

smashwilson:
> Stupid Question: How do I test out the module? Just pull it locally or something?

The easiest way would be to clone my fork and use `ansible-playbook -M` to point to the local copy:

``` bash
cd ${SOMEWHERE}
git clone https://github.com/smashwilson/ansible-modules-core.git

cd ${PLAYBOOK_DIR}
ansible-playbook -M ${SOMEWHERE}/ansible-modules-core/cloud/docker site.yml -v
```

You can see some usage examples in my [test repository](https://github.com/smashwilson/docker-module-test) for the moment.

> Slight update on timeframes -- looks like we're shooting to release rc1 on Monday. So we would want to get whatever portions of the code we can into that. takes one day away from the time to get the code for reloaded done.

Noted. Had a few crises come up today to deal with first (ironically, one of which would have been made much simpler to resolve with this functionality) but this is next on my list. Thanks for keeping me up to date :smile:

bcoca:
use the hacking/test-module script in the repo

MrMMorris:
Alright, I will be testing this on my own today and tomorrow :)

MrMMorris:
@smashwilson for questions/comments, should I post here or are you on IRC?

smashwilson:
@MrMMorris Just post here. I'm being yanked around a lot right now, so the more asynchronous, the better. (Plus, this way others have a chance to chime in with opinions.)

MrMMorris:
@smashwilson am I correct in assuming that something like restart_policy should be working?

`ansible-playbook -M ~/Work/ansible-modules-core/cloud/docker -i hosts site.yml`

`msg: unsupported parameter for module: restart_policy`

smashwilson:
Hmm, I didn't touch that, and it still looks like it's there...

Make sure your `~/Work/ansible-modules-core/` repo is on the right branch - it should be `docker-pull-option`.

MrMMorris:
@smashwilson that was it, thanks. An issue I have hit:

The docs say that `present` is the default state, but it seems that when I don't include a state, it uses `running` which is not supported:

```
- name: run data container
  docker:
    name: data
    command: echo
    image: mongo:2.6
    volumes: /data/db
```

```
TASK: [run data container] *****************************************
msg: value of state must be one of: present,started,reloaded,restarted,stopped,killed,absent, got: running
```

There is also a reference to `state=running` here: http://goo.gl/IhC7ft

MrMMorris:
Another issue:

Task:

```
- name: run registrator container
  docker:
    name: registrator
    image: progrium/registrator:latest
    memory_limit: 50MB
    volumes: /var/run/docker.sock:/tmp/docker.sock
    command: "consul://{{ ansible_default_ipv4['address'] }}:8500"
    publish_all_ports: True
    restart_policy: always
    state: started
```

Error:

```
SUDO-SUCCESS-usyuuadyysctkmnhrwxxjbkwtnlevvxm
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1425594836.44-177322795885462/docker", line 3045, in <module>
    main()
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1425594836.44-177322795885462/docker", line 1448, in main
    started(manager, containers, count, name)
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1425594836.44-177322795885462/docker", line 1307, in started
    containers.refresh()
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1425594836.44-177322795885462/docker", line 1268, in refresh
    self.deployed = self.manager.get_deployed_containers()
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1425594836.44-177322795885462/docker", line 1095, in get_deployed_containers
    image, tag = get_split_image_tag(image)
UnboundLocalError: local variable 'image' referenced before assignment
```

Rerunning the playbook does not resolve the error.

smashwilson:
@MrMMorris Awesome, thanks :grin: Keep 'em coming. I made good progress on `reloaded` today and I'm hoping to have it functional tomorrow. I'll use your comments as TODOs. You've already caught two cases I didn't hit with my testing!

MrMMorris:
@smashwilson any ideas what the problem is with my last error? I want to continue with testing, but I can't do much without fixing it. If it's just a line change or something, I can do that locally :bowtie:

smashwilson:
@MrMMorris The image doesn't exist on the host to begin with - try logging in and running a manual `docker pull` for it first.

smashwilson:
@abadger The reloaded state has been implemented, added, and tested for a few common cases. :sparkles: There may still be dragons lurking in reload detection for some options (like lxc_conf, say) because the docker API isn't always consistent in how options are round-tripped through the API, but it works for the common ones: image, command, volumes, port bindings, environment.

Both of the bugs @MrMMorris found have been reproduced locally and addressed. I've also touched up the documentation and examples.

I did try the merge against #854 to see how bad the conflicts would be and the answer is unfortunately "messy". It's especially complicated because mine honors the `DOCKER_` environment variables, which lets us integrate with boot2docker, docker swarm, or docker machine, but his adds more fine-grained control over TLS strictness and lets you use system TLS settings, so... really, I think we want a solution that's somewhere in the middle. I'm not sure what you want to do about that.

Also, I have a lot of commits here. Did you want me to rebase them together somehow? A single commit, or several logical ones?

MrMMorris:
@smashwilson not sure if this was the case before, but I assumed it should mirror docker-compose.yml:

When you set `environment:` with just the key, it looks to the host environmental variables for the value. For example:

`env: "DB_HOST=mongo"` should create the `DB_HOST` env var with a value of `mongo`
`env: "SECRET"` should create the `SECRET` env var, with the value being whatever the value is for `SECRET` on the host running the container.

https://docs.docker.com/compose/yml/#environment

MrMMorris:
and a question: you can't use `count` with `name`, which makes sense. Is is still possible then to link to containers which aren't named?

abadger:
@smashwilson Yay!

If it's not too much trouble a couple logical commits are best.  But I know how hard that can be when you have a few commits that touch one thing and then a few commits that touch something else.  I think the ideal would be one commit for the state reorganization + pull option.  And a second one for the tls stuff.

smashwilson:
@MrMMorris:

> @smashwilson not sure if this was the case before, but I assumed it should mirror docker-compose.yml:
> 
> When you set environment: with just the key, it looks to the host environmental variables for the value.

Hmm. Ansible has several other mechanisms for doing that, though: `{{ lookup('env') }}` and using `ansible-vault` to store secrets. I would be cautious about deviating from Ansible convention to do this, although I see what you're saying.

Also, I had some inkling of using blank environment variables to ensure that a specific variable was _not_ set in the container. As written, I had to use a subset relationship to determine if an environment has changed sufficiently to reload the container, because variables added by parent images or the Dockerfile also appear in the "inspect" output. As a consequence, deleting a key from an `env:` dict won't be sufficient to trigger a container reload. Unfortunate, but workable to get things moving.

Let's punt on that until we see some feedback from people using this in anger.

> and a question: you can't use count with name, which makes sense. Is is still possible then to link to containers which aren't named?

I believe it is, but it's difficult! You'll need to register to result of the task that creates them and muck around inside the `containers` list to extract an identifier. It'll be a list with dict items whose structures match the output of `docker inspect` -- or you can use a `debug: var=registered_thing` task to view the full thing pretty-printed.

> If it's not too much trouble a couple logical commits are best. But I know how hard that can be when you have a few commits that touch one thing and then a few commits that touch something else. I think the ideal would be one commit for the state reorganization + pull option. And a second one for the tls stuff.

I have it down to four: one for the docs; one for the pull option; one for the state reorganization; and one for TLS. There may still be a little crossover between them.

Also, I should probably run it through my testing playbook one more time to make sure I didn't muck up the rebase, but `git diff` against the old ref and a quick eye-scan didn't reveal any extraneous changes.

MrMMorris:
> I would be cautious about deviating from Ansible convention to do this, although I see what you're saying.

Yea I quickly realized I could just pass the secrets straight to the `env` declaration instead of through env vars on the host. I was stuck in the non-ansible headspace :wink: It may not make sense to mirror docker-compose.yml and I don't think we lose much by not going that route.

smashwilson:
@abadger Verified that my testing playbook is still green:

```
< PLAY RECAP >
 ------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||


localhost                  : ok=73   changed=36   unreachable=0    failed=0 
```

abadger:
Alright, merged.  I'm going to try to make the auth parameters compatible with https://github.com/ansible/ansible-modules-core/pull/854 (maybe some renaming involved) so that even if that doesn't get committed for 1.9 we still have a compatible path forward for it.

bobrik:
Host binds always trigger reload:

``` yaml
    - name: run logstash for syslog
      docker:
        command: "some args"
        image: bobrik/logstash:1.5.0.rc1.1
        name: logstash-syslog
        state: reloaded
        memory_limit: 256MB
        restart_policy: always
        net: host
        volumes:
          - /var/log/logstash:/var/log/logstash:rw
      tags:
        - logstash
```

``` json
{
    "changed": true,
    "containers": [
        {
            "Id": "7edb17d51933c98307fbd7fae74cb29a530bfe2d1a0f5b6cb887bf278ed142a9",
            "Warnings": null
        }
    ],
    "msg": "removed 1 container, stopped 1 container, started 1 container, created 1 container.",
    "reload_reasons": "volumes (set([u'/var/log/logstash']) => set([]))",
    "summary": {
        "created": 1,
        "killed": 0,
        "pulled": 0,
        "removed": 1,
        "restarted": 0,
        "started": 1,
        "stopped": 1
    }
}
```

Is there a reason for this?

bobrik:
Exposed ports do the same thing:

``` yaml
    - name: run diplodocus for logstash
      docker:
        command: -listen 0.0.0.0:8000
        image: diplodocus:2015-02-26.1
        name: logstash-diplodocus
        memory_limit: 128MB
        docker_api_version: "{{ docker_api_version | default(None) }}"
        restart_policy: always
        state: reloaded
        net: bridge
        ports:
          - 127.0.0.1:8989:8000
        volumes:
          - /var/log/logstash:/var/log/diplodocus:ro
      tags:
        - diplodocus
```

``` json
{
    "changed": true,
    "containers": [
        {
            "Id": "506dbc9e63617ed49ec19d0a87dd7b4805d6f56f8c2e57bd2cf207fc89731a96",
            "Warnings": null
        }
    ],
    "msg": "removed 1 container, stopped 1 container, started 1 container, created 1 container.",
    "reload_reasons": "exposed_ports (set([u'8000/tcp']) => set([]))",
    "summary": {
        "created": 1,
        "killed": 0,
        "pulled": 0,
        "removed": 1,
        "restarted": 0,
        "started": 1,
        "stopped": 1
    }
}
```

Image has `EXPOSE 8000` in `Dockerfile`.

smashwilson:
@bobrik Good finds, thanks!

For the host binds, I'm guessing that I'm not normalizing the "actual" side of the comparison properly.

The exposed port is a little trickier. Because it's exposed in the Dockerfile, which Ansible doesn't have visibility into, we can't know which ports are exposed from the playbook and which are exposed from the Dockerfile. There are two directions we can go there:
1. Add an `expose: [ "8000" ]` to your task as well. The downside here is that it's redundant information, and if your task and your Dockerfile become out of sync, you'll have unnecessary reloads.
2. We can change the exposed ports comparison to a subset one, like I did for the environment. The downside here is that you lose the ability to reload when a port is exposed but should no longer be.

Actually, now that I think about it, there may be a third option: I wonder if the inspected image contains information about the exposed ports... that would be much better. I could use that for the environment and for non-binding volumes, as well.

I'll investigate in a day or so and see what I can find. :mag:

bobrik:
You can get exposed ports from docker image, at least with docker cli:

```
# docker inspect -f '{{ .ContainerConfig.ExposedPorts }}' diplodocus:2015-02-26.1
map[8000/tcp:map[]]
```

I think this info should be available from docker-py as well.

smashwilson:
Yeah! Env vars and volumes are there, too. Okay, excellent, I can use that to make reload detection much better in a few cases.

abadger:
Hmm... from some recent discussion and thoughts on a different ticket I think we probably want to be able to control whether we're doing a subset or exact state.  https://github.com/ansible/ansible-modules-core/pull/123#issuecomment-76226338

The proposal here is slightly different than #123 because it will  allow establishing what was built into the container's image and then ansible will build on top of that.  However, it will still suffer from "ansible task 1 wants to expose a port; ansible task 2 wants to set an env_var; have to change those into a single task otherwise they'll stomp on each other".

I think an additional parameter to allow the user to select one behaviour or the other would work.

bobrik:
> "ansible task 1 wants to expose a port; ansible task 2 wants to set an env_var; have to change those into a single task otherwise they'll stomp on each other".

Not sure if it make sense to change settings for docker container that was launched by another task. Can you provide a real world example where this is needed?

You also cannot have "exact state". If I set `ENV SOMETHING 8` in my `Dockerfile`, there's no way to unset that variable, as far as I know. And there could be many env variables that ansible shouldn't care about, like [this one](https://github.com/bobrik/docker-logstash/blob/2fb8c7d1246480e40636c015e90b577dcfe1d9ff/Dockerfile#L5). I bet you can't un-expose port too.

bobrik:
@smashwilson looks like this is incorrect, volumes should include binds:

``` python
            # VOLUMES
            # not including bind modes.

            expected_volume_keys = set()
            if self.volumes:
                for key, config in self.volumes.iteritems():
                    if not config and key not in self.binds:
                        expected_volume_keys.add(key)
            actual_volume_keys = set((container['Config']['Volumes'] or {}).keys())
```

This works:

``` python
            # VOLUMES

            expected_volume_keys = set(self.volumes.keys())
            actual_volume_keys = set((container['Config']['Volumes'] or {}).keys())
```

smashwilson:
> Hmm... from some recent discussion and thoughts on a different ticket I think we probably want to be able to control whether we're doing a subset or exact state. #123 (comment)
> 
> The proposal here is slightly different than #123 because it will allow establishing what was built into the container's image and then ansible will build on top of that. However, it will still suffer from "ansible task 1 wants to expose a port; ansible task 2 wants to set an env_var; have to change those into a single task otherwise they'll stomp on each other".
> 
> I think an additional parameter to allow the user to select one behaviour or the other would work.

That feels strange to me. It comes back to something I [alluded to way back in the comment thread](https://github.com/ansible/ansible-modules-core/pull/502#issuecomment-67639990): an Ansible task can be either a desired state to assert, or a specific action to perform. Comparing with a subset operator (like env does now) feels like an imperfect middle ground: it puts us somewhere in between saying "add these environment variables" and "ensure the container's environment is exactly this."

It's especially awkward because Docker itself treats a launched container's configuration as immutable: you can't mount new volumes, expose new ports, or mutate its environment.

To me, `state=started` and the others let you use the module as an action -- "start this container" -- while `state=reloaded` lets you treat it as a state -- "ensure the container has exactly this configuration." Personally, I'd rather keep `reloaded` as close to that as possible.

bobrik:
Below is the patch that fixes both of my problems. Feel free to submit PR with these changes if the patch makes sense.

``` diff
--- /tmp/docker.py  2015-03-08 19:34:42.000000000 +0300
+++ ../library/docker/docker.py 2015-03-08 19:45:58.000000000 +0300
@@ -725,16 +725,18 @@ class DockerManager(object):
         return False

     def get_inspect_image(self):
+        return self.client.inspect_image(self.module.params.get('image'))
+
+    def get_image_repo_tags(self):
         image, tag = get_split_image_tag(self.module.params.get('image'))
         if tag is None:
             tag = 'latest'
         resource = '%s:%s' % (image, tag)

-        matching_image = None
         for image in self.client.images(name=image):
             if resource in image.get('RepoTags', []):
-                matching_image = image
-        return matching_image
+                return image['RepoTags']
+        return None

     def get_inspect_containers(self, containers):
         inspect = []
@@ -796,13 +798,22 @@ class DockerManager(object):
                     continue

             # EXPOSED PORTS
-            # Note that ports that are bound at container run are also exposed
-            # implicitly.
-            expected_exposed_ports = set()
+
+            image_exposed_ports = set((image["ContainerConfig"]["ExposedPorts"] or {}).keys())
+
+            expected_exposed_ports = set(image_exposed_ports)
             for p in (self.exposed_ports or []):
                 expected_exposed_ports.add("/".join(p))

-            actually_exposed_ports = set((container["Config"]["ExposedPorts"] or {}).keys())
+            if self.port_bindings:
+                for container_port in self.port_bindings.keys():
+                    if isinstance(container_port, int):
+                        container_port = "{}/tcp".format(container_port)
+
+                    expected_exposed_ports.add(container_port)
+
+            container_exposed_ports = set((container["Config"]["ExposedPorts"] or {}).keys())
+            actually_exposed_ports = image_exposed_ports.union(container_exposed_ports)

             if actually_exposed_ports != expected_exposed_ports:
                 self.reload_reasons.append('exposed_ports ({} => {})'.format(actually_exposed_ports, expected_exposed_ports))
@@ -810,13 +821,8 @@ class DockerManager(object):
                 continue

             # VOLUMES
-            # not including bind modes.

-            expected_volume_keys = set()
-            if self.volumes:
-                for key, config in self.volumes.iteritems():
-                    if not config and key not in self.binds:
-                        expected_volume_keys.add(key)
+            expected_volume_keys = set((self.volumes or {}).keys())
             actual_volume_keys = set((container['Config']['Volumes'] or {}).keys())

             if actual_volume_keys != expected_volume_keys:
@@ -1032,7 +1038,7 @@ class DockerManager(object):
         # that map to the same Docker image.
         inspected = self.get_inspect_image()
         if inspected:
-            images = inspected.get('RepoTags', [])
+            images = self.get_image_repo_tags()
         else:
             image, tag = get_split_image_tag(self.module.params.get('image'))
             images = [':'.join([image, tag])]

```

smashwilson:
> Below is the patch that fixes both of my problems.

Beat me to it! Thanks.

> Feel free to submit PR with these changes if the patch makes sense.

Why not submit your own PR? You have the changes right there, and that way you get credit :wink:

bobrik:
I opened #910.

abadger:
> Ansible task can be either a desired state to assert, or a specific action to perform. 

So the thing is -- I don't think seeing reloaded as a desired state is in conflict with treating these as additive (and unfortunately, we need subtractive as well).  it's about how fine-grained the state is, not about whether it's an action or a state.  We're asserting that in docker container foo, the state of envX=Y.  If we run the playbook multiple times one right after the other, the first time, there may be a change if the env var is set/modified to bring it in line with the playbook.  On the subsequent runs there should be no change because the state of that env var is already correctly set.

> It's especially awkward because Docker itself treats a launched container's configuration as immutable: you can't mount new volumes, expose new ports, or mutate its environment.

Now this may be a problem but I'm not sure.  Say that we have additive env var handling (like we do now).  If we have two subsequent tasks in the playbook that first add X=Y and then add Z=A what happens?  I think that the docker container's env gets updated and restarted twice.  Then on subsequent runs the docker container does not get updated or restarted at all.  Is this correct?  If so, then I think it's acceptable.  And seems like the only way to enable the "container needs the union of configs between two logical groups" situation.  the user may want to optimize their playbook so that on the initial run the container is only updated and restarted once but that's an optimization.

If that's not correct then please correct my understanding :-)

smashwilson:
> Say that we have additive env var handling (like we do now). If we have two subsequent tasks in the playbook that first add X=Y and then add Z=A what happens? I think that the docker container's env gets updated and restarted twice. Then on subsequent runs the docker container does not get updated or restarted at all. Is this correct?

Not quite. What will happen is that the task that sets X=Y will trigger a container restart, with the environment variable X=Y; then the task that sets Z=A will also trigger a container restart, which will lose X=Y entirely and launch a new container with Z=A but not X=Y. Subsequent runs will always restart the container twice the same way.

The subset comparison for env right now wasn't meant to support additivity, it was a hack to work around not having #912 yet - I was hitting situations where there were extra environment variables in the container that weren't in the module parameters.

The Docker API (and therefore the `docker` client and docker-py) doesn't support modifications to existing containers at all. You really can't do something like "add an environment variable to a running container", or even map a new port, after the container is running. To support that we'd need to fake it with restarts, and like I said it would feel awkward to me as a Docker user.

abadger:
So if we can fake it with multiple restarts (within the play, subsequent runs of the same play will not restart again) then I think we will need to do that -- but we could have a parameter to control that.  So it won't block #912.  Reviewing the code once more, the reason that the current code would flip-flop on env vars with two tasks is not the issuperset() comparison in get_differing_containers(self), it's that we set  environment to self.env in create_containers() right?  So to create such a parameter we'd need to:

if ansible_param says to add to env, when we restart_container(): save current container's env.  stop_container(). create_container(saved_env).  container.create( union(self.env, saved_env)).  And same for ports, volumes, etc.

So a good deal of work.  Unfortunately, it will probably hit a number of people who are treating containers as lightweight virtual machines.  It's common to configure these aspects of hosts in different tasks depending on the groups that the machines are in  (as in my example, you could need to open a port for access to a vpn if you're in one datacenter and set an environment  var for an http_proxy if you are running certain applications).  If no one mentions it during 1.9rc testing, I probably won't work on it until v2.  But if someone does I guess we'll need to implement it before 1.9 final.

MrMMorris:
@abadger @smashwilson I'm back with what seems like a bug in `pull: always`. Here is the task:

```
- name: run node client container
  docker:
    name: "client-{{ item }}"
    image: "company/client:{{ item }}-latest"
    links:
      - "server-{{ item }}"
    env: "SERVICE_NAME=client,SERVICE_TAGS={{ item }}"
    ports: 80
    restart_policy: always
    pull: always
    state: started
  with_items: node_env
  register: client
```

I have ansible-pull running this task every 5 minutes. My CI build pushes to docker hub and ansible-pull runs which tells me that this task has changed. However, it looks like all it is doing is pulling the latest image and then that somehow results in the image name reverting to the ID and now new image is run.

Before ansible-pull:

```
CONTAINER ID        IMAGE                      COMMAND                CREATED             STATUS              PORTS                            NAMES
0796a5c62056        company/client:dev-latest  "nginx -g 'daemon of   31 minutes ago      Up 12 minutes       443/tcp, 0.0.0.0:49162->80/tcp   client-dev
```

After:

```
CONTAINER ID        IMAGE              COMMAND                CREATED             STATUS              PORTS                            NAMES
0796a5c62056        87ae5a963313       "nginx -g 'daemon of   35 minutes ago      Up 16 minutes       443/tcp, 0.0.0.0:49162->80/tcp   client-dev
```

Do I have something wrong in my task? or is this a bug

MrMMorris:
Nevermind, I assume I need to change the state to `reloaded` :smile:

smashwilson:
> Nevermind, I assume I need to change the state to reloaded

Bingo :grin:

MrMMorris:
@smashwilson I may have hit a similar issue, but the problem does not seem as obvious to me. Looks like the newer container is run but the old ones stay up:

Before deploy:

```
CONTAINER ID        IMAGE                                  COMMAND                CREATED             STATUS              PORTS                            NAMES
05ef9af927ee        company/tv-client:dev-latest           "nginx -g 'daemon of   17 minutes ago      Up 17 minutes       443/tcp, 0.0.0.0:49202->80/tcp   clever_pike
0bdeb3ef8f37        company/tv-client:dev-latest           "nginx -g 'daemon of   17 minutes ago      Up 17 minutes       443/tcp, 0.0.0.0:49201->80/tcp   lonely_nobel
```

After deploy:

```
CONTAINER ID        IMAGE                                  COMMAND                CREATED             STATUS              PORTS                            NAMES
eda5d790bbc1        company/tv-client:dev-latest           "nginx -g 'daemon of   5 seconds ago       Up 5 seconds        443/tcp, 0.0.0.0:49204->80/tcp   serene_blackwell
fce2c32876eb        company/tv-client:dev-latest           "nginx -g 'daemon of   6 seconds ago       Up 5 seconds        443/tcp, 0.0.0.0:49203->80/tcp   agitated_lumiere
05ef9af927ee        147b87b45e35                           "nginx -g 'daemon of   20 minutes ago      Up 20 minutes       443/tcp, 0.0.0.0:49202->80/tcp   clever_pike
0bdeb3ef8f37        147b87b45e35                           "nginx -g 'daemon of   20 minutes ago      Up 20 minutes       443/tcp, 0.0.0.0:49201->80/tcp   lonely_nobel
```

Code:

```
- name: run node client container
  docker:
    image: "company/tv-client:dev-latest"
    env: "SERVICE_NAME=client,SERVICE_TAGS=dev"
    ports: 80
    restart_policy: always
    pull: always
    state: reloaded
    count: 2
```

I just added the `count: 2` and removed the container naming so I assume it's something to do with that.

Any ideas?

smashwilson:
@MrMMorris Ah. I'm guessing that this is related to the module's identification of containers to target based on image alone. After the pull completes, the containers that were already running no longer have an image in the `repo_tags` set, because the tag has moved, so they no longer appear in `get_deployed_containers()`, and the module thinks that nothing is running.

Can you open a new issue for this, and cc me on it? I should be able to have a look tomorrow.

djdefi:
+1 for the issue @MrMMorris is describing.

MrMMorris:
@rtrauntvein see https://github.com/ansible/ansible-modules-core/pull/993

You will have to use the devel branch for wait till 1.9.1

smashwilson:
@rtrauntvein As another workaround, I believe you can do something like this:

``` yaml
- name: launch four containers
  docker:
    image: whatever
    command: sleep 1d
    name: whatever_{{ item }}
    pull: always
    state: reloaded
 with_sequence: count=4
```

(I should put this in the other issue, too, in case anyone finds it there.)

Issue #388, opened 2014-11-25
Subversion: Export fails because `svn info` is called after export


dailyherold:
I have a task that does the following:
- name: Export latest templates from SVN
  subversion: repo=https://repo/path/to/templates
              dest=/opt/templates
              export=True
              username=user
              password=pass

If the destination folder **does not** exist then the task fails as I get the following msg:
`svn: '/opt/templates' is not a working copy`
Looking at the failed command output by Ansible I see:
`"cmd": "/usr/bin/svn --non-interactive --trust-server-cert --no-auth-cache --username user --password ******** info /opt/templates"`
Confirming in the source code I find on [line 225](https://github.com/ansible/ansible-modules-core/blob/devel/source_control/subversion.py#L225) that `get_revision()` will be run no matter if export is true or false. [`get_revision()`](https://github.com/ansible/ansible-modules-core/blob/devel/source_control/subversion.py#L141) calls `svn info` which will fail if `export=True` since there is no `.svn` file. 

**Note:** The repo path _is_ in fact exported on the remote machine despite this message. So the export action itself works, it is just the call to `get_revision()` that fails.

Any further attempts to run fail as well, and I get the same `ERROR: /opt/templates folder already exists, but its not a subversion repository.` as the existing issue #370.

abadger:
I just merged https://github.com/ansible/ansible-modules-core/pull/506 which should fix this issue as well.  Please open a new bug if the problem still occurs with the new version of the module.

PR #506, opened 2014-12-13
Fix #370

### LINKED ISSUE #370: subversion: export fails if dest directory exists
Using the following syntax, the task always fails if /tmp/foo already exists.

`- name: (testing) svn export of desired callflow.xml
      subversion: repo=http://repo.local/repos/target/ dest=/tmp/foo/ export=yes`

The error shown is:
`msg: ERROR: /tmp/foo/ folder already exists, but its not a subversion repository.`

This appears to be happening because the subversion module does not account for exporting if the target directory exists (but is not a working copy):

`if not os.path.exists(dest):`
...export and checkout both handled
`elif os.path.exists("%s/.svn" % (dest, )):`
...checkout handled. export presumably not handled because this is a working directory
`else:
        module.fail_json(msg="ERROR: %s folder already exists, but its not a subversion repository." % (dest, ))`

I think if we insert another case before the `else` of `elif os.path.exists(dest) and export:` and handle export to a preexisting directory there, it should work.

mscherer:
Use the currently ignored and unused force parameters to explicitly say we want to push over a existing repository, following subversion convention. Fix issue #370

### DIFF
--- a/source_control/subversion.py
+++ b/source_control/subversion.py
@@ -199,7 +199,7 @@ def main():
     os.environ['LANG'] = 'C'
     svn = Subversion(module, dest, repo, revision, username, password, svn_path)
 
-    if not os.path.exists(dest):
+    if export or not os.path.exists(dest):
         before = None
         local_mods = False
         if module.check_mode:
@@ -227,9 +227,12 @@ def main():
     else:
         module.fail_json(msg="ERROR: %s folder already exists, but its not a subversion repository." % (dest, ))
 
-    after = svn.get_revision()
-    changed = before != after or local_mods
-    module.exit_json(changed=changed, before=before, after=after)
+    if export:
+        module.exit_json(changed=True)
+    else:
+        after = svn.get_revision()
+        changed = before != after or local_mods
+        module.exit_json(changed=changed, before=before, after=after)
 
 # import module snippets
 from ansible.module_utils.basic import *
--- a/source_control/subversion.py
+++ b/source_control/subversion.py
@@ -123,7 +123,12 @@ class Subversion(object):
 		
     def export(self, force=False):
         '''Export svn repo to directory'''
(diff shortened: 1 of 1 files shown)


abadger:
This looks good to me except for changing the default value of force.  Not sure what to do there -- force is not currently used in combination with export so changing the default to no would keep the same behaviour for export.  But force is currently used when updating an already existing repository.  So if we changed the default value we could cause errors for some people.

Maybe we should change the default value and just document that the default has changed?  (Both in the module documentation and the ansible/ansible/ Changelog file?)

What's your opinion?

abadger:
This should also fix https://github.com/ansible/ansible-modules-core/issues/388

mscherer:
Indeed, didn't see force was used later in the code. So we could have a different default for each command ? It would be a bit hard to explain in the doc however.

abadger:
I agree, that seems to be hard to document.

justintaft:
+1 for leaving force to false for export.

If wanting to change the default value of force to  `true`, add a warning message to the build process so people KNOW they should explicitly set the default, because it will change. Later on, make the actual change to `true`.

Alternatively, make the force field required in the future, and not have a default value.

For documentation

```
no for export, yes otherwise
```

isn't too confusing. It's better then breaking people's builds.

abadger:
@mscherer: jimi-c, bcoca, and I talked it over and we're good to change the default of force to false as long as the change gets documented (so modify ChangeLog in the ansible/ansible repo to explain the effect on the subversion checkout functionality.) We'll try to review the other modules for the next major release and change other instances where force defaults to True to default to False as a safety measure as well.

abadger:
Thanks @mscherer, I'vve merged this and will make sure the Changelog is updated.

# Merged

Hi!

issue is resolved.

Thank you!

PR #507, opened 2014-12-13
Fixes #468 - file. Can't set permissions on hardlinks

### LINKED ISSUE #468: Ansible 1.8.2, module: file. Can't set permissions on hardlinks.
##### Issue Type:

Bug Report
##### Ansible Version:

1.8.2
##### Environment:

Gentoo amd64, Debian wheezy amd64
##### Summary:

Can't set permissions on hardlinks using file module. follow=yes doesn't help
##### Steps To Reproduce:

file: path=<some hardlink path> owner=root group=root mode=4710
##### Expected Results:

Permissions will be set correctly
##### Actual Results:

error:
msg: src and dest are required for creating links

I think it's totally wrong cause hardlinks are also regular files which permissions and ownership should be able to be changed.

sysadmin75:
Fixes #468 

Hopefully this isn't too much of a hack.


sysadmin75:
I wasn't satisfied with this patch so I created a different one.

PR #509, opened 2014-12-14
Fix volume bindings for docker 1.4.0

ghost:
Related to 
https://github.com/docker/docker-py/issues/419
https://github.com/docker/docker/issues/9628


steveej:
Do I understand correctly that your commit disables regular volumes for docker >= 1.4.0?

ghost:
Well, it seems to be so. I think you should discard this commit. We need to check client api version instead of docker version.

steveej:
I don't think that any version checking is needed. I think that the handling of binds/volumes has been wrong in the first place. I've tried to improve it with #547.

gregdek:
Closing in favor of #547.

PR #510, opened 2014-12-14
Add insecure_registry to docker-py pull method

ghost:
Starting from docker-py>=0.5.0 it is impossible to work with private registries based on HTTP.
So we need additional parameter to allow pull from insecure registry
Related to ansible/ansible#9111

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -210,6 +210,13 @@ options:
     default: ''
     aliases: []
     version_added: "1.8"
+  insecure_registry:
+    description:
+      - Use insecure private registry by HTTP instead of HTTPS (needed for docker-py >= 0.5.0).
+    required: false
+    default: false
+    aliases: []
+    version_added: "1.8"
 
 author: Cove Schneider, Joshua Conner, Pavel Antonov
 requirements: [ "docker-py >= 0.3.0", "docker >= 0.10.0" ]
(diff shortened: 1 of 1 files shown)


abadger:
It looks like insecure_registry was not a valid parameter before docker-py 0.5.0.  Is that correct?  If so we should either guard against using it in docker-py < 0.5.0 and return an error if the parameter is specified with an old docker-py or always return an error if docker-py < 0.5.0.  Since it appears insecure_registry is not needed with docker-py < 0.5.0, the guarding case is probably more appropriate.

abadger:
I made the changes mentioned above to check that the docker-py version was recent enough and merged the changes:
- https://github.com/ansible/ansible-modules-core/commit/4172d445d3bd8699add652f3b7dc4693397f8d92
- https://github.com/ansible/ansible-modules-core/commit/f65d9ab79326a67430beac4ea4671f9a0e6e7097

Let me know if my changes break things.  Thanks for the pull request!

# Merged

Hi!

issue is resolved.

Thank you!

drbild:
@abadger Is there some reason these commits aren't included in the latest release, 1.8.4?

abadger:
@drbild they were added after 1.8.0 was out.  1.8.[1-4] are bugfix releases so this wasn't added in any of those.  We're putting out a 1.9 release soon (1.9rc1 was tagged in the repository, no tarball yet because of some things that needed work to create an official tarball using our jenkins instance).  We're going to have an rc2 around tuesday.  You can take a look at what's going into 1.9 by clonign the repository and following the instructions in this thread:

https://groups.google.com/forum/#!topic/ansible-project/iURJ7cpp5Q0

drbild:
@abadger Makes sense!  Thanks for the quick response and the link.

Issue #438, opened 2014-12-03
hostnamectl Not Used When Setting Hostname Using hostname Module On CentOS 7


jwbrown77:
##### Issue Type:

Bug Report
##### Ansible Version:

1.8.1
##### Environment:

Running from: Ubuntu 14.04.1.  Destination host: CentOS 7.0.1406
##### Summary:

CentOS 7 uses systemd and should use the "hostnamectl" utility to set the hostname rather than adding it to /etc/sysconfig/network
##### Steps To Reproduce:

```
- name: Set the hostname
  hostname: name={{ hostname }}
```
##### Expected Results:

For the hostname to be permanently set to what the playbook defines.
##### Actual Results:

It appears that it's using behavior from previous non-systemd versions of CentOS: Using the "hostname" binary to set the current hostname, then adding "HOSTNAME=xxx" to /etc/sysconfig/network in an attempt to set it permanently.  The temporary change works, but the changes are lost on reboot, as CentOS 7 doesn't seem to respect the "HOSTNAME" value in /etc/sysconfig/network.  The correct way is to use "hostnamectl set-hostname xxx"
##### My Debugging Effort:

It appears that there is a "Redhat Strategy" and a "Fedora Strategy" in the source code.  The Fedora Strategy does what's required; calling hostnamectl.  But it never gets triggered, because in lines 52-53 of hostname.py, the version is set to "0" if a string is returned from Python's platform module:

```
 if type(distribution_version) is str:
    distribution_version = 0
```

Running this:

```
import platform; type(platform.linux_distribution()[1])
```

On Ubuntu 14.04, CentOS 6, and CentOS 7 always returns a "str" type, but lines 52-53 check for "str" type and set the version to 0.  I don't know why it would set to 0 instead of attempting to cast to an integer/float where line 384-385 could use it in its comparison:

```
if distribution_version and LooseVersion(distribution_version) >= LooseVersion("7"):
    strategy_class = FedoraStrategy
```

Apologies if I'm misunderstanding the problem; I'm not a highly skilled Python programmer.

Thanks

jwbrown77:
I forked the repo to attempt a fix, but found that casting to an int doesn't work, because the "LooseVersion" class _expects_ a string, and fails if fed an int.

Actually, I wound up simply commenting out all the logic in _get_distribution_version in hostname.py and simply return the raw result of:

```
distribution_version = get_distribution_version()
```

[...]

PR #512, opened 2014-12-14
Revert commit cbc417c, as the code is broken, see #438

mscherer:
LooseVersion expect a string, so filtering to return a int
is incorrect. Thanks to Jeremy Brown for the analysis.

### DIFF
--- a/system/hostname.py
+++ b/system/hostname.py
@@ -45,18 +45,6 @@ from distutils.version import LooseVersion
 from ansible.module_utils.basic import *
 
 
-# wrap get_distribution_version in case it returns a string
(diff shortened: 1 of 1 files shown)


abadger:
Did some research and found that @neersighted created the first version of this code before the switch from float comparisons to LooseVersion but no record of why the code was subsequently changed and then merged.  I agree that it looks like it should be reverted as using LooseVersion for the comparisons should take care of the initial problem of trying to turn Debian unstable versions (jessie/sid) into a float.

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

PR #513, opened 2014-12-14
Allow globbing in creates= and removes= directives

fgsch:
Fixes 1904

### DIFF
--- a/commands/command.py
+++ b/commands/command.py
@@ -20,6 +20,7 @@
 
 import sys
 import datetime
+import glob
 import traceback
 import re
 import shlex
@@ -161,7 +162,7 @@ def main():
         # and the filename already exists.  This allows idempotence
         # of command executions.
         v = os.path.expanduser(creates)
-        if os.path.exists(v):
+        if glob.glob(v):
             module.exit_json(
                 cmd=args,
                 stdout="skipped, since %s exists" % v,
@@ -175,7 +176,7 @@ def main():
     # and the filename does not exist.  This allows idempotence
     # of command executions.
         v = os.path.expanduser(removes)
-        if not os.path.exists(v):
+        if not glob.glob(v):
             module.exit_json(
                 cmd=args,
                 stdout="skipped, since %s does not exist" % v,
--- a/commands/command.py
+++ b/commands/command.py
@@ -47,12 +47,12 @@ options:
     aliases: []
   creates:
     description:
-      - a filename, when it already exists, this step will B(not) be run.
+      - a filename or glob pattern, when it already exists, this step will B(not) be run.
     required: no
     default: null
   removes:
     description:
-      - a filename, when it does not exist, this step will B(not) be run.
+      - a filename or glob pattern, when it does not exist, this step will B(not) be run.
     version_added: "0.8"
     required: no
     default: null


abadger:
https://github.com/ansible/ansible/issues/1904

Did this get discussed on the mailing list?  I'm not so certain that globbing is a good idea here as globs may pick up more than intended.  On the other hand, it is just a tool that the user can use so if it was discussed already and a decision was reached that globbing was fine, the code itself looks good.

fgsch:
@abadger I don't know if it was discussed. I do however have the need for such feature and searching around took me to https://github.com/ansible/ansible/issues/1904.

I appreciate this could pickup multiple files but whether is more than intended depends on how the globbing is constructed and what you want to do. The alternative would be to use regex but I found globbing more natural in this context and something easier to understand for most users.

gildegoma:
:+1: It works for me (I tested globbing with `creates=` on a patched v1.8.2). Thanks @fgsch :) 

Note: the documentation should be adapted according to this addition.

fgsch:
@gildegoma Documentation updated. Thanks for reminding me.

gildegoma:
@fgsch thanks for quick reply, you rock :heart: !

abadger:
Thanks for all the work on this -- the code changes look fine to me.  If someone cares to go the final mile on this, I'd just like someone to quickly post about this change to the mailing list.  If there are no objections to using globs in response, then I'll be happy to merge this.  Thanks!

bcoca:
there are several modules that support creates=, they should all behave the same

gildegoma:
> there are several modules that support creates=, they should all behave the same

@bcoca thanks for raising that point. I also thought about that, but prefered not to ask to keep the PR scope as simple as possible. That said, I also think that we better should align these `creates|removes` behaviours (except if does not make sense). A few thoughts about that point:
- I think that the [uri module](https://github.com/ansible/ansible-modules-core/blob/9c6826e9286f2e683c583ff11ccd562bfb5eed8c/network/basics/uri.py#L378-L391) could be adapted the same way as for `command`.
- `shell` uses `command` behind the scene
- it seems that in `ansible/ansible` main git repository there are some `runner/action_plugins` bits that should be updated: 
  - [unarchive](https://github.com/ansible/ansible/blob/4dfd86d8478d36f1e774c8770a0a8eb610d9ffb1/lib/ansible/runner/action_plugins/unarchive.py#L58-L75), which provides [`creates`](http://docs.ansible.com/unarchive_module.html) option, but not `removes`.
  - [script](https://github.com/ansible/ansible/blob/4dfd86d8478d36f1e774c8770a0a8eb610d9ffb1/lib/ansible/runner/action_plugins/script.py#L56-L89), which provides both `creates` and `removes`.

I don't know exactly how these `action_plugins` are used in ansible (module) guts, but otherwise I don't think there is something else to adapt.

I am not sure if it really makes sense to add globbing to module like `unarchive`, but following a "who can do more, can do less" principle, I don't see any drawbacks to extend its behaviour as well.

bcoca:
I'm not asking to add support to all in this PR, but we'll need to address it with other updates, centralizing the functionality into module common code might be a good idea.

fgsch:
@bcoca I did review all the creates and removes before I submitted this.

I opened this PR to specifically address #1904 - I didn't want to add things outside that issue to make it easier to review and/or accept.

As for the modules that support creates and removes directives:
- `command`, `shell` and `script` are covered by this PR
- `uri`: I was not entirely sure globbing made sense here. My reasoning was that if you are fetching something you know the full name of the resource but I realiase now there are other cases that didn't think at the time.
- `unarchive`: Can't remember about this one TBH.

I'm happy to review the other modules when this is accepted and ensure the behaviour is the same across all of them.

fgsch:
@bcoca Is there anything else missing here that would help moving things forward?
I'm happy to extend and/or look more into this if needed.

bcoca:
no, mostly time, we area about to start v2 switch so we probably won't get back to this until that is done.

gregdek:
So looks like @abadger gave an OK on this module on 1/7. Are we ok accepting this narrow patch, and what should we do to ensure we're checking globbing behavior elsewhere?

Promoting to "P1_reviewed" so we can push a decision on this.

abadger:
I think I'm okay to merge this.  Anyone want to advance an option for checking the globbing behaviour elsewhere?  (Maybe someone would like to open bugs for the modules mentioned in https://github.com/ansible/ansible-modules-core/pull/513#issuecomment-69130331 ?)

fgsch:
As I said before I'm happy to look after the other modules once there is a final decision :wink:

gregdek:
OK. Recommend we merge, and @fgsch feel free to open other PRs as needed and reference this one so we remember the rationale. Thanks!

abadger:
So @jimi-c had the idea that instead of adding this code to each module we could add it to basic.py and then adding a parameter when instantiating an AnsibleModule().  Something like `supports_create`.  This would be similar to what we do with `add_file_common_args`.

The code to implement `supports_create` would go here: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L338

You can look at how `add_file_common_args` is implemented there for ideas.

Then we'd need a method that expands the checks whether the file specified by creates and removes already exists/doesn't exist.  It would be analogous to `load_file_common_arguments()` in basic.py
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L398  Maybe call it something like `check_creates_common_arguments()`

Modules like command.py would then change.  Instead of specifying `creates` and `removes` on their own they would add `supports_create=True` to the AnsibleModule() constructor.  Instead of checking for the file specified by creates/remove themselves they'd call `module.check_creates_common_arguments()` which would do the checking for them and then we could module.exit_json() depending on whether `check_creates_common_arguments()` found that the file existed/did not exist.

We'd also want to make it easy to documentat creates and removes so we'd want the documentation to go in a fragment in  here: https://github.com/ansible/ansible/tree/devel/lib/ansible/utils/module_docs_fragments  (you can look at how files.py is implemented again for an example).  Then modules can use `extends_documentation_fragment: create`  to add documentation for those arguments.

This is a more ambitious implementation but should make it easier to generalize to all of the modules that use it.

@fgsch are you up for this?  I think I've laid out every place in the code that needs to be updated to do things this way and you have most of the algorithm to do it in command.py... it would just need someone to put all the pieces into their proper place to make it work.

abadger:
@fgsch We talked a little more and also decided if you don't have the time at the moment,  we could also merge this PR now and make it general when someone wants to push the changes to the other modules.... after all, the creates/removes feature is currently a part of each individual module so we aren't adding more maintainance burden to do this... we just want to subtract some of that burden going forward.

gregdek:
@abadger just so I understand: how does that impact this PR?
1. This PR should be merged, and ultimately replaced when the proposed changes around "supports_create" are implemented?

Or 2. This PR should be tabled, and we should move to "supports_create" immediately?

abadger:
@gregdek 2 is the ideal but depending on how @fgsch is feeling about making those changes at this moment in time 1 is perfectly acceptable.

fgsch:
@abadger I'm happy to take a look but considering how long this been outstanding I'd prefer if this PR is merged and then I look into changing the implementation to the more generic approach.

My fear is that a new PR / diff will take another 6 months to be reviewed and people wanting this feature will have to wait even longer.

abadger:
@fgsch Definitely understandable!  Sorry for the long review time and we'll try to do better with the next one as we have more of a design plan outlined for it.

# Merged

Hi!

issue is resolved.

Thank you!

PR #524, opened 2014-12-16
re-enable AIX password setting

zitterbacke:
the AIX class uses a unsafe shell for setting the user password (containing a pipe in the command). This patch adopts to the new behavior of module_utils/basic.py (since somewhere around 1.7).

besides it changes the qoutes for the echo command from double to single, because password-hashes contain $-signs and one would not have this variables expanded.

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -262,12 +262,12 @@ class User(object):
         # select whether we dump additional debug info through syslog
         self.syslogging = False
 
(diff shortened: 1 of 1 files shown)


abadger:
Looks good.  Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #531, opened 2014-12-17
fixes #529 ec2_group module bug

### LINKED ISSUE #529: ec2_group module failures with ICMP protocol
##### Issue Type:

Bug Report
##### Ansible Version:

1.8.2
##### Environment:

localhost / OSX / *
##### Summary:

The `ec2_group` module fails on ICMP protocol operations that should normally be unchanged. That is, the first time the module is called to create and/or change a group, the operation succeeds.  The second and subsequent times the module is called, where the operation should normally be idempotent, the module instead fails.
##### Steps To Reproduce:

Call the `ec2_group` module twice in a row where one of the `rules` **or** `egress_rules` has `proto=ICMP`.
##### Expected Results:

According to the [boto documentation](http://boto.readthedocs.org/en/latest/security_groups.html), the correct way to allow all ICMP types is to use the range -1:-1.  I would expect the following task to idempotently allow the instance to respond to all ICMP types

```
- ec2_group:
    name: IcmpAllInboundSG
    description: respond to all inbound ICMP traffic
    region: us-west-2
    rules:
      - proto: ICMP
        from_port: -1
        to_port: -1
        cidr_ip: 0.0.0.0/0
```

Likewise, I would expect the following task to idempotently allow the instance to respond only to traceroute requests

```
- ec2_group:
    name: EX2
    description: respond to all inbound traceroute requests
    region: us-west-2
    rules:
      - proto: ICMP
        from_port: 30
        to_port: 30
        cidr_ip: 0.0.0.0/0
```

And likewise again, only to ping requests

```
- ec2_group:
    name: EX3
    description: respond to all inbound ping requests
    region: us-west-2
    rules:
      - proto: ICMP
        from_port: 0
        to_port: 0
        cidr_ip: 0.0.0.0/0
```
##### Actual Results:

In the above three examples, the first time the operation is ran (and thus the results "changed"), the operation succeeds.  In subsequent playbook runs, the operation fails with the following output.

```
TASK: [security group to allow ping] ******************************************
  File "/Users/kfranz/.ansible/tmp/ansible-tmp-1418809244.68-137009819118878/ec2_group", line 2145, in <module>
    main()
  File "/Users/kfranz/.ansible/tmp/ansible-tmp-1418809244.68-137009819118878/ec2_group", line 293, in main
    group.authorize(rule['proto'], rule['from_port'], rule['to_port'], ip, grantGroup)
  File "/Library/Python/2.7/site-packages/boto-2.34.0-py2.7.egg/boto/ec2/securitygroup.py", line 203, in authorize
    dry_run=dry_run)
  File "/Library/Python/2.7/site-packages/boto-2.34.0-py2.7.egg/boto/ec2/connection.py", line 3192, in authorize_security_group
    params, verb='POST')
  File "/Library/Python/2.7/site-packages/boto-2.34.0-py2.7.egg/boto/connection.py", line 1223, in get_status
<Response><Errors><Error><Code>InvalidPermission.Duplicate</Code><Message>the specified rule "peer: 0.0.0.0/0, ICMP, type: 30, code: 30, ALLOW" already exists</Message></Error></Errors><RequestID>0e8274ad-7884-4050-adca-5c742fe0c697</RequestID></Response>


```

kalefranz:
The problem was ultimately various capitalization mismatches between the rules lookup dictionary keys.  The least-invasive fix I think was to factor out the key creation into a function, and then make sure it's consistent for all possible inputs.

As it turns out, using a lowercase protocol (`icmp` in my case), instead of uppercase in the playbook also fixes the problem.  But boto accepts both upper and lower case so ansible should too.

fix for #529

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -114,11 +114,21 @@ except ImportError:
     sys.exit(1)
 
 
+def make_rule_key(prefix, rule, group_id, cidr_ip):
+    """Creates a unique key for an individual group rule"""
+    if isinstance(rule, dict):
+        proto, from_port, to_port = (rule.get(x, None) for x in ('proto', 'from_port', 'to_port'))
+    else:  # isinstance boto.ec2.securitygroup.IPPermissions
+        proto, from_port, to_port = (getattr(rule, x, None) for x in ('ip_protocol', 'from_port', 'to_port'))
+
+    key = "%s-%s-%s-%s-%s-%s" % (prefix, proto, from_port, to_port, group_id, cidr_ip)
+    return key.lower().replace('-none', '-None')
+
+
 def addRulesToLookup(rules, prefix, dict):
     for rule in rules:
         for grant in rule.grants:
-            dict["%s-%s-%s-%s-%s-%s" % (prefix, rule.ip_protocol, rule.from_port, rule.to_port,
-                                        grant.group_id, grant.cidr_ip)] = rule
+            dict[make_rule_key(prefix, rule, grant.group_id, grant.cidr_ip)] = rule
 
 
 def get_target_from_rule(module, ec2, rule, name, group, groups, vpc_id):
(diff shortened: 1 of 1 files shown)


juliedavila:
Tested, it works.

PR #532, opened 2014-12-17
Add warning to docs about parallel use of add_host

quentinsf:
add_host can behave unpredictably (at least from a user's point of view) when run on multiple machines simultaneously - not all instances will succeed in making a change that persists into the future.

After spending many hours yesterday discovering this the hard way, I thought adding a warning note to the documentation was worthwhile, to the effect that 'serial:1' is a good idea.

### DIFF
--- a/inventory/add_host.py
+++ b/inventory/add_host.py
@@ -7,6 +7,9 @@ short_description: add a host (and alternatively a group) to the ansible-playboo
 description:
   - Use variables to create new hosts and groups in inventory for use in later plays of the same playbook. 
     Takes variables so you can define the new hosts more fully.
+notes:
+  - Behaviour may be unpredictable if run on multiple hosts simultaneously. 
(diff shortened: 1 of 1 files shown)


bcoca:
# Is This A Bug?

Hi!

We're not sure this is a bug, and we don't mean for this to be confrontational.  Let's explain what we're thinking:
- add_host is non persistent, it only adds hosts to the in memory inventory for the current run, also it bypasses the host loop so it does NOT run on multiple hosts. Due to it's nature it makes little sense to run other than on the master.

As such, we're going to close this ticket.  However, we're open to being corrected, should you wish to discuss.  You can stop by one of our two mailing lists 
to talk about this and we might be persuaded otherwise.

Comments on closed tickets aren't something we monitor, so if you do disagree with this, a mailing list thread is probably appropriate.

sivel:
There is another documentation PR that tries to address this a bit more correctly at https://github.com/ansible/ansible-modules-core/pull/104

PR #533, opened 2014-12-17
Updating rubygems --no-document param

thataustin:
`--no-rdoc` and `--no-ri` are deprecated: http://guides.rubygems.org/command-reference/#gem-install

### DIFF
--- a/packaging/language/gem.py
+++ b/packaging/language/gem.py
@@ -182,8 +182,7 @@ def install(module):
         cmd.append('--no-user-install')
     if module.params['pre_release']:
         cmd.append('--pre')
-    cmd.append('--no-rdoc')
-    cmd.append('--no-ri')
+    cmd.append('--no-document')
     cmd.append(module.params['gem_source'])
     module.run_command(cmd, check_rc=True)
 
--- a/packaging/language/gem.py
+++ b/packaging/language/gem.py
@@ -84,7 +84,7 @@ options:
       - Allow adding build flags for gem compilation
     required: false
     version_added: "2.0"
-author: 
+author:
     - "Ansible Core Team"
     - "Johan Wiren"
 '''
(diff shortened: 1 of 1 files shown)


bcoca:
how long has --no-document option existed? would it make sense to check the gems version before using one or the other?

thataustin:
@drbrain, I'm going to ping you on this one.  I believe you added the `--[no-]document` option a couple of years back, no?  Would you say it's safe to switch exclusively to `--no-document` by now?

thataustin:
Trying to dig in a little to refresh us all... It looks like it was added in https://github.com/rubygems/rubygems/commit/589309d46e619fcac0f192cdae7a89ed29a8c5c2 and made it's first appearance for end-users in the 2.0 release.  I'm not a huge ruby-ista, so I don't know if most people would be beyond 2.0 by now.

gregdek:
@thataustin no response from @drbrain... so do we have an opinion on this?

If you think this patch still applies, please rebase and we will review.

thataustin:
@gregdek -  I honestly still don't know what versions of rubygems people use.  I'd assume it would be fine given that the 2.0 release was on Feb 24, 2013.  I'd also assume that the sorts of people using ansible would probably be fine to update their ruby gems install in the rare case they are that far behind.

But I'd trust your experience on these matters.  I've not maintained a heavily depended-upon project like this one before.

drbrain:
Sorry, I missed the original message. Most people are on RubyGens 2.0 or newer now.

gregdek:
Not having a lot of knowledge here :) I would go back to @bcoca comment. Because this is essentially choosing arguments for the user, seems like the sensible thing would be to check what version of gems is being used, right? That's the safest route: choose the arguments that you know are supported for the appropriate gems version, which can be inferred.  Does that work, or no?

bcoca:
Actually, there are plenty of people on 1.8/1.9 versions, I personally know of a few that have not migrated to 2.0+ yet. Not everyone upgrades to the latest  _cough_ python3 _cough_ and we have to deal with legacy targets, so I ask you again to try detecting the version before using the new flag.

thataustin:
@bcoca - thanks for the feedback!  Knowing that it's not just considered archaic is quite helpful.  And you're right @gregdek, that does make the most sense.  My only hesitancy was just that I didn't want to add any cruft if it was already out of date.  I'll look into adding some version detection.

gregdek:
Looks like the core team currently owns this module, actually. We should consider finding a community maintainer. But since we don't have one right now: cc @bcoca @abadger @jimi-c @bennojoy

gregdek:
@bcoca why is this in needs_revision? Looks like the requested version detection was added in bb816f0?

PR #534, opened 2014-12-17
ec2_ami can update an ami's launch_permissions

ferrants:


juliedavila:
So one issue I just discovered. Using the following task:

``` yaml
---
 - name: Testing ec2_ami
   ec2_ami:
     aws_access_key: "{{ aws_access }}"
     aws_secret_key: "{{ aws_secret }}"
     instance_id: i-67086448
     name: Testing PR534 II
     region: us-east-1
     state: present
     launch_permissions:
        user_ids: [ jdavila, otheruser ]
        group_names: [TestingGroup, SolArch]
   register: stuff
```

Which the errs out with following:

```
failed: [localhost] => {"failed": true, "image_id": "ami-8e3f50e6"}
msg: InvalidAMIAttributeItemValue: Invalid attribute item value "TestingGroup, SolArch " for UserGroup item type.

```

This is expected as I forgot to add quotes. However, it still proceeded to create the AMI which should not happen.

I'll resume testing once it's been tweaked accordingly.

juliedavila:
@ferrants are you still working on this?

ferrants:
@defionscode - I haven't touched it, I will this week.

ferrants:
@defionscode - Your use is incorrect, I have updated the documentation in this PR.

The only valid group, for now, is "all". See: https://boto.readthedocs.org/en/2.0/ec2_tut.html#images-instances "The only group available at the moment is the group “all” which would allow any valid EC2 user to launch your image."

Users should be account ids, not other users within your account.

```
 - name: Testing ec2_ami
   ec2_ami:
     aws_access_key: "{{ aws_access }}"
     aws_secret_key: "{{ aws_secret }}"
     instance_id: i-67086448
     name: Testing PR534 II
     region: us-east-1
     state: present
     launch_permissions:
        user_ids: [ '123456789012' ] # if you have another aws account in use, you should put that here
        group_names: ['all'] # this makes the ami public
   register: stuff
```

juliedavila:
I see. More on the topic of style, if `all` is the only valid option why make it a parameter? I feel it implies there is more than one choice. Judging boto from its past, I would gamble that other options won't really be available until boto3. Thoughts?

ferrants:
I don't think it's a boto thing, I think it's an aws thing. Even in the AWS console, there's no option to set the groups, just to make it public. Most of the aws parameters for ansible modules translate directly to the boto calls, so i think what we have is fine, but maybe a `public` parameter is more appropriate.

gregdek:
@ferrants looks like this needs to be rebased -- if it's still applicable, please rebase and we'll move it into community review.

ferrants:
I will update this and https://github.com/ansible/ansible-modules-core/pull/1006 next week

ferrants:
@gregdek - updated and squashed

gregdek:
Thanks @ferrants. Much appreciated.

@scicoin-project Please review this PR to make sure it adheres to the following guidelines:

juliedavila:
@ferrants followup on my previous comment. if 'all' is the only option, why is it not simply a boolean param like `make_public`? Just trying to understand the reason to use a module param that only works with a single specific input in a non-boolean way.

ferrants:
@defionscode - Most of the aws parameters for ansible modules translate directly to the boto calls, so boto documentation maps nicely to ansible documentation. I can change it, but I think the direct mapping is cleaner.

juliedavila:
@ferrants I agree that it's cleaner on the code side but not necessarily from the UX as an end user of the module. I won't call this a blocker for acceptance so I'll give a +1.

ferrants:
@defionscode - my understanding is that AWS will someday open support for more groups, which is why it's set this way in the api/boto. This should be more future-proof.

juliedavila:
That's fine with me then. +1

ferrants:
@defionscode - trying to do some PR cleanup. Is this waiting on anything to be merged?

juliedavila:
nope.

@gregdek @bcoca 
shipit
+1
etc
etc 
;)

gregdek:
Thanks @ferrants for the PR and @defionscode for the review. Sorry for the delay.

bcoca:
was missing version_added for new feature

PR #537, opened 2014-12-17
fix uri modul for JSON-escape quotation marks

zitterbacke:
consider the following response body (content) of a REST/JSON webservice containing escaped quotation marks:

``` json
{ "key": "\"works\"" }
```

decoding this string not as raw will lose the backslash as JSON escape. later json.loads will fail to parse.

Inspired by [this thread](https://groups.google.com/forum/#!topic/ansible-project/kymtiloDme4) on the mailing list and the following python shell code:

``` python
import json
string=r'{ "key": "\"works\"" }'
json.loads(string)
json.loads(string.decode('raw_unicode_escape'))
json.loads(string.decode('unicode_escape'))
```

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -304,7 +304,7 @@ def uri(module, url, dest, user, password, body, method, headers, redirects, soc
         r.update(resp_redir)
         r.update(resp)
         try:
-            return r, unicode(content.decode('unicode_escape')), dest
+            return r, unicode(content.decode('raw_unicode_escape')), dest
         except:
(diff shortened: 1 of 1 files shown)


adematte:
any plans on merging this soon?

abadger:
The uri module had some additional issues in this same code. I've just submitted https://github.com/ansible/ansible-modules-core/pull/1331 to try to fix those issues. If anyone is still interested in this, please test that PR to be sure it doesn't cause a regression in this fix.

PR #538, opened 2014-12-17
Update to allow volume type gp2

s33dunda:
AWS has added a new ebs volume type called gp2 which is their ssd with standard iops. I would like the ability to add this type along with the default standard.`


jxn:
Does this PR still need to modify [ec2-module](/ansible/ansible-modules-core/blob/devel/cloud/amazon/ec2.py) to work with the ec2 module?

jimi-c:
Hi @s33dunda, this appears to have been already implemented in b37a2328c0cba611c73ad5d1094a7ec4d1cfcdf1, so we're going to go ahead and close this issue.

Thanks!

PR #539, opened 2014-12-17
Added external_ips parameter to GCE

ZuluPro:
Allow user to use a static IP.
 The parameter is a list, so user gives a list of IP's name or 'ephemeral' for a new (as in libcloud).

If several instances are created and there are not enough element in `external_ips`, missing will be 'ephemeral'.

Module will fail if a unknown IP's name is given.

By default the parameter is an empty list, so the module hold its behavior and attach new IP to VM.

### DIFF
--- a/cloud/google/gce.py
+++ b/cloud/google/gce.py
@@ -96,6 +96,13 @@ options:
     default: null
     aliases: []
     version_added: "1.7"
+  external_ips:
+    description:
+      - a list of IP to attach to the instances; each element will be used for a new instance. Pass in a static IP's name for attach it or 'ephemeral' (default) for a new non-static address. If there is not enough specidifed IP, 'ephemeral' will be used.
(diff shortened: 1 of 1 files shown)


ZuluPro:
Do you want me to do something for facilitate your works?

gregdek:
@ZuluPro Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

gregdek:
Hi @ZuluPro -- we'd love to evaluate this for inclusion if it still applies, but it needs to be rebased.

ZuluPro:
Ok, I will make it, do you think I can make
`git pull git@github.com:ansible/ansible-modules-core.git master`
(from my current branch)

Or do you have an another solution ?

ZuluPro:
It has been implemented with another PR.

PR #544, opened 2014-12-18
Create SSH Keys always after creating $HOME

encbladexp:
##### Issue Type:

Bugfix Pull Request
##### Ansible Version:

ansible 1.8.2
##### Environment:

Arch Linux 64-Bit
##### Summary:

Ansible should always create a $HOME before trying to create an SSH Key.
##### Steps To Reproduce:

Change $HOME of an existing User and enable generate_ssh_key for this User, ansible will fail with this simple task because we try to generate ans SSH Key before creating $HOME.
##### Expected Results:

Ansible creates $HOME before SSH Key.
##### Actual Results:

Changing a Users Name or $HOME fails.

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -1917,6 +1917,16 @@ def main():
         if user.groups is not None:
             result['groups'] = user.groups
 
+        # handle missing homedirs
+        info = user.user_info()
+        if user.home is None:
+            user.home = info[5]
+        if not os.path.exists(user.home) and user.createhome:
+            if not module.check_mode:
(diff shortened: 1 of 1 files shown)


MarkusH:
Looks good.

encbladexp:
After 19 days, this is still an Bug and not an Feature Pull Request. If we try to create SSH Keys before creating an $HOME we will run into an error. ;)

encbladexp:
Rebased for current devel Branch.

PR #547, opened 2014-12-21
Fix bind-volumes on docker >= 1.4.0

steveej:
If bind-volumes are submitted to docker >= 1.4.0 with the volumes set in addition to the binds, docker will create a regular volume and not bind-mount the specified path.


lorin:
I tried out this patch locally since I needed the functionality, and I hit a case where this fix didn't work.

If the image specifies a volume directory, and you try to bind a local directory to the volume directory, Docker won't bind the container directory to the host directory.

For example, the [Dockerfile](https://github.com/nginxinc/docker-nginx/blob/57da11369acbec3256b0c2704a50282eeabb684f/Dockerfile#L16) of the [official nginx image](https://registry.hub.docker.com/_/nginx/) contains:

```
VOLUME ["/var/cache/nginx"]
```

If you try to bind this directory to a local directory using the docker module, Docker still won't do the bind correctly:

```
- docker:
    image: nginx
    name: mynginx
    volumes: /home/lorin/nginx-cache:/var/cache/nginx
```

I worked around it by passing the bind info in the `host_config` parameter that gets passed to the `create_container` method.

```
diff --git a/cloud/docker/docker.py b/cloud/docker/docker.py
index f71bad4..4c74b1c 100644
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -721,6 +721,7 @@ class DockerManager(object):
                   'volumes':      self.volumes,
                   'mem_limit':    _human_to_bytes(self.module.params.get('memory_limit')),
                   'environment':  self.env,
+                  'host_config':  docker.utils.create_host_config(binds=self.binds),
                   'hostname':     self.module.params.get('hostname'),
                   'detach':       self.module.params.get('detach'),
                   'name':         self.module.params.get('name'),
```

This seems to be discussed in https://github.com/docker/docker-py/issues/419 as bug in docker-py.

steveej:
I've updated my PR with additional changes.

Things I've tested successfully with the official nginx image

#### running the image without additional binds

`$ ansible -i 'localhost,' localhost -m docker -a 'image=nginx state=running'`

```
                "Volumes": {
                    "/var/cache/nginx": "/var/lib/docker/vfs/dir/4442d05a9d1b51eb82bb27f293809d60c0ad5eecfbe047f7e28db91c45e36c84"
                },
                "VolumesRW": {
                    "/var/cache/nginx": true
                }
```

#### running the image and overriding the VOLUME from the Dockerfile

`$ ansible -i 'localhost,' localhost -m docker -a 'image=nginx state=running volumes=/var/tmp/nginx-cache:/var/cache/nginx'`

```
                "Volumes": {
                    "/var/cache/nginx": "/var/tmp/nginx-cache"
                },
                "VolumesRW": {
                    "/var/cache/nginx": true
                }
```

`$ ls -l /var/tmp/nginx-cache/`

```
total 0
drwx------ 1 messagebus root 0 12. Mär 13:27 client_temp
drwx------ 1 messagebus root 0 12. Mär 13:27 fastcgi_temp
drwx------ 1 messagebus root 0 12. Mär 13:27 proxy_temp
drwx------ 1 messagebus root 0 12. Mär 13:27 scgi_temp
drwx------ 1 messagebus root 0 12. Mär 13:27 uwsgi_temp
```

This looks like correct behavior to me.

@lorin 
Could you please verify that the problem is gone with my latest commits?

lorin:
@steveeJ Just tested it out, this commit resolves the issue. :+1:

magicrobotmonkey:
This would be really nice to see fixed,makes it difficult to use ansible to orchestrate containers

stefanvanwouw:
Have the same problem. Please accept this PR. +1

maliskovik:
same issue here.

lchorbadjiev:
+1

gregdek:
Lots of love for this PR, including a +1 from @lorin -- promoting to P1_reviewed.

PR #554, opened 2014-12-23
Added ability to use url as key source

likwid:
I have a feeling this first implementation is too naive, but I figured I would give it a shot.

Github provides the ability to pull public keys associated with a user via https://github.com/username.keys

I have wanted to use this multiple times, but didn't find it to be possible with the current implementation of authorized_key module. I probably should update the documentation to reflect that a url is allowed, but I wanted to see what the feedback was like, in case I need to change the implementation significantly.

### DIFF
--- a/system/authorized_key.py
+++ b/system/authorized_key.py
@@ -118,6 +118,7 @@ import os.path
 import tempfile
 import re
 import shlex
+import urllib2
 
 class keydict(dict):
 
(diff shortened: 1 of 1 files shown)


likwid:
Taking no news as good news here...

Since there hasn't been any discussion around the implementation, I updated the documentation to reflect using a URL as the key source.

bcoca:
looks good, but instead of re-implementing URI fetching can you use the shared module fetch_url function?

likwid:
Refactored it per feedback.

PR #556, opened 2014-12-23
follow does not resolve to linked directory (introduced by #9518)

redbaron:
https://github.com/ansible/ansible/issues/9518 issue fix added a check for the case when path is already exsits and it is not a directory. This patch extends check to treat "symlink to directory" case as  regular directory for the purpose of this check.

In the nutshell it fixes following case:

```
- hosts: all
  gather_facts: no
  tasks:
    # prepare
    - file: path=/tmp/dir state=directory
    - file: path=/tmp/link src=/tmp/dir state=link


    # master branch  breaks on following, 
    - file: path=/tmp/link state=directory follow=yes
```


abadger:
I see follow in your example.  Should follow also be in the code?  I'm not certain of whether it should be part of the check myself, this is a genuine question.

redbaron:
follow=yes is there to indicate that we operate on a symlink target. Both before and after the patch behavior of "follow=yes"equal to "follow=no", which is IMHO not correct, but that is a question of separate bugreport/discussion.

bcoca:
this patch just makes it work as follow=yes is always set, so the behaviors are still equal and the same, just changing the default.

This should fail, but doesn't with your patch

```
    # master branch  breaks on following,
    - file: path=/tmp/link state=directory follow=no
```

redbaron:
this patch fixes follow=yes case when symlink points to a dir which is broken after #9518 , follow=no case is left untouched as it is separate issue

bcoca:
check my PR, i believe this fixes for all cases, yours just ignored follow but defaulted to opposite behavior.

bcoca:
closing in favor of merged #596

PR #558, opened 2014-12-23
forcing git fetch if sha1 are equal between repos but branch exists only...

dahpgjgamgan:
forcing git fetch if sha1 are equal between repos but branch exists only at remote end

A fix for issue reported in ansible engine repo
https://github.com/ansible/ansible/issues/9876

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -718,7 +718,8 @@ def main():
                 if version in get_tags(git_path, module, dest):
                     repo_updated = False
             else:
-                repo_updated = False
+                if version in get_branches(git_path, module, dest):
+                    repo_updated = False
         if repo_updated is None:
             if module.check_mode:
                 module.exit_json(changed=True, before=before, after=remote_head)


gregdek:
Owned by core team: @bcoca @abadger @jimi-c @bennojoy please review

abadger:
The code itself looks correct.  Thanks!  However, because this is in the middle of a big conditional that it's easy to get lost in I'd like to see two additions:
- the case needs a comment to say what case this fixes (the commits are the same but the branch name is only in the remote repo then we need to fetch the remote repo to get updated branch names.)
- and an integration test (basically, a series of plays that setup a git repo which will expose the bug) in https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_git/tasks/main.yml  This will hopefully prevent regressions if someone refactors the code and doesn't understand what this test is doing.

gregdek:
@dahpgjgamgan A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

dahpgjgamgan:
Ok, I'll try to get back to the issue in the following weeks. Just to
rephrase the previous messages - you want additional comments and an
integration test with the reproduction of the issue?

2015-09-11 19:42 GMT+02:00 Greg DeKoenigsberg notifications@github.com:

gregdek:
Yes, @dahpgjgamgan, that's how I interpret @abadger's comments as well. 

The tricky bit is that it'll require PRs to two different repos, so when you submit your PR for the integration test in ansible/ansible, please cross-reference the PRs so we can evaluate them together. Thanks!

abadger:
I added the comment I requested and created an integration test.  Merged both to devel.  Also added to the stable-2.0.1 branch so that it will get into that release.

PR #560, opened 2014-12-24
Fixes version check for docker-py

weitzj:
get_docker_py_versioninfo() did not always return a tuple as a version number, but sometimes a list. The comparison of a list and tuple fails, which causes newer docker-py versions than (0, 3, 0) to fail.

This bugfix makes sure that always a tuple is returned.

Tested on OSX 10.9.5


abadger:
Thanks for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #561, opened 2014-12-24
Fixes invalid dictionary access.

weitzj:
Fixes an invalid dictionary access, which threw a KeyError. Now it defaults to None.

Also optionals should be a dictionary instead of an array, since we got strings as keys.


abadger:
Was volumes_from throwing an error or only dns?  It might make more sense to set dns to default to None in the argument_spec at the beginning of the main function.

weitzj:
It happens with dns and volumes_from, since they are not defined in the local params dictionary.

I updated the pull request.

abadger:
Thanks for cleaning up after my mistakes ;-)

# Merged

Hi!

issue is resolved.

Thank you!

PR #564, opened 2014-12-24
Fix compatibility issues with python-apt < 0.7.9

FabioBatSilva:
### This patch fix the apt module using python-apt 0.7.4 (ubutun 8.04 - hardy)

When installing a specific package version ansible will fail while try to access `apt.package.Package#versions` witch is only available after 0.7.9.

see :
http://apt.alioth.debian.org/python-apt-doc/library/apt.package.html#apt.package.Package.versions
#### apt: name=foo=1.00 state=present

```
Traceback (most recent call last):
  File "/root/.ansible/tmp/ansible-tmp-1419437723.99-201671988715310/apt", line 2167, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1419437723.99-201671988715310/apt", line 586, in main
    dpkg_options=dpkg_options)
  File "/root/.ansible/tmp/ansible-tmp-1419437723.99-201671988715310/apt", line 283, in install
    installed, upgradable, has_files = package_status(m, name, version, cache, state='install')
  File "/root/.ansible/tmp/ansible-tmp-1419437723.99-201671988715310/apt", line 209, in package_status
    avail_upgrades = fnmatch.filter((p.version for p in pkg.versions), version)
AttributeError: 'Package' object has no attribute 'versions'
```
#### lsb_release -a

```
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 8.04
Release:    8.04
Codename:   hardy
```
#### dpkg -l | grep python

```
ii  python                 2.5.2-0ubuntu1
ii  python-apt             0.7.4ubuntu7
ii  python-central         0.6.5ubuntu1
ii  python-gdbm            2.5.2-0ubuntu2
ii  python-gnupginterface  0.3.2-9ubuntu1
ii  python-minimal         2.5.2-0ubuntu1
ii  python-simplejson      1.7.3-1
ii  python-support         0.7.5ubuntu1
ii  python2.5              2.5.2-2ubuntu4
ii  python2.5-minimal      2.5.2-2ubuntu4
```

PR #567, opened 2014-12-25
handle list of lists - python-apt < 0.7.9 compatibility

FabioBatSilva:
See : https://github.com/ansible/ansible-modules-core/pull/564

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -180,7 +180,8 @@ def package_versions(pkgname, pkg, pkg_cache):
         # assume older version of python-apt is installed
         # apt.package.Package#versions require python-apt >= 0.7.9.
         pkg_cache_list = (p for p in pkg_cache.Packages if p.Name == pkgname)
-        pkg_versions = (p.VersionList for p in pkg_cache_list)
+        pkg_list_of_lists = (p.VersionList for p in pkg_cache_list)
+        pkg_versions = (p for l in pkg_list_of_lists for p in l)
         versions = set(p.VerStr for p in pkg_versions)
 
     return versions
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -202,8 +202,12 @@ def package_status(m, pkgname, version, cache, state):
         ll_pkg = cache._cache[pkgname] # the low-level package object
     except KeyError:
         if state == 'install':
-            if cache.get_providing_packages(pkgname):
-                return False, True, False
+            try:
+                if cache.get_providing_packages(pkgname):
+                    return False, True, False
+            except AttributeError:
+                # older python-apt providing packages cannot be used
+                pass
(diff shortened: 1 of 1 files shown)


abadger:
For the first portion of the change, I think that itertools.chain would be more readable:

``` python
        pkg_cache_list = (p for p in pkg_cache.Packages if p.Name == pkgname)
        pkg_versions = (p.VersionList for p in pkg_cache_list)
        versions = set(p.VerStr for p in itertools.chain(pkg_versions))
```

(and an appropriate import itertools earlier in the file)

For the second part, I had to read the code a few times but I think this is what the original code is doing:  When we are asked to install a pkgname and the pkgname isn't a real package we try to detect whether there's a virtual package that can satisfy the requirement.  If we successfully detect a virtual package we return a value that tells the install() function that the package is upgradable (may be a slight lie... the package may already be at the desired version.  That will cause a false positive for check mode and possibly a little extra work but shouldn't cause other issues.)  

If the pkgname is not a virtual provide then we return an error to the user that the pkgname is invalid.

If that's a correct reading of the code then I think for that portion if get_providing_packages() does not exist we need to return an error instead of passing.  Maybe something like this:

``` python
try:
    if cache.get_providing_packages(pkgname):
        return False, True, False
except AttributeError:
    # older python-apt providing packages cannot be used
    m.fail_json(msg="No package matching '%s' is available (python-apt version too old to detect virtual packages)" % pkgname)
```

FabioBatSilva:
@abadger I've done the changes you suggested,  
But i was wondering if we should just mark the package as upgradable and let `apt-get install foo=1.0.0` deal with it.

something like : 

``` python
try:
    if cache.get_providing_packages(pkgname):
        return False, True, False
    m.fail_json(msg="No package matching '%s' is available" % pkgname)
except AttributeError:
    # older python-apt providing packages cannot be used
    # mark as upgradable and let apt-get install deal with it
    return False, True, False
```

abadger:
@FabioBatSilva Hmm... I don't generally like making the function return something that isn't quite true because future changes to the code or features that we don't always think about (for instance, checkmode) may be impacted by the slight lie.

But AFAICS, your proposal to have apt-get install deal with this problem would work for all of the existing locations in the code which use package_status().  Additionally, this will only be triggered for older python-apt versions which we don't support anymore -- it doesn't touch the code path for recent python-apt.

So I'd be okay with your proposal if you still like it better.  Just be aware that you might have to propose more python-apt<0.7.x  fixes in the future if the module code changes and there's more places that use package_status() :-)  Maybe just add the python-apt version that we're hardcoding this return value for to the proposed comment as well so that people know in what circumstance that code will be invoked.

FabioBatSilva:
@abadger Done !!

abadger:
Merged!   Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #572, opened 2014-12-26
[hostname] rename FedoraStratgy to SystemdStrategy

ghost:


abadger:
This looks simple and will be more organizationally consistent.  @bcoca, okay if I merge this?

PR #588, opened 2015-01-01
Docker module: add wait parameter

lorin:
Add a `wait` parameter to the _docker_ module.

When this parameter is specified, when starting or restarting a container, the module will block until the container has stopped. If the return value of the process is non-zero, the module will fail, using the process's output as the error message.

Example use case: initializing a database or running a database migration inside of a container. This task must complete before going to the next task in the playbook, and if it fails you need a meaningful error message.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -231,6 +231,11 @@ options:
     default: false
     aliases: []
     version_added: "1.9"
+  wait:
+    description:
+      - Wait until process terminates.
+      - Fails on non-zero return code
+    version_added: "1.9"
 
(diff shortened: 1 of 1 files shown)


vgrossmann:
+1

mikz:
:+1:

AdamCmiel:
+1

gregdek:
@lorin if this still applies, please rebase and we will get it into the community review process asap.

anthcourtney:
+1

lorin:
@gregdek this has been obsoleted by #1031, which just got merged.

PR #589, opened 2015-01-02
ec2_asg: honor lc_check parameter fully

robbwagoner:
Sometimes it is useful to replace some/all instances in an ASG with the
same Launch Configuration. Parameter `lc_check` is inteded for this
purpose, but is only used in the `replace_batch()` method. This fixes a
small oversight in `replace()`.


bennojoy:
@robbwagoner  shoudl be fixed via #1030 can you please check

robbwagoner:
@bennojoy confirmed fixed by #1030.

PR #590, opened 2015-01-02
iam module

juliedavila:
There is currently no IAM module. This is part 1 of 3 that I plan to make.

Part 1: Create/Update/Delete users, groups and roles
Part 2: Create/Update/Delete policy objects for users/groups/roles
Part 3: Upload/Remove/Update server and signing certificates


cloud/amazon/iam.py
```diff
@@ -0,0 +1,354 @@
...
+  access_key_state:
+    description:
+      - when type is user: creates, removes, deactivates or activates a user's access key. Note that remove and deactivate will apply to all keys.
+    required: "false"
+    default: null
+    choices: ["create", "remove", "activate", "deactivate"]
+    aliases: []
+  group:
+    description:
+      - when type is user, adds user to the specified group. When 'update' is specified, it will add the user to groups, preserving original group membership.
```
mpavlov: The distinction between `state == present` and `state == update` is a bit unclear. Is it that `state == present` will remove unspecified group memberships for this user?

cloud/amazon/iam.py
```diff
@@ -0,0 +1,354 @@
...
+
+## Create, or Delete Roles. There is no update as boto does not currently provide a role_update function
+    if iam_type == 'role' and state == 'present':
+        if name not in role_list:
+            result = create_role(iam, name, path)
+    elif iam_type == 'role' and state == 'absent':
+        if name in role_list:
+            result = delete_role(iam, name)
+
+    module.exit_json(**result)
```
mpavlov: To increase readability and maintainability of `main()`, could factor out the `iam_type`-specific code into dedicated functions. E.g., `work_on_role()`, `work_on_group()`, and `work_on_user()`.

cloud/amazon/iam.py
juliedavila: Yeah, will tweak that soon.

cloud/amazon/iam.py
juliedavila: I'll have to think about this, I don't really see too much complexity here.

juliedavila:
still some work to be done

juliedavila:
Going to resubmit with cleaner commit history. Now at https://github.com/ansible/ansible-modules-core/pull/756

PR #591, opened 2015-01-03
Fix 'wait_for' doesn't work with ipv6only host

atrubachev:
If host ipv6only module 'wait_for' works incorrectly, because `socket.AF_INET` is used.
The function `socket.create_connection()` is used instead of `socket.connect()` in this patch.

### DIFF
--- a/utilities/logic/wait_for.py
+++ b/utilities/logic/wait_for.py
@@ -351,10 +351,8 @@ def main():
                 except IOError:
                     break
             elif port:
-                s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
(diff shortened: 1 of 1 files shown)


bcoca:
one issue is that create_connection was added in python 2.6 and we want modules to be compatible with 2.4

gregdek:
@jhoekx @jarv @AnderEnder please review; if you give +1 to this PR, we will promote to P1_reviewed for faster inclusion.

gregdek:
@jhoekx @jarv @AnderEnder are none of you available to review this PR?

jarv:
+1

AnderEnder:
+1

PR #592, opened 2015-01-03
nova_compute: fix mess of ipv6 and ipv4 addresses in private_ip

atrubachev:
When we use ipv4 and ipv6 addresses as private at the same time, we get a mess in 'private_ip':

```
"info": {
    "addresses": {
        "MYNETS": [
            {
                "OS-EXT-IPS-MAC:mac_addr": "fa:16:3e:81:6d:ba",
                "OS-EXT-IPS:type": "fixed",
                "addr": "2a02:6b8:0:1a5d:1df6:344b:1786:db66",
                "version": 6
            },
            {
                "OS-EXT-IPS-MAC:mac_addr": "fa:16:3e:81:6d:ba",
                "OS-EXT-IPS:type": "fixed",
                "addr": "141.8.181.75",
                "version": 4
            }
        ]
    },
...
"private_ip": "2a02:6b8:0:1a5d:1df6:344b:1786:db66141.8.181.75",
"public_ip": ""
...
```

Value of function `openstack_find_nova_addresses()` is used without applies `join()` in this patch.

```
...
"private_ip":
    [
        "2a02:6b8:0:1a5d:1df6:344b:1786:db67",
        "141.8.181.150"
    ],
"public_ip": ""
...
```


gregdek:
@emonty feedback here?

emonty:
I believe this code is better. HOWEVER - it is an interface change. As such, I think we should defer to the new os_server module

gregdek:
If the new os_server module is ready or close to ready, we should close this PR, then, right?

emonty:
@gregdek yup. os_server is landed.

PR #595, opened 2015-01-06
glance_image: Allow for specification of kernel, ramdisk and extra properties

gandelman-a:
Glance image properties are used to do many useful things.  Most commonly, they are used to associate an image with existing kernel and ramdisk images, which may be required to boot an image.  This updates the module to accept a 'kernel' and 'ramdisk' paramters, as well as dict of arbitrary k=v pairs.  When specified and differing from the image returned from the server, the image will be updated.


emonty:
I believe this is all covered in #1042 now.

Issue #423, opened 2014-12-02
EC2 module spot-pricing broken in 1.8.*


josephtate:
##### Issue Type:

Regression
##### Ansible Version:

1.8.2
##### Environment:

N/A
##### Summary:

Commit 7f11c3d7cc9b4b71d72ad6a59e2b7aa22ecf93cf broke ec2 spot instance launching after 1.7.2. I'll work on a pull request to fix.
##### Steps To Reproduce:

```
    - name: Launch EC2 instances
      register: ec2
      local_action:
        module: ec2
        keypair: "{{ssh_key}}"
        instance_type: "{{instance_type}}"
        spot_price: "{{spot_price}}"
        count: 1
        group: "{{security_group | default('ZZ9-Benchmark-Group')}}"
        image: "{{ami}}"
        kernel: "{{kernel}}"
        instance_tags: { Name: "{{instance_name}}" }
        region: us-west-2
        wait: true
        vpc_subnet_id: "{{vpc_subnet| default(None)}}"
        assign_public_ip: "{{ 'on' if vpc_subnet|default(None) else 'off'}}"
        zone: us-west-2c
        volumes:
          - device_name: /dev/sda1
            volume_size: "{{ebs_vol_size}}"
            device_type: gp2
            delete_on_termination: true
          - device_name: /dev/xvdb
            ephemeral: ephemeral0
          - device_name: /dev/xvdc
            ephemeral: ephemeral1
          - device_name: /dev/xvdd
            ephemeral: ephemeral2
          - device_name: /dev/xvde
            ephemeral: ephemeral3
```
##### Expected Results:

The task should complete successfully instead of failing with a traceback.
##### Actual Results:

```
TASK: [Launch EC2 instances] ************************************************** 
<127.0.0.1> REMOTE_MODULE ec2 region=us-west-2 group=vpc-scratch-system,vpc-crunchapp-web-server assign_public_ip=on keypair='Jenkins Home Key' instance_type=r3.xlarge vpc_subnet_id=subnet-e3ac82a5 image=ami-3daad70d zone=us-west-2c spot_price=0.35 kernel=''
[...]

PR #597, opened 2015-01-06
EC2: move logic about terminated instances up (#423)

jbbarth:
As stated in #423, the commit 7f11c3d broke ec2 spot instance launching
after 1.7.2. This is because it acts on the 'res' variable which have 2
different types in the method, and in case we request spot instances,
the resulting object is not a result of `ec2.run_instances()` but
`ec2.request_spot_instances()`. Actually this fix doesn't seem to be
relevant in the spot instances case, because by construction we won't
retrieve 'terminated' instances in the end.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -915,6 +915,17 @@ def create_instances(module, ec2, override_count=None):
                             continue
                         else:
                             module.fail_json(msg = str(e))
+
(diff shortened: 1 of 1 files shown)


bcoca:
@jbbarth  can you rebase this request?

jbbarth:
Done! Let me know if I should squash the commits together.

PR #598, opened 2015-01-06
vpc_zone_identifier must be a csv string when an asg is updated.

### LINKED ISSUE #486: ec2_asg insisting that subnet id does not exist
#### Issue Type:

Bug Report
#### Ansible Version:

ansible 1.9 installed from source develop branch
#### Environment:

OSX 10.10
#### Summary:

Passing in a subnet in the same way provided in the module examples `['subnet-1234']` returns an AWS API error when the ASG has already been created. 
#### Steps To Reproduce:

Create a playbook similar to the following:

``` yaml

---
- name: debug aws stuff
  hosts: all
  gather_facts: no
  tasks:
    - name: create Auto Scaling Group
      ec2_asg:
        name: "somename"
        launch_config_name: "someconfig"
        aws_access_key: KEY
        aws_secret_key: KEY
        min_size: 1
        max_size: 2
        region: us-east-1
        vpc_zone_identifier: [ 'subnet-1234' ]
        state: present
```
#### Expected Results:

OK
#### Actual Results:

```
msg: BotoServerError: 400 Bad Request
<ErrorResponse xmlns="http://autoscaling.amazonaws.com/doc/2011-01-01/">
  <Error>
    <Type>Sender</Type>
    <Code>ValidationError</Code>
    <Message>The subnet ID '['subnet-1234']' does not exist</Message>
  </Error>
  <RequestId>7061d8b0-807f-11e4-a663-115fef5ac273</RequestId>
</ErrorResponse>


```

jsmartin:
This fixes a bug introduced in cf24e7d56c6981cf489f931e580f5d5673687401, and fixes #486.  That commit makes vpc_zone_identifier a list parameter.  It works fine on an initial creation of an ASG, but subsequent runs would fail:

```
TASK: [rolling_asg | create autoscale groups] *********************************
<fancyapp-rolling> REMOTE_MODULE ec2_asg health_check_type=ELB max_size=10 launch_config_name=rollingAMI-2 vpc_zone_identifier=subnet-6812da43 desired_capacity=5 min_size=5 region=us-east-1 name=rollingAMI
failed: [fancyapp-rolling] => {"failed": true}
msg: BotoServerError: 400 Bad Request
<ErrorResponse xmlns="http://autoscaling.amazonaws.com/doc/2011-01-01/">
  <Error>
    <Type>Sender</Type>
    <Code>ValidationError</Code>
    <Message>The subnet ID '['subnet-6812da43']' does not exist</Message>
  </Error>
  <RequestId>339b08e9-95cd-11e4-a3e5-5771b0508bdd</RequestId>
</ErrorResponse>
```

This patch fixes that problem.

### DIFF
--- a/cloud/amazon/ec2_asg.py
+++ b/cloud/amazon/ec2_asg.py
@@ -326,6 +326,8 @@ def create_autoscaling_group(connection, module):
         for attr in ASG_ATTRIBUTES:
             if module.params.get(attr):
                 module_attr = module.params.get(attr)
+                if attr == 'vpc_zone_identifier':
+                    module_attr = ','.join(module_attr)
                 group_attr = getattr(as_group, attr)
                 # we do this because AWS and the module may return the same list
                 # sorted differently


abadger:
The bugfix looks good.  The error message change introduces information that's going to be meaningless to non-coding endusers though.  Perhaps you can change the messages so that we can tell which check failed but not confuse end users?  Like changing one of them to be "Timeout exceeded while waiting for instance to appear: %s"

jsmartin:
I just reverted the messages to their previous state.  They are going to be removed in a forthcoming PR anyhow.

mpdehaan:
Merged, thanks @jsmartin !

ilya-epifanov:
Now it fails with:

The subnet ID 'u' does not exist, in fact it tries to create an ASG with groups ['s', 'u', 'b', 'n', 'e', 't', '-', ...]

Reverting this change fixes both initial creation and an update of an ASG

dgromov:
The latest version of Ansible (1.8.4) seems to include this change but not the one introduced in https://github.com/ansible/ansible-modules-core/commit/cf24e7d56c6981cf489f931e580f5d5673687401

The above case,  "['s', 'u', 'b', 'n', 'e', 't', '-', ...]" wouldn't happen if both commits were there. The code now says to parse the subnets as a string and then to do join(',') on them.

lostmimic:
On 1.8.4 I just got this error when trying to update an asg:

```
<Message>The subnet ID 'f' does not exist</Message>
```

looking at the string I pass into vpc_zone_identifier there is only 1 instance of the 'f' char: 
subnet-########, subnet-########, subnet-###f####

PR #606, opened 2015-01-07
fix for allowing permissions on hard links and soft links + follow=yes

bcoca:
corner case in which you specify state=file and want to change permissions  and used to fail if state was 'hard' or 'link', 'link' requires follow=yes to work, 'hard' is treated as 'file' and will set permissions.


abadger:
+1

PR #611, opened 2015-01-08
Remove skipped=True when using creates and removes

fgsch:
Based on #8645

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -381,7 +381,7 @@ def main():
         # of uri executions.
         creates = os.path.expanduser(creates)
         if os.path.exists(creates):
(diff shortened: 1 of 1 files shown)


gregdek:
@romeotheriault please review; if you give +1 to this PR, we will promote to P1_reviewed.

rrtj3:
Sorry it's been a while since I've worked on this module, but I guess I'm not sure I understand the desire to remove the skipped=True when a file exists or is non-existent, with the creates and removes checks. 

Why exactly do we not want to have the 'skipped=True'? The stdout says we're skipping....

fgsch:
@romeotheriault it's been a while but IIRC same reason as https://github.com/ansible/ansible/issues/8645

rrtj3:
I see. That makes sense to me. 

+1

gregdek:
Thanks @romeotheriault -- promoting to P1_reviewed.

PR #566, opened 2014-12-25
Cloud Modules STS Support Redux

semifocused:
Add support for AWS Security Token Service (temporary credentials) to all AWS cloud modules.

Rationalize sources and ordering of security credentials for consistency across modules.

Update documentation to use shared doc fragment and consistent usage information across modules.

PR #618, opened 2015-01-08
enable mfa authentication for cloudformation module

haraldsk:
the cloudformation module doesn't support MFA through boto the way the ec2 module does, this simple fix will enable it.


haraldsk:
I'm going to close this PR as I can found  #566 allready addresses the issue in a more general manner.

PR #619, opened 2015-01-08
mysql_user module: Added support for 'REQUIRE SSL' grant option

bpennypacker:
##### Ansible Version:

ansible 1.8 (devel 43eb821d3f) last updated 2015/01/07 20:16:03 (GMT +000)
  lib/ansible/modules/core: (detached HEAD db5668b84c) last updated 2015/01/07 20:16:14 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 110250d344) last updated 2015/01/07 20:16:20 (GMT +000)
##### Environment:

N/A
##### Summary:

This pull request adds the ability to enforce SSL on user connections.  
##### Steps To Reproduce:

```
    - local_action: mysql_user
                user=foo
                host='%'
                password=bar
                state=present
                append_privs=yes
                login_host=mysqlhost
                login_user=mysqlroot
                login_password=mysqlpass
                priv='*.*:REQUIRESSL'
```
##### Expected Results:

When invoked as above then the mysql command "SHOW GRANTS FOR 'foo'@'%';" should return something like this:

GRANT USAGE ON *.* TO 'foo'@'%' IDENTIFIED BY PASSWORD <REDACTED> REQUIRE SSL
##### Actual Results:

Currently the mysql_user module has no way of adding the "REQUIRE SSL" option to users.

### DIFF
--- a/database/mysql/mysql_user.py
+++ b/database/mysql/mysql_user.py
@@ -117,6 +117,9 @@ EXAMPLES = """
 # Creates database user 'bob' and password '12345' with all database privileges and 'WITH GRANT OPTION'
 - mysql_user: name=bob password=12345 priv=*.*:ALL,GRANT state=present
 
+# Modifiy user Bob to require SSL connections. Note that REQUIRESSL is a special privilege that should only apply to *.* by itself.
+- mysql_user: name=bob append=true priv=*.*:REQUIRESSL state=present
+
 # Ensure no user named 'sally' exists, also passing in the auth credentials.
 - mysql_user: login_user=root login_password=123456 name=sally state=absent
 
@@ -159,7 +162,7 @@ VALID_PRIVS = frozenset(('CREATE', 'DROP', 'GRANT', 'GRANT OPTION',
                          'EXECUTE', 'FILE', 'CREATE TABLESPACE', 'CREATE USER',
                          'PROCESS', 'PROXY', 'RELOAD', 'REPLICATION CLIENT',
                          'REPLICATION SLAVE', 'SHOW DATABASES', 'SHUTDOWN',
-                         'SUPER', 'ALL', 'ALL PRIVILEGES', 'USAGE',))
+                         'SUPER', 'ALL', 'ALL PRIVILEGES', 'USAGE', 'REQUIRESSL'))
 
 class InvalidPrivsError(Exception):
     pass
@@ -261,6 +264,8 @@ def privileges_get(cursor, user,host):
         privileges = [ pick(x) for x in privileges]
         if "WITH GRANT OPTION" in res.group(4):
             privileges.append('GRANT')
+        if "REQUIRE SSL" in res.group(4):
+            privileges.append('REQUIRESSL')
         db = res.group(2)
         output[db] = privileges
     return output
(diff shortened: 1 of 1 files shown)


database/mysql/mysql_user.py
```diff
@@ -307,19 +317,28 @@ def privileges_revoke(cursor, user,host,db_table,grant_option):
...
     db_table = db_table.replace('%', '%%')
-    priv_string = ",".join(filter(lambda x: x != 'GRANT', priv))
+    priv_string = ",".join(filter(lambda x: x not in [ 'GRANT', 'REQUIRESSL' ], priv))
     query = ["GRANT %s ON %s" % (priv_string, mysql_quote_identifier(db_table, 'table'))]
     query.append("TO %s@%s")
     if 'GRANT' in priv:
-        query.append("WITH GRANT OPTION")
+        query.append(" WITH GRANT OPTION")
+    if 'REQUIRESSL' in priv:
+        query.append(" REQUIRE SSL")
```
abadger: No need to prepend spaces -- We use " ".join(query) to combine the strings.

database/mysql/mysql_user.py
```diff
@@ -307,19 +317,28 @@ def privileges_revoke(cursor, user,host,db_table,grant_option):
     query = ["REVOKE ALL PRIVILEGES ON %s" % mysql_quote_identifier(db_table, 'table')]
     query.append("FROM %s@%s")
     query = ' '.join(query)
-    cursor.execute(query, (user, host))
+    try:
+        cursor.execute(query, (user, host))
+    except Exception, e:
+        raise Exception("%s. Query=\"%s\"" % (str(e), query % (user, host)))
```
abadger: The exceptions that we throw here we want to catch later when we use module.fail_json() (in the main function).  So don't raise Exception, instead either raise one of the Exceptions we already catch or create a new exception class and add that as something that we catch there.

database/mysql/mysql_user.py
```diff
@@ -307,19 +317,28 @@ def privileges_revoke(cursor, user,host,db_table,grant_option):
...
+        query.append(" WITH GRANT OPTION")
+    if 'REQUIRESSL' in priv:
+        query.append(" REQUIRE SSL")
     query = ' '.join(query)
-    cursor.execute(query, (user, host))
+    try:
+        cursor.execute(query, (user, host))
+    except Exception, e:
+        raise Exception("%s. Query=\"%s\"" % (str(e), query % (user, host)))
+
```
abadger: Same note about Exception and module.fail_json() here as above.

abadger:
Made some code comments but overall this looks good.  Let me know when you've updated and I'll take a look at merging.

database/mysql/mysql_user.py
bpennypacker: Exceptions from cursor.execute actually aren't being fully handled properly.  If you look where user_mod is called (around line 508) it only catches SQLParseError and InvalidPrivsError errors.  If cursor.execute throws any other exception (it was throwing a ProgrammingError when I was working on this) then it just crashes ansible. Perhaps the exception handling in the main section needs to be improved upon, but I didn't want this to turn into a rewrite of all the exception handling.

abadger:
Merged thanks!  I'll look into adding ProgrammingError to the exceptions we're catching in the main portion of the module.

# Merged

Hi!

issue is resolved.

Thank you!

PR #623, opened 2015-01-11
Remove unused option in yum's repoquery call

mscherer:
Citing the man page:
       -q, --query
              For rpmquery compatibility, doesn't do anything.


mchlumsky:
The commit looks good. I can confirm the option does nothing on EL 5,6,7 and Fedora 21.

+1

Issue #586, opened 2014-12-31
[pip] no support for specifying the python version to use for a virtualenv?


wbolster:
It seems the `pip` module doesn't allow me to specify which Python version to use for a virtualenv.

When I use something like this:

```
pip: name=myproject virtualenv=/path/to/virtualenv/
```

... the system Python version is used, which is Python 2.7 on my system. However, my project requires Python 3, and Python 3.4 is installed properly on the system. The problem is that there doesn't seem to be a way to tell the `pip` module to create the virtualenv using Python 3.4.

Fwiw, when manually creating a virtualenv, I would use `virtualenv -ppython3.4 myvenv` to create it.

cchurch:
One approach you could use would be to create the virtualenv first and specify the Python version:

```
command: creates=/path/to/virtual virtualenv -p python3.4 /path/to/virtualenv/
```

Then, since the virtualenv already exists, the pip module won't try to create it and _should_ use the version of Python in the virtualenv.

wbolster:
Thanks, I had a work-around like that in place already.

However, I still think this needs to be in Ansible's `pip` module. virtualenvs are intended for isolated python environments, and that includes not only python packages, but also the python version used.

Proposal:  add a `virtualenv_python` (or maybe `python_version`?) argument.

wbolster:
Just to be clear: some systems have separate `virtualenv-2.7` and `virtualenv-3.4` commands that may do the right thing, but these are not guaranteed to be installed, since it requires multiple `virtualenv` installations, and distros may ship only a single version. Since `virtualenv` is designed to create environments for different Python versions (using the `-p` flag), this is perfectly fine. It just means that Ansible's current `virtualenv_command` does not cover all situations, hence the `virtualenv_python` argument I've added in PR #629, which, together with `virtualenv_command`, does cover all situations.

bivald:
This would help me as well, specially since it appears as newer Debian (which can apt-get python3) doesn't get virtualenv-3.4 link, just virtualenv.

wbolster:
Unfortunately my PR #629 has been lying around for months, slowly gathering dust and bitrot. :-(

bivald:
I would assume they are working on it, but that they simply have quite the backlog.

glassresistor:
this would help me as well. the upside of ansible is cleanly specifying something, having to use two tasks makes it less clean.

lacion:
[...]

PR #629, opened 2015-01-12
Add virtualenv_python arg to pip module

wbolster:
See #586 for details.

### DIFF
--- a/packaging/language/pip.py
+++ b/packaging/language/pip.py
@@ -63,7 +63,7 @@ options:
     default: "no"
     choices: [ "yes", "no" ]
   virtualenv_command:
-    version_aded: "1.1"
+    version_added: "1.1"
     description:
       - The command or a pathname to the command to create the virtual
         environment with. For example C(pyvenv), C(virtualenv),
--- a/packaging/language/pip.py
+++ b/packaging/language/pip.py
@@ -70,6 +70,14 @@ options:
         C(virtualenv2), C(~/bin/virtualenv), C(/usr/local/bin/virtualenv).
     required: false
     default: virtualenv
+  virtualenv_python:
+    version_added: "FIXME"
+    description:
+      - The Python executable used for creating the virtual environment.
+        For example C(python3.4), C(python2.7). When not specified, the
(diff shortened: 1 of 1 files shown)


wbolster:
Note to reviewer: there's a FIXME value in the `versionadded` field that should be changed prior to merging.

wbolster:
Is there anything I can do to move this forward? (and close #586)

/cc @bcoca

bcoca:
@wbolster for one, this does not merge cleanly, after that i can only ask for patience as we go through the backlog

wbolster:
Okay, I've updated this PR to apply cleanly against the current `devel` branch, which was updated in the mean time. No more merge conflicts now.

gregdek:
@mattupstate please review; if you give +1 we will promote to P1_reviewed.

mattupstate:
Is this not already supported by using the `command` option?

gregdek:
@mattupstate Yes; that's actually addressed in #586.

mattupstate:
@gregdek I don't understand how something can be addressed without a pull request.

gregdek:
Sorry, didn't make myself clear.

Yes, there's an example for using the command option for doing this in issue #586. That's also where @wbolster provides the rationale for including this in the pip module directly.

wbolster:
@mattupstate No, the `command` option specifies which `virtualenv` command to use. This PR is about adding support to the Ansible `pip` module for a flag that will translate to the `-p` option to `virtualenv`.The end result would be that something like `virtualenv -ppython3.4` is executed; the first part comes from the `command` option, and the second part from the proposed flag. See https://github.com/ansible/ansible-modules-core/issues/586#issuecomment-69732132 for a detailed explanation why this is needed.

gregdek:
@mattupstate a friendly ping to weigh in on this PR when you can.

mattupstate:
@gregdek It looks OK to me. And for future reference, I'm not super concerned about reviewing the modules I originally wrote. I contributed them when Ansible was first open sourced and things have changed quite a bit since then, so I'm not exactly the best person to review anymore.

PR #639, opened 2015-01-14
distutils is not available on some non-Linux OS's

tgolly:
##### Summary:

The "service" module is broken on Solaris since 30d6713.

distutils.LooseService() is only referenced in the LinuxService class, this ensures it's only loaded on Linux platforms.

After this patch the service module works again on Solaris.
##### Steps To Reproduce:

Use the service module on a Solaris machine:

```
service: name=/network/ntp state=restarted
```
##### Expected Results:

ntp daemon is restarted
##### Actual Results:

```
ImportError: No module named distutils.version
```

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -106,7 +106,8 @@ import select
 import time
 import string
 
-if platform.system() == 'Linux':
+# The distutils module is not shipped with SUNWPython on Solaris.
+if platform.system() != 'SunOS':
     from distutils.version import LooseVersion
 
 class Service(object):
--- a/system/service.py
+++ b/system/service.py
@@ -106,7 +106,8 @@ import select
 import time
 import string
 
-from distutils.version import LooseVersion
+if platform.system() == 'Linux':
+    from distutils.version import LooseVersion
 
 class Service(object):
     """


abadger:
Where are you getting your python version from?  Which version of python is it? Is distutils present but not distutils.version?  ( python -c 'import distutils' works but python -c 'import distutils.version' does not)  distutils is a part of python's stdlib so it should be there unless the python version on your OS is being crippled by your vendor.  Without distutils, installing third party python packages from pypi wouldn't work as well as several means of querying for information about the system. Trying to google for whether solaris is lacking distuitls in general turns up only one relevant page from stackoverflow so this doesn't seem quite right.... perhaps there's a separate Solaris python-libs or python-dev or something package that needs to be installed to put distutils on the system?

tgolly:
Hi abadger, you are right!  It appears distutils is available in the devel package.  

```
# grep distutils/__init__.py SUNWPython-devel/pkgmap
1 f none lib/python2.6/distutils/__init__.py 0644 root bin 670 57754 1340772982
1 f none lib/python2.6/distutils/__init__.pyc 0644 root bin 437 30473 1340772985
```

Unfortunately this has a chilling effect on our ability to use Ansible, we can't install devel packages on the production servers.

I'd really like my one-line patch to be considered, as distutils is only required for the Linux platform.  It means we wouldn't need to have SUNWPython-devel installed on our prod servers which keeps the security folks happy.

abadger:
<nod>   Yeah, I see your dilemma but I also don't want to put us in a position where we're avoiding making use of stdlib modules because some vendors have crippled their packages unnecessarily.  (For what it's worth, older Ubuntu Linux also split distutils into a separate, non-default devel package and users had no choice but to install that package until Ubuntu moved distutils into the main package). 

How about this -- if you rework your patch so that it avoids the import on Solaris instead of only importing on Linux and add a comment about the split packages on Solaris, I'll merge it.  But I can't guarantee that we won't find we need to use distutils unconditionally somewhere in the future.

Sound fair?

tgolly:
Yep, sounds fair and I agree.  Patch updated.  Thanks!

abadger:
Thanks for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #643, opened 2015-01-14
keep authorized_keys in order

ghost:
Problem:
The order in which keys are added to the authorized_keys file gets shuffled around when using dicts.
This can lead to problems with other deploymenttools where comments are used as markers.


ghost:
problems in my version (not fully migrated to lists), no time to fix them .. sorry

PR #653, opened 2015-01-16
nova_compute: Added optional parameter scheduler_hints.

Jorge-Rodriguez:
This parameter allows to pass scheduler hints to the nova boot which enables, among other things, the use of affinity features.


Jorge-Rodriguez:
Crap! I moved these commits to a new branch on my fork and now this is (of course) closed.
@bcoca Do I re-submit from the new branch?

PR #658, opened 2015-01-19
Added new module win_computername to set a Windows computer name

wimnat:
This module is similar to the hostname module.  It is used to set the Windows computer name.  If the name is already set to the same value then consistency is maintained.  In order to complete the name change a reboot of the system is required but this is not part of the module.


wimnat:
Closed PR as functionality already submitted in PR in extras.

Issue #551, opened 2014-12-22
Can't delete Route53 CNAME record


smiller171:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8.2
  configured module search path = None
##### Environment:

Running from Ubuntu 14.04 Ansible Tower against AWS API
##### Summary:

When attempting to delete a Route53 CNAME record, (get and then delete) using the route53 module, it always fails saying the values provided do not match.
##### Steps To Reproduce:

```

---
- name: get record
  route53:
    command: get
    zone: foo.com
    record: "{{stack}}.foo.com"
    type: CNAME
  register: rec

- name: debug rec var
  debug: var=rec

- name: Delete URL
  route53:
    command: delete
    zone: foo.com
    record: "{{ rec.set.record }}"
    type: "{{ rec.set.type }}"
    value: "{{ rec.set.value }}"
```
##### Expected Results:

```
TASK: [delete_workbench | get record] ***************************************** 
ok: [localhost]

TASK: [delete_workbench | debug rec var] ************************************** 
ok: [localhost] => {
    "rec": {
        "changed": false, 
        "invocation": {
            "module_args": "", 
            "module_name": "route53"
        }, 
        "set": {
            "record": "ansible-2.projectcobe.com.", 
            "ttl": "300", 
            "type": "CNAME", 
            "value": "ansible-2-workbench-ELB-752533913.us-east-1.elb.amazonaws.com", 
            "values": [
                "ansible-2-workbench-ELB-752533913.us-east-1.elb.amazonaws.com"
            ], 
            "zone": "projectcobe.com."
        }
    }
}

TASK: [delete_workbench | Delete URL] ***************************************** 
changed: [localhost]

PLAY RECAP ******************************************************************** 

localhost                  : ok=2    changed=1    unreachable=0    failed=0

##### Actual Results:

TASK: [delete_workbench | get record] ***************************************** 
ok: [localhost]

TASK: [delete_workbench | debug rec var] ************************************** 
ok: [localhost] => {
    "rec": {
        "changed": false, 
        "invocation": {
            "module_args": "", 
            "module_name": "route53"
        }, 
        "set": {
            "record": "ansible-2.projectcobe.com.", 
            "ttl": "300", 
            "type": "CNAME", 
            "value": "ansible-2-workbench-ELB-752533913.us-east-1.elb.amazonaws.com", 
            "values": [
[...]

PR #659, opened 2015-01-19
We don't need (or want) to specify the TTL for route53 delete command

alexclifford:


alexclifford:
After further discussion on https://github.com/ansible/ansible-modules-core/issues/551 we determined the docs need adjusting instead of a module code change.

PR #670, opened 2015-01-20
Docker, pulling image before attempting to create

marqc:
Added option "pull_latest" that pulls docker image from registry before attempting to create container. Usable when creating containers from images that has been pushed to registry under same tag as earlier (already pulled) builds of image (eg. for tags latest, staging, production).

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -231,6 +231,13 @@ options:
     default: false
     aliases: []
     version_added: "1.9"
+  pull_latest:
+    description:
+      - Pull current version of image from registry before container create. Use when using single tag for different versions (eg. latest, production)
+    required: false
+    default: false
+    aliases: []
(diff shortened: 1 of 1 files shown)


bcoca:
would this not make more sense in the docker_image module?

sivel:
I believe `docker_image` is deprecated now.

fabiosussetto:
+1 for this, to me it's not clear that when you use the docker module with state=running or present the new image is not pulled automatically. I was assuming that it was pulling the :latest implicit version (as Docker does)

abadger:
Does the pull=always parameter of this pull request satisfy your needs? https://github.com/ansible/ansible-modules-core/pull/502

It's now merged into devel and the stable-1.9 branch

marqc:
Yes, pull always is what i needed. Thanks.

PR #671, opened 2015-01-20
Docker, extended checking of changes in named conainers

marqc:
Checking if running named container has changed (is different than defined in ansible playbook docker module entry). Checks image repository+name+tag, declared envs, mapped ports and mounted volumes.

In my current working environment, when i have over 15 different docker images and ansible playbooks to manage them, it saves a lot of time, when i'm only changing envs of container or change port mapping.

Repository+name+tag - must be exact match
Envs - running container must have all envs defined in playbook
Mapped ports - running container must have all mappings defined in playbook
Volumes - must be exact match

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -821,7 +821,7 @@ class DockerManager(object):
             if conf_in_container_port in container_ports:
                 container_host_info = container_ports[conf_in_container_port]
                 #nullcheck when port is exposed in Dockerfile but not mapped to host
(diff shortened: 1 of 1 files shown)


abadger:
Does the state=reloaded option from https://github.com/ansible/ansible-modules-core/pull/502 (merged into devel and the stable-1.9 branch) suit your needs here?

marqc:
Yes, that is exactly what i needed.

PR #675, opened 2015-01-20
now handles non string values for sysctl

bcoca:
should fix issues with passing value: # and other non string types.


abadger:
+1 good to merge.

PR #678, opened 2015-01-21
Handle the case where a Cloudformation stack rollback fails.

bracki:
In rare cases a stack rollback might fail.
Currently this will lead to ansible sleeping indefinitely.
This PR fixes this issue.


tedder:
How is this not [covered by the condition directly above it](https://github.com/Jimdo/ansible-modules-core/blob/e2abbbefd7d654f7ee05e18828fc1d0d1427a67b/cloud/amazon/cloudformation.py#L183)?

bracki:
Because `UPDATE_FAILED != UPDATE_ROLLBACK_FAILED`.

tedder:
aha, yep, I misread the match on _FAILED above it. Thought it was a regex.

PR #686, opened 2015-01-22
Remove --force-yes from install_deb options

### LINKED ISSUE #327: apt: Cannot downgrade a deb= package
Issue Type: Bug report
Ansible Version: 1.6 (but module doesn't appear to have changed on latest) ubuntu 12.04
Summary:
If running the following code to downgrade a package, it fails with the error below:

Steps To Reproduce:
Install a newer version of package.deb, then use ansible to install an older one directly from the deb file.

```
- apt: deb=/path/to/old package.deb state=installed force=yes dpkg_options=force-confdef,force-confold,force-downgrade
```

Expected Results:
It installs the debian package, as force-downgrade is specified

Actual Results:

```
failed: [host] -> {"failed": true, "item": ""}
msg: A later version is already installed
```

I think the check for if the package is a newer version happens before the force-downgrade is passed onto dpkg. In apt.debfile.DebPackage(deb_file).check() it looks to check if the version is older, and returns false if it is. I can't see an option to override this in ansible

Rizzly:
Dpkg does not have the option "--force-yes", so don't supply it. Fixes https://github.com/ansible/ansible-modules-core/issues/327.

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -382,8 +382,6 @@ def install_deb(m, debs, cache, force, install_recommends, dpkg_options):
         options = ' '.join(["--%s"% x for x in dpkg_options.split(",")])
         if m.check_mode:
             options += " --simulate"
-        if force:
-            options += " --force-yes"
 
         cmd = "dpkg %s -i %s" % (options, " ".join(pkgs_to_install))
         rc, out, err = m.run_command(cmd)


gregdek:
Hi @Rizzly. hanks for submitting this pull request. Please accept our apologies for taking so long to review it.

Rizzly:
@gregdek No problem! It seems as though the change in this commit https://github.com/ansible/ansible-modules-core/pull/53 fixed the problem I was experiencing and also does it in a better way.

It's ok to close this PR and as far as I can understand https://github.com/ansible/ansible-modules-core/issues/327 can be marked as resolved.

gregdek:
Thanks for your patience, @Rizzly.

PR #687, opened 2015-01-22
get_url: Force redownload if sha256sum does not match

### LINKED ISSUE #625: sha256sum in get_url doesn't check hash of files that already exist
If a sha256sum is specified, I would assume that the file's digest will be checked even if the file already exists on the server prior to downloading. At present, get_url makes no attempt to check if the digest of the pre-existing file matches the provided digest or not and will only re-download if force is specified.

Using force=yes isn't really an acceptable workaround because force=yes will cause extra network traffic in every ansible run, when we should only need to go to the network if the expected sha256sum differs from the current sha256sum.

This is a migration of ansible/ansible#9097.

abackstrom:
This patch to `get_url` forces a redownload if a sha256sum is provided and a dest file exists that does not match that checksum. The 'force' argument is still respected and will force regardless of checksum. Fixes #625.

### DIFF
--- a/network/basics/get_url.py
+++ b/network/basics/get_url.py
@@ -217,8 +217,29 @@ def main():
     dest_is_dir = os.path.isdir(dest)
     last_mod_time = None
 
+    # Remove any non-alphanumeric characters, including the infamous
+    # Unicode zero-width space
+    stripped_sha256sum = re.sub(r'\W+', '', sha256sum)
+
+    # Fail early if sha256 is not supported
+    if sha256sum != '' and not HAS_HASHLIB:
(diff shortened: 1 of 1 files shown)


muffl0n:
+1 for this

bts:
I currently have to work around this bug in `get_url`. +1!

sylvaingi:
+1

phobologic:
:+1:

gregdek:
Lots of folks like this one. @jpmens can you review? If you give +1 we will promote to P1_reviewed.

jpmens:
Untested but looks ok to me. +1

PR #691, opened 2015-01-23
Openbsd rcctl

jarmani:
Update OpenBSD rcctl handling to use new syntax, special thanks to @eest


eest:
This change has already made it into the OpenBSD port of ansible: http://marc.info/?l=openbsd-ports-cvs&m=142206037411721&w=2

PR #696, opened 2015-01-24
extend ec2_elb module to view LB health.

tedder:
see previous ugly merge and discussion #25 

and previous to that https://github.com/ansible/ansible/pull/7577

tag @bcoca to continue discussion


bennojoy:
@tedder sorry for the delay here, can you please resolve conflicts.

tedder:
@bcoca rebased, no conflicts now.

bcoca:
for returning facts lets make a facts module instead of overloading state in a bad way

tedder:
@bcoca your previous thoughts on that seemed to vacillate: https://github.com/ansible/ansible-modules-core/pull/25#issuecomment-70940423

I'm tired of babysitting this pull request. It's been over a year. I'll close it and continue to use it locally.

bcoca:
I'm sorry you feel that way,  we do have a backlog and many requests that have been living here for long. There is just so much we can do about that, things have been improving and our merge rate has gone up a lot and we try to do more and better things every day, this has been discussed elsewhere so I'm not going to dwell on it.

As for vacillating, I was trying to reach a compromise, which you disagreed with, that took me back to my original position of moving this into a fact module. Specially since there is another PR that also adds elb data gathering to this module, my thought was to merge both into ec2_elb_facts.

Yes, there are modules that have this view/query/list pattern already, from which we learnt this is not a good one and why we are trying to get away from it.

tedder:
"I'm sorry you feel that way" .. "this has been discussed elsewhere so I'm not going to dwell on it."

Message received.

bcoca:
@tedder don't rearrange my response to say something I did not. I just tried to keep it succinct as we have been commenting left and right the initiatives we have undertaken to get through the backlog and mitigate wait times, I did not think they needed repeating here. That does not mean I'm ignoring you or this PR, it only reflected my concern on being too repetitive.

I understand you are frustrated, but I don't think you should try this hard to look for a dismisiveness  that does not exist. If we really did not care this ticket would not have any responses and I would not still be trying to merge your work with #1244 even after you closed the ticket.

PR #703, opened 2015-01-26
postgresql_privs: Fix quoting of function identifiers

johaness:
For `type=function`, there is already specific quoting code further up the function. Input like `objs=function_name(int:varchar)` is correctly transformed into `"schema_name"."function_name"(int,varchar)`.

Quoting again with `pg_quote_identifier` breaks with "User escaped identifiers must escape extra quotes" and is unnecessary, so removing in this PR.

### DIFF
--- a/database/postgresql/postgresql_privs.py
+++ b/database/postgresql/postgresql_privs.py
@@ -474,10 +474,13 @@ class Connection(object):
         if obj_type == 'group':
             set_what = ','.join(pg_quote_identifier(i, 'role') for i in obj_ids)
         else:
+            # function types are already quoted above
+            if obj_type != 'function':
+                obj_ids = [pg_quote_identifier(i, 'table') for i in obj_ids]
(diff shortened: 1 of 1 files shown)


abadger:
This bug would seem to affect other types than just functions wouldn't it?  (tables and such as well)?  Perhaps the two if-then-else's should be merged together so that quoting of obj_ids and setting set_what are both done in the same if-then-else?

johaness:
The bug seems to be triggered by the trailing function parameters -- `(int:varchar)` in the example above -- so other types should not be affected.

abadger:
okay.  Add a comment about the function types being quoted in the code just above being the reason that we're skipping it now and I think this would be good to merge then.

johaness:
added comment as suggested

abadger:
Thanks for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

Issue #349, opened 2014-11-18
Add support for installing apt build dependencies. 


graingert:
@nathanhruby created a PR to add support for installing apt build dependencies. 

Is it worth attempting to reintegrate this https://github.com/ansible/ansible/pull/8583 ?

abadger:
Yes, if you're willing to take a look at doing it.  It was somewhat lower priority than other things we had to get merged in but it does seem like a useful feature to support.

abadger:
I'll go ahead and close this issue but feel free to open a pull request if you get the code rebased.

PR #704, opened 2015-01-26
added: build-dep to apt module #349

conrado:
Original code from @nathanhruby on ansible/ansible#8583 rebased yet again

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -511,7 +511,7 @@ def main():
             dpkg_options = dict(default=DPKG_OPTIONS)
         ),
         mutually_exclusive = [['package', 'upgrade', 'deb']],
-        required_one_of = [['package', 'upgrade', 'update_cache', 'build-dep', 'deb']],
+        required_one_of = [['package', 'upgrade', 'update_cache', 'deb']],
         supports_check_mode = True
     )
 
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -82,6 +82,12 @@ options:
     required: false
     default: "yes"
     choices: [ "yes", "safe", "full", "dist"]
+  build_dep:
+    description:
+      - Instead, install the build dependencies for the named pkg (equivalent to 'apt-get build-dep foo')
+    required: false
+    default: "no"
+    choises: [ "yes", "no" ]
   dpkg_options:
     description:
       - Add dpkg options to apt command. Defaults to '-o "Dpkg::Options::=--force-confdef" -o "Dpkg::Options::=--force-confold"'
@@ -133,6 +139,9 @@ EXAMPLES = '''
 
 # Install a .deb package
 - apt: deb=/tmp/mypackage.deb
+
+# Install the build dependencies for package "foo"
+- apt: pkg=foo build_dep=yes
 '''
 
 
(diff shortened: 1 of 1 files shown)


abadger:
Thanks @conrado, @nathanhruby /me puts this on his list of things to review for merging.

packaging/os/apt.py
```diff
@@ -82,6 +82,12 @@
     required: false
     default: "yes"
     choices: [ "yes", "safe", "full", "dist"]
+  build_dep:
+    description:
+      - Instead, install the build dependencies for the named pkg (equivalent to 'apt-get build-dep foo')
+    required: false
+    default: "no"
+    choises: [ "yes", "no" ]
```
abadger: s/choises/choices/

packaging/os/apt.py
```diff
@@ -497,10 +537,11 @@ def main():
             install_recommends = dict(default='yes', aliases=['install-recommends'], type='bool'),
             force = dict(default='no', type='bool'),
             upgrade = dict(choices=['yes', 'safe', 'full', 'dist']),
+            build_dep = dict(default='no', type='bool'),
             dpkg_options = dict(default=DPKG_OPTIONS)
         ),
         mutually_exclusive = [['package', 'upgrade', 'deb']],
-        required_one_of = [['package', 'upgrade', 'update_cache', 'deb']],
+        required_one_of = [['package', 'upgrade', 'update_cache', 'build-dep', 'deb']],
```
abadger: I don't think we want build-dep in here because specifying build-dep by itself is not sufficient for the module to do anything.

packaging/os/apt.py
```diff
@@ -82,6 +82,12 @@
     required: false
     default: "yes"
     choices: [ "yes", "safe", "full", "dist"]
+  build_dep:
+    description:
+      - Instead, install the build dependencies for the named pkg (equivalent to 'apt-get build-dep foo')
```
abadger: s/Instead, install/Install/

packaging/os/apt.py
```diff
@@ -484,6 +493,37 @@ def upgrade(m, mode="yes", force=False, default_release=None,
...
+    else:
+        force_yes = ''
+
+    packages = ''
+    pkgspec = expand_pkgspec_from_fnmatches(m, pkgspec, cache)
+    for package in pkgspec:
+        name, version = package_split(package)
+        packages += "'%s' " % package
+    if len(packages) == 0:
+        m.exit_json(changed=False)
```
abadger: With the current parameter setup, using exit_json() in this function would be a bug.  The current parameter setup is that installing build-deps is an optional addition to whatever else we're doing.  So we would have to return to the main function to continue to the next step unless build-dep fails.

abadger:
Two higher level comments:
- Looking at the build-dep function, I can see that it shares most of its code in common with def install().  Could we refactor this to share the common code (either via one function with conditionals and an extra parameter to select whether we're installing packages or installing build-deps of packages or by pulling common pieces of code into their own functions)?
- Thinking about the use case for build-dep, do we want people to be installing or updating packages and installing the build-deps of those packages in one ansible task?  Since apt-get has build-dep and install be two separate steps I kinda think this should be two tasks as well.  That would mean that instead of a new toplevel parameter (build_dep=yes/no) we should instead make this a new value of state.  So you would do:

```
ansible all -m apt -a 'pkg=foo state=build-dep'
```

To install build deps.  Does that sound reasonable?

conrado:
@abadger proposed changes made

packaging/os/apt.py
abadger: Still need to remove build-dep from the required_one_of line.

packaging/os/apt.py
```diff
@@ -595,20 +606,13 @@ def main():
             if latest and '=' in package:
                 module.fail_json(msg='version number inconsistent with state=latest: %s' % package)
 
-        if p['state'] == 'latest':
+        if p['state'] in ('latest', 'present', 'build-dep'):
+            build_dep = p.get('build-dep', False)
```
abadger: Need to set build_dep depending on state.  Also need to set upgrade depending on whether we use latest or present.  Probable some code like this:

``` python
build_dep = False
upgrade = False
if p['state'] == 'build-dep':
    build_dep = True
elif p['state'] == 'latest':
    upgrade = True
```

Then make use of build_dep and upgrade in the call to install().

abadger:
Cool.  Just one last thing to fix here.

Oh and one other thing -- the original pull request had some additional tests to go along with the new functionality.  If you want to create a pull request in the ansible/ansible repository for those I'd be happy to merge them at the same time as I merge this.

conrado:
I have looked at that, the original tests seem to still need to go into the ansible/ansible repository.

conrado:
Have looked into that, some other integration tests seem broken. Just out of curiosity, are there any plans for moving them into these repositories?

abadger:
@conrado, i think everyone's agreed that it would be desirable for integration tests to be near the modules they test but no one's looked into how we might best make that happen yet.  I'm not even sure of what all of the concerns regarding that would be yet.

abadger:
@conrado and  @nathanhruby thank you for your work and for your patience!  Merged!

# Merged

Hi!

issue is resolved.

Thank you!

nathanhruby:
much thanks @conrado and @abadger !

PR #707, opened 2015-01-27
Document a way to delete an EC2 group

alexclifford:
I launch test environments and then delete them a lot and found this was a quick and easy way to delete the security groups (which is required before the VPCs will delete). I wouldn't always know what the VPC id was but I'd have the region, VPC cidr block, and group name all set in variables prior to creation.

Alter it if you'd like, but I think some sort of delete documentation would be handy for others.


herbygillot:
Should documentation reference a module that does not exist in Ansible official (`get_vpc_ids`)?  Correct me if I'm wrong, thanks.

PR #720, opened 2015-01-29
keystone_user role improvements

gandelman-a:
This adds some missing functionality around managing roles.  Specifically,
it fills in the unimplemented functions that allow roles to be added/removed
to user+tenant pairs.  This is distinct from managing roles themselves, so
those functions have been renamed to ensure_user_role_{exists, absent}.
The ability to manage roles themselves has been added here as well via
new ensure_role_{exists, absent} functions.  Doc examples updated.


sivel:
cc @emonty

emonty:
Closing in favor of new modules, which @gandelman-a is already working on

PR #721, opened 2015-01-29
update git remote url before ls-remote

jinowolski:
related to https://github.com/ansible/ansible/issues/8177

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -473,9 +473,19 @@ def get_head_branch(git_path, module, dest, remote, bare=False):
     f.close()
     return branch
 
-def fetch(git_path, module, repo, dest, version, remote, bare, refspec):
+def set_remote_url(git_path, module, repo, dest, remote):
     ''' updates repo from remote sources '''
     commands = [("set a new url %s for %s" % (repo, remote), [git_path, 'remote', 'set-url', remote, repo])]
+
+    for (label,command) in commands:
+        (rc,out,err) = module.run_command(command, cwd=dest)
+        if rc != 0:
+            module.fail_json(msg="Failed to %s: %s %s" % (label, out, err))
+
+def fetch(git_path, module, repo, dest, version, remote, bare, refspec):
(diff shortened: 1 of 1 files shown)


gregdek:
This module is maintained by the core team. @abadger @bcoca @jimi-c @bennojoy please review when you can.

robinro:
:+1:
This fix is really useful, please merge it. People change repo-urls all the time and in the current state this is broken.

bcoca:
cc @abadger

abadger:
This is an area of git that I don't usually work with but the code looks good.  Would anyone be interested in writing an integration test for this?  The git module in general is filled with complex code for corner cases so integration tests are very useful to be sure it doesn't regress.  (an integration test is basically just a few playbook tasks that set up the conditions for the error to occur and then a task that would provoke the error if the problem recurs.)  they live in the main ansible repo at https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_git/tasks/main.yml

abadger:
@robinro Would you be willing to write an integration test?  Or even just give us a series of playbook tasks that trigger this behaviour that someone else can adapt when they have time?

gregdek:
@abadger is lack of an integration test preventing us from merging this?

abadger:
@gregdek It is what I'm waiting on... Since no one's replied since I asked for it... I'm not sure whether we should merge or continue waiting.  All we need is a set of playbook tasks that fails with the current code and succeed with this patch applied.  Right now I'm not sure what the precise reproducer is so I'm afraid this code could accidentally be rebroken when we make changes to the module in the future.

robinro:
@abadger @gregdek I can look into writing an integration test, but probably not within the next 3 weeks. It should be rather straight forward:
- clone repo from URL1
- clone repo from URL2 
- make sure .git/config contains URL2 and not URL1
- maybe also check that the checkout changed

minglecm:
Not sure on Ansible release cycle specifics, but any idea when this fix will make it into a non 2.0 release?

PR #729, opened 2015-01-30
[ec2_vol] Find or Create volume by name

Jell:
This PR is a working proposal for a "Find or create volume by name". The change is straightforward: if volume name is given, but no volume match that name, create a volume with the name then proceed as usual.

This makes idempotency more trivial, and avoids unnecessary complexity to idempotently create a volume before attaching it to an instance.

### DIFF
--- a/cloud/amazon/ec2_vol.py
+++ b/cloud/amazon/ec2_vol.py
@@ -143,9 +143,9 @@ EXAMPLES = '''
     with_items: ec2.instances
   register: ec2_vol
 
-# Example: Launch an instance and then add a volume if not already present
+# Example: Launch an instance and then add a volume if not already attached
+#   * Volume will be created with the given name if not already created.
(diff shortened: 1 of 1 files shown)


luontola:
This feature would be excellent. For now I had to avoid using ec2_vol, because it's not idempotent.

My use case would have been to use a separate volume for storing our CI's data and configuration, so that we could easily recreate the CI server without losing the data. But since it's too hard with ec2_vol, I guess I'll have to resort to manually attaching the old volume if I recreate the server (I'm creating a new volume with `ec2` module's `volumes` parameter with `delete_on_termination` disabled).

bcoca:
fyi, this will need a rebase

bennojoy:
@Jell Could you please rebase this..

Jell:
rebased!

bennojoy:
Merged thanks

PR #738, opened 2015-01-31
Vsphere linkedclone

NicCOConnor:
Added the ability in the vsphere_guest module to linked_clone from a snapshot. all tests on my end worked, Would love some feedback or a merge if everything is good.


NicCOConnor:
There are other commits in this pull request that are unintended. however, I'm not really sure how to get rid of them.

PR #739, opened 2015-01-31
Added the ability to Linked_clone from snapshot in vsphere_guest module

NicCOConnor:


gregdek:
@rhoop please review; if you +1 we will tag P1_reviewed for faster inclusion.

rhoop:
I'll review today.

rhoop:
@NicCOConnor Which version of vSphere was this tested on?

NicCOConnor:
5.5 but I have the ability to test on 6.

rhoop:
I only have 5.5 and it works. If you can test on 6 I will approve.

NicCOConnor:
Tested on vSphere 6 this morning, works like a charm. no issues.

abadger:
- From the code it looks like if linked_clone is set to True then snapshot should be set too.  We need to check that and return a good error message if it's not set together.  Probably best to check it in main soon after we instantiate the AnsibleModule().  Since this won't be available until Ansible-2.0 You can probably use required_if to do this.  Something like:
  `required_if=[('linked_clone', True, ['snapshot'])],`
- Is there a reason for using `choices=BOOLEANS` instead of type='bool' ?  If not go ahead and change it; either you or I could update the unrelated parameters to use type='bool'.
- The default for linked_clone should probably be False.  That seems to match the logic and would make better documentation for end users.
- version_added in documentation needs to be updated to "2.0"

NicCOConnor:
- There is a check starting at line 1187

```
mutually_exclusive=[['state', 'vmware_guest_facts'],['state', 'from_template']],
        required_together=[
            ['state', 'force'],
            [
                'state',
                'vm_disk',
                'vm_nic',
                'vm_hardware',
                'esxi'
            ],
            ['resource_pool', 'cluster'],
            ['from_template', 'resource_pool', 'template_src'],
            ['linked_clone', 'snapshot']
        ],
    )
```

that is a `required_together` which provides a nice error message if one is defined without the other. I thought this would suffice 
- `choices=BOOLEANS` was used simply becuase that is what the original author used and I just copied it, I have no issues changing it. 
- Assuming your talking about the documentation portion, Made the change and will be included as soon as I push 
- version_added changed and will be included with my next push

abadger:
- required_together: I had missed that.  However, I'm not sure if required_together works in this case.  Could you test whether specifying linked_clone=False explicitly and no snapshot parameter works as expected? (ie: that combination should not error out due to linked_clone being set without also setting snapshot).
- choices=BOOLEAN.  Cool.  Yeah, it's likely old coding style vs new one.  type='bool' is the newer, hopefully clearer, way to do it.
- For the linked_clone default -- I was thinking both in the documentation and in the argument_spec.

NicCOConnor:
@abadger so looking into this further. I've done some test on the code. Your third bullet point "setting default value to false" seems to break the mutually exclusive check. Meaning that if one is set and the other is not then the default value would take over nullifying the check. I've not tried it with `required_if=[('linked_clone', True, ['snapshot'])],` but I think it might be the same. thoughts?

abadger:
Note: mutually_exclusive should be separate from required_together and required_if.

I'm about to board a plane but from looking at the code in basic.py that implements these checks I think required_together will break as you say but required_if should work.

abadger:
@NicCOConnor Confirmed -- required_together fails with a default value other than None but required_if works because it is checking whether the value is True.

NicCOConnor:
I could not get required_if to work. It kept failing saying that the snapshot value was missing. perhaps my syntax was wrong. Would like to see your code so I can make the appropriate changes.

abadger:
@NicCOConnor Ah, I found a bug in the required_if code.  Let me push a fix for that.

abadger:
@NicCOConnor fixed the required_if code here: https://github.com/ansible/ansible/commit/dd058a1dc283ae6b4fd627ef14225be73d6bd5b8

If you test required_if now it should work.

NicCOConnor:
Everything is working. I'm just trying to figure out how to rebase and push these changes up cleanly.

NicCOConnor:
@abadger Looks like everything is good to go and the changes you requested have been added. Let me know if there is anything else I can do for this pull request.

abadger:
@NicCOConnor Sorry about this -- I took another look and I have a question -- is there a reason that we have both linked_clone and snapshot?  The logic seems like we could remove linked_clone and just decide what to do based on whether the user specified a snapshot or not( `if snapshot is None:`.  (Perhaps we should name the snapshot parameter linked_clone or snapshot_to_clone though, to make it clearer what it's doing).

If you think that's fine, I can make the change if you want -- I feel bad about noticing this so late in the review of the feature.

NicCOConnor:
My logic was based mostly on the template portion of the code. Where you specify  `from_template: true` and then `template_src: some_template` I like the idea of using snapshot_to_clone: which seems more more clean and one less variable to identify. let me see if I can work it in real quick.

NicCOConnor:
@abadger, ok let me know what you think of these changes.

NicCOConnor:
@abadger, Just pinging you to see if you've had a chance to look at this yet.

abadger:
@NicCOConnor Looks great!  I'm going to merge.  I'll make some comments on the changes in case you decide to make more module changes in the future but for this PR I'm going to make the changes after merging so you don't have to wait any longer ;-)

cloud/vmware/vsphere_guest.py
```diff
@@ -79,6 +79,12 @@
     description:
       - Name of the source template to deploy from
     default: None
+  snapshot_to_clone:
+    version_added 2.0
```
abadger, line 83: Need to put the 2.0 in quotes otherwise the doc formatter thinks that 2.0 is a float instead of a string.

cloud/vmware/vsphere_guest.py
```diff
@@ -545,9 +551,14 @@ def deploy_template(vsphere_client, guest, resource_pool, template_src, esxi, mo
     try:
         if vmTarget:
             changed = False
+        elif snapshot_to_clone != None:
```
abadger, line 554: When testing whether a variable is None idiomatic python uses `if snapshot_to_clone is None`.  The reason is that None is a singleton (https://en.wikipedia.org/wiki/Singleton_pattern ) so we use a test to see if they are the same object instead of a test that the values are equal.

abadger:
@NicCOConnor Thanks for your work on this and your patience!

# Merged

Hi!

issue is resolved.

Thank you!

PR #747, opened 2015-02-04
Add ec2 instance tags and groups info

barraponto:
### Issue Type

Feature
### Summary

Just leaving these details in would make it easier to setup `add_host` tasks to mimic `ec2.py` dynamic inventory or other tasks using registered variables from the `ec2` module.


jimi-c:
Merged, thank you!

PR #750, opened 2015-02-05
add wait_for state

tedder:
much like the wait_for module, state=wait_for in ec2_elb waits for the ELB health check on a given instance to reach a given state. This is different than waiting for the instance to be connectable.

This includes state=view, which has been pending since May 2014. Its pull request history:
- originally in a pull request here: https://github.com/ansible/ansible/pull/7577
- in a carryover pull request here: #25
- finally added cleanly in this pull request: #696


tedder:
hmm. on second thought, this isn't great, as you can't wait for the health status of an instance that isn't in the ELB yet. So it's not very helpful; the `wait` and `wait_timeout` do more.

PR #753, opened 2015-02-06
Don't use 'tenancy' in ec2 spot requests

skyluc:
The problem was introduced in commit f5789e8. 'tenancy' is a parameter of ec2.run_instances, but not in ec2.request_spot_instances. So it was breaking the support for spot requests.

This pull request is based on #597, which is fixing a different problem with ec2 spot requests.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -834,7 +834,8 @@ def create_instances(module, ec2, override_count=None):
             if ebs_optimized:
               params['ebs_optimized'] = ebs_optimized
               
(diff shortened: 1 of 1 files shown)


jbbarth:
+1 for that, I confirm the problem and that the solution works.

Fwiw you will note that "wait" mode is also totally broken with spot instances, I have some work on that side going on.

skyluc:
Rebased on devel, after the merge of #597

sanchom:
This isn't included in pulls from ansible:devel yet because of old submodule references.

Issue #207, opened 2014-10-21
add support for routing policy to route53 module


wenottingham:
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.7.2
##### Environment:

Ubuntu 14.04
##### Summary:

route53 module does not let you set routing policy. http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy.html

(Filing on behalf of user).

j1n6:
+1

j1n6:
@jimi-c would you accept PR?

bcoca:
we promise to review it, if you submit it

lamielle:
This would be awesome.  I'd like to update a weighted routing policy with the hosts of number of load balancers.  The number may change over time, so not having to hit the AWS console to update this config is pretty important.

tcouper:
+1

fstehle:
:+1:

unixunion:
+1

zimbatm:
Just a heads up that I'm working on this. I'll be submitting a PR when I have something working. In the mean time, feel free to ping me if you want to help out.

zimbatm:
Okay it's done. The PR also contains changes that improves the module's performance.

bennojoy:
@zimbatm  can you please squash to a singe commit thanks,,,

zimbatm:
@bennojoy I've squashed the first commit so the last commit is all about introducing the routing policies. The previous commits all logical commits that stand on their own and introduce optimisations

Issue #473, opened 2014-12-09
route53 module does not differentiate between Private and Public hosted zones


dkaufman-irise:
# Issue Type:

Bug Report
# Ansible Version:

ansible 1.9 (devel 2bf269568b) last updated 2014/12/03 07:59:53 (GMT +000)
lib/ansible/modules/core: (detached HEAD 3a80b73) last updated 2014/12/03 07:59:53 (GMT +000)
lib/ansible/modules/extras: (detached HEAD 317654dba5) last updated 2014/12/03 07:59:54 (GMT +000)
v2/ansible/modules/core: (detached HEAD cb69744) last updated 2014/10/28 01:26:59 (GMT +000)
v2/ansible/modules/extras: (detached HEAD 8a4f07eecd) last updated 2014/10/28 01:27:04 (GMT +000)
configured module search path = $HOME/repos/ansible/lib/
# Environment:

Amazon Linux AMI 2014.09
# Summary:

route53 modules fails to correctly create/modify/delete DNS records when both a public and private zone exist for the same DNS zone.
# Steps To Reproduce:

create public and private zone in Route53 for same domain, attempt to update one or the other using the route53 module.
# Expected Results:

zone file is created/updated in correct zone.
# Actual Results:

private zone seems to be preferentially chosen, no way of specifying private or public zone.

ansibot:
# Can You Help Us Out?

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

It looks like you might not have filled out the issue description based on our standard issue template.  You might not have known about that, and that's ok too, we'll tell you how to do it.

We have a standard template because Ansible is a really busy project and it helps to have some standard information in each ticket, and GitHub doesn't yet provide a standard facility to do this like some other bug trackers.  We hope you understand as this is really valuable to us!. 

Solving this is simple: please copy the contents of this [template](https://raw.githubusercontent.com/ansible/ansible/devel/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket.  That's it!

# If You Had A Question To Ask Instead

If you happened to have a "how do I do this in Ansible" type of question, that's probably more of a user-list question than a bug report, and you should probably ask this question on the project mailing list instead.  

However, if you think you have a bug, the report is the way to go!  We definitely want all the bugs filed :)  Just trying to help!

# About Priority Tags

Since you're here, we'll also share some useful information at this time.

[...]

PR #754, opened 2015-02-06
Support private zones in route53

jjshoe:


jjshoe:
fixes #207 and #473

robbiet480:
:+1:

cloud/amazon/route53.py
```diff
@@ -84,6 +84,11 @@
     required: false
     default: 500
     aliases: []
+  private_zone:
+    description: 
+      - When you have both a public and private zone you need to pass in a true value to modify any records in the private zone.
+    required: false
+    default: false
```
abadger, line 91: More clear to say something like "Set this to True to operate on records within a private zone.  False, the default, operates on public zones."

Also need to have a version_added entry to say that this was added in ansible 1.9.

cloud/amazon/route53.py
```diff
@@ -218,8 +224,9 @@ def main():
     zones = {}
     results = conn.get_all_hosted_zones()
     for r53zone in results['ListHostedZonesResponse']['HostedZones']:
-        zone_id = r53zone['Id'].replace('/hostedzone/', '')
-        zones[r53zone['Name']] = zone_id
+        if (r53zone['Config']['PrivateZone'] == 'false' and not module.params['private_zone']) or (r53zone['Config']['PrivateZone'] == 'true' and module.params['private_zone']):
+            zone_id = r53zone['Id'].replace('/hostedzone/', '')
```
abadger, line 228: It would fit the style of ansible modules more to set a variable around line 189 that we could compare here. Something like:

``` python
     retry_interval_in     = module.params.get('retry_interval')
     if module.params.get('private_zone'):
         private_zone_in = 'true'
     else:
         private_zone_in = 'false'
[...]
         if r53zone['Config']['PrivateZone'] == private_zone_in:
[...]
```

abadger:
Does this really fix #207?  It loks to me like it only implements private zone selection (#473)

jjshoe:
What I have works for my company, I'll let someone else battle it out for now.

srgvg:
That is not how you build up credibility, Joel.

jjshoe:
@srvg ok mom

srgvg:
Q.E.D.

jjshoe:
@srvg ok mom

jjshoe:
You could take this PR, clean it up, and re-submit it, or you could continue to try to impose yourself on me. I guess you don't lead by example, though.

PR #756, opened 2015-02-06
iam_module

juliedavila:
This is a resubmission of https://github.com/ansible/ansible-modules-core/pull/590 in order to have a clean commit history. 

There is currently no IAM module. This is part 1 of 3 that I plan to make.

Part 1: Create/Update/Delete users, groups and roles
Part 2: Create/Update/Delete policy objects for users/groups/roles See https://github.com/ansible/ansible-modules-core/pull/887
Part 3: Upload/Remove/Update server certificates. See #918 

### DIFF
--- a/cloud/amazon/iam.py
+++ b/cloud/amazon/iam.py
@@ -1,0 +1,726 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: iam
+short_description: Manage IAM users, groups, roles and keys
+description:
+     - Allows for the management of IAM users, groups, roles and access keys.
+version_added: "2.0"
+options:
+  iam_type:
+    description:
+      - Type of IAM resource
+    required: true
+    default: null
+    choices: [ "user", "group", "role"]
+    aliases: []
+  name:
+    description:
+      - Name of IAM resource to create or identify
+    required: true
+    aliases: []
+  new_name:
+    description:
+      - When state is update, will replace name with new_name on IAM resource
+    required: false
+    aliases: []
+  new_path:
+    description:
+      - When state is update, will replace the path with new_path on the IAM resource
+    required: false
+    aliases: []
+  state:
+    description:
+      - Whether to create, delete or update the IAM resource. Note, roles cannot be updated.
+    required: true
+    default: null
+    choices: [ "present", "absent", "update" ]
+    aliases: []
+  path:
+    description:
+      - When creating or updating, specify the desired path of the resource. If state is present, it will replace the current path to match what is passed in when they do not match.
+    required: false
+    default: "/"
+    aliases: []
+  access_key_state:
+    description:
+      - When type is user, it creates, removes, deactivates or activates a user's access key(s). Note that actions apply only to keys specified.
+    required: false
+    default: null
+    choices: [ "create", "remove", "active", "inactive"]
+    aliases: []
+  key_count:
+    description:
+      - When access_key_state is create it will ensure this quantity of keys are present. Defaults to 1.
+    required: false
+    default: '1'
+    aliases: []
+  access_key_ids:
+    description:
+      - A list of the keys that you want impacted by the access_key_state paramter.
+  groups:
+    description:
+      - A list of groups the user should belong to. When update, will gracefully remove groups not listed.
+    required: false
+    default: null
+    aliases: []
+  password:
+    description:
+      - When type is user and state is present, define the users login password. Also works with update. Note that always returns changed.
+    required: false
+    default: null
+    aliases: []
+  update_password:
+    required: false
+    default: always
+    choices: ['always', 'on_create']
+    description:
+     - C(always) will update passwords if they differ.  C(on_create) will only set the password for newly created users.
+  aws_secret_key:
+    description:
+      - AWS secret key. If not set then the value of the AWS_SECRET_KEY environment variable is used.
+    required: false
+    default: null
+    aliases: [ 'ec2_secret_key', 'secret_key' ]
+  aws_access_key:
+    description:
+      - AWS access key. If not set then the value of the AWS_ACCESS_KEY environment variable is used.
+    required: false
+    default: null
+    aliases: [ 'ec2_access_key', 'access_key' ]
+
+
+requirements: [ "boto" ]
+notes:
+  - 'Currently boto does not support the removal of Managed Policies, the module will error out if your user/group/role has managed policies when you try to do state=absent. They will need to be removed manually.'
+author: Jonathan I. Davila and Paul Seiffert
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Basic user creation example
+tasks:
+- name: Create two new IAM users with API keys
+  iam:
+    iam_type: user
+    name: "{{ item }}"
+    state: present
+    password: "{{ temp_pass }}"
+    access_key_state: create
+  with_items:
+    - jcleese
+    - mpython
+
+# Advanced example, create two new groups and add the pre-existing user
+# jdavila to both groups.
+task:
+- name: Create Two Groups, Mario and Luigi
+  iam:
+    iam_type: group
+    name: "{{ item }}"
+    state: present
+  with_items:
+     - Mario
+     - Luigi
+  register: new_groups
+
+- name:
+  iam:
+    iam_type: user
+    name: jdavila
+    state: update
+    group: "{{ item.created_group.group_name }}"
+  with_items: new_groups.results
+
+'''
+
+import json
+import itertools
+import sys
+try:
+    import boto
+    import boto.iam
+    HAS_BOTO = True
+except ImportError:
+   HAS_BOTO = False
+
+def boto_exception(err):
+    '''generic error message handler'''
+    if hasattr(err, 'error_message'):
+        error = err.error_message
+    elif hasattr(err, 'message'):
+        error = err.message
+    else:
+        error = '%s: %s' % (Exception, err)
+
+    return error
+
+
+def create_user(module, iam, name, pwd, path, key_state, key_count):
+    key_qty = 0
+    keys = []
+    try:
+        user_meta = iam.create_user(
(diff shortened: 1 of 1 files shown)


cloud/amazon/iam.py
```diff
@@ -0,0 +1,539 @@
...
+    required: "false"
+    default: null
+    choices: ["create", "remove", "Active", "Inactive"]
+    aliases: []
+  access_key_ids:
+    description:
+      - a list of the keys that you want impacted by the access_key_state paramter.
+  groups:
+    description:
+      - A list of groups the user should belong to. When update, will gracefully the remove groups not listed.
```
brian-brazil: I think there's an extra 'the' here

bcoca:
several things:
- doc parsing has issues, mostly because all descriptions have to be lists ( start with -) and : is a special character and must be quoted or avoided.
- aws errors are not captured and exceptions are allowed to fall through, these should return in a fail_json call to the module.

```
failed: [localhost] => (item=test42) => {"failed": true, "item": "test42", "parsed": false}
  File "/home/bcoca/.ansible/tmp/ansible-tmp-1424109800.44-46630851071663/iam", line 2335, in <module>                                                                            
    main()                                                                                                                                                                        
  File "/home/bcoca/.ansible/tmp/ansible-tmp-1424109800.44-46630851071663/iam", line 426, in main                                                                                 
    iam, name, password, path, key_state)                                                                                                                                         
  File "/home/bcoca/.ansible/tmp/ansible-tmp-1424109800.44-46630851071663/iam", line 165, in create_user                                                                          
    pwd = iam.create_login_profile(name, pwd)                                                                                                                                     
  File "/usr/lib64/python2.7/site-packages/boto/iam/connection.py", line 954, in create_login_profile                                                                             
    return self.get_response('CreateLoginProfile', params)                                                                                                                        
  File "/usr/lib64/python2.7/site-packages/boto/iam/connection.py", line 102, in get_response                                                                                     
boto.exception.BotoServerError: BotoServerError: 400 Bad Request                                                                                                                  
<ErrorResponse xmlns="https://iam.amazonaws.com/doc/2010-05-08/">                                                                                                                 
  <Error>                                                                                                                                                                         
    <Type>Sender</Type>                                                                                                                                                           
    <Code>PasswordPolicyViolation</Code>                                                                                                                                          
    <Message>Password does not conform to the account password policy.</Message>                                                                                                  
  </Error>                                                                                                                                                                        
  <RequestId>d391166e-b605-11e4-9256-31147537fd9a</RequestId>                                                                                                                     
</ErrorResponse>   
```

also there seems to be an issue with user_groups

```
failed: [localhost] => (item=test42) => {"failed": true, "item": "test42", "parsed": false}
  File "/home/bcoca/.ansible/tmp/ansible-tmp-1424109955.91-207947638917259/iam", line 2340, in <module>
    main()
  File "/home/bcoca/.ansible/tmp/ansible-tmp-1424109955.91-207947638917259/iam", line 447, in main
    user_meta=user, keys=keys, groups=user_groups, changed=changed)
UnboundLocalError: local variable 'user_groups' referenced before assignment
```

this seems mostly bad return when groups don't exist as it is outside the "if" that checks for groups

bcoca:
also, 'delete_all_keys' function is called but it is not defined anywhere

cloud/amazon/iam.py
juliedavila: Maybe I'm going crazy but I dont see the extra 'the' which line specifically are you referring to?

brian-brazil: "will gracefully the remove groups not listed" should probably be "will gracefully remove groups not listed"

juliedavila: Ah! I was looking for 'the the'. Thanks. Fixed. Will update the P/R within the hour with this change and the stuff mentioned below.

juliedavila:
Patched up. Identified some other issues as well which have also been fixed.

juliedavila:
One last issue to address.  With boto, a role/user/group cannot be deleted until all policy objects have been detached from the resource. A bit of a challenge given that boto doesn't support detaching Managed policies. Would need to add a disclaimer in the docs.

bcoca:
ok, add disclaimer (note might work) and I'll be able to give this another look by EOW

juliedavila:
Updated.

rahulmehrotra1:
@defionscode Can you please provide an example for iam role creation using the module in the documentation ?
I am not able to create a role using the following

> name: Creating eventlogger role
> local_action:
> module: iam
> iam_type: role
> name: eventlogger
> state: present

getting the following error

```
TASK: [Creating eventlogger role] ********************************************* 
failed: [localhost -> 127.0.0.1] => {"failed": true, "parsed": false}
  File "/home/rahul/.ansible/tmp/ansible-tmp-1426115965.92-241507198509407/iam", line 2499, in <module>
    main()
  File "/home/rahul/.ansible/tmp/ansible-tmp-1426115965.92-241507198509407/iam", line 694, in main
    iam, name, path, orig_role_list, orig_prof_list)
TypeError: create_role() takes exactly 6 arguments (5 given)
```

juliedavila:
@rhlmhrtr my fault. The way you used it should have created that role. I'll fix that now.

juliedavila:
@rhlmhrtr yep it was a one word fix. Your example should work now.

rahulmehrotra1:
@defionscode awsm, just tried it out. Thanks

Constantin07:
@defionscode 
The update doesn't work properly. How to reproduce:
1. create a user with path: /division1/
2. change the path to /division2/
3. run playbook once again. It reports "changed=0" but in AWS console the path is old one.
no matter what state is specified "present" or "update"

Constantin07:
@defionscode 

```
- name: Test IAM
  iam:
   iam_type: role
   name: aaa
   state: present
```

Creation of role fails with such error:
TASK: [Query IAM] *************************************************************
<localhost> REMOTE_MODULE iam iam_type=role name=aaa state=present
  File "/root/.ansible/tmp/ansible-tmp-1427922366.45-15755177960783/iam", line 2325, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1427922366.45-15755177960783/iam", line 379, in main
    key_state = module.params.get('access_key_state').lower()
AttributeError: 'NoneType' object has no attribute 'lower'

           to retry, use: --limit @/root/aws-iam.retry

localhost                  : ok=0    changed=0    unreachable=0    failed=1

PS: AWS access and secret keys are defined as variables and work fine with other AWS modules.

juliedavila:
@Constantin07 I am unable to replicate your role creation issue with your exact test case. Have you refetched this PR?

Checking the path change issue now.

juliedavila:
@Constantin07 I see the issue, the way I designed it was to require `new_path` if you wanted to change the path. This makes sense for the username, but for path I think I should probably refactor it so it can be done the way you attempted. I'll patch it.

juliedavila:
@Constantin07 I've added that functionality and tested. Now you can switch paths via `state=update` and `new_path` or just by specifying a path that doesn't match the user's current path.

bpesics:
@defionscode great stuff, but why don't you submit these modules instead of core to ansible-modules-extras first so that the barrier to get used and shipped is lower?

juliedavila:
@bpesics right now the team is set on waiting for v2 before merging new modules. Whether I placed it here or extras, the wait would be the same. Then there is the consistency bit of keeping AWS modules together.

sivel:
Despite the desire for consistency, _all_ new modules should be submitted to -extras as per the instructions in the README.md for this repo:

> New module submissions for modules that do not yet exist should be submitted to ansible-modules-extras, rather than this repo.

juliedavila:
@sivel an internal discussion took place, these particular modules (IAM) will go directly into this repo.

adamchainz:
I've just tried this out. Our codebase has a strict flake8 linter + 120 chars policy so I had to de-lint it. In doing so I found a couple unnecessary imports, unclear lines, etc. and also one apparent bug, which I'll report on the appropriate PR.

My de-linted _iam.py_ is at https://gist.github.com/adamchainz/7f5834fbbfa8e72c48da#file-iam-py

The only other change asides from flake8 compatibility I made was to make `state` default to `present`, like all other ansible modules.

bcoca:
actually we prefer that state is required and not to default to anything

adamchainz:
Okay, I just found it very surprising since every one of `apt`, `yum`, `ec2`, etc. defaults to `present`.

echiu64:
Would be nice to have a way to manage IAM policies vs inline policies attached to a user/group/role.

Wondering if iam_policy should maybe be a key in iam vs on it's own, since it's not really a standalone policy but an inline policy, so may as well inline it into iam? Looks like it's a limitation in boto right now (https://github.com/boto/boto/issues/2956)

But moving it into iam paves way to have an "iam_policy" that manages standalone policy objects.

Edit: Looks like support will be there soon: https://github.com/boto/boto/pull/3185

juliedavila:
@echiu64 yep, agreed. Managed policies are also non supported in boto. Boto3 should be a godsend.

Constantin07:
Hi,

When creating the iam role

```
  - name: Create IAM roles
    iam:
     iam_type: role
     name: "{{ item }}"
     state: present
    with_items:
      - webservers
      - apservers
```

 the below error is thrown if 'access_key_state' is omitted:

```
TASK: [Create IAM roles] ******************************************************
<localhost> REMOTE_MODULE iam iam_type=role name=webservers state=present
failed: [localhost] => (item=webservers) => {"failed": true, "item": "webservers", "parsed": false}
  File "/root/.ansible/tmp/ansible-tmp-1433171112.85-143211962364645/iam", line 2330, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1433171112.85-143211962364645/iam", line 379, in main
    key_state = module.params.get('access_key_state').lower()
AttributeError: 'NoneType' object has no attribute 'lower'
```

The user has nothing to do with roles. I think this can fix the problem:

```
@@ -376,7 +376,8 @@ def main():
     password = module.params.get('password')
     path = module.params.get('path')
     new_path = module.params.get('new_path')
-    key_state = module.params.get('access_key_state').lower()
+    if module.params.get('access_key_state') is not None:
+        key_state = module.params.get('access_key_state').lower()
     key_ids = module.params.get('access_key_ids')

     if iam_type == 'user' and module.params.get('password') is not None:
```

Kind Regards,
C

juliedavila:
@Constantin07 

The current code I see is:

``` python
key_state = module.params.get('access_key_state')
    if key_state:
        key_state = key_state.lower()
        if any([n in key_state for n in ['active', 'inactive']]) and not key_ids:
            module.fail_json(changed=False, msg="At least one access key has to be defined in order"
                                                " to use 'active' or 'inactive'")
```

Which should only be attempting the `lower()` if the key_state is not None. Can you confirm that you have the latest PR code?

Constantin07:
Thanks @defionscode. It turns out I used an outdated version of code.
Please ignore my previous message.

evanccnyc:
Any update on when this and iam_policy will be checked in?

bcoca:
@defionscode are all issues resolved?

juliedavila:
Waiting on @bennojoy to finish up with some testing. He found a bug when attempting to remove a user. It's been patched. Not sure if he's gotten around to iam_policy yet. AFAIK, all known issues have been squashed.

bcoca:
bennojoy gave his OK

PR #167, opened 2014-10-09
Adds alias record support to route53 (migrated from https://github.com/ansible/ansible/pull/8974)

jmound:
...new module repo layout. The actual code changes are copied from https://github.com/tepwil/ansible/commit/260ab8
- I saw the [PR](https://github.com/ansible/ansible/pull/8974) in ansible, and mpdehaan's comment, and thought I'd do the work to get the changes over in to the core modules.
- I forked tepwil's fork, made the change, committed it to my fork, and put that in this pull request.

```
 cp ansible-tepwil/ansible/library/cloud/ec2_elb_lb mindelini-ansible-core/ansible-modules-core/cloud/ec2_elb_lb.py 
  cp ansible-tepwil/ansible/library/cloud/route53 mindelini-ansible-core/ansible-modules-core/cloud/route53.py
```

PR #758, opened 2015-02-07
Add alias support to Route53 module.

Coderah:
This is the rebase that the pull request https://github.com/ansible/ansible-modules-core/pull/167 needs.

Code is largely the same, I was in need of this feature and that pull request had clearly stagnated itself. After setting it up in my fork and testing it decided I'd submit it as it seemed a rebase was all that was holding things up.

A summary of why this exists:

A pull request was made to the `ansible` repo before `ansible-modules-core` was split out, intending to add alias support to the route53 module allowing for `A records` to point to an ELB, it makes ELB return a hosted zone id as well (which is required when making an alias). Someone made a pull request directly copying that code to `ansible-modules-core` conversation died out on Nov 24th of 2014 leaving the pull request in need of a rebase and stuck open.

Obviously credit is due to original creator.

I will maintain this as is needed until it gets merged in.

### DIFF
--- a/cloud/amazon/ec2_elb_lb.py
+++ b/cloud/amazon/ec2_elb_lb.py
@@ -328,7 +328,9 @@ class ElbManager(object):
                 'security_group_ids': check_elb.security_groups,
                 'status': self.status,
                 'subnets': self.subnets,
-                'scheme': check_elb.scheme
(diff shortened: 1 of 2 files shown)


jimi-c:
Merged, thank you!

PR #760, opened 2015-02-08
Adding running and stopped states to nova compute

tkinz27:
This commit adds the ability to power on and power off a VM
managed by openstack. This capability s very similar to the
ec2 module.

This is the same PR submitted on Oct 26, but has been updated
so there are no conflicts (PR #231). This PR has been updated
to include a "restarted" state since the power off function
in the nova api does a hard shutdown ( at least until the Juno
release... see https://blueprints.launchpad.net/nova/+spec/user-defined-shutdown )
The reboot function tries a soft reboot.

Sorry I forgot to turn off my editors auto deletion of trailing
whitespace. Add ?w=1 to the diff url to ignore those diffs.


sivel:
cc @emonty

emonty:
This is great - however we're not taking new features on these openstack modules because they're fundamentally broken wrt modern keystone. The new modules are in the process of landing - I'd love to see this ported over to them. However - I know that's a rude thing to ask. SO - I'm happy to do that porting work and give you credit for it, or I can point you there if you'd rather do it.

tkinz27:
Let me know where the new modules are and I can take a look. If its fairly easy to work with I can probably do the port.

PR #762, opened 2015-02-09
Add exclusive option to authorized_keys

omgjlk:
This option allows the module to ensure that ONLY the specified keys
exist in the authorized_keys file. All others will be removed. This is
quite useful when rotating keys and ensuring no other key will be
accepted.

( this is a re-creation of #7680 )

### DIFF
--- a/system/authorized_key.py
+++ b/system/authorized_key.py
@@ -70,6 +70,15 @@ options:
     required: false
     default: null
     version_added: "1.4"
+  exclusive:
+    description:
+      - Whether to remove all other non-specified keys from the
+        authorized_keys file. Multiple keys can be specified in a single
+        key= string value by separating them by newlines.
+    required: false
+    choices: [ yes", "no" ]
+    default: "no"
+    version_added: "1.9"
 description:
     - "Adds or removes authorized keys for particular user accounts"
 author: Brad Olson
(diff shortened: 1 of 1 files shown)


abadger:
I like the concept.  A few thoughts on implementation:
- version in the documentation should be changed to 1.9
- Would "state=absent exclusive=True" make sense to clear all keys or is there another way to do that?
- Doing this with a set is probably clearer (and gets rid of the call to .items() that requires a comment):

``` python
to_remove = frozenset(existing_keys).difference(keys_to_exist)
for key in to_remove:
    del existing_keys[key]
    to_write = True
```

omgjlk:
I would say clearing all keys is better accomplished by copy with content='' and force=true. Creating an empty file is a "solved problem" within Ansible, I'd rather not add more ways to skin that particular cat.

I forgot to update the version when doing the copy/paste somehow. Oops.

I'll use a set like you suggest. Changes coming soon.

abadger:
s/different/difference/ :-)

system/authorized_key.py
```diff
@@ -402,6 +420,13 @@ def enforce_state(module, params):
             del existing_keys[parsed_new_key[0]]
             do_write = True
 
+    # remove all other keys to honor exclusive
+    if state == "present" and exclusive:
+        to_remove = frozenset(existing_keys).different(keys_to_exist)
+        for key in to_remove:
+            del existing_keys[existing[0]]
```
abadger: s/existing[0]/key/

system/authorized_key.py
```diff
@@ -402,6 +420,13 @@ def enforce_state(module, params):
             del existing_keys[parsed_new_key[0]]
             do_write = True
 
+    # remove all other keys to honor exclusive
+    if state == "present" and exclusive:
+        to_remove = frozenset(existing_keys).different(keys_to_exist)
```
abadger: s/different/difference/

system/authorized_key.py
omgjlk: Haha, oops. Obviously I rushed this. My bad.

abadger:
Thank you, sir!

# Merged

Hi!

issue is resolved.

Thank you!

PR #780, opened 2015-02-12
Added target_tags to gce_module.

rmarsagoogle:
Added the target_tags keyword to the gce_net module for creating firewall rules.


erjohnso:
LGTM, thanks @rmarsagoogle !

PR #783, opened 2015-02-13
Fix subversion module for repos with svn:externals files

Gugli:
When a SVN repository has some svn:externals properties, the module always report changes.

Long story :
When a SVN repository has some svn:externals properties, "svn status" will report files with the X attribute, and lines will be added at the end to list externals statuses with a text looking like
"Performing status on external item at ....".
Such lines were counted as a local modification by the regex, and the module returned a change, even though they were none.

To have a clean (and parsable) "svn status" output, it is recommended to use the --quiet option. The externals will only appear if they have been modified. With this option on, it seems even safer to consider there are local modifications when "svn status" outputs anything.

### DIFF
--- a/source_control/subversion.py
+++ b/source_control/subversion.py
@@ -153,11 +153,10 @@ class Subversion(object):
 
     def has_local_mods(self):
         '''True if revisioned files have been added or modified. Unrevisioned files are ignored.'''
-        lines = self._exec(["status", self.dest])
-        # Match only revisioned files, i.e. ignore status '?'.
-        regex = re.compile(r'^[^?]')
+        lines = self._exec(["status", "--quiet", self.dest])
+        # The --quiet option will return only modified files.
         # Has local mods if more than 0 modifed revisioned files.
-        return len(filter(regex.match, lines)) > 0
+        return len(filter(len, lines)) > 0
 
     def needs_update(self):
         curr, url = self.get_revision()


Gugli:
Makes the subversion module more robust when detecting local changes. 
My use case was using a SVN repo with externals : module returned "changed":true even though there was no local changes.

Gugli:
I think I misnamed the request, and it has been flagged as a feature even though it looks more like a bugfix.

I called the request "Add support...", but I should probably have called it "Fix SVN module for repos with externals", since it's not correctly working a the moment, and externals are a standard SVN feature.

Gugli:
Anyone else using externals to help test this patch ?

This patch still works with my infra (prevent the module form always reporting changes for repos with externals) and does not seem to break anything, but it would be nice to have this changes validated by someone else.

Gugli:
Related bug report : 
https://github.com/ansible/ansible-modules-core/issues/471

patphobos:
This patch work for us, we dont need to do a force=yes anymore on checkouts that contains eXternals.

renard:
Hi guys,

Any new about that bug fix? This issue is a really PITA here and the fix works for several weeks without any side effect.
Thanks in advance.

patphobos:
up ?

Skeeve:
It seems just adding "--quit" does not work.

For me it's failing with "ERROR: modified files exist in the repository".

The output without --quit is:

```
    X      lib

    Performing status on external item at 'lib'
```

Without it, just the "X-line" is removed. The rest remains, which seems to be sufficient for ansible to consider it a change.

Proposal: Change the last line of "has_local_mods" to

```
    return len(filter(lambda s:re.search("^M\s", s), lines)) > 0
```

This should reduce the lines to just those starting with an "M" followed by whitespace.

Petrox:
still failing:

msg: ERROR: modified files exist in the repository.

no local or remote changes happened, although the repo has 3 externals (with no changes)

ubuntu 14.04, ansible 1.9.4, subversion 1.8.8

Skeeve:
Sorry, but I seem to remeber I removed my pull-request as my changes are no longer required it seems.

PR #789, opened 2015-02-13
Added the ability to set an instance to forward ip.

rmarsagoogle:
This adds the ability to create an instance with can-forward-ip enabled.


erjohnso:
LGTM - thank you @rmarsagoogle !

PR #796, opened 2015-02-16
Useful log output.

mcameron:


bcoca:
Just in case the return is not a specific nova exception, would it not be better to use str(e) ?

mcameron:
Done. In my case I was getting floating ip pool not found error which was not surfaced before this fix.

PR #798, opened 2015-02-16
rax: add boot from volume support

sivel:
This pull request adds boot from volume support to the `rax` module and the ability to create a bootable volume via `rax_cbs`.

This PR requires an the accompanying https://github.com/ansible/ansible/pull/10259

### DIFF
--- a/cloud/rackspace/rax.py
+++ b/cloud/rackspace/rax.py
@@ -35,6 +35,34 @@ options:
       - "yes"
       - "no"
     version_added: 1.5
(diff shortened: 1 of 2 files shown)


sivel:
All integrations for `rax` still pass with this change.

As for `rax_cbs` there was a previous change that is causing the integration tests to fail, although it is very minor.  I'll submit a PR for the integration test fix soon.  Other than that failure, all actual functionality tests succeed for `rax_cbs`

PR #807, opened 2015-02-18
apt: use dpkg_option "sourcelist" for update.

SkaveRat:
Older python-apt versions do not support the `sources_list` parameter, use normal update as fallback

I'm neither a python nor an apt pro - it would be awesome if someone could check this properly before merge. It does work for my case

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -270,6 +270,21 @@ def expand_dpkg_options(dpkg_options_compressed):
                        % (dpkg_options, dpkg_option)
     return dpkg_options.strip()
 
+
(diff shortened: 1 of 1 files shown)


SkaveRat:
The way the apt module handles the dpkg options, this needs to be handled differently, as this sometimes causes problems.

is there objection to use a new action parameter `sources_list` for this?

bcoca:
I'm thinking this makes more sense as a default in apt_repository as you only want to update the new repo you added. It should be an option for apt also, it is fine as a new parameter.

Also you can check the library version, we do this in many places, for docker and ec2 modules, for example.

PR #815, opened 2015-02-18
fix for mount w/o opts (bug introduced when fixing bind mounts)

### LINKED ISSUE #812: Mount module throws KeyError exception when "opts" parameter is not provided
##### Issue Type:

Bug Report
##### Ansible Version:

1.8.3
##### Environment:

Amazon Linux AMI release 2014.09
##### Summary:

Not specifying a the "opts" parameter in the "mount" module causes a KeyError to be thrown.
This looks like it was introduced as part of PR: https://github.com/ansible/ansible-modules-core/pull/49
The following line should probably check for the presence of the "opts" key in the args dictionary before trying to access it: https://github.com/ansible/ansible-modules-core/blob/a75bef5bc6ae3ca23e312f9aec9c56b31811f4d0/system/mount.py#L323
##### Steps To Reproduce:

The following mount task throws the exception (note the absence of the optional opts parameter)
- name: Mount NFS volume
  mount: name=/media/storage fstype=nfs4 src={{storage_instance_ip.stdout}}:/ state=mounted
##### Expected Results:

Mount module should mount the drive.
##### Actual Results:

Python KeyError exception is thrown

bcoca:
should fix #812

### DIFF
--- a/system/mount.py
+++ b/system/mount.py
@@ -320,7 +320,7 @@ def main():
             if os.path.ismount(name):
                 if changed:
                     res,msg = mount(module, **args)
-            elif "bind" in args['opts']:
+            elif 'bind' in args.get('opts', []):
                 changed = True
                 cmd = 'mount -l'
                 rc, out, err = module.run_command(cmd)


abadger:
+1

PR #823, opened 2015-02-19
draft for documenting module returns

bcoca:
documented copy, acl and stats as examples, currently only viewable through modified ansible-doc in https://github.com/ansible/ansible/pull/10295

### DIFF
--- a/files/stat.py
+++ b/files/stat.py
@@ -69,6 +69,179 @@ EXAMPLES = '''
 - stat: path=/path/to/myhugefile get_md5=no
 '''
 
+RETURN = '''
+stat:
+    description: dictionary containing all the stat data
+    returned: success
+    type: dictionary
+    contains:
+        exists:
+            description: if the destination path actually exists or not
+            returned: success
+            type: boolean
+            sample: True
+        path:
+            description: The full path of the file/object to get the facts of
+            returned: success
+            type: boolean
+            sample: '/path/to/file'
+        mode:
+            description: Unix permissions of the file in octal
+            returned: success, path exists and user can read stats
+            type: octal
+            sample: 1755
+        isdir:
+            description: Tells you if the path is a directory
+            returned: success, path exists and user can read stats
+            type: boolean
+            sample: False
+        ischr:
+            description: Tells you if the path is a character device
+            returned: success, path exists and user can read stats
+            type: boolean
+            sample: False
+        isblk:
+            description: Tells you if the path is a block device
+            returned: success, path exists and user can read stats
+            type: boolean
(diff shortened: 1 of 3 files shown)


files/stat.py
```diff
@@ -69,6 +69,179 @@
...
+    type: dictionary
+    contains:
+        exists:
+            description: if the destination path actually exists or not
+            returned: success
+            type: boolean
+            sample: True
+        path:
+            description: The full path of the file/object to get the facts of
+            returned: success
```
twekberg: It looks like the path is returned only if 'success and path exists'. Look at the code for the case where the path doesn't exist. A dictionary with one element ('exists') is returned.

files/stat.py
```diff
@@ -69,6 +69,179 @@
...
+            description: Number of links to the inode (hard links)
+            returned: success, path exists and user can read stats
+            type: int
+            sample: 1
+        atime:
+            description: Time of last access
+            returned: success, path exists and user can read stats
+            type: float
+            sample: 1424348972.575
+        mtime: st.st_mtime,
```
twekberg: Please consider removing ' st.st_mtime,'

files/stat.py
```diff
@@ -69,6 +69,179 @@
...
+            description: User name of owner
+            returned: success, path exists and user can read stats and installed python supports it
+            type: string
+            sample: httpd
+        gr_name:
+            description:
+            returned: success, path exists and user can read stats and installed python supports it
+            type: string
+            sample: www-data
+'''
```
twekberg, line 253: What about isuid and isgid? I had these documented:
isuid - invoking user's id matches path's user id
isgid - invoking group's id matches path's group id

files/stat.py
```diff
@@ -69,6 +69,179 @@
...
+            returned: success, path exists and user can read stats and path supports hashing
+            type: boolean
+            sample: True
+        pw_name:
+            description: User name of owner
+            returned: success, path exists and user can read stats and installed python supports it
+            type: string
+            sample: httpd
+        gr_name:
+            description:
```
twekberg: Needs a description

files/stat.py
```diff
@@ -69,6 +69,189 @@
...
+            description: Tells you if others have execute permission
+            returned: success, path exists and user can read stats
+            type: boolean
+            sample: True
+        isuid:
+            description: Tells you if the invoking user's id matches the owner's id
+            returned: success, path exists and user can read stats
+            type: boolean
+            sample: False
+        isrid:
```
twekberg: Should be isgid:

PR #824, opened 2015-02-19
Now correctly gzip/bzips file back up in case of import failure

Jmainguy:


bcoca:
I thought i had fixed this with alternate approach, but not unziping the file but by piping the uncompressed stream directly to mysql.

Jmainguy:
Do you mean the original implementation?

https://github.com/ansible/ansible/commit/bb73819c43bfb0069c5459924e26014ef0a4fe35

The original implementation worked, when everything worked. However it would hide some failures and report OK even when it failed, due to the return codes being swapped around when piping commands into each other.

bcoca:
ah, probably, this is why pipefail is A killer feature in bash

abadger:
Took a look, one minor thing to change:

Instead of looking for both gzip and gunzip (and having to use separate error messages if they aren't found), just use gzip_path with the -d option for decompressing.  same with bunzip2.

Other than that, this looks good to merge.

Jmainguy:
@abadger updated to remove bunzip2 and gunzip dependency. Great suggestion.

abadger:
Thanks @Jmainguy !

# Merged

Hi!

issue is resolved.

Thank you!

PR #825, opened 2015-02-20
Fixes #172 Updated ec2_vpc module to return created subnets in the same order as they were defined at the playbook

### LINKED ISSUE #172: ec2_vpc subnets in seemingly random order
When building VPCs using the ec2_vpc module the subnets are returned in a seemingly arbitrary order. It would be nice if they were returned in the order specified or ordered by cidr block to simplify deploying EC2 instances into specific subnets consistently.

daniilyar:


dskyberg-Ping:
+1

0x736861:
+1, we need that, too

abadger:
I think I would do this by using python's sort function instead of looping over subnet once and current_subnets every time through the subnet loop.  I can write that code really quickly if you'd want to review that it works.

abadger:
Okay, here's code for the alternative: https://github.com/ansible/ansible-modules-core/pull/904

I don't use ec2 so please review and test that it works.

One thing I wondered about when I wrote this was whether current_subnets would be a subset of subnets or if it could contain extra subnets.  The code you had seems like it doesn't anticipate that being a problem but since I wasn't sure I added handling for that.

abadger:
#904 merged so closing this one.

Thanks for submitting such a nice proof of concept!

# Possible Duplicated Efforts

Hi!

It looks like the work from this pull request is a duplicate of the following PR(s):
- #904 

Based on this, we are going to close this PR in favor of the above.

However, we're absolutely always up for discussion.  Since this is a really busy project, we don't always see comments on closed tickets, but want to encourage open dialog.  You can stop by the development list, and we'd be glad to talk about it - and we might even be persuaded otherwise!

PR #831, opened 2015-02-20
List the name servers of a zone. Fixes #801

### LINKED ISSUE #801: route53 module: list the associated name servers of zone
##### Issue Type:

Feature Idea
##### Ansible Version:

1.8.2
##### Environment:

OS running: "Ubuntu 14.04", OS managing: “N/A”
##### Summary:

List the name servers which amazon associated to the zone when using the `get` command.

j-carl:
Fixes #801

Also list the name servers associated to a zone, when command `get` is used.

After I made the change to the code, I found out that you can get the name servers already with current code. So this would be only an addition to make things a little bit more convenient for the users.

Just for the curios ones here is an example to get the name servers with existing code:

```
route53:
    command: get
    zone: foo.com
    record: foo.com
    type: NS
register: rec
```

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -248,7 +248,13 @@ def main():
                 module.exit_json(changed=False)
 
     if command_in == 'get':
-        module.exit_json(changed=False, set=record)
+        if type_in == 'NS':
+            ns = record['values']
+        else:
+            # Retrieve name servers associated to the zone.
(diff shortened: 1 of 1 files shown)


gregdek:
@bpennypacker please review; if you +1 we will tag P1_reviewed for faster inclusion.

bpennypacker:
Nice little feature to add.  Thanks, @j-carl!   Looks good to me.  +1

PR #837, opened 2015-02-23
Added authentication capabilities to win_get_url

cerowain:
I found an issue providing authentication details to win_url_get so I added fields to match the standard get_url module. It was an on site hack, looking for anyone who wants to clean it up as required.


cerowain:
Not sure what I did exactly to trigger the second commit being pushed to the upstream.  Shall have to put it down to user error.

gregdek:
@cerowain looks like this fails travis; can you make sure this PR is still relevant, and if so, check the travis error and correct any issue? Happy to review at that point.

PR #838, opened 2015-02-23
Added check mode support to the ini_file module

blxd:
There was no support for check mode in the ini_file module so I added it. 

I also fixed a typo, two indentation mistakes, and a PEP8 complaint - I can break these in to a separate merge if required.


jimi-c:
Merged, thank you!

PR #839, opened 2015-02-23
Do not insert extra newline if line already contains it

gimoh:
When using YAML multi-line strings, e.g.:
- lineinfile:
  dest: /tmp/foo
  line: >
    foo
    bar

the line already ends with a newline.  If an extra newline is appended unconditionally it will lead to inserting an extra newline on each run.


gregdek:
@dhozac @ahtik can you review? If either of you give +1 (and no -1) we will flag P1_reviewed for faster inclusion.

PR #841, opened 2015-02-24
Encrypt s3 object

billwanjohi:
S3 as a service provides server-side encryption of objects as an option. This adds an option on the s3 module to PUT files with this encryption enabled. This is a requirement for many of the buckets that I use.

PR #842, opened 2015-02-24
s3 module: reformat for PEP-0008, PEP-3101

billwanjohi:
This has no functional changes, just cleans things up for read/maintain-ability.

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -19,7 +19,11 @@ DOCUMENTATION = '''
 module: s3
 short_description: S3 module putting a file into S3.
 description:
-    - This module allows the user to dictate the presence of a given file in an S3 bucket. If or once the key (file) exists in the bucket, it returns a time-expired download URL. This module has a dependency on python-boto.
+    - This module allows the user to dictate the presence of a
+      given file in an S3 bucket.
+      If or once the key (file) exists in the bucket,
+      it returns a time-expired download URL.
+      This module has a dependency on python-boto.
 version_added: "1.1"
 options:
   bucket:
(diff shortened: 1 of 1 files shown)


bcoca:
we don't take  pure formating PRs as they break attribution.

billwanjohi:
Can you elaborate on this? I made these changes while I was working on #841, but didn't want to muddy that pull request with things that might slow down a review (a particular concern with a project so overwhelmed with submissions, I would think you agree).

I also wanted to come back and break up main() into a set of "mode" functions; e.g. put_mode(), get_mode(), etc. and simplify the branching logic therein.

Is all of that literally just disallowed in the Ansible project? What's your strategy for improving code quality, if small improvements are rejected out of hand? Is an existing module considered sacrosanct until its functionality breaks?

bcoca:
There is a difference between re-factoring and stylistic changes, pep8 falls into the latter. 

As for string formatting, modules are required to be 2.4 compatible, so anything that adds a 2.6 dependency is a problem, though the boto dependency makes that a mute point here.

Also I'm not a fan of the 80 column limitation, IMHO it actually makes the code less readable as you can fit less in the screen. I have not used a 80 col terminal in 20 years, I believe most people have never had "the pleasure". 

We do try to avoid messing with working modules and tend to leave re-factoring to when it clearly makes it a problem to maintain. This is based, both because of the lack of bandwidth for this type of request and a general dislike of fixing things that are not broken. 

Code can always be made better, any change that has a quantifiable production effect (performance, bugs, features) will be more likely to be prioritized over others. We do value maintainability, but the limited scope of modules makes it easy to keep less maintainable code well isolated.

billwanjohi:
Thanks, that is helpful context. If I were to endeavor to break different s3 modes into separate function calls with less branching, would that sort of refactor be considered?

Also, is there a coding style manual for Ansible? I am happy to, for example, target a higher column count (100? 120?), but in the absence of some project standard, I usually follow Guido's rules.

bcoca:
I have my vimrc at 132, but I'm not religious about it

PR #846, opened 2015-02-25
Change s3 to use connect_to_region

mlaccetti:
- Allows us to use Ansible/s3 for GovCloud when accessing a pre-existing bucket

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -339,7 +339,8 @@ def main():
             module.fail_json(msg = str(e))
     else:
         try:
-            s3 = boto.connect_s3(aws_access_key, aws_secret_key)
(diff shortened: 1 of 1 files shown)


mlaccetti:
Related to the original issue to consume the 'region' setting: https://github.com/ansible/ansible/issues/8783

mlaccetti:
Is there any idea as to the ETA on this? Can't really use Ansible in production until this comes in.  :(

bcoca:
I would move the import to the top with the rest, otherwise the change looks good

mlaccetti:
I cloned the way it was done in the rest of the code; I can move the import to the top and remove the redundancy.

mlaccetti:
Okie dokie, cleaned up the imports.

PR #853, opened 2015-02-26
Private zones route53

### LINKED ISSUE #473: route53 module does not differentiate between Private and Public hosted zones

bcoca:
should fix #473


bcoca:
dupe

PR #854, opened 2015-02-26
Add TLS support to docker client connection

dguerri:


abadger:
Hi @dguerri , @smashwilson mentioned in #502 that he thought this implementation had some features that were better than his implementation in #502 .  I'm taking a look at it to see what I think... I haven't decided yet, though, so probably should answer high level questions before revising the code.

high level:

I see that tls_verify and tls_assert_hostname mirror the docker api but I'm not sure I understand the rationale.  Is  there any reason not to just have one ansible-level setting to make sure tls is good and then map that setting to set both of those to True for the docker API?

implementation:

We'd want four space indent for all python code rather than two space indent.

What happens when Docker.Client(tls=False|True)?  I can't seem to find an answer to that in the docker-py documentation.

dguerri:
Hi @abadger and @smashwilson , thanks for considering my PR:
- high level:

about `tls_verify` and `tls_assert_hostname`: `tls_assert_hostname=False` just skips subject verification while `tls_verify=False` skips CA verification altogether.

Anyway, I have just noticed that `assert_hostname` in docker API (as other parameters) is not a real boolean: it could be None, False or a string. If it is a string, it will be used as a hostname to validate the one in the certificate. If you pass True, docker client will crash. 
I need to update the PR, using a boolean (say `tls_verify_hostname`) and an optional `tls_assert_hostname`.
- implementation:

About 4 spaces indent: I will resubmit a PR (found it now)

About Docker API... that's another funny thing. 

`tls=False` is the default (https://github.com/docker/docker-py/blob/master/docker/client.py#L44) and that means 'do not use x509 authentication'.

If `tls=True` is used the API try to use systems settings to validate certificates (using a class named SSLAdapter). This is equivalent to the CLI command `docker --tls ...`

If `isinstance(tls, TLSConfig)`, then it uses the configuration provided in the TLSConfig instance passed as the tls argument.

It is documented here: http://docker-py.readthedocs.org/en/latest/tls/

dguerri:
@abadger @smashwilson PR updated.

abadger:
regarding tls_verification and hostname checking -- I'm just wondering why we wouldn't want to verify the hostname if we're requesting that we use tls to verify the server.

Similarly, for tls_assert_hostname, when would we not want it to be the hostname from docker_url?

dguerri:
Honestly I find it hard to provide a _legit_ use case for that... I do have a specific use case with boot2docker from this guy https://github.com/wearableintelligence/boot2docker-vagrant-box/releases/tag/docker%2Fv1.5.0 . 
The certificate used there uses just boot2docker as subject name. That is the hostname of the box, which is not resolvable outside, so my client fails to validate that certificate,

Anyhow validating a certificate ignoring the subject name is better than just skip verification altogether as verification involves other checks, like the expiry date and the CA certificate chain.

EDIT
To further clarify: using TLS with client certificate, we are checking client and server identity (on server and client respectively). Skipping server hostname verification, means assuming that the specific name of the server is not known (we can't trust the name on cert), but we verify that it is trusted by the CA that emitted its certificate.

smashwilson:
> Honestly I find it hard to provide a legit use case for that... I do have a specific use case with boot2docker from this guy https://github.com/wearableintelligence/boot2docker-vagrant-box/releases/tag/docker%2Fv1.5.0 . 
> The certificate used there uses just boot2docker as subject name. That is the hostname of the box, which is not resolvable outside, so my client fails to validate that certificate,

Yeah, I had to disable it to connect to boot2docker. Unfortunate.

abadger:
@dguerri, I've merged @smashwilson's #502 but am working on changing some of the parameters to try to be compatible with this PR.  That way we can merge this later without forcing people to update their playbooks.  I think that's doable but the more I dig into it, the more nervous I feel about how docker is using tls for authenticating the server.  It seems like the strategy that they propose here: http://docs.docker.com/articles/https/, for instance, means that you should never ever use docker (and this strategy) with a public CA.  You have to use it with a single-purpose CA that you run (as any certificate issued by the CA will give access to A) utilize the docker server and B) MITM a docker client trying to connect to a docker server).

That's kind of a scary caveat to me so I'm not sure how far I want to accomodate it -- on the one hand, it's very dangerous if people running these docker installs don't understand the risk.  On the other, docker is officially telling people to do it this way so lots of people may run into this.  I'll have to sleep on that portion (and I welcome any clarifications you might make :-)

abadger:
Okay, going through this, here's a way that parameters could be structured.  Does this seem sufficient and doable with docker-py API?
- use_tls encrypt|verify: Default None means do not use tls unless required for one of the other parameters.  Encrypt means that we'll use tls to connect to docker_url rather than unencrypted tls.  verify means that we'll attempt to verify that the server is who they claim to be.
- tls_client_key and tls_client_cert: These must both be specified or neither specified.  Presence of these parameters means that we're going to use tls certificates to authenticate the client to the server.
- tls_ca_cert: If specified, then a ca certificate to check the server's credentials.  Note that checking is not performed if use_tls == encrypt
- tls_hostname: If specified, then a hostname to check is the server's hostname.  If unspecified, then tls_hostname is taken from docker_url.

Some pseudo code for the relation between use_tls and checking a server against its cert

```
if use_tls == encrypt:
    Do not verify that the server's cert is valid.
else:
    if tls_ca_cert:
        verify that the server's cert is valid and belongs to hostname.
        may need to specify tls_hostname to override what's in docker_url
    if not tls_ca_cert:
        verify that the server's cert is valid against system ca bundle
        and belongs to the hostname.
       May need to specify tls_hostname to override what's in docer_url
```

dguerri:
Hi @abadger 

> It seems like the strategy that they propose here: http://docs.docker.com/articles/https/, for instance, means that you should never ever use docker (and this strategy) with a public CA. You have to use it with a single-purpose CA that you run (as any certificate issued by the CA will give access to A) utilize the docker server and B) MITM a docker client trying to connect to a docker server).

I think that example just shows the quickest possible way to create and use a PKI. 
When it comes to certificates generated by that PKI, you are quite right: you can use what-they-called-client-certificate as a server certificate. 
That is because they haven't used the extended key usage extension. 
In a real production environment, I would use a real CA, manage CRLs, ... and of course I would limit the usage of certificates so that a client couldn't authenticate another client.

To double check that this management is actually possible and useful with docker, I have just created PKI, and used the extended key usage extension.

When I tried to use a server certificate as a client, I get this in docker log file:

```
http: TLS handshake error from 10.211.55.2:57494: tls: client's certificate's extended key usage doesn't permit it to be used for client authentication
```

And, viceversa, when I try to autenticate the server with a client certificate:

```
~# docker ps -a
FATA[0000] An error occurred trying to connect: Get https://10.211.55.154:2376/v1.17/containers/json?all=1: x509: certificate specifies an incompatible key usage
```

Of course, as it happens, the security is all about the CA. CA key must be stored securely.
On the other hand, local unix sockets are as secure as your box and your system accounts are.

dguerri:
> use_tls encrypt|verify: Default None means do not use tls unless required for one of the other parameters. Encrypt means that we'll use tls to connect to docker_url rather than unencrypted tls. verify means that we'll attempt to verify that the server is who they claim to be.
> tls_client_key and tls_client_cert: These must both be specified or neither specified. Presence of these parameters means that we're going to use tls certificates to authenticate the client to the server.
> tls_ca_cert: If specified, then a ca certificate to check the server's credentials. Note that checking is not performed if use_tls == encrypt
> tls_hostname: If specified, then a hostname to check is the server's hostname. If unspecified, then tls_hostname is taken from docker_url.

It seems doable.
What about the environment? Shall we assume that the environment always override these settings?
I understand that we must be backward compatible with #502, so we should probably skip all these settings if TLS is configured using the environment.
What is your take about that?

abadger:
I want these settings to override the environment.  We don't have to maintain backwards compat with #502 until it goes out in a release which is one reason I'm thinking so much about this ticket :-)  (1.9rc1 is scheduled for Monday)

abadger:
I'm leaving the house soon and probably won't be back until tonight... I'd started playing around with this this morning; originally just to try to adapt it so that things would be compatible with merging this but then I got sucked into figuring out how to make the docker-py api do what we needed with the least amount of new module parameters.  Here's the code I have so far in case you are interested in taking it farther today (warning -- kinda a brain dump so there's probably a bunch of issues and unfinished things): https://gist.github.com/abadger/db04252fc594934139a5

abadger:
Alternative has now been merged to devel and stable-1.9.  Thanks for all the help there!

Closing this ticket.

Issue #852, opened 2015-02-26
file module dereferences relative symlinks incorrectly


ngaya-ll:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.8.4
##### Environment:

Ubuntu 12.04
##### Summary:

The file module dereferences symlinks with relative paths incorrectly when follow=yes.
##### Steps To Reproduce:

The following playbook demonstrates the issue:

```

---
- hosts: 127.0.0.1
  connection: local
  tasks:
    # Create test directory
    - file: name=foo state=directory
    - file: name=foo/bar state=touch

    # Create relative symlink
    - file: state=link src=bar dest=foo/baz

    # Test that symlink target exists
    - file: name=foo/baz state=file follow=yes
```
##### Expected Results:

All tasks should succeed.
##### Actual Results:

The last task fails with message "file (bar) is absent, cannot continue".
##### Discussion:

The problem is due to a misuse of [`os.readlink()`](https://docs.python.org/2/library/os.html#os.readlink).  The docs suggest using `os.path.join(os.path.dirname(path), os.readlink(path))` to get a correct path (this will work for both absolute and relative symlinks).  This does not deal with symlinks that point to other symlinks, however.  It might be better to avoid the problem by using [`os.path.realpath()`](https://docs.python.org/2/library/os.path.html#os.path.realpath) to get a canonical filename.

One place in the code where this bug occurs is on [line 247 of files/file.py](https://github.com/ansible/ansible-modules-core/blob/release1.8.4/files/file.py#L247), but there may be other occurrences as well.

abadger:
Thanks for the report!  @bcoca's commit should fix this issue.

# Closing This Ticket

Hi!

We believe recent commits (likely detailed above) should resolve this question or problem for you.

This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular
issue is resolved.

Thank you!

PR #855, opened 2015-02-26
fixed file module to use realpath when following links,

bcoca:
readlink could be relative and fail in corner cases as per #852


abadger:
+1

abadger:
@bcoca there are two more instances of readlink in file.py -- were those okay?

bcoca:
kind of, the logic would work as the path was either prepended or not used, I only modified the ones when we had to follow the symlink.

PR #872, opened 2015-03-01
redhat_subscription: add the `org_id` parameter

nixpanic:
When subscribing a system with an activationkey, it seems (sometimes?)
required to pass the "--org <number>" parameter to subscription-manager.
Activation Keys can be created through the Red Hat Customer Portal, and
a subscription can be attached to those. This makes is easy to register
systems without passing username/passwords around.

The organisation ID can be retrieved by executing the following command
on a registered system (_not_ the account number):

  # subscription-manager identity

URL: https://access.redhat.com/management/activation_keys
Signed-off-by: Niels de Vos ndevos@redhat.com

### DIFF
--- a/packaging/os/redhat_subscription.py
+++ b/packaging/os/redhat_subscription.py
@@ -54,6 +54,11 @@ options:
             - supply an activation key for use with registration
         required: False
         default: null
+    org_id:
+        description:
+            - Organisation ID to use in conjunction with activationkey
+        required: False
+        default: null
     pool:
         description:
             - Specify a subscription pool name to consume.  Regular expressions accepted.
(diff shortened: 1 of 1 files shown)


ktdreyer:
This is extremely close to what my team and I needed. Thank you!

I tried your patch and discovered that it suffers from the same problem in https://github.com/ansible/ansible-modules-core/pull/1070 . Could you switch it to use `args.extend` instead of `args.append`?

nixpanic:
Now with `args.extend`. Thanks!

ktdreyer:
Thank you! This works great (along with #1070 :)

`Reviewed-by: Ken Dreyer <kdreyer@redhat.com>`

ktdreyer:
@nixpanic , would you mind rebasing your branch onto the tip of master? #1070 was just merged.

jimi-c:
Merged, thanks!

PR #220, opened 2014-10-23
OpenStack Identity (keystone) service module

berendt:
With this module it is possible to create and delete keystone services.
It is not possible to update keystone services because this is not supported
in version 2.0 of the keystone API.

PR #874, opened 2015-03-01
Add keystone_endpoint and keystone_service modules

dguerri:


bcoca:
@emonty any comments?

emonty:
Spoke with @dguerri briefly. These should go into the new openstack module structure as os_keystone_endpoint and os_keystone_service. There is another keystone PR going up in just a bit, as soon as it's there, I'll post a reference in here so that these can be modified to match that one. Forces should also be combined with @berendt and #220 which also provides a service module. Will follow up with both later today.

dguerri:
Meanwhile, to those who are interested, these two modules are already available in a keystone role I am maintaining here: https://github.com/dguerri/openstack-keystone/tree/development (in the dev branch, currently)

dguerri:
Hey @bcoca @emonty, sorry for the silly question... But, where can I find the new openstack module structure?
I will be happy to move the code there.

dguerri:
I see...

emonty:
@dguerri I just pushed up ansible/ansible-modules-extras#315 which has 2 of the first three end-user facing modules. The two you've done here are admin-facing, so I think they want to be called os_keystone_endpoint and os_keystone_service. You're probably going to need to add some more methods to shade - but you can start out just by instantiatiing an OpenStackCloud like the other modules do, and then grabbing cloud.keystone_client as your keystone client and your current code should work mostly as-is - making shifting functionality into shade and replacing direct usage of keystone_client in these modules refactoring later rather than blocking now.

dguerri:
@emonty @bcoca I have just opened a PR against Monty's repo: https://github.com/emonty/ansible-modules-extras/pull/10
Cheers!

Issue #267, opened 2014-11-04
synchronize module: rsync_opts problem


m-emelchenkov:
Issue Type:

Bug Report

Ansible Version:

ansible 1.7.2

Environment:

Mac OS 10.10 (with rsync 3.1.1 from Homebrew) / Ubuntu 14.04

Summary:

I've found a problem with rsync_opts in module synchronize. There is an option in rsync which contains comma, and this option will not be handled by the module correctly due to module design.

Steps To Reproduce:

Add rsync_opts="--iconv=UTF8-MAC,UTF-8"

Expected Results:

Comma should not be replaced by whitespace.

Actual Results:

Comma is replaced by whitespace and command halts.

Synchronize module contains these:
if rsync_opts:
cmd = cmd + " " + " ".join(rsync_opts)

That means in the result command we get space character instead comma and the command will halt due to this:
--iconv=UTF8-MAC UTF-8

Maybe I do not understand something, but why not to pass rsync_opts as string, not as a list? Then there will not be this problem exists, and separating parameters will be performed using space character.

I use workaround by patching "synchronize" file:
if rsync_opts:
cmd = cmd + " " + ",".join(rsync_opts)

Bug was opened on Jul 9, 2014 and time passed since then. I hope someone will take it into work.
https://github.com/ansible/ansible/issues/8074

jody-frankowski:
--chmod=D700,F600 for example, seems to be problematic too.

m-emelchenkov:
Please assign someone to this issue. If we just make workaround this may break something. i'm tired patching locally every new version.

bcoca:
@mikhailguru has a PR been submitted?

m-emelchenkov:
@bcoca What is PR?

bcoca:
Pull Request, sorry, too used to the abbrevs ...

m-emelchenkov:
@bcoca Ah, no. I don't feel it is a good patch, because it needs further investigation that it does not have any drawbacks, so I called it "workaround" in this issue instead submitting to the codebase. It will be nice to connect the module's author to this discussion to get feedback from him.

bcoca:
the issue might not be with the module itself but with how module arguments get parsed in general

davidvasandani:
Running into the same issue when trying to run the synchronize module though a bastion host.

```
[...]

PR #877, opened 2015-03-02
Fix for UTF-8 iconv processing in Synchronize module

m-emelchenkov:
Workaround for bugreport https://github.com/ansible/ansible-modules-core/issues/267

### DIFF
--- a/files/synchronize.py
+++ b/files/synchronize.py
@@ -316,7 +316,7 @@ def main():
         cmd = cmd + " --rsync-path=%s" % (rsync_path)
 
     if rsync_opts:
-        cmd = cmd + " " +  " ".join(rsync_opts)
+        cmd = cmd + " " +  ",".join(rsync_opts)
 
     changed_marker = '<<CHANGED>>'
     cmd = cmd + " --out-format='" + changed_marker + "%i %n%L'"


gregdek:
@tima please review; if you give +1 we will tag P1_reviewed for faster inclusion.

tima:
@gregdek I looked at this and I'm pretty sure it's not the right thing to do and can't recommend it be merged in. 

As @bcoca surmised (https://github.com/ansible/ansible-modules-core/issues/267#issuecomment-69429470) it's a parsing error of module params happening at a lower level. This synchronize module use case just so happens to expose it. Another module that accepts a comma in a param value would have the same issue.

This PR will not work if a user defines multiple rsync_opts like is shown in the module docs you get a comma instead of a space between `--no-motd` and `--exclude` that rsync can't handle and throw and error.

`rsync --delay-updates -F --compress --archive --rsh 'ssh  -S none -o StrictHostKeyChecking=no' --no-motd,--exclude=.git --out-format='<<CHANGED>>%i %n%L' "/tmp/helloworld" "/var/www/helloworld"'`

Now if I reformat my task differently without this PR applied so that "native" YAML list parsing is done, like so...

```
    -  synchronize:
        src: "/tmp/helloworld"
        dest: "/var/www/helloworld"
        rsync_opts: ['--no-motd', '--exclude=.git']
```

Ansible generates the correct rsync command.

`rsync --delay-updates -F --compress --archive --rsh 'ssh  -S none -o StrictHostKeyChecking=no' --no-motd --exclude=.git --out-format='<<CHANGED>>%i %n%L' "/tmp/helloworld" "/var/www/helloworld"'`

gregdek:
OK, @tima, thanks. Am I to surmise that any workaround to the issue in #267 would be similarly inappropriate? In other words, should we just close this PR? Is there an open issue that we can use to track the underlying bug here?

gregdek:
Closing due to -1 from @tima and lack of response from original submitter.

m-emelchenkov:
@tima, you give me an idea how to pass the parameter without the problem. Thanks!!!

```
- synchronize:
    src: "../../../Sources/www/"
    dest: "/var/www/{{ domain }}/"
    delete: "yes"
    perms: "no"
    rsync_path: "sudo -u www-{{ host }} rsync"
    rsync_opts: ["--iconv=UTF8-MAC,UTF-8"]
```

PR #881, opened 2015-03-03
Fix 'Group does not exist' when adding groups

BartVB:
Creating a user with 'groups=sudo, adm' resulted in an error because the group ' adm' (with space) doesn't exist. Strip group names to fix this.

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -505,7 +505,7 @@ class User(object):
         if self.groups is None:
             return None
         info = self.user_info()
-        groups = set(filter(None, self.groups.split(',')))
+        groups = set([x.strip() for x in self.groups.split(',') if x])
         for g in set(groups):
             if not self.group_exists(g):
                 self.module.fail_json(msg="Group %s does not exist" % (g))


gregdek:
@BartVB Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

system/user.py
```diff
@@ -502,7 +502,7 @@ def get_groups_set(self, remove_existing=True):
         if self.groups is None:
             return None
         info = self.user_info()
-        groups = set(filter(None, self.groups.split(',')))
+        groups = [x.strip() for x in self.groups.split(',') if x]
```
bcoca: this should still be a set() to avoid duplicates

gregdek:
@BartVB A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

BartVB:
Sorry that took a while. The fix is still needed. I'm hardly a Python expert but AFAIK the proposed fix (with the addition by bcoca) should do what is says on the tin.

gregdek:
Thanks @BartVB. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

davidbath:
Apologies in advance for likely hijacking an inappropriate thread, but I wanted to check if this would also fix a problem I see in version 1.9.2 on ubuntu. However, I've been bashing my head against the wall for a few hours, and stumbled on this thread which does sound hopeful!

If groups= are provided as a list, e.g.

```
- name: Adding sysadmin users
  user:
    name: "{{ item.key }}"
    comment: "{{ item.value.fullname }}"
    groups: "{{ item.value.groups }}"
    state: present
with_dict: users_sysadmin
```

where users_sysadmin looks like

```
users_sysadmin:
  someuser:
    fullname: Some User
    groups:
      - adm
      - plugdev
      - admin
```

then an error is returned that group adm can't be found.

If you put the list `adm,plugdev,admin` (no spaces) directly into the user module without the lookup it's fine.

On the surface, it looks like this fix could be related, but I also can't quite believe that. Surely others have used groups in a list (as an item in a dict) before?

Please kick of off into the right place to raise this question if it's not here.

Thanks!

gregdek:
@BartVB Thanks again for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

gregdek:
Thanks @BartVB. @sfromm please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
@ansible This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

mscherer:
seems fine to me, so 'shipit'

gregdek:
Thanks @BartVB for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

jctanner:
Related PR https://github.com/ansible/ansible-modules-core/pull/1176 which I think is a better approach since it does not change the current default behavior.

PR #884, opened 2015-03-03
elasticache: add 'info' command, and 'update' state parameter to 'command'

Rob-Johnson:
This change request makes a few of changes to the elasticache module:
- Add the ability to gather facts/information about a cluster with an 'info' option available to the state parameter.
- Rename the 'state' parameter to 'command'. There are no behavioural changes associated with it, but 'command' allows for other, non idempotent commands such as 'info' and 'reboot' to be performed.
- Change the 'present', 'absent' and 'rebooted' noun states to appropriate verbs 'create', 'delete' and 'reboot'.

There seems to be mixed precedent for the use of state vs commands and the corresponding use of nouns vs verbs to reflect the actions performed: 
- The [route53](http://docs.ansible.com/route53_module.html) and [rds](http://docs.ansible.com/rds_module.html) module use command.
- Many other AWS modules use state.
- [ec2](http://docs.ansible.com/ec2_module.html) has a separate [ec2_facts](http://docs.ansible.com/ec2_facts_module.html) module for gathering facts.

I'm interested in hearing your thoughts on what is considered 'the ansible way' to achieve this.


bcoca:
Ansible modules should be declarative and not imperative as this is what the project strives for.

Command should not be used, state is the preferred 'ansible way', on our wishlist we have a 'module normalization' that will eventually address that. 

For fact gathering we prefer a separate _facts module.

Rob-Johnson:
Thanks for your response @bcoca.

> For fact gathering we prefer a separate _facts module.

In that case, I'll close this PR and work on creating a new _facts module for elasticache.

PR #885, opened 2015-03-03
ec2: Apply all security groups at instance-creation

quodlibetor:
The code was picking out the first instance from the security groups
specified, even when multiple groups were specified. Now we use all of
them.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -806,8 +806,7 @@ def create_instances(module, ec2, vpc, override_count=None):
             if type(group_id) == str:
                 group_id = [group_id]
             grp_details = ec2.get_all_security_groups(group_ids=group_id)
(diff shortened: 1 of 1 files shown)


abadger:
Ugh -- we've got some ugly named variables in that module don't we? :-(  Took me a while to figure out what the original code was doing but your fix certainly looks right to apply on top of that (mimics what we're doing if group_name is given and is a list).  Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

quodlibetor:
Yeah, I wanted to change as little as possible otherwise I would have fixed the pluralizations.

Thanks so much for the quick response! :hammer:

abadger:
yeah, definitely understood.  I try to stick to changing as little as possible within one pull request too.  Makes reviewing much easier :-)

PR #886, opened 2015-03-03
Small code cleanup

abadger:
- isinstance instead of comparing types
- Change group_name to a list rather than having separate code paths if it's a string


abadger:
@quodlibetor would you mind having a look at this?  It's a small cleanup to the same portion of code you just fixed.

quodlibetor:
This looks much saner. :+1:

abadger:
Thanks for the second pair of eyes :-)

quodlibetor:
Glad to be of service!

PR #887, opened 2015-03-03
iam_policy

juliedavila:
There is currently no IAM module. This is part 2 of 3 that I plan to make.

Part 1: Create/Update/Delete users, groups and roles. See: https://github.com/ansible/ansible-modules-core/pull/756
Part 2: Create/Update/Delete policy objects for users/groups/roles
Part 3: Upload/Remove/Update server certificates. See: #918 

### DIFF
--- a/cloud/amazon/iam_policy.py
+++ b/cloud/amazon/iam_policy.py
@@ -1,0 +1,325 @@
+#!/usr/bin/env python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


adamchainz:
As I reported on #756 , I've had to de-lint this module to try it in my codebase. I found an apparent bug in doing so!

My de-linted _iam_policy.py_ is at https://gist.github.com/adamchainz/7f5834fbbfa8e72c48da#file-iam_policy-py

The apparent bug is on lines 162-163, and 200-201. Both create a `msg` that is then not used - I think it should be sent to `module.fail_json`, but I haven't fixed it because it doesn't affect me at current.

I also made two enhancements:
1. `state` defaults to `present` like all ansible modules
2. `policy_document` can take a dict inline, which means fewer files to manage, especially with simple policies. The DOCUMENTATION and EXAMPLES were updated in accordance with this.

bcoca:
> 1. state defaults to present like all ansible modules

not true, some do do so but we prefer state to be required and explicit.

evanccnyc:
Does this need more testing before check in?

PR #889, opened 2015-03-03
GCE module: allow to keep instance boot disk after instance deletion

nabam:
Useful to create images from boot disks with gce_img

### DIFF
--- a/cloud/google/gce.py
+++ b/cloud/google/gce.py
@@ -129,6 +129,13 @@ options:
     required: false
     default: "ephemeral"
     aliases: []
+  disk_auto_delete:
+    version_added: "1.9"
+    description:
+      - if set boot disk will be removed after instance destruction
+    required: false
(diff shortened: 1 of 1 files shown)


nabam:
@erjohnso please take a look at this pull request also

cloud/google/gce.py
```diff
@@ -129,6 +129,12 @@
     required: false
     default: "ephemeral"
     aliases: []
+  disk_auto_delete:
```
erjohnso, line 132: Should we have a `version_added` here?

erjohnso:
One quick suggestion, but otherwise looks great, thanks for this @nabam.

cloud/google/gce.py
nabam, line 132: @erjohnso done

erjohnso:
Thanks @nabam, lgtm.

PR #890, opened 2015-03-04
Add pid mode support to docker module

ghost:
This allows setting the pid namespace for a container. Currently only
the 'host' pid namespace is supported.

This requires Docker 1.4.1 and docker-py 1.0.0

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -209,6 +209,13 @@ options:
     default: false
     aliases: []
     version_added: "1.8"
+  pid:
+    description:
+      - Set the PID namespace mode for the container (currently only supports 'host'). Requires docker-py >= 1.0.0 and docker >= 1.4.1.
+    required: false
+    default: false
+    aliases: []
+    version_added: "1.8"
   registry:
     description:
       - The remote registry URL to use for pulling images.
@@ -790,6 +797,7 @@ class DockerManager(object):
             'privileged':   self.module.params.get('privileged'),
             'links': self.links,
             'network_mode': self.module.params.get('net'),
+            'pid_mode': self.module.params.get('pid'),
         }
 
         optionals = {}
@@ -872,6 +880,7 @@ def main():
             lxc_conf        = dict(default=None, type='list'),
             name            = dict(default=None),
             net             = dict(default=None),
+            pid             = dict(default=None),
             insecure_registry = dict(default=False, type='bool'),
         )
     )
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -213,9 +213,9 @@ options:
     description:
       - Set the PID namespace mode for the container (currently only supports 'host'). Requires docker-py >= 1.0.0 and docker >= 1.4.1.
     required: false
-    default: false
+    default: None
     aliases: []
-    version_added: "1.8"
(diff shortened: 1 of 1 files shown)


ghost:
First time contributing to ansible. Let me know if I missed some process I needed to follow.

bobrik:
:+1: for this functionality

bcoca:
The default should either be made to be False or match the documentation to the actual default 'None', also version added will at best be 1.9

ghost:
Matching the config of 'net' it would be false. I can set default to 'null' to match the other defaults in the documentation. Which is prefered, consistency in the module or setting it to 'None'?

bcoca:
Adding correcting the rest of the module to my list, for now lets just make sure new additions are correct.

bobrik:
@bcoca any chance to get it in 1.9?

abadger:
@SamYaple Since this requires a more recent version of docker and docker-py, this needs to be handled like the other optional_params.  I can put together the code to do that but I'd need to know the docker versions that are needed to make this work.

docker-py 1.0.0 works for the client side check.  What does Docker 1.4.1 equate to in terms of the Docker server APIVersion?

Without this we'll have to revert this and bring it back in 1.10/2.0.

abadger:
If I'm reading the source correctly, docker-1.4.0 and docker-1.4.1 both have APIVersion 1.16

https://github.com/docker/docker/blob/v1.4.1/api/common.go#L18
https://github.com/docker/docker/blob/v1.4.0/api/common.go#L18

Does that mean that pid_mode works with docker server 1.4.0?  Or do we need to use Docker APIVersion 1.17 which means docker server 1.5.0 and higher?

bobrik:
Setting pid namespace is only supported in docker 1.5

On Friday, March 13, 2015, Toshio Kuratomi notifications@github.com wrote:

> If I'm reading the source correctly, docker-1.4.0 and docker-1.4.1 both
> have APIVersion 1.16
> 
> https://github.com/docker/docker/blob/v1.4.1/api/common.go#L18
> https://github.com/docker/docker/blob/v1.4.0/api/common.go#L18
> 
> Does that mean that pid_mode works with docker server 1.4.0? Or do we need
> to use Docker APIVersion 1.17 which means docker server 1.5.0 and higher?
> 
> —
> https://github.com/ansible/ansible-modules-core/pull/890#issuecomment-78813632
> .

## 

Regards, Ian Babrou
http://bobrik.name http://twitter.com/ibobrik skype:i.babrou

abadger:
@bobrik thanks!

bobrik:
I'm handling different docker versions this way:
1. Having docker-py 1.0.0 on each docker host
2. Setting `docker_api_version=1.16` in inventory for each docker host (this is for 1.4)
3. Using `docker_api_version: "{{ docker_api_version | default(None) }}"` in playbooks

Is there a better/recommended way to deal with docker version incompatibilities?

ghost:
Sorry guys didn't verify the version number.

@abadger You are right. It is in APIVersion 1.17, >=docker-1.5.0. Thanks for the catch (and patch!) Ill let you take it from here.

abadger:
Cool -- If you guys can, please test out this pull request: https://github.com/ansible/ansible-modules-core/pull/936

PR #892, opened 2015-03-04
Fix misreporting of changed state for update-rc.d managed sysv services.

jonhadfield:
The update-rc.d code used to manage SysV services reports changed as true when enabling an already enabled service or disabling a disabled service. This is because self.changed is set to a default of true in the 'service_enable' function and never set to false, under any conditions, in the code for update-rc.d.


abadger:
Thanks for the PR but I think I just fixed this yesterday: https://github.com/ansible/ansible-modules-core/commit/939ecd33d4a6706e7ed3d021043143ebff177f8b

(Used an else instead of setting a new default).

# Possible Duplicated Efforts

Hi!

It looks like the work from this pull request is a duplicate of the following PR(s):
- https://github.com/ansible/ansible-modules-core/commit/939ecd33d4a6706e7ed3d021043143ebff177f8b

Based on this, we are going to close this PR in favor of the above.

However, we're absolutely always up for discussion.  Since this is a really busy project, we don't always see comments on closed tickets, but want to encourage open dialog.  You can stop by the development list, and we'd be glad to talk about it - and we might even be persuaded otherwise!

jonhadfield:
Ah, your fix works for me too. Thanks.

PR #897, opened 2015-03-05
openstack nics by name

overquota:
add nic to instance by name (label in nova) besides net-id

### DIFF
--- a/cloud/openstack/_nova_compute.py
+++ b/cloud/openstack/_nova_compute.py
@@ -115,7 +115,7 @@ options:
      default: None
    nics:
      description:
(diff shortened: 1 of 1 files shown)


overquota:
We need this feature because we have several regions with 1 share network which has the same name everywhere, but different net-id

xsuchy:
I would like to have this feature as well.

emonty:
We're not taking new features to the old nova_compute modue in favor of the new os_server module. However, the feature you have here is handled in the new module in #1055

overquota:
cool, thanks!

PR #903, opened 2015-03-06
Allow vpc_subnet_id to remain unspecified (see bc0c169)

bretmartin:
@cwarner-mdsol alerted me to this problem in a previous commit of mine. `vpc_subnet_id` must be allowed to be unset, in which case `vpc_id` still needs to be set, at least to `None`.

@cwarner-mdsol, can you test that this fixes your issue?


cwarner-mdsol:
worksforme.

PR #809, opened 2015-02-18
Update ec2_ami_search module lto be able to search for different amis on ec2

haad:
Change ec2_ami_search  to be able to search amazon for given images based on tags, name, uuid, or any other option described in describe images manual page.

```
- name: Search for ami
  ami_search:
   region: us-east-1
   owner: 'self'
   ami_tags: {"ami_uuid": '{{build_ami_id}}' }
  register: image
```

PR #905, opened 2015-03-07
Replacement for ec2_ami_search

manicminer:
This is a replacement for the current ec2_ami_search module. It makes use of the boto get_all_images method to search for AMIs matching a number of criteria. Results can be sorted and sliced.

Note that the options are not compatible with the current version of this module, although the current version of this module doesn't really have anything to do with EC2 at all, it just fetches and parses a text file from an Ubuntu project server. However one can achieve a near drop-in replacement with the new syntax (example given). 

Potential uses for this module might be
- Search for the latest and greatest AMI of your favourite distro or publisher
- Search your own AMIs and launch instances with them
- Search your own AMIs and perform actions on them, such as deleting older AMIs

### DIFF
--- a/cloud/amazon/ec2_ami_search.py
+++ b/cloud/amazon/ec2_ami_search.py
@@ -1,7 +1,5 @@
 #!/usr/bin/python
 #
-# (c) 2013, Nimbis Services
-#
 # This file is part of Ansible
 #
 # Ansible is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


bcoca:
~~seems to be a dupe of #809~~
competing PR #809

manicminer:
@bcoca Not really a dupe, this version is substantially different to #809 and offers more options. I structured the filters differently to better match the EC2 documentation and allow more fine grained searches. You can also choose the desired action when no AMIs are matched, either fail or return successful with an empty list of results.

Note also the inclusion of sorting/slicing options for the search results. This enables you to find the latest/oldest AMI (or number of AMIs) matching your criteria. I based this partially on the method described in [this askubuntu answer](http://askubuntu.com/a/53586) for finding the latest Ubuntu AMI and also found it particularly useful for other workflows. For example, in a process which automatically bakes AMIs, it offers a way to 'clean up' the older images that you no longer need.

``` yaml
- ec2_ami_search:
    owner: self
    ami_tags:
      - foo: bar
    sort: name
    sort_end: "-{{ ami_retention }}"
  register: old_ami_search

- ec2_ami:
    region: "{{ region }}"
    image_id: "{{ item.ami_id }}"
    delete_snapshot: yes
    state: absent
  with_items: old_ami_search.results
```

I understand if you'd want to rename this module given the incompatibility with the current module of the same name. Also happy to make that change, and open to any naming suggestions.

bcoca:
Sorry, 'dupe' might be misleading, it was more an indication for myself and the core team to make sure to examine both efforts and choose the one we think is best (or even combine them).

haad:
As a autor of #809 I think this module should be included as it's better than mine.

bcoca:
closed #809 in favor of this one, will review this week

robbwagoner:
I've used @tombamford's ec2_ami_search module to great extent. It works very well. Well done @tombamford. +1 for accepting this PR.

bcoca:
i still think it should be renamed as it is not backwards compatible and will break playbooks, i'm fine with deprecating the existing one but not just breaking it

bcoca:
ec2_find_ami ?

haad:
sounds good to me.

On Thu, Apr 16, 2015 at 5:25 PM, Brian Coca notifications@github.com
wrote:

> ec2_find_ami ?
> 
> —
> https://github.com/ansible/ansible-modules-core/pull/905#issuecomment-93762930
> .

## 

Regards.

Adam

bcoca:
so did a few fixes myself and added as ec2_ami_find

PR #907, opened 2015-03-08
[git] Determine branch name more reliable

tverlaan:
A branch name may contain slashes. The current way is not reliable. My suggestion should be sufficient (I think).

`split('/',2)` will make sure to only remove the first two entries (split by slahses of course)
`replace(remote,'')` will remove the remote from the name if the branch name is taken from the remote subdir in the gitdit
`lstrip('/')` is used to remove a preleading slash.


gregdek:
This module is maintained by the Ansible core team, so it may take a while longer to get a review; we appreciate your patience. attn: @bcoca @abadger @jimi-c @bennojoy

source_control/git.py
```diff
@@ -469,7 +469,7 @@ def get_head_branch(git_path, module, dest, remote, bare=False):
     if is_not_a_branch(git_path, module, dest):
         f.close()
         f = open(os.path.join(repo_path, 'refs', 'remotes', remote, 'HEAD'))
-    branch = f.readline().split('/')[-1].rstrip("\n")
+    branch = f.readline().split('/',2)[-1].replace(remote,'').lstrip("/").rstrip("\n")
```
mgedmin: `.replace(remote, '')` is dangerous, since it replaces all occurrences. What if my branch is named `original-idea` and the remote is named `origin`?  This new code would return a branch name of `al-idea`.

gregdek:
Thanks for the PR @tverlaan. Since git functionality is central and requires review from the core team, we won't have time to evaluate for inclusion in 2.0.  Marking with the "next" release tag accordingly.

PR #910, opened 2015-03-08
Less reloads in docker module

bobrik:
Related to #502. I haven't used python in ages, feel free to correct me.

This patch removes unnecessary reloads when:
- Bind mounts are used
- Port bindings are used

cc @smashwilson @abadger

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -725,16 +725,24 @@ class DockerManager(object):
         return False
 
     def get_inspect_image(self):
+        try:
+            return self.client.inspect_image(self.module.params.get('image'))
+        except DockerAPIError as e:
+            if e.response.status_code == 404:
+                return None
+            else:
+                raise e
+
+    def get_image_repo_tags(self):
         image, tag = get_split_image_tag(self.module.params.get('image'))
         if tag is None:
             tag = 'latest'
         resource = '%s:%s' % (image, tag)
 
-        matching_image = None
         for image in self.client.images(name=image):
             if resource in image.get('RepoTags', []):
-                matching_image = image
-        return matching_image
+                return image['RepoTags']
+        return None
 
     def get_inspect_containers(self, containers):
         inspect = []
@@ -796,13 +804,21 @@ class DockerManager(object):
                     continue
 
             # EXPOSED PORTS
-            # Note that ports that are bound at container run are also exposed
-            # implicitly.
-            expected_exposed_ports = set()
+
+            image_exposed_ports = set((image["ContainerConfig"]["ExposedPorts"] or {}).keys())
+
+            expected_exposed_ports = set(image_exposed_ports)
             for p in (self.exposed_ports or []):
                 expected_exposed_ports.add("/".join(p))
 
-            actually_exposed_ports = set((container["Config"]["ExposedPorts"] or {}).keys())
+            if self.port_bindings:
+                for container_port in self.port_bindings.keys():
+                    if isinstance(container_port, int):
+                        container_port = "{}/tcp".format(container_port)
+
+                    expected_exposed_ports.add(container_port)
+
+            actually_exposed_ports = set((container["NetworkSettings"]["Ports"] or {}).keys())
 
             if actually_exposed_ports != expected_exposed_ports:
                 self.reload_reasons.append('exposed_ports ({} => {})'.format(actually_exposed_ports, expected_exposed_ports))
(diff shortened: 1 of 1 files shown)


cloud/docker/docker.py
```diff
@@ -725,16 +725,18 @@ def has_changed(self):
         return False
 
     def get_inspect_image(self):
+        return self.client.inspect_image(self.module.params.get('image'))
+
+    def get_image_repo_tags(self):
```
smashwilson, line 736: I'm a bit curious about the `get_inspect_image` / `get_image_repo_tags` split here. Did this solve a specific problem you encountered, or is it just a better client method to call?

Also, does `self.client.inspect_image` still work if you leave off the image tag?

smashwilson:
:sparkles: Thanks! I'll take this for a spin momentarily, but it looks perfectly reasonable to me at first read.

cloud/docker/docker.py
bobrik, line 736: `get_inspect_image` is essentially `docker inspect image-name`, it does not have `RepoTags` key.

`self.client.inspect_image` works without a tag, but fails if image does not exist locally, I'm pushing fix right now.

smashwilson, line 736: :+1:

smashwilson:
By the way, I'm augmenting my test playbook to include [an image with pre-exported ports, volumes, and environment variables](https://github.com/smashwilson/docker-module-test/commit/bef7a3cd323eb4b259268ca920af64d39deee1d7). It's currently failing, but it'll help me catch the rest of these cases.

cloud/docker/docker.py
```diff
@@ -796,13 +804,22 @@ def get_differing_containers(self):
...
             for p in (self.exposed_ports or []):
                 expected_exposed_ports.add("/".join(p))
 
-            actually_exposed_ports = set((container["Config"]["ExposedPorts"] or {}).keys())
+            if self.port_bindings:
+                for container_port in self.port_bindings.keys():
+                    if isinstance(container_port, int):
+                        container_port = "{}/tcp".format(container_port)
+
+                    expected_exposed_ports.add(container_port)
```
smashwilson, line 819: The problem with including port bindings in the exposed ports section here is that the reload_reason will be incorrect if you change a port binding. You'll never have a chance to reach the actual `# PORT BINDINGS` section later in the function; it'll always short-circuit here at `exposed_ports` if the port bindings have changed.

As much as possible, I tried to keep a one-to-one correspondence between the sections in `get_differing_containers` and the actual settings that users set on the module. Yes, ports that are bound at container start time are also "exposed"; but the module parameter that's used to configure them is `ports:`, not `expose:`.

cloud/docker/docker.py
```diff
@@ -796,13 +804,22 @@ def get_differing_containers(self):
...
+
+            container_exposed_ports = set((container["NetworkSettings"]["Ports"] or {}).keys())
+            actually_exposed_ports = image_exposed_ports.union(container_exposed_ports)
```
smashwilson: Also, now that I think about it, it's strange to union `image_exposed_ports` into both `expected_exposed_ports` _and_ `actually_exposed_ports`. Doesn't that make it not do anything? You should only need to include it in `expected_exposed_ports`, I would think; they should all be in `actually_exposed_ports` anyway. If they aren't, there's something wrong and we probably need to reboot anyway :wink:

bobrik: You are right, `container_exposed_ports` is `actually_exposed_ports` anyway.

cloud/docker/docker.py
bobrik, line 819: Let's check exposed ports after port bindings. Would that solve the issue?

smashwilson, line 819: Uh, kind of? It seems a lot easier to me to only check explicitly exposed ports in the exposed ports section. What does adding `self.port_bindings` actually gain?

I'm playing with this myself right now in my test playbook, let's see if I can put something together.

bobrik, line 819: Hm, looks like you're right, it works as expected even without adding port bindings. Do you have an example when it could break?

bobrik, line 819: Huh, no, it won't work if you have port bindings without explicitly exposed ports. Not sure if there's a better way than checking exposed ports after port bindings then.

smashwilson, line 819: Do you mean checking against `container["NetworkSettings"]["Ports"]`? That won't work - I mean comparing:

``` python
expected_exposed_ports = set((image['ContainerConfig']['ExposedPorts'] or {}).keys())
for p in (self.exposed_ports or []):
    expected_exposed_ports.add("/".join(p))

actually_exposed_ports = set((container["Config"]["ExposedPorts"] or {}).keys())
```

Image exposed ports + module parameter exposed ports = container exposed ports. Comparing apples to apples, right?

I actually have a PR ready that handles this, environment variables, and volumes from the image configuration. Just a moment, I'll push it up and submit it.

bobrik:
Closing in favor of #912.

PR #912, opened 2015-03-09
[docker] Account for configuration performed in the image for state=reloaded

smashwilson:
This modifies the container diffing algorithm used by the Docker module's `state=reloaded` container diffing to take into account exposed ports, environment variables, and volumes from the Dockerfile, like so:

``` Dockerfile
EXPOSE 8000
ENV THING0 wat
ENV THING1 wat
VOLUME /var/one
VOLUME /var/two
```
- [x] Handle overriding Dockerfile environment variables in the module
- [x] Test overriding Dockerfile volume mounts with module parameters that are:
  - [x] non-bind volumes
  - [x] bind volumes
- [x] Use Python 2.6-compatible `.format()` calls

/cc @abadger, @bobrik

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -804,9 +804,7 @@ class DockerManager(object):
                     continue
 
             # EXPOSED PORTS
-            # Note that ports that are bound at container run are also exposed
-            # implicitly.
-            expected_exposed_ports = set()
+            expected_exposed_ports = set((image['ContainerConfig']['ExposedPorts'] or {}).keys())
             for p in (self.exposed_ports or []):
                 expected_exposed_ports.add("/".join(p))
 
@@ -820,7 +818,7 @@ class DockerManager(object):
             # VOLUMES
             # not including bind modes.
 
-            expected_volume_keys = set()
+            expected_volume_keys = set((image['ContainerConfig']['Volumes'] or {}).keys())
             if self.volumes:
                 for key, config in self.volumes.iteritems():
                     if not config and key not in self.binds:
@@ -846,13 +844,13 @@ class DockerManager(object):
             # actual_env is likely to include environment variables injected by
             # the Dockerfile.
 
-            expected_env = set()
+            expected_env = set(image['ContainerConfig']['Env'] or [])
             if self.env:
                 for name, value in self.env.iteritems():
                     expected_env.add("{}={}".format(name, value))
             actual_env = set(container['Config']['Env'] or [])
 
-            if not actual_env.issuperset(expected_env):
+            if actual_env != expected_env:
                 # Don't include the environment difference in the output.
                 self.reload_reasons.append('environment')
                 differing.append(container)
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -816,13 +816,11 @@ class DockerManager(object):
                 continue
 
             # VOLUMES
-            # not including bind modes.
 
             expected_volume_keys = set((image['ContainerConfig']['Volumes'] or {}).keys())
             if self.volumes:
-                for key, config in self.volumes.iteritems():
-                    if not config and key not in self.binds:
-                        expected_volume_keys.add(key)
+                expected_volume_keys.update(self.volumes.keys())
+
             actual_volume_keys = set((container['Config']['Volumes'] or {}).keys())
 
             if actual_volume_keys != expected_volume_keys:
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -842,11 +842,20 @@ class DockerManager(object):
             # actual_env is likely to include environment variables injected by
             # the Dockerfile.
 
-            expected_env = set(image['ContainerConfig']['Env'] or [])
+            expected_env = {}
+
+            for image_env in image['ContainerConfig']['Env'] or []:
+                name, value = image_env.split('=', 1)
+                expected_env[name] = value
+
             if self.env:
                 for name, value in self.env.iteritems():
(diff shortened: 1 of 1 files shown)


bobrik:
:+1: for exposed ports.

However, bind-mounts without corresponding `VOLUME` in `Dockerfile` constantly trigger reloads. #910 didn't have this problem.

smashwilson:
> However, bind-mounts without corresponding VOLUME in Dockerfile constantly trigger reloads. #910 didn't have this problem.

Ah! I missed that bit. Just a moment, I'll test it and pull in that change, too.

smashwilson:
@bobrik Give this a try. Took me a while to get a proper test case that failed for this one.

smashwilson:
Oh - ha. My `teardown` playbook was leaving `testimage` containers around between executions. Well, that explains why things were wonky.

bobrik:
Seems to be working! :ship:

abadger:
So if I understand what you are saying about Dockerfile config settings correctly, if the Dockerfile sets ENV['foo'] = 'bar' and the ansible task tries to set ENV['foo'] = 'baz', the state of the container after ansible is run will be ENV['foo'] == bar.  If that's correct we should also do some checking that none of the ansible params are trying to change something that are already set.  If they are, then module.fail_json() so that the user gets an error instead of the Docker container silently having a different state than they requested.

smashwilson:
> So if I understand what you are saying about Dockerfile config settings correctly, if the Dockerfile sets ENV['foo'] = 'bar' and the ansible task tries to set ENV['foo'] = 'baz', the state of the container after ansible is run will be ENV['foo'] == bar.

Nope. The state of the container in this case would be `ENV['foo'] == 'baz'`, because the parameters used to launch the container take precedence over the ones built into the image by the Dockerfile. It's a convenient way to specify defaults, but let users override then at launch time.

The issue that this fixes is that if an image is built with certain common settings in its Dockerfile -- `EXPOSE`, `VOLUME`, or `ENV` -- using `state=reloaded` would always cause extraneous restarts, because the container inspect result will include those as well as any that were provided manually at launch.

However, it's true that I'm not accounting for environment variables that are present in both places properly... what I need to do is parse the image's env into a dict, then apply module parameters over it, then reformat that into a form that's comparable to the container inspect results. Good catch there :smile:

abadger:
Yeah, env being a dict makes it trickier than the rest of the values :-)

Another thing you can clean up if you touch it in your cleanup -- modules need to be either python2.4 compatible OR whatever minimum python version the modules it requires need.  For docker-py that appears to be python2.6.  In 2.6, '{} {}'.format() doesn't work.  '{0} {1}'.format() is necessary.  I was going to clean that up after but since you're already in that section of code I don't want to make merging harder ;-)

smashwilson:
> Yeah, env being a dict makes it trickier than the rest of the values :-)

A dict stored as a list of strings to parse, no less!

> Another thing you can clean up if you touch it in your cleanup -- modules need to be either python2.4 compatible OR whatever minimum python version the modules it requires need. For docker-py that appears to be python2.6. In 2.6, '{} {}'.format() doesn't work. '{0} {1}'.format() is necessary. I was going to clean that up after but since you're already in that section of code I don't want to make merging harder ;-)

Ah! Noted and added to a checklist on the PR description :point_up:.

smashwilson:
Done and tested. I was worried that overriding volumes from the Dockerfile may have a similar problem, but because "volumes" (which is a simple list) and "binds" are handled separately, they do not!

abadger:
Thanks!  I'm going to have a hectic day today, so if you have more fixes you need in feel free to ping me (abadger1999) on irc.freenode.net if you need to get my attention (if I'm not on irc today then I'm driving today :-)

# Merged

Hi!

issue is resolved.

Thank you!

bobrik:
Dynamic typing and yaml strike back:

```
>>> {"a": 5} == {"a": "5"}
False
```

Integer/float env variables that are not explicitly converted to strings in playbooks trigger reload.

bobrik:
Links also trigger reloads. @smashwilson, how do you feel about this one? Not sure what was `running` needed for, is it because you named your test container `running`? :)

``` diff
diff --git a/library/docker/docker.py b/library/docker/docker.py
index 7a83694..208cc45 100644
--- a/library/docker/docker.py
+++ b/library/docker/docker.py
@@ -859,7 +859,7 @@ class DockerManager(object):

             if actual_env != expected_env:
                 # Don't include the environment difference in the output.
-                self.reload_reasons.append('environment')
+                self.reload_reasons.append('environment {} => {}'.format(actual_env, expected_env))
                 differing.append(container)
                 continue

@@ -992,7 +992,7 @@ class DockerManager(object):

             expected_links = set()
             for link, alias in (self.links or {}).iteritems():
-                expected_links.add("/{0}:/running/{1}".format(link, alias))
+                expected_links.add("/{0}:{1}/{2}".format(link, container["Name"], alias))

             actual_links = set(container['HostConfig']['Links'] or [])
             if actual_links != expected_links:
```

smashwilson:
> Integer/float env variables that are not explicitly converted to strings in playbooks trigger reload.

Hah - good catch.

That's not so bad, you can always quote it in the playbook. But we should coerce to string, I agree.

> Links also trigger reloads. @smashwilson, how do you feel about this one? Not sure what was running needed for, is it because you named your test container running? :)

D'oh, I bet you're right! I was puzzled where that one was coming from. Mind opening an issue and cc'ing me?

bobrik:
@smashwilson, I opened #916. Fixes are pretty trivial, I decided to skip issue and fast-forward with PR.

PR #914, opened 2015-03-09
Add support for SNS notification ARNs in CloudFormation

s0enke:
This adds the possibility to push stack events to SNS topics.

cc @bracki

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -54,6 +54,12 @@ options:
     required: false
     default: null
     aliases: []
+  notification_arns:
+    description:
+      - The Simple Notification Service (SNS) topic ARNs to publish stack related events.
+    required: false
+    default: null
+    version_added: "2.0"
   stack_policy:
     description:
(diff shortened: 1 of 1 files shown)


bennojoy:
@s0enke  could you please resolve the conflicts .. thanks

s0enke:
@bennojoy Done! Would also be good to know if there is a place where I could add integration tests if wanted :)

bennojoy:
@s0enke  thanks and sorry i should have mentioned this earlier ,can you please add version_added: 2.0 in the module doc for the new parameter and an emplty alias[] is not needed. and tests can be added in https://github.com/ansible/ansible/tree/devel/test it does need some work for all modules , it is in my todo list too

s0enke:
> @s0enke thanks and sorry i should have mentioned this earlier ,can you please add version_added: 2.0 in the module doc for the new parameter and an emplty alias[] is not needed.

Ok, done!

bennojoy:
merged thanks

Issue #919, opened 2015-03-10
{ failed: "true" } error when running basic rax module from samples


kenperkins:
I'm using a slightly modified playbook from the [examples](http://docs.ansible.com/rax_module.html#examples). I'm expliciting adding my username and api_key, as well as region.

Ansible Version 1.8.4
Pyrax version 1.9.3

```
- name: Build a Cloud Server
  hosts: all
  gather_facts: False
  tasks:
    - name: Server build request
      failed_when: false
      ignore_errors: true
      local_action:
        module: rax
        username: myusername
        api_key: xxxxx
        name: rax-test1
        region: IAD
        flavor: general1-1
        image: fed66ee7-fb4b-4464-b8c6-609c1fde218f
        key_name: ken-mbp
        wait: yes
        state: present
        networks:
          - private
          - public
      register: rax
```

```
$ ansible-playbook -i localhost, playbook.yml

PLAY [Build a Cloud Server] ***************************************************

TASK: [Server build request] **************************************************
failed: [localhost -> 127.0.0.1] => {"failed": true}

FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
           to retry, use: --limit @/Users/ken/playbook.retry

localhost                  : ok=0    changed=0    unreachable=0    failed=1
```

sivel:
You should have gotten additional information below the 'failed'.  

Something to the effect of:

```
TASK: [Server build request] **************************************************
failed: [localhost -> 127.0.0.1] => {"failed": true}
msg: iad is not a valid region, must be one of: SYD,IAD,DFW,ORD,HKG
```

To note, that is indeed 1 of your errors, 'iad' is not a valid region, it should be 'IAD'.  I'll also assume it is a typo that you have no 'hosts' listed in your play.

On another note, I am unable to confirm whether that image exists, but that should also lead to `msg: Can not find requested image`.

It may also help to run with extra verbosity such as using `-vv` in your ansible-playbook run.  If you can provide the output from that run, it may be useful in determining what your issue is.

kenperkins:
@sivel you're correct about the missing hosts; I updated my example above to be accurate with what I ran.

Regarding the error you listed above, that does not appear. I see nothing beyond the `{"failed": true}` output.

I'll re-run with `-vv`.

kenperkins:
Ok with `-vv` it doesn't materially change the output, just adds a log line above:

```
[...]

PR #920, opened 2015-03-10
rax: Prevent an empty error message

sivel:
Along with https://github.com/ansible/ansible/pull/10421 this PR tries to avoid a situation where we catch an exception that does not contain a message.


sivel:
See also https://github.com/ansible/ansible-modules-core/issues/919

PR #926, opened 2015-03-12
Alternative implementation of tls for docker

abadger:
Alternative to #854 
@dguerri, @smashwilson Could you guys take a look at this?  If it looks good and tests good I'll see if this is okay to merge for 1.9.  If we can't get enough testing or the other committers are leary of merging it now then I'll have to work on something that just changes the parameter names for 1.9.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -567,19 +567,18 @@ class DockerManager(object):
         if not tls_ca_cert and env_cert_path:
             tls_ca_cert = os.path.join(env_cert_path, 'ca.pem')
 
-        if tls_ca_cert:
-            tls_hostname = module.params.get('tls_hostname')
-            if tls_hostname is None:
-                if env_docker_hostname:
-                    tls_hostname = env_docker_hostname
+        tls_hostname = module.params.get('tls_hostname')
+        if tls_hostname is None:
+            if env_docker_hostname:
+                tls_hostname = env_docker_hostname
+            else:
+                parsed_url = urlparse(docker_url)
+                if ':' in parsed_url.netloc:
+                    tls_hostname = parsed_url.netloc[:parsed_url.netloc.rindex(':')]
                 else:
-                    parsed_url = urlparse(docker_url)
-                    if ':' in parsed_url.netloc:
-                        tls_hostname = parsed_url.netloc[:parsed_url.netloc.rindex(':')]
-                    else:
-                        tls_hostname = parsed_url
-            if not tls_hostname:
-                tls_hostname = True
+                    tls_hostname = parsed_url
+        if not tls_hostname:
+            tls_hostname = True
 
         # use_tls can be one of four values:
         # no: Do not use tls
@@ -610,8 +609,10 @@ class DockerManager(object):
                 else:
                     params['verify'] = True
                     params['assert_hostname'] = tls_hostname
+            elif use_tls == 'encrpyt':
+                params['verify'] = False
 
-            if params or use_tls == 'encrypt':
+            if params:
                 # See https://github.com/docker/docker-py/blob/d39da11/docker/utils/utils.py#L279-L296
                 docker_url = docker_url.replace('tcp://', 'https://')
                 tls_config = docker.tls.TLSConfig(**params)
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -102,18 +102,41 @@ options:
       - URL of the host running the docker daemon. This will default to the env
         var DOCKER_HOST if unspecified.
     default: ${DOCKER_HOST} or unix://var/run/docker.sock
-  docker_tls_cert:
+  use_tls:
     description:
-      - Path to a PEM-encoded client certificate to secure the Docker connection.
+      - Whether to use tls to connect to the docker server.  "no" means not to
+        use tls (and ignore any other tls related parameters). "encrypt" means
+        to use tls to encrypt the connection to the server.  "verify" means to
+        also verify that the server's certificate is valid for the server
+        (this both verifies the certificate against the CA and that the
+        certificate was issued for that host. If this is unspecified, tls will
+        only be used if one of the other tls options require it.
+    choices: [ "no", "encrypt", "verify" ]
+    version_added: "1.9"
+  tls_client_cert:
+    description:
+      - Path to the PEM-encoded certificate used to authenticate docker client.
+        If specified tls_client_key must be valid
     default: ${DOCKER_CERT_PATH}/cert.pem
-  docker_tls_key:
+    version_added: "1.9"
+  tls_client_key:
     description:
-      - Path to a PEM-encoded client key to secure the Docker connection.
+      - Path to the PEM-encoded key used to authenticate docker client. If
+        specified tls_client_cert must be valid
     default: ${DOCKER_CERT_PATH}/key.pem
-  docker_tls_cacert:
+    version_added: "1.9"
+  tls_ca_cert:
     description:
       - Path to a PEM-encoded certificate authority to secure the Docker connection.
+        This has no effect if use_tls is encrypt.
     default: ${DOCKER_CERT_PATH}/ca.pem
+    version_added: "1.9"
+  tls_hostname:
+    description:
+      - A hostname to check matches what's supplied in the docker server's
+        certificate.  If unspecified, the hostname is taken from the docker_url.
+    default: Taken from docker_url
+    version_added: "1.9"
   docker_api_version:
     description:
       - Remote API version to use. This defaults to the current default as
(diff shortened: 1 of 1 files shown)


dguerri:
I will test it asap (long day of work ahead, tho)... 
BTW: I love the `required_together` thing :+1:

cloud/docker/docker.py
```diff
@@ -102,18 +102,41 @@
...
+  use_tls:
     description:
-      - Path to a PEM-encoded client certificate to secure the Docker connection.
+      - Whether to use tls to connect to the docker server.  "no" means not to
+        use tls (and ignore any other tls related parameters). "encrypt" means
+        to use tls to encrypt the connection to the server.  "verify" means to
+        also verify that the server's certificate is valid for the server
+        (this both verifies the certificate against the CA and that the
+        certificate was issued for that host. If this is unspecified, tls will
+        only be used if one of the other tls options require it.
```
smashwilson, line 113: Hmm, do the docs render right like this? I always had to prefix each line with a `-` to make the parser happy.

cloud/docker/docker.py
```diff
@@ -1347,9 +1402,11 @@ def main():
             memory_limit    = dict(default=0),
             memory_swap     = dict(default=0),
             docker_url      = dict(),
-            docker_tls_cert = dict(),
-            docker_tls_key  = dict(),
-            docker_tls_cacert = dict(),
+            use_tls         = dict(default=None, choices=['no', 'encrypt', 'verify']),
+            tls_client_cert = dict(required=False, default=None, type='str'),
```
smashwilson, line 1412: I'd been assuming that `required=False, default=None, type='str'` were the defaults. Any reason not to just use `dict()` instead?

cloud/docker/docker.py
```diff
@@ -1372,7 +1429,10 @@ def main():
             net             = dict(default=None),
             pid             = dict(default=None),
             insecure_registry = dict(default=False, type='bool'),
-        )
+        ),
+        required_together = (
+            ['tls_client_cert', 'tls_client_key'],
+        ),
```
smashwilson, line 1441: :+1:

cloud/docker/docker.py
smashwilson, line 113: I just tested and they're fine. Good to know :+1:

abadger, line 113: Yep, there's certain formatting (for instance colon) that you have to use an initial  dash with.  If those aren't in the docstring, this formatting works.

cloud/docker/docker.py
abadger, line 1412: You are correct about the default values.  I like explicitness as it's self documenting (especially for default.  type and required I am more indifferent to).  I can get rid of required and type if you want.

smashwilson:
Okay, I was able to use this and connect with boot2docker, but only by adding parameters to my docker tasks:

``` yaml
- name: remove all containers
  docker: state=absent image=debian:jessie use_tls=verify tls_hostname=boot2docker
  tags: ["teardown"]
```

That's probably okay. :smile: It would be nice to structure things so that the same docker tasks can execute against localhost across platforms, though. To that end I would propose:
1. Default `use_tls` to `"verify"` if it's unspecified but `tls_ca_cert` is provided (by module parameter or derived from environment variable).
2. Accept another `DOCKER_` environment variable to set `tls_hostname`. `DOCKER_TLS_HOSTNAME` probably?

It's a little less than optimal to make Mac users take some extra step to be able to use this... but I much prefer it to the current code, which defaults to lessened security!

dguerri:
hey guys, so we dropped the option to skip hostname verification.
I am ok with that, I understand that it is something rarely useful.

abadger:
@smashwilson Re: use_tls and tls_ca_cert, yeah, I went back and forth on how that should be implemented several times.  I can see changing that.  It would also be fine with me to add use of another docker env variable.  I was under the impression that these were standard env variables, though (from docker, docker-py, or boot2docker).  Is that not true?

@dguerri Yeah.  I don't really want to skip hostname verification as it can lead to extreme insecurity.  However, I was hoping that using tls_hostname would be a workaround for any place that you would otherwise be tempted to skip hostname verification.  Is that the case?

dguerri:
I like that approach. :+1: 
I am testing now. Will let you know asap

smashwilson:
> @smashwilson Re: use_tls and tls_ca_cert, yeah, I went back and forth on how that should be implemented several times. I can see changing that. It would also be fine with me to add use of another docker env variable. I was under the impression that these were standard env variables, though (from docker, docker-py, or boot2docker). Is that not true?

Yes, they're the ones we're already accepting, and you can automatically configure them with `$(boot2docker shellinit)`:

```
ashl6947@MMN1F3FD58 ~ $ boot2docker shellinit
Writing /Users/ashl6947/.boot2docker/certs/boot2docker-vm/ca.pem
Writing /Users/ashl6947/.boot2docker/certs/boot2docker-vm/cert.pem
Writing /Users/ashl6947/.boot2docker/certs/boot2docker-vm/key.pem
    export DOCKER_CERT_PATH=/Users/ashl6947/.boot2docker/certs/boot2docker-vm
    export DOCKER_TLS_VERIFY=1
    export DOCKER_HOST=tcp://192.168.59.103:2376
```

`DOCKER_TLS_HOSTNAME` would be a new one that Mac users would have to set manually, but I don't think that's _too_ bad. (It would be nice to document this somehow in the error message.)

> @dguerri Yeah. I don't really want to skip hostname verification as it can lead to extreme insecurity. However, I was hoping that using tls_hostname would be a workaround for any place that you would otherwise be tempted to skip hostname verification. Is that the case?

:+1: :100: Absolutely. I much prefer this to just disabling hostname verification. I wish Docker itself let you do this, too - or, better yet, if boot2docker signed the certificate to the correct hostname.

abadger:
@smashwilson, okay, Pushed a commit with your ideas implemented.  test it out and make sure I didn't flub anything :-)

dguerri:
Ok, sorry for the delay buy I had some problem setting up a development env for ansible 1.9

`tls_hostname` works as expected. Without it I can't authenticate the server.
Environment variables work as expected, if explicit parameters are not specified. (I got confused by this initially =D)

With `use_tls=encrypt` (and nothing else) I get this:

```
msg: SSLError(SSLError(SSLError(1, '_ssl.c:507: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed'),),)
```

I think this is beacuse we must set tls_verify to False. For instance:

``` diff
diff --git a/cloud/docker/docker.py b/cloud/docker/docker.py
index 1014fc1..e945b2c 100644
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -609,6 +611,7 @@ class DockerManager(object):
             if params or use_tls == 'encrypt':
                 # See https://github.com/docker/docker-py/blob/d39da11/docker/utils/utils.py#L279-L296
                 docker_url = docker_url.replace('tcp://', 'https://')
+                params['verify'] = use_tls != 'encrypt'
                 tls_config = docker.tls.TLSConfig(**params)

         self.client = docker.Client(base_url=docker_url,
```

Last, using just these two: `use_tls=verify tls_hostname=boot2docker` it exits for the wrong reason:

```
TASK: [Start MySQL container] ************************************************* 
  File "/Users/davide/.ansible/tmp/ansible-tmp-1426185465.11-224963353501429/docker", line 3092, in <module>
    main()
  File "/Users/davide/.ansible/tmp/ansible-tmp-1426185465.11-224963353501429/docker", line 1441, in main
    manager = DockerManager(module)
  File "/Users/davide/.ansible/tmp/ansible-tmp-1426185465.11-224963353501429/docker", line 607, in __init__
    params['assert_hostname'] = tls_hostname
UnboundLocalError: local variable 'tls_hostname' referenced before assignment


```

This fixes the issue:

``` diff
diff --git a/cloud/docker/docker.py b/cloud/docker/docker.py
index 1014fc1..0a827c9 100644
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -576,6 +576,8 @@ class DockerManager(object):
                     tls_hostname = parsed_url
             if not tls_hostname:
                 tls_hostname = True
+        else:
+            tls_hostname = None

         # use_tls can be one of four values:
         # no: Do not use tls
```

With this patch the error is:

```
msg: SSLError(SSLError(SSLError(1, '_ssl.c:507: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed'),),)
```

which makes more sense. But you may want to warn the user even before as `use_tls=verify` should require CA at least.

dguerri:
I will keep testing and I will keep you posted! 
Anyway: great job I like this much more than my first PR :)

abadger:
@dguerri tls_verify to False, good catch. I mistakenly thought that if we sent True instead of a TLSConfig, it would simply do encryption but I was wrong.  I've corrected that with an else: clause.

For tls_hostname, I think that we want it to be set to boot2docker in your example.  So Instead of defaulting to None, I removed the topmost if tls_ca_cert condition.  Does that look right?

For use_tls=verify without tls_ca_cert being specified... I think what we want is for docker-py to check that the docker server's tls certificate is valid according to the system's CA bundle.  This will be secure as long as hostname checking is also performed (we never turn that off so it should be secure).  This allows people to run a docker daemon on a host that has a valid, non-self signed certificate.  The certificate is validated by the public CAs and we know that it belongs to the server we intend because the certificate matches with the hostname that we're trying to reach.  I don't know the docker-py api very well, though, so if the code doesn't match the intentions, feel free to tell me :-)

dguerri:
> For use_tls=verify without tls_ca_cert being specified... I think what we want is for docker-py to check that the docker server's tls certificate is valid according to the system's CA bundle.

You are absolutely right, my bad. Even if there is no matching example here http://docker-py.readthedocs.org/en/latest/tls/, the first example confirms that.

dguerri:
> For tls_hostname, I think that we want it to be set to boot2docker in your example. So Instead of defaulting to None, I removed the topmost if tls_ca_cert condition. Does that look right?

Yep! That's correct again :)

abadger:
Cool :-)  If you guys don't find anymore problems I'll push to devel and stable-1.9 tomorrow and get some wider testing.

abadger:
Merged to devel and stable-1.9.  Thanks guys for the original code, feedback, and testing!

# Merged

Hi!

issue is resolved.

Thank you!

PR #934, opened 2015-03-12
git: set remote branch before switching versions when using depth arg

jbgo:
Below is an example playbook demonstrating the failure.

```
- name: switching branches fails when using depth arg
  hosts: localhost
  connection: local
  tasks:
    - git: >
        depth=1
        dest=test-clone
        repo=git@github.com:ansible/ansible-modules-core.git
        version=devel
        accept_hostkey=yes
    - git: >
        depth=1
        dest=test-clone
        repo=git@github.com:ansible/ansible-modules-core.git
        version=stable-1.9
        accept_hostkey=yes
```

This is because when creating a shallow, the --no-single-branch option is implied. Here's the relevant excerpt from `man git-clone`.

```
       --[no-]single-branch
           Clone only the history leading to the tip of a single branch, either specified by the --branch option or the primary branch remote's HEAD points at. When
           creating a shallow clone with the --depth option, this is the default, unless --no-single-branch is given to fetch the histories near the tips of all
           branches. Further fetches into the resulting repository will only update the remote-tracking branch for the branch this option was used for the initial
           cloning. If the HEAD at the remote did not point at any branch when --single-branch clone was made, no remote-tracking branch is created.
```

This is my first time working with ansible modules, so if I'm doing something wrong or there is a better way to accomplish this, please let me know. Thanks!

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -573,12 +573,26 @@ def submodule_update(git_path, module, dest, track_submodules):
         module.fail_json(msg="Failed to init/update submodules: %s" % out + err)
     return (rc, out, err)
 
+def set_remote_branch(git_path, module, dest, remote, version, depth):
+    cmd = "%s remote set-branches %s %s" % (git_path, remote, version)
+    (rc, out, err) = module.run_command(cmd, cwd=dest)
+    if rc != 0:
+        module.fail_json(msg="Failed to set remote branch: %s" % version)
+    cmd = "%s fetch --depth=%s %s %s" % (git_path, depth, remote, version)
+    (rc, out, err) = module.run_command(cmd, cwd=dest)
+    if rc != 0:
+        module.fail_json(msg="Failed to fetch branch from remote: %s" % version)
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks for your pull request. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

jbgo:
@gregdek @bcoca is there still interest in incorporating into Ansible? I haven't touched it since Ansible 2.0, but would be happy to test it/update it if there is still interest.

robinro:
@jbgo just to let you know: your change did not work on centos6 (set-branches is not available there, git version 1.7.1)
please try to add test cases whenever you have them to see this earlier

fix now in https://github.com/ansible/ansible-modules-core/pull/3492
test in https://github.com/ansible/ansible/pull/15490

git --depth is a quite unstable thing, please add more test cases if you have ideas for them

PR #941, opened 2015-03-13
Make state=running work for backwards compatibility with 1.8

abadger:
@smashwilson -- I noticed that we no longer have running and now have started (because of a comment from @bcoca in #502 ).  Is it a straightforward rename?  If so, I think having running still be valid and mapping it directly to state='started' is a good idea.  I don't think it needs to be documented as it's just for backwards compatibility, what do you think?

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1423,7 +1423,7 @@ def main():
             env             = dict(type='dict'),
             dns             = dict(),
             detach          = dict(default=True, type='bool'),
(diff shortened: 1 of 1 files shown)


smashwilson:
@abadger Ah! Yes, you're absolutely right. Mapping `running` to `started` should be sufficient, yes.

smashwilson:
Yeah, this looks good. :+1:

abadger:
Thanks for reviewing it!

PR #942, opened 2015-03-13
New files/chmod module

bpennypacker:
##### Issue Type:

New Module Pull Request
##### Ansible Version:

ansible 1.8 (devel 43eb821d3f) last updated 2015/01/07 20:16:03 (GMT +000)
  lib/ansible/modules/core: (detached HEAD db5668b84c) last updated 2015/01/07 20:16:14 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 110250d344) last updated 2015/01/07 20:16:20 (GMT +000)
##### Environment:

N/A
##### Summary:

This module wraps chmod/chown in a nice way.  Unlike calling chmod or chown directly via a shell, this module will accurately report if any file attributes were changed or not. It also returns a list of files that are modified so they can be addressed by other tasks if desired.  Some examples:

# Ensure user Tom has write permission to all the files in his home directory
- chmod: filespec=/home/tom/\*  mode=u+w

# Make all the directories under /usr/local/foo readable by the root user only
- chmod: path=/usr/local/foo mode=0700 owner=root recurse=true type=directory

# Make /usr/local/etc/_.sh executable
- chmod: filespec=/usr/local/etc/_.sh mode=0755

# Update all symlinks under /usr/foo that begin with the letter X
- chmod: path=/usr/foo filespec=X\* recurse=yes type=link mode=755
##### Steps To Reproduce:

N/A
##### Expected Results:

N/A
##### Actual Results:

N/A


tedder:
Isn't this (mostly) done by the `file` module?

bpennypacker:
The problem with the file module is that it doesn't give you very much granularity in what it can do. For example, suppose you want to recursively change the permissions on all the symlinks in a directory. There's no way to tell the file module to recurse through a directory but only make changes to symlinks found within those directories.

Or suppose you want to change the permissions of /usr/local/bin/*.sh to be executable. The file module doesn't accept wildcards in it's path, so you can't do this:

```
- file: path=/usr/local/bin/*.sh mode=0755
```

Right now to do this in ansible you would need to get a directory listing by doing something like spawning an `ls` command then using with_items to iterate over each file.

Or simply changing ownership of all files under a directory, similar to `chown -R joe /home/joe`, or finding all the files in a subdirectory that has the same extension and modifying them. The file module doesn't let you do anything like these. 

I suppose  these features could be added to the file module, but I feel it would just overload it and make it that much more confusing to people.

sivel:
I would say that "duplicating" functionality would be more confusing

bcoca:
The file module can be fed this data, either through a previous task or using a lookup plugin (if local files).  I would prefer to add more and better ways to feed the file module over duplicating it's functionality.

PR #945, opened 2015-03-14
get_url now downloads to dest dir instead of system default

### LINKED ISSUE #9512: get_url fails when you try to download large file to separate mount
I tried to download a 16gb db snapshot to a separate drive (/data), but it failed because the file was buffered on a different drive, I believe in /tmp.  This seems like a design flaw when curl has no problem (see below).

This fails:
get_url:
  url: "{{snapshot_url}}"
  dest: /data/latest-snapshot.tar.gz

But this works:
command: curl -o  /data/latest-snapshot.tar.gz {{snapshot_url}}

sysadmin75:
Fixes https://github.com/ansible/ansible/issues/9512


bcoca:
It would be nicer to make this an option instead of changing the existing behavior w/o giving the user any choice.

sysadmin75:
I would agree the user needs a choice only if this has the potential to break playbooks. However (unless I'm missing something), I don't think changing the temporary download directory will break anything because the end result is still the same. So is there really a need to give the user a choice? Maybe someone knows of a use case.

bcoca:
My only thought is cleanup, tmp dirs normally get cleaned up, so any
partial/failed downloads would eventually disappear, not so in most other
dirs.

## 

Brian Coca

sysadmin75:
Yes tmp dirs are normally cleaned but aren't they usually cleaned during boot time? The servers I handle are rarely rebooted. And being the "good" sysadmin that I am, if the get_url task fails, I am investigating why, which would result in deleting any file left over files from a partial download.

bcoca:
no tmpreaper running?

## 

Brian Coca

sysadmin75:
No, never heard of it. I just searched the centos repos for it and it isn't available.

gregdek:
I second @bcoca here. This should clearly be an option, with /tmp as the default.

gregdek:
@sysadmin75 A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

sysadmin75:
Ok, I'll work on this as an option. Hopefully I can have this in the next couple of days.

bcoca:
on RHEL/Centos you have tmpwatch by default, which does very similar thing, periodically cleans up tmp dirs

sysadmin75:
I am closing this PR because I have opened a new PR that provides this as an option.
https://github.com/ansible/ansible-modules-core/pull/1806

Issue #940, opened 2015-03-13
Ansible not pulling images from private Docker repository


abadger:
Report from @soupdiver Here: https://github.com/ansible/ansible/issues/10209

I've problems with Ansible automatically pulling docker images from my private repository.
I do login first via:

``` yaml
- name: login to registry
  shell: "docker login -e {{ registry_mail }} -p {{ registry_pass }} -u {{ registry_user }} {{ registry_host }}"
```

and then I tried the following instructions:

``` yaml
- name: start foo container
  docker:
    image: registry.foo:8888/app
    name: foo-app
    env: "DB_NAME=foo,APP_HOST=foo.foo,NODE_ENV=docker"
    ports: 127.0.0.1:1337:1337
    state: running

- name: start foo container
  docker:
    image: app
    registry: registry.foo:8888
    name: foo-app
    env: "DB_NAME=foo,APP_HOST=foo.foo,NODE_ENV=docker"
    ports: 127.0.0.1:1337:1337
    state: running

- name: start foo container
  docker:
    image: app
    registry: registry.foo:8888
    username: "{{ registry_user }}"
    password: "{{ registry_pass }}"
    name: foo-app
    env: "DB_NAME=foo,APP_HOST=foo.foo,NODE_ENV=docker"
    ports: 127.0.0.1:1337:1337
    state: running
```

everyone gave me an error: `msg: Docker API error: No such image`

Doing manually `docker run registry.foo:8888/app` did pull and start the image.

I had to manually insert

``` yaml
- name: pull foo container manually
  shell: docker pull registry.foo.com:8888/app
```

Any ideas?

abadger:
Are you able to try again with a checkout from the devel branch?  There's been some large code changes in the docker module in the past few weeks.  I don't think it fixes your issue but I do think the exact error message we get now will help to narrow down where this is going wrong.

abadger:
Mm... also, I see a report from someone else where they have a private registry working with a playbook like this:

```
  - name: main.yml | start our container
    docker:
      image: app
      name: my app
      registry: 'https://our.registry.com'
      username: "{{ docker_username }}"
      email: "{{ docker_email }}"
      password: "{{ docker_password }}"
      publish_all_ports: yes
      docker_api_version: "{{ client_api_version.stdout }}"
      state: running
```

perhaps the registry needs to have "https://" at the start?  Or perhaps the credentials for logging in need to be part of each task?

soupdiver:
[...]

PR #947, opened 2015-03-15
When no use_tls is not specified and no tls parameters are given we need to not attempt tls.

abadger:
From the error in #940 it looks like we need to set tls_config=None in more cases.  This is a fix for one common case in the tls features introduced with #926

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -589,12 +589,8 @@ class DockerManager(object):
         #   or tls_ca_cert (which requests verifying the server with
         #   a specific ca certificate)
         use_tls = module.params.get('use_tls')
-        if use_tls == 'no':
-            tls_config = None
-        else:
-            # If this stays True, we'll do encryption.  If something overrides
-            # it, then we'll have a TLSConfig obj instead
-            tls_config = True
+        tls_config = None
+        if use_tls != 'no':
             params = {}
 
             # Setup client auth


abadger:
I think this is a necessary fix so I'm going to merge this to devel and stable-1.9 now.  @smashwilson, @dguerri, @lorin if this causes problem with your usage of boot2docker (or any other tls stuff), please let me know ASAP.

dguerri:
Working as expected here with `use_tls=verify`

dguerri:
And this is what I am getting without environment and explicit settings about TLS verification:

Using tcp:// uri

```
TASK: [Start Auto-DNS container] ********************************************** 
failed: [localhost] => {"changed": false, "failed": true}
msg: ConnectionError(ProtocolError('Connection aborted.', BadStatusLine('\x15\x03\x01\x00\x02\x02\n',)),)
```

Using https:// uri

```
TASK: [Start Auto-DNS container] ********************************************** 
failed: [localhost] => {"changed": false, "failed": true}
msg: SSLError(SSLError(SSLError(1, '_ssl.c:507: error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version'),),)
```

abadger:
@dguerri Okay, I've now merged this fix to devel and stable-1.9 and updated the module repo pointers in the ansible/ansible repository.  Do you still get those Connection/SSL errors?

dguerri:
Sorry I expressed myself badly. Those errors were expected because I disabled tls and tried to connect to a tls-enabled docker!

abadger:
Ah.  Cool, so it all works :-)

PR #949, opened 2015-03-16
added "block_device" option to nova_compute module

marszall87:
Simple change that allows to boot an instance with attached block devices.

### DIFF
--- a/cloud/openstack/_nova_compute.py
+++ b/cloud/openstack/_nova_compute.py
@@ -175,6 +175,11 @@ options:
      required: false
      default: None
      version_added: "1.9"
+   block_device:
+     description:
+        - A list of block devices that should be attached
+     required: false
+     default: None
 requirements: ["novaclient"]
 '''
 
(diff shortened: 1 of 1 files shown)


sivel:
cc @emonty 

@marszall87 with the deprecation of the current modules, I don't believe new features or bugs fixes are being accepted.  See https://github.com/ansible/ansible-modules-extras/pull/92 for a list of the new modules.

marszall87:
Oh, I didn't know about that. However, I still don't see the possibility to create and attach a new volume with single operation in new modules and it is possible with OpenStack Python API. Probably should make pull request there.

cdmitri:
@marszall87, @sivel cc @emonty

I've had this pull request there for a while https://github.com/ansible/ansible-modules-core/pull/394 it does the block devices with the same syntax as ec2.

We've been using it in my team ever since. I'll take a look at the new modules.

emonty:
Mentioned this in #394 - but you're right, the new modules don't have attaching volumes as part of a single play in them... it's sounding like that's a thing that is strongly desired, so if you don't get to adding it, I'll add it to my list. FWIW - it sounds like a thing then that should be an option to create_server in shade - similar to auto_ip.

marszall87:
I guess that it is pretty common use case to boot image from external volume, so doing it in single step would be useful. It seems like OpenStack devs are going that way, hence the new "block_device_mapping_v2" option to Nova API that supports auto-created volumes.

emonty:
Added issue ansible/ansible-modules-extras#320 to track the need for this in the new modules.

PR #956, opened 2015-03-17
Added multiple allowed ranges.

rmarsagoogle:
Updated to support multiple protocols and port ranges in the allowed parameter.


erjohnso:
lgtm - thanks for contributing @rmarsagoogle

Issue #297, opened 2014-11-10
postgres_user module doesn't work with AWS RDS databases


dan-mcdonald:
##### Issue Type:

Bug report
##### Ansible Version:

ansible 1.7.0
##### Environment:

OSX Mavericks 10.9.4
##### Summary:

When running against an AWS RDS Postgresql instance, the postgres_user module can't set up new users.
##### Steps To Reproduce:

Run this task:

```
- name: Ensure user has access to the database
  postgresql_user: login_host={{ db_host }}
                   port={{ db_port }}
                   login_user={{ db_admin_user }}
                   login_password={{ db_admin_password }}
                   db={{ db_name }}
                   name={{ db_user }}
                   password={{ db_password }}
                   priv=ALL
                   state=present 
```
##### Expected Results:

```
ok: [...] => {"changed": true, "db": "database"}
```
##### Actual Results:

```
failed: [...] => {"failed": true, "parsed": false}
invalid output was: SUDO-SUCCESS-qwpjepgvenunnlewzwjddpnrbfjyptxo
Traceback (most recent call last):
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1407775799.05-32923950289495/postgresql_user", line 1869, in <module>
    main()
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1407775799.05-32923950289495/postgresql_user", line 497, in main
    changed = user_alter(cursor, module, user, password, role_attr_flags, encrypted, expires)
  File "/home/ubuntu/.ansible/tmp/ansible-tmp-1407775799.05-32923950289495/postgresql_user", line 201, in user_alter
    cursor.execute(select, {"user": user})
psycopg2.ProgrammingError: permission denied for relation pg_authid
```

Apparently the `pg_authid` relation [is not available in RDS](https://forums.aws.amazon.com/thread.jspa?threadID=151248).

Possible workaround: if access denied for `pg_authid`, then always set the password.

dan-mcdonald:
This is a copy of ansible/ansible#8547, a victim of the module re-organization.

smiller171:
It appears that the problem is not that a user cannot be created. It seems to work only if the user does not already exist.

mattwillsher:
I've see the opposite - it worked first run as long as I created the database first. Subsequent runs fail. 
My code block is 

``` yaml
- name: Create PostgreSQL database
  postgresql_db:
    login_host: "{{ crowd_db_server }}"
[...]

PR #957, opened 2015-03-17
postgres_user: skip password changes so pg_authid isn't needed

tedder:
Some places ([AWS RDS](https://forums.aws.amazon.com/thread.jspa?threadID=151248)) don't have, or don't allow, access to the `pg_authid` table. The only reason that is necessary is to check for a password change.

This flag is a workaround so passwords can only be set at creation time. It isn't as elegant as changing the password down the line, but it fixes the longstanding issue #297 that prevented this from being useful on AWS RDS.

### DIFF
--- a/database/postgresql/postgresql_user.py
+++ b/database/postgresql/postgresql_user.py
@@ -113,6 +113,12 @@ options:
     required: false
     default: null
     version_added: '1.4'
+  no_password_changes:
+    description:
+      - if C(yes), don't inspect database for password changes. Effective when C(pg_authid) is not accessible (such as AWS RDS). Otherwise, make password changes as necessary.
+    required: false
+    default: 'yes'
+    choices: [ "yes", "no" ]
 notes:
    - The default authentication assumes that you are either logging in as or
      sudo'ing to the postgres account on the host.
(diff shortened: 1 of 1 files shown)


mtotheikle:
+1

database/postgresql/postgresql_user.py
```diff
@@ -113,6 +113,12 @@
     required: false
     default: null
     version_added: '1.4'
+  no_password_changes:
+    description:
+      - if C(yes), don't inspect database for password changes. Effective when C(pg_authid) is not accessible (such as AWS RDS). Otherwise, make password changes as necessary.
+    required: false
+    default: 'yes'
```
bcoca: default is incorrectly stated as 'yes', code has 'no' which keeps backwards compatibility

database/postgresql/postgresql_user.py
```diff
@@ -485,6 +492,7 @@ def main():
         module.fail_json(msg="privileges require a database to be specified")
     privs = parse_privs(module.params["priv"], db)
     port = module.params["port"]
+    no_password_changes = module.params.get("no_password_changes", False)
```
bcoca: no need to get, False as setting default in #480 guarantees as value

database/postgresql/postgresql_user.py
```diff
@@ -113,6 +113,12 @@
     required: false
     default: null
     version_added: '1.4'
+  no_password_changes:
+    description:
```
bcoca, line 117: needs version_added: 2.0

tedder:
fixed those, thanks.

mtotheikle:
Thank you for the quick merge @bcoca!

PR #963, opened 2015-03-18
Allow to add build flags for gem installation

u2ix:
This allows to add gem module specific build flags. This will make it possible to install some gems without problems, if they need build flags to compile native code. 
#### A specific example

Nokogiri is a widely used xml library for ruby. It ships with a packaged libxml2, which is not compatible to some systems (for me it was on Fedora). To make the compilation work it needs the build flags `--use-system-libraries`
**Without this change it is not possible to install gems using nokogiri (or any other gem needing specific build flags)**

With this change:

```
gem: name=nokogiri state=latest user_install=no build_flags='--use-system-libraries'
```

and it works.

### DIFF
--- a/packaging/language/gem.py
+++ b/packaging/language/gem.py
@@ -73,6 +73,11 @@ options:
     required: false
     default: "no"
     version_added: "1.6"
+  build_flags:
+    description:
+      - Allow adding build flags for gem compilation
+    required: false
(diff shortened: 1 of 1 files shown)


swozey:
This is definitely important for IOS/Xcode build farms. Thanks u2ix.

PR #968, opened 2015-03-19
Default limit changed

0dataloss:
If SATA is the defaut volume_type the default size can be reduced from 100 GB to 75 GB


sivel:
Due to the majority of pyrax versions having a hardcoded value of 100 as the minimum size, I'm going to keep the current default to prevent issues.

0dataloss:
This has been a problem while I was updating the code to understand the new limits, I have created a pull request for the pyrax project as well: https://github.com/rackspace/pyrax/pull/550

sivel:
You may also be interested in https://github.com/ansible/ansible-modules-core/pull/797 where I already removed the limits from this module.

PR #972, opened 2015-03-19
Add missing import of apt.debfile

caphrim007:
In cases when the python-apt package is not installed, ansible will
attempt to install it. After this attempt, it tries to import the
needed apt modules, but forgets to import the apt.debfile module.

The result is that playbooks that use the deb argument on a machine
that does not initially have the python-apt package available will
fail with the following error

AttributeError: 'module' object has no attribute 'debfile'

This patch adds the appropriate import to the apt module to ensure
that necessary libraries are available in cases when the deb argument
is being used on a system that does not initially have the python-apt
package installed


abadger:
# Merged

Hi!

issue is resolved.

Thank you!

abadger:
Thanks much!

PR #978, opened 2015-03-22
[docker] pulling missing docker image before doing anything

bobrik:
This actually fixes #670.

Fixed scenarios:
- First
  1. Run container from 10gb image
  2. Run newer version that requires download of 10 more gb
  3. Boom, running container is nuked and nothing works before newer image is downloaded
- Second
  1. Run some super-duper important container
  2. Change image to something that does not exist, deploy
  3. Boom, running container is nuked and error is presented to a user

In both scenarios error would happen before nuking current container.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1460,10 +1460,14 @@ def main():
         if count > 1 and name:
             module.fail_json(msg="Count and name must not be used together")
 
-        # Explicitly pull new container images, if requested.
-        # Do this before noticing running and deployed containers so that the image names will differ
-        # if a newer image has been pulled.
-        if pull == "always":
+        # Explicitly pull new container images, if requested. Do this before
+        # noticing running and deployed containers so that the image names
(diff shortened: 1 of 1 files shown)


PikachuEXE:
What if the state is `absent`?

bobrik:
This only applies to `reloaded` state, PR is updated.

PikachuEXE:
Should `manager.get_inspect_image()` be moved to last?
Not sure if `manager.get_inspect_image()` is more expensive than checking state

bobrik:
Yep, updated again.

gregdek:
@cove @joshuaconner @softzilla @smashwilson please review; if any +1 (and no -1) we will promote to P1_reviewed for faster inclusion.

gregdek:
@cove @joshuaconner @softzilla @smashwilson any opinions on this Docker PR?

joshuaconner:
+1 lgtm!

(also sorry for not having more time to review these recently - bad timing with work)

Issue #980, opened 2015-03-22
vsphere_guest.py does not change vm disk


tapanswain-zz:
##### ISSUE TYPE

Bug Report
##### COMPONENT NAME

vsphere_guest module
##### ANSIBLE VERSION

N/A
##### SUMMARY

i am a perl developer and use vmware module for our day to day activity, 
When i am going through  vsphere_guest.py module, i could not get the logic  for
 def reconfigure_vm(vsphere_client, vm, module, esxi, resource_pool, cluster_name, guest, vm_extra_config, vm_hardware, vm_disk, vm_nic, state, force):

if vm_hardware['memory_mb']:  -- it works  
 if vm_hardware['num_cpus']:  -- it works  

but where you are  call for disk edit , i could see vm_disk parameter  in the function.  There is no call for this.  

Logic for add disk to vm 

1) get the find scsi controlle name  
2) get length  of the array 
3) then call  add_virtualdisk function with  backingtype,diskMode,fileName,controllerKey,unitNumber,size   parameter  

 please find the link for my code, 
https://github.com/tapanswain/vmware-sdk/blob/master/VMUtil.pm
https://github.com/tapanswain/vmware-sdk/blob/master/vdiskcreate-updated.pl 
 please reach me on my email id if you need more info.

aioue:
Proposed documentation fix until full reconfigure functionality is added to the module.
https://github.com/ansible/ansible-modules-core/pull/1403

sysreq0:
Hi,

I've created a pull request to add disk resizing to the 'reconfigured' state:

https://github.com/ansible/ansible-modules-core/pull/2728

ansibot:
@tapanswain we are unable to validate that "

##### " is an existing ansible module.

We are expecting a component name that matches something that can be found in:
- http://docs.ansible.com/ansible/list_of_all_modules.html

Please revise the component name in the description.

[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@chrrrles, @rhoop, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

jesusch:
##### COMPONENT NAME

vsphere_guest

I guess that's what the bot expects in the description?

ansibot:
[...]

PR #982, opened 2015-03-23
add vm disk editing

tapanswain-zz:
need to add new state for adding disk to exiting  vm .


bcoca:
what is this?

tapanswain-zz:
please  find the issue here. 

https://github.com/ansible/ansible-modules-core/issues/980

bcoca:
it seems you got a bunch of unrelated commits in the request, clean up and resubmit

abadger:
@tapanswain Perhaps you opened this pull request for the wrong branch?  It appears to be asking to merge our stable-1.9 branch into our devel branch (which we wouldn't do... we cherry-pick commits in devel into 1.9 not the other way around).  I looked at the list of commits and didn't see any from you which is what leads me to believe you submitted the PR for the wrong branch.

tapanswain-zz:
Hi  All,

I am from perl background and use perl sdk for automating the things.

I am recently started  using ansible , while going through  vsphere_guest.py . I could not get the logic how to add disk to exciting vm.

def reconfigure_vm(vsphere_client, vm, module, esxi, resource_pool, cluster_name, guest, vm_extra_config, vm_hardware, vm_disk, vm_nic, state, force):

if vm_hardware['memory_mb']: -- it works

if vm_hardware['num_cpus']: -- it works

but where you are call for disk edit , i could see vm_disk parameter in the function. There is no call for this.

please find the link for  perl code,  Need  your help here  to make this working in  vsphere_guest module.
https://github.com/tapanswain/vmware-sdk/blob/master/VMUtil.pm
https://github.com/tapanswain/vmware-sdk/blob/master/vdiskcreate-updated.pl https://github.com/tapanswain/vmware-sdk/blob/master/vdiskcreate-updated.pl

Tapan

bcoca:
so I'm closing this request as it seems to be a bad branch, @tapanswain feel free to reopen with a branch with your actual changes.

aioue:
As of 1.9.1 nothing apart from CPU and RAM can be changed when reconfiguring or launching from a template. This would help bring some of that functionality into Ansible.

PR #990, opened 2015-03-24
Added os_*_manage modules

retr0h:
These modules are used by the operator of OpenStack when deploying a cloud.  We would like to incorporate these into core, rather than fragmenting the community with various repos performing similar functionality.

Moving functionality from openstack-ansible/openstack-ansible-modules into core.

### DIFF
--- a/cloud/openstack/cinder_manage
+++ b/cloud/openstack/cinder_manage
@@ -1,0 +1,105 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+DOCUMENTATION = '''
+---
+module: cinder_manage
+short_description: Initialize OpenStack Block Storage (cinder) database
+description: Create the tables for the database backend used by cinder
+options:
+  action:
+    description:
+      - action to perform. Currently only dbysnc is supported
+    required: true
+  conf:
+    description:
+      - path to cinder config file.
+    required: false
+    default: /etc/cinder/cinder.conf
+requirements: [ cinder ]
+author: Lorin Hochstein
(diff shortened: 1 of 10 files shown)


emonty:
@retr0h could you rename this prefixing with os_ to match the agreed naming scheme.

Also - I know this is a direct port in from os-ansible-modules (thanks for starting on that!) Do you envision this module growing the need to talk to the heat API for admin functions? Or is this purely about tasks needed to be performed on the host like db-sync?

retr0h:
@emonty prefix the filename to `os_` or the module's usage?

I would think this module will replicate the functionality of `heat-manage`.

abadger:
@retr0h Hey, looks like @emonty had some questions about this set of modules.  Did the conversation go to a mailing list and then get resolved?

omgjlk:
These all need a license statement in the files.

omgjlk:
Instead of importing and running subprocess in each module, they should take advantage of the `run_command` function in the basic module_utils code within Ansible. https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1462

omgjlk:
When bringing these in, the docs should be updated to get rid of the action: syntax, and instead just use:

```
- name: name of task
  module_name:
    args_go_here: value
    other_args: other_value
```

gregdek:
@retr0h Please make the suggested revisions suggested by @j2sol. When you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

PR #993, opened 2015-03-24
[docker] Compare container's configured image instead of running image

### LINKED ISSUE #989: [Docker] Old containers aren't cleaned up when using count
**Issue Type:**

Bug Report 

**Ansible Version:**

ansible 1.9 (devel 85cfe1bd52)

**Environment:**

Ubuntu 14.04

**Summary:**

When using `count`, old containers are not removed

**Steps To Reproduce:**

Create containers with `count: 2`, `pull: always`, and `state: reloaded`. Run task to pull an updated image.

```
- name: run node client container
  docker:
    image: "company/tv-client:dev-latest"
    env: "SERVICE_NAME=client,SERVICE_TAGS=dev"
    ports: 80
    restart_policy: always
    pull: always
    state: reloaded
    count: 2
```

**Expected Results:**

The 2 original containers are gone and 2 new ones with the latest image are running in their place

**Actual Results:**

The 2 original containers are still running alongside the new containers, but without image tags

```
CONTAINER ID        IMAGE                                  COMMAND                CREATED             STATUS              PORTS                            NAMES
eda5d790bbc1        company/tv-client:dev-latest           "nginx -g 'daemon of   5 seconds ago       Up 5 seconds        443/tcp, 0.0.0.0:49204->80/tcp   serene_blackwell
fce2c32876eb        company/tv-client:dev-latest           "nginx -g 'daemon of   6 seconds ago       Up 5 seconds        443/tcp, 0.0.0.0:49203->80/tcp   agitated_lumiere
05ef9af927ee        147b87b45e35                           "nginx -g 'daemon of   20 minutes ago      Up 20 minutes       443/tcp, 0.0.0.0:49202->80/tcp   clever_pike
0bdeb3ef8f37        147b87b45e35                           "nginx -g 'daemon of   20 minutes ago      Up 20 minutes       443/tcp, 0.0.0.0:49201->80/tcp   lonely_nobel
```

smashwilson:
The "Image" field returned by `client.containers(all=True)` is the same one that's used for the output of `docker ps`. It's chosen from the list of image and tag names currently known by Docker, which means that if the image and tag have been changed by a pull since the container was launched, it will be a SHA rather than a real name, and `get_deployed_containers()` will no longer recognize it as a matching container.

This performs a full inspect on each container and compares against a normalized version of "Config/Image" instead, which is the image name that was used when the container was launched. It _is_ somewhat more expensive to do this, because we need to perform an inspect call to the Docker API for each container rather than just the matching ones, but it's necessary for image matching to work properly.

Fixes #989.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -420,6 +420,13 @@ def get_split_image_tag(image):
 
     return resource, tag
 
+def normalize_image(image):
+    """
+    Normalize a Docker image name to include the implied :latest tag.
+    """
+
+    return ":".join(get_split_image_tag(image))
+
 
 def is_running(container):
     '''Return True if an inspected container is in a state we consider "running."'''
@@ -1109,11 +1116,13 @@ class DockerManager(object):
         if inspected:
             repo_tags = self.get_image_repo_tags()
         else:
-            image, tag = get_split_image_tag(self.module.params.get('image'))
-            repo_tags = [':'.join([image, tag])]
+            repo_tags = [normalize_image(self.module.params.get('image'))]
 
         for i in self.client.containers(all=True):
-            running_image = i['Image']
+            details = self.client.inspect_container(i['Id'])
+            details = _docker_id_quirk(details)
+
+            running_image = normalize_image(details['Config']['Image'])
             running_command = i['Command'].strip()
 
             if name:
@@ -1128,9 +1137,6 @@ class DockerManager(object):
                 matches = image_matches and command_matches
 
             if matches:
-                details = self.client.inspect_container(i['Id'])
-                details = _docker_id_quirk(details)
-
                 deployed.append(details)
 
         return deployed
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1119,15 +1119,17 @@ class DockerManager(object):
             repo_tags = [normalize_image(self.module.params.get('image'))]
 
         for i in self.client.containers(all=True):
-            details = self.client.inspect_container(i['Id'])
-            details = _docker_id_quirk(details)
-
-            running_image = normalize_image(details['Config']['Image'])
-            running_command = i['Command'].strip()
+            details = None
 
             if name:
                 matches = name in i.get('Names', [])
             else:
+                details = self.client.inspect_container(i['Id'])
+                details = _docker_id_quirk(details)
+
+                running_image = normalize_image(details['Config']['Image'])
+                running_command = i['Command'].strip()
+
                 image_matches = running_image in repo_tags
 
                 # if a container has an entrypoint, `command` will actually equal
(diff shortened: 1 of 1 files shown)


smashwilson:
I tweaked things a little to save a few unnecessary inspect calls. It's a little uglier this way, but it'll save some work when you're matching containers by name.

MrMMorris:
@smashwilson looks good! thanks so much

And just to expand my testing, I tried increasing and decreasing the `count` which also worked as expected :+1:

abadger:
Merged to devel.  It won't be in the 1.9.0 release but I think it's a candidate for any eventual 1.9.1 release.

# Merged

Hi!

issue is resolved.

Thank you!

smashwilson:
@abadger :bow: Thanks!

1.9.1 works for me - I kind of want to revisit the `Command` comparison, too, so that will give me a little time to get to that, too.

MrMMorris:
@smashwilson not sure if this is related, but I am running into a similar problem as my original issue (https://github.com/ansible/ansible-modules-core/issues/989), but this time, it's not stopped containers that are leftover, but images:

before deploy:

```
REPOSITORY                         TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
company/tv-client                  dev-latest          cd58f33a781d        15 minutes ago      122 MB
```

after:

```
REPOSITORY                         TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
company/tv-client                  dev-latest          b22b44c135d5        9 minutes ago       122 MB
<none>                             <none>              cd58f33a781d        27 minutes ago      122 MB
```

```
ansible 1.9 (devel 1ae5b9292c) last updated 2015/03/30 22:53:15 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 5d776936cc) last updated 2015/03/30 22:53:17 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 400166a655) last updated 2015/03/30 22:53:18 (GMT +000)
  v2/ansible/modules/core: (detached HEAD 34784b7a61) last updated 2015/03/30 22:53:19 (GMT +000)
  v2/ansible/modules/extras: (detached HEAD 400166a655) last updated 2015/03/30 22:53:20 (GMT +000)
```

Also, a  related question: When old containers are removed after a deploy, are the volumes removed too? equivalent to `docker rm -v blah`?

smashwilson:
@MrMMorris: Ah - the module doesn't actually touch images at all, yet, beyond pulling at the appropriate times. It wasn't clear to me when it would actually be safe to do cleanup, or if the `docker_image` module (or whatever replaces it) would be a better fit, or what.

MrMMorris:
I see. I think it makes sense to remove dangling images after a deploy (`docker rmi `docker images -q -f dangling=true``). Unless someone has a use case for keeping them around?

Not really a pressing issue, just noticed orphaned images piling up.

And what about my volumes question? (I edited my comment as you were commenting :p)

abadger:
Just updating to note that it looks like this got left out of 1.9.1.  We had some miscommunication internally about our stable branch workflow and that lead to failing to push this to stable-1.9 once 1.9.0 was released.  I've pushed it to the stable-1.9 branch now to make sure that it will go into 1.9.2 if we end up creating that. https://github.com/ansible/ansible-modules-core/commits/stable-1.9  (May 4th, 2015)

PR #995, opened 2015-03-25
Add dhcp_options to VPC creation in cloud/amazon

ghost:
Add ability to create/associate dhcp_options

### DIFF
--- a/cloud/amazon/ec2_vpc.py
+++ b/cloud/amazon/ec2_vpc.py
@@ -76,6 +76,11 @@ options:
     required: false
     default: null
     aliases: []
+  dhcp_options:
+    description:
+     - 'A dictionary of dhcp_options for the vpc - either id (to associate existing dhcp_options) or dns_servers list (to create a new set of dhcp_options and associate with the vpc)'
+     required: false
+     version_added: "1.9"
   wait:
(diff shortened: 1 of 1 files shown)


bennojoy:
@michal-jora  could you please add version_added to the new paramter in the doc string.

mpisanko:
@bennojoy done

bennojoy:
@michal-jora  can you please make version_added: 2.0 as 1.9 is already released

bennojoy:
@michal-jora  seems like you are deleting the dhcpoptions after the vpc is deleted , shouldn't this be done before deleting the vpc ? vpc_conn.delete_vpc(vpc.id)

bcoca:
cc @defionscode , does it make sense to keep adding to this module since there is a push to deprecate it in favor of several smaller more specialized modules?

juliedavila:
@bcoca good call, I think only bug fixes should be added to this now, feature type PRs should go to the new ones. @michal-jora this would be the `ec2_vpc_net` module which is now in devel for core. I believe DHCP options is already supported there.

gregdek:
So should we close this one then?

juliedavila:
I would recommend a close

PR #998, opened 2015-03-25
DigitalOcean API version 2 with user_data option

linefeedse:
I needed user_data for a coreos playbook, so I made the minimal change possible to be able to run V2 of the API. When using V2, dopy v >= 0.3.2 supports the user data option.


telemakhos:
When do you guys plan to merge this?

skottler:
Thanks, looks great!

PR #1001, opened 2015-03-26
Validate variable, return only the found variable value instead of tuple

ospii:
Docs imply the mysql_variables is used to operate a single variable therefore
- fail before making any db connections if variable is not set
- validate chars for mysql variable name with re.match(^[a-z0-9_]+)
- use "SHOW VARIABLE WHERE Variable_name" instead of LIKE search
- getvariable() returns only the value or None if variable is not found
- the module returns only the found variable value instead of tuple for easier operation eg. as registere variable in tasks

### DIFF
--- a/database/mysql/mysql_variables.py
+++ b/database/mysql/mysql_variables.py
@@ -67,6 +67,7 @@ EXAMPLES = '''
 import ConfigParser
 import os
 import warnings
+from re import match
 
 try:
     import MySQLdb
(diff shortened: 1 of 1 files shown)


gregdek:
@banyek can you review? If you can +1 we will promote to P1_reviewed for faster inclusion.

gregdek:
@banyek a friendly ping -- got time to review this PR?

banyek:
LGTM +1

PR #1010, opened 2015-03-26
gc_storage allow setting headers

richtier:
It would be super useful to allow specifying "headers" when uploading a file to google cloud storage.

For example, when uploading a gzipped file to gc_storage I would like to set "Content-Encoding" to "gzip", or specify "Cache-Control".

my task has the following:

``` yml
- name: uploading built files to cloud
  gc_storage: >
    bucket=my-awesome-bucket
    src='{{ item }}'
    object='{{ item | basename }}'
    mode=put
    permission=public-read
    headers='{"Content-Encoding": "gzip", "Cache-Control": "max-age=31536000"}'
  with_fileglob:
    - '{{ repository_path }}/ask-peers/build/production/*'
```

and this is the outcome:

image

### DIFF
--- a/cloud/google/gc_storage.py
+++ b/cloud/google/gc_storage.py
@@ -57,6 +57,11 @@ options:
       - This option let's the user set the canned permissions on the object/bucket that are created. The permissions that can be set are 'private', 'public-read', 'authenticated-read'.
     required: false
     default: private 
+  headers:
+    description:
+      - Headers to attach to object.
+    required: false
+    default: {}
   expiration:
     description:
       - Time limit (in seconds) for the URL generated and returned by GCA when performing a mode=put or mode=get_url operation. This url is only avaialbe when public-read is the acl for the object.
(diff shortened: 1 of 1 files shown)


bcoca:
cc @erjohnso

cloud/google/gc_storage.py
```diff
@@ -57,6 +57,11 @@
       - This option let's the user set the canned permissions on the object/bucket that are created. The permissions that can be set are 'private', 'public-read', 'authenticated-read'.
     required: false
     default: private 
+  headers:
+    description:
```
erjohnso, line 62: Could you add a `version_added: x.y.z` here?
@bcoca - what version should @richtier use - 1.9.1?

bcoca, line 62: 2.0 at this point, minor versions only add bugfixes, not new features

richtier:
sure, I will do that

cloud/google/gc_storage.py
richtier, line 62: done :)

erjohnso:
Great, thanks @richtier!

PR #1012, opened 2015-03-27
[yum] Set disabled repos before enabled repos

DavidWittman:
The ordering of disabling/enabling yum repositories matters, and the yum module was mixing and matching the order. Specifically, when yum-utils isn't installed, the codepath which uses the yum python module was incorrectly ordering enabling and disabling.

The preferred order is to disable repositories and then enable them to prevent clobbering. This was previously discussed in ansible/ansible#5255 and incompletely addressed in ansible/ansible@0cca4a3.

### DIFF
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -187,10 +187,10 @@ def is_installed(module, repoq, pkgspec, conf_file, qf=def_qf, en_repos=[], dis_
         pkgs = []
         try:
             my = yum_base(conf_file)
-            for rid in en_repos:
-                my.repos.enableRepo(rid)
             for rid in dis_repos:
                 my.repos.disableRepo(rid)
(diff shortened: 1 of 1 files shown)


mchlumsky:
I reviewed and tested the change. It works.

+1

franmrl:
Isn't this breaking the (very uncommon) usage of disablerepo=myrepo enablerepo=\*  ? Shouldn't we just keep the same order as specified in the yaml, or are the yaml dicts unordered and we can't do it like that?

DavidWittman:
@franmrl It is, but as you indicated yourself, that behavior is not quite as common as `--disablerepo=* --enablerepo=foo`. I believe the YAML dicts are unordered, but feel free to try to tackle this in a PR if you have a need for both.

devops-senpai:
👍

PR #1014, opened 2015-03-27
[Docker] Add signal parameter when killing container

nemunaire:
A signal parameter can be given to the kill function ([already present in the 0.3.0](https://github.com/docker/docker-py/blob/0.3.0/docker/client.py#L538)). This pull-request add an optional `signal` parameter in order to define the signal sent to the container, instead of the default KILL.

With a already running container `debian`, you can send a SIGTERM with the following playbook:

``` yaml
- docker:
    image=debian
    state=killed
    signal=TERM
```

The `signal` parameter can't be a choice because signal differs from an OS to another. If an invalid signal is given, the following error is displayed when the playbook is run:

```
msg: Docker API Error: Invalid signal: TOTO
```

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -176,6 +176,12 @@ options:
     description:
       - Enable detached mode to leave the container running in background.
     default: true
+  signal:
+    description:
+      - With the state "killed", you can alter the signal sent to the
+        container.
+    required: false
+    default: KILL
   state:
     description:
       - Assert the container's desired state. "present" only asserts that the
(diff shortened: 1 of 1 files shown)


gregdek:
@cove @joshuaconner @softzilla @smashwilson please review; if any of you +1 (with no -1) we will promote to P1_reviewed for faster inclusion.

gregdek:
@cove @joshuaconner @softzilla @smashwilson ping -- do any of you have time to take a look at this PR?

joshuaconner:
Sorry to be slow on this! Does the documentation for `signal` need a `version_added` attribute? Otherwise LGTM.

bcoca:
yes, but i'll add it in subsequent commit

PR #1015, opened 2015-03-27
Make ec2_eip idempotent

jml:
This is exactly https://github.com/ansible/ansible-modules-core/pull/300/ but with the `allocate` state renamed to `allocated` and the `allocate_eip` function extracted from `main` into a top-level function.

### DIFF
--- a/cloud/amazon/ec2_eip.py
+++ b/cloud/amazon/ec2_eip.py
@@ -74,6 +74,9 @@ EXAMPLES = '''
 - name: output the IP
   debug: msg="Allocated IP is {{ eip.public_ip }}"
 
+- name: another way of allocating an elastic IP without associating it to anything
+  ec2_eip: state='present'
+
 - name: provision new instances with ec2
   ec2: keypair=mykey instance_type=c1.medium image=ami-40603AD1 wait=yes group=webserver count=3
   register: ec2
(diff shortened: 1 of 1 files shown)


juliedavila:
@jml this looks good (havent tested yet) but I'd like to avoid the merge commit 9c275fcbbaaf84a0fd05f3fbd19625c0cf92cbe7. Could you fix it?

jml:
@defionscode Thanks. I've got rid of the merge commit. Let me know if it needs further work.

jml:
@defionscode friendly ping

jml:
Is this OK to be merged, or would you like me to do some more work?

bcoca:
So since instance ID is not required, I would prefer to keep this as 'present' and just allocate it to the account when no instance id is supplied. This would not require a 3rd state

jml:
Done. Please take a look.

jml:
Is this good to merge?

bcoca:
the commit history still looks messy, can you rebase this to one commit?

jml:
Done. Thanks.

sruon:
This was apparently reverted in commit c36e8947bcfae48d07347d3fd32f1aa345958bb7

I will open a new separate issue.

PR #1017, opened 2015-03-27
Supports creation of vpc instances with boto profile

tepwil:
The current ec2 module has a bug whereby it doesn't support creating ec2 instances in a VPC when using a boto profile. If you try to create an instance in a VPC using a boto profile, the default credentials end up being used.  In my setup this results in an 'InvalidSubnetID.NotFound' error message being returned.  

This pull request resolves the bug by adding support for boto profiles.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -1205,14 +1205,13 @@ def main():
 
     ec2 = ec2_connect(module)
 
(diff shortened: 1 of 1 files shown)


bennojoy:
@tepwil i think this already available upstream, can you please try with the latest devel.

tepwil:
Yes, I confirm that the bug has been resolved upstream in the latest devel.  Therefore, I'm closing this PR since it's no longer needed.  Thanks for resolving this.

PR #1030, opened 2015-03-30
ec2_asg: Streamlined rolling udpate algorithm. 

jsmartin:
This is a re-working of the rolling update algorithm.  It reduces the number of scaling actions.  Also adds a ton of logging, which can be turned on by uncomment line 199.

Prior to this if you had a 3 node ASG, and wanted to replace each instance, with a batch size of 1,  it would try to maintain a minimum size of 4, even after the number of new instances were equal to the original desired capacity.  Now if the new number of new instances are equal to the desired capacity, it will terminate any remaining old instances.

In the case of a 3 node ASG with a batch size of 3:

desired capacity, min size, max size are bumped up by 3
We wait until the instances are ready/healthy
We start the termination loop
We notice that we don't need any more instances since we already have 3 new
We terminate the rest of the instances
We set the desired capacity, min size, max size back to their original size.

This also covers the functionality/bugfix in PR #589

### DIFF
--- a/cloud/amazon/ec2_asg.py
+++ b/cloud/amazon/ec2_asg.py
@@ -191,9 +191,13 @@ to "replace_instances":
 
 import sys
 import time
+import logging as log
 
 from ansible.module_utils.basic import *
 from ansible.module_utils.ec2 import *
+log.getLogger('boto').setLevel(log.CRITICAL)
+#log.basicConfig(filename='/tmp/ansible_ec2_asg.log',level=log.DEBUG, format='%(asctime)s: %(message)s', datefmt='%m/%d/%Y %I:%M:%S %p')
+
 
 try:
     import boto.ec2.autoscale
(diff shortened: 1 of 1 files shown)


parzival418:
Just finished testing this with a new rolling deploy process I am working on and it works quite smoothly.  I was having some issues with the provided ASG module, and this helped me to get it all rolling out the way I wanted.

jsmartin:
Just a note, will need to de-register instances from ELB (if they belong) before terminating the instance in the ASG in order for connection draining to work:

https://aws.amazon.com/blogs/aws/elb-connection-draining-remove-instances-from-service-with-care/
https://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/TerminologyandKeyConcepts.html#conn-drain

jsmartin:
Now supports deregistering instance from ELB to ensure connection draining is honored as per above comment.  Rebased.

haad:
+1

PR #239, opened 2014-10-30
New parameter supported in cloudformation: template_format.

JoseLuisMD:
Added support to a new parameter for cloudformation module: template_format. 

This parameter allows the user to choose the format of the template specified file: 'json' or 'yaml'. The parameter is not required, and it has 'json' as default value. 

If the 'yaml' option is chosen, the module loads the yaml file and converts it to .json format.

PR #1034, opened 2015-03-30
add cloudformation_assemble, yml->json.

tedder:
This is complementary to the cloudformation module itself, it allows a tree of YML files to be kept, then does a last-minute transformation to JSON. I've attempted to document it, and I wrote about it here too:
- http://awsadvent.tumblr.com/post/106058327586/aws-advent-2014-repeatable-infrastructure-with
- https://github.com/tedder/aws-advent-2014-yml-cloudformation/blob/master/article.md (alt link)

I can't figure out how to generate/test the documentation. `make docsite` doesn't work anymore.

Paging @defionscode as requested on #239.


cloud/amazon/cloudformation_assemble.py
```diff
@@ -0,0 +1,149 @@
...
+    if f:
+        return yaml.load(f) or dict()
+    else:
+        raise Exception("Could not find file: " + f)
+
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            directory=dict(default="cfbuild"),
+            destination=dict(default="/tmp/cfbuild.json"),
```
juliedavila, line 100: I don't agree with having a default for this parameter or directory. Also, may need to be specified as required?

juliedavila:
I'm trying to think of the best way to deal with the generated json file. Placing it in /tmp as you have there might be reasonable. My initial thoughts (just brainstorming):
1. Have you tested this by placing the fragments in a $role/files folder?
2. There doesn't seem to be support for idempotency. Looks like it would report changed every time, though you don't have changed defined at all so it may just return OK every time.
3. There should be exception handling to account for potentially poorly-formatted YAML/JSON. Also, exceptions should be captured and outputted into a `module.fail_json(msg='Exception Message')` type of syntax.

tedder:
1. I hadn't ever tried that. It doesn't work. Can you point me at code that does the right thing? path_dwim seems to expect the path stuff to already be worked out; I'm a little unsure what to do. I'm on IRC if that's easier.
2. I'll have it return changed=true for now.
3. Yep. Old module that I'm cleaning up, so it wasn't very "modern" in that way. Fixed.
4. I should have really put this in ansible-modules-extras.

tedder:
I'm going to maintain this outside of the Ansible castle.

PR #1045, opened 2015-03-31
Add OpenStack Keypair module

emonty:
Also deprecate old nova_keypair module.

### DIFF
--- a/cloud/openstack/os_keypair.py
+++ b/cloud/openstack/os_keypair.py
@@ -84,17 +84,15 @@ def main():
         mutually_exclusive=[['public_key', 'public_key_file']])
     module = AnsibleModule(argument_spec, **module_kwargs)
 
-    if module.params['public_key_file']:
-        public_key = open(module.params['public_key_file']).read()
-    else:
-        public_key = module.params['public_key']
-
-    if not HAS_SHADE:
-        module.fail_json(msg='shade is required for this module')
-
     state = module.params['state']
     name = module.params['name']
     public_key = module.params['public_key']
+
+    if module.params['public_key_file']:
+        public_key = open(module.params['public_key_file']).read()
+
+    if not HAS_SHADE:
+        module.fail_json(msg='shade is required for this module')
 
     try:
         cloud = shade.openstack_cloud(**module.params)
--- a/cloud/openstack/os_keypair.py
+++ b/cloud/openstack/os_keypair.py
@@ -33,6 +33,10 @@ extends_documentation_fragment: openstack
 version_added: "2.0"
 description:
    - Add or Remove key pair from OpenStack
+notes:
+   - The module returns a dictionary describing the keypair, with
+     keys including: id, name, public_key. A private_key entry may
+     also be included if a keypair was generated for you.
 options:
    name:
      description:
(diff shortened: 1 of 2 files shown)


bcoca:
requires rebase

emonty:
@gregdek @j2sol this one is ready to go - we cut the shade release this morning which supports it.

Shrews:
:thumbsup:

cloud/openstack/os_keypair.py
```diff
@@ -0,0 +1,146 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+
+try:
+    import shade
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
```
omgjlk, line 25: Since the version of shade is specifically required for this functionality, should there be an attempt to use said functionality and exit nicely if that version isn't installed?

cloud/openstack/os_keypair.py
```diff
@@ -0,0 +1,146 @@
...
+      state: present
+      name: ansible_key
+      public_key_file: /home/me/.ssh/id_rsa.pub
+
+# Creates a new key pair and the private key returned after the run.
+- os_keypair:
+      cloud: rax-dfw
+      state: present
+      name: ansible_key
+'''
```
omgjlk, line 75: Can you document the return data?

omgjlk:
Just a couple minor comments, not blockers. Would be fine to merge as is.

cloud/openstack/os_keypair.py
Shrews, line 25: I think that's an excellent idea, and we have considered this briefly among ourselves in the past. I think we need to put some more thought into this as a more broad solution for all of the modules, not just this one. So don't expect it added to this module with this change, but yes, we will add something to all of them to address this.

cloud/openstack/os_keypair.py
Shrews, line 75: Added as a note in the docstring.

omgjlk, line 75: That helps, but I meant in the actual doc structure: https://github.com/ansible/ansible-modules-core/blob/devel/files/stat.py#L72

Shrews, line 75: Done! Thanks for the documentation structure tip.

omgjlk:
Looks all good to me, please to be merging.

gregdek:
@bcoca @abadger @jimi-c @bennojoy can we get a merge?

abadger:
Since I'm sure you'll be adding more modules :-)  Some things to note (I'll fix these for this module):
- when a library requires a python version later than python 2.4 we like to add it to requirements so that we know that it's okay for the module to use python features added after python-2.4.  (with statement, catching exceptions with as, etc) .  We sometimes have to add these to
- Don't have to add requirements to documentation if they're in the common openstack doc fragment.  (In this case, shade and python >= 2.7 is there.)
- Eventually we might figure out a good way to run unittests with modules.  We should put the call to main() in a conditional so it doesn't run on import like this:

``` python
if __name__ == '__main__':
    main()
```

PR #1046, opened 2015-03-31
Add OpenStack Network module

emonty:
Also, deprecate the old quantum network module

### DIFF
--- a/cloud/openstack/_quantum_network.py
+++ b/cloud/openstack/_quantum_network.py
@@ -30,6 +30,7 @@ DOCUMENTATION = '''
 ---
 module: quantum_network
 version_added: "1.4"
+deprecated: Deprecated in 2.0. Use os_network instead
 short_description: Creates/Removes networks from OpenStack
 description:
    - Add or Remove network from OpenStack.
--- a/cloud/openstack/os_network.py
+++ b/cloud/openstack/os_network.py
@@ -1,0 +1,107 @@
+#!/usr/bin/python
+
+# Copyright (c) 2014 Hewlett-Packard Development Company, L.P.
+# Copyright (c) 2013, Benno Joy <benno@ansible.com>
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


cloud/openstack/os_network.py
```diff
@@ -0,0 +1,107 @@
...
+
+    if not HAS_SHADE:
+        module.fail_json(msg='shade is required for this module')
+
+    state = module.params['state']
+    name = module.params['name']
+    shared = module.params['shared']
+    admin_state_up = module.params['admin_state_up']
+
+    try:
```
bcoca, line 82: exception handling is a bit too 'catchall' it would be nice to capture and return more meaningful messages from module 'failted to connect: <exception reason>', 'failed to create network', etc.

cloud/openstack/os_network.py
```diff
@@ -0,0 +1,107 @@
...
+   state:
+     description:
+        - Indicate desired state of the resource.
+     choices: ['present', 'absent']
+     default: present
+   name:
+     description:
+        - Name to be assigned to the network.
+     required: true
+     default: None
```
bcoca: no default when required

cloud/openstack/os_network.py
```diff
@@ -0,0 +1,107 @@
...
+DOCUMENTATION = '''
+---
+module: os_network
+short_description: Creates/Removes networks from OpenStack
+extends_documentation_fragment: openstack
+version_added: "1.10"
+description:
+   - Add or Remove network from OpenStack.
+options:
+   state:
```
bcoca: missing required: false

cloud/openstack/os_network.py
```diff
@@ -0,0 +1,107 @@
...
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
+
+DOCUMENTATION = '''
+---
+module: os_network
+short_description: Creates/Removes networks from OpenStack
+extends_documentation_fragment: openstack
+version_added: "1.10"
```
bcoca: 2.0

cloud/openstack/_quantum_network.py
```diff
@@ -29,6 +29,7 @@
 ---
 module: quantum_network
 version_added: "1.4"
+deprecated: Deprecated in 1.10. Use os_network instead
```
bcoca: 2.0

cloud/openstack/os_network.py
emonty, line 82: Happy to update this to split up the exceptions - however, the messages returned in OpenStackCloudException are exactly what you're suggesting, so the error messages _should_ be user informative. ... I'll follow up with a patch that splits the try/except up and see what it looks like.

PR #1048, opened 2015-03-31
Add OpenStack Router module

emonty:
Also deprecate old quantum_router module

### DIFF
--- a/cloud/openstack/os_router.py
+++ b/cloud/openstack/os_router.py
@@ -26,6 +26,7 @@ module: os_router
 short_description: Create or delete routers from OpenStack
 extends_documentation_fragment: openstack
 version_added: "2.0"
+author: "David Shrewsbury (@Shrews)"
 description:
    - Create or Delete routers from OpenStack. Although Neutron allows
      routers to share the same name, this module enforces name uniqueness
--- a/cloud/openstack/os_router.py
+++ b/cloud/openstack/os_router.py
@@ -28,7 +28,7 @@ DOCUMENTATION = '''
 module: os_router
 short_description: Create or Delete routers from OpenStack
 extends_documentation_fragment: openstack
-version_added: "1.10"
+version_added: "2.0"
 description:
    - Create or Delete routers from OpenStack. Although Neutron allows
      routers to share the same name, this module enforces name uniqueness
@@ -45,7 +45,7 @@ options:
      required: true
    admin_state_up:
      description:
-        - Desired admin state of the created router.
+        - Desired admin state of the created or existing router.
      required: false
      default: true
 requirements: ["shade"]
@@ -57,6 +57,13 @@ EXAMPLES = '''
     state=present
     name=router1
     admin_state_up=True
+'''
+
+RETURN = '''
+id:
+    description: Router ID
+    returned: On success when I(state) is 'present'.
+    type: string
 '''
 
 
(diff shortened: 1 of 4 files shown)


bcoca:
requries rebase

emonty:
Let's hold on this one for a sec - I think we can make it and the other two router modules take better advantage of underlying library much better.

cloud/openstack/os_router.py
```diff
@@ -0,0 +1,138 @@
...
+except ImportError:
+    HAS_SHADE = False
+
+
+DOCUMENTATION = '''
+---
+module: os_router
+short_description: Create or Delete routers from OpenStack
+extends_documentation_fragment: openstack
+version_added: "1.10"
```
omgjlk: This needs to be 2.0

cloud/openstack/os_router.py
```diff
@@ -0,0 +1,138 @@
...
+
+EXAMPLES = '''
+# Creates a router for tenant admin
+- os_router:
+    state=present
+    name=router1
+    admin_state_up=True
+'''
+
+
```
omgjlk, line 150: Return data should be documented.

cloud/openstack/os_router.py
```diff
@@ -0,0 +1,138 @@
...
+        - Indicate desired state of the resource
+     choices: ['present', 'absent']
+     default: present
+   name:
+     description:
+        - Name to be give to the router
+     required: true
+   admin_state_up:
+     description:
+        - Desired admin state of the created router.
```
omgjlk: This should be updated to reflect that it can change state of an existing router.

cloud/openstack/os_router.py
```diff
@@ -0,0 +1,138 @@
...
+                                 id=router['id'])
+            else:
+                if _needs_update(router, admin_state_up):
+                    cloud.update_router(router['id'],
+                                        admin_state_up=admin_state_up)
+                    module.exit_json(changed=True, result="updated",
+                                     id=router['id'])
+                else:
+                    module.exit_json(changed=False, result="success",
+                                     id=router['id'])
```
omgjlk: result= in these exit statements seem unnecessary. Changed is changed, success is assumed unless error.

cloud/openstack/os_router.py
```diff
@@ -0,0 +1,138 @@
...
+                cloud.delete_router(name)
+                module.exit_json(changed=True, result="deleted")
+
+    except shade.OpenStackCloudException as e:
+        module.fail_json(msg=e.message)
+
+# this is magic, see lib/ansible/module_common.py
+from ansible.module_utils.basic import *
+from ansible.module_utils.openstack import *
+main()
```
omgjlk: I think we want to do if **name** blah blah here for getting the main function to run.

gregdek:
@emonty thoughts on @j2sol proposed revisions here?

gregdek:
ping @emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz

Shrews:
Agree whole-heartedly with @j2sol's comments. And I actually thought this one was already merged for some reason. Will see about getting it cleaned up.

Shrews:
@j2sol Addressed all of your comments in the latest commit. Thanks!

Shrews:
Let's not merge this just yet. @emonty and I were discussing possibly merging this module with the modules from PRs #1052 and #1051 because, realistically, one wouldn't create a router without adding an interface and gateway. I will begin looking at that and will update here.

emonty:
lgtm - @j2sol ?

omgjlk:
I think this is good. I'll want to follow it up with another PR to add the ability to create the router for a specific tenant, instead of assuming the login tenant, but that shouldn't be part of the initial merge MVP.

Shrews:
@j2sol That sounds like something good to add later. Shade does not currently support that, mostly because the neutron API docs do not show that as being possible, though it clearly is now that I look at the neutron command line options.

emonty:
I'd love to have a deeper conversation about this concept and also pull 
in our keystone friend, because conceptually there's weird things going 
on here. But yes - as a follow on with a plan.

Shrews:
@gregdek @emonty With this newest module, we can close #1051 and #1052 since those are no longer needed. I can't close those myself.

Shrews:
@emonty @j2sol Added one commit to deprecate the quantum_router_interface and quantum_router_gateway modules since you both +1'd this. This lets us properly close #1051 and #1052 which had those deprecations.

Shrews:
And added missing author line.  :)

emonty:
shipit

gregdek:
Thanks @emonty and @Shrews for the work here. Marking for inclusion.

PR #1053, opened 2015-03-31
Add OpenStack Security Group module

emonty:


cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,113 @@
...
+        name              = dict(required=True),
+        description       = dict(default=None),
+        state             = dict(default='present', choices=['absent', 'present']),
+    )
+    module_kwargs = openstack_module_kwargs()
+    module = AnsibleModule(argument_spec, **module_kwargs)
+
+    try:
+        cloud = shade.openstack_cloud(**module.params)
+        nova_client = cloud.nova_client
```
omgjlk: Is nova client being used here to mask the difference between nova-network and neutron for the group creation?

cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,113 @@
...
+        changed = False
+        secgroup = cloud.get_security_group(module.params['name'])
```
omgjlk: Since security groups can have non-unique names, is there anything in shade that protects against finding multiple groups with the same name here?

cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if module.params['state'] == 'present':
+            secgroup = cloud.get_security_group(module.params['name'])
```
omgjlk: Why a second call to get the secgroup? This call just happened on line 82.

cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,113 @@
...
+            if not secgroup:
+                _security_group(module, nova_client, action='create',
+                                name=module.params['name'],
+                                description=module.params['description'])
+                changed = True
+
+        if secgroup and secgroup.description != module.params['description']:
```
omgjlk: This functionality of being able to update the description should be demonstrated in the docstring, along with details that idempotence is on the name of group only.

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,156 @@
...
+module: os_security_group_rule
+short_description: Add/Delete rule from an existing security group
+extends_documentation_fragment: openstack
+version_added: "2.0"
+description:
+   - Add or Remove rule from an existing security group
+options:
+   security_group:
+     description:
+        - Name of the security group
```
omgjlk: Can this be an ID as well, if we have the secgroup ID?

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,156 @@
...
+        - Starting port
+      required: true
+   port_range_max:
+      description:
+        - Ending port
+     required: true
+   remote_ip_prefix:
+      description:
+        - Source IP address(es) in CIDR notation (exclusive with remote_group)
+     required: false
```
omgjlk: If you don't supply a remote_ip_prefix, what is the default behavior?

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,156 @@
...
+        ],
+    )
+    module = AnsibleModule(argument_spec, **module_kwargs)
+
+    try:
+        cloud = shade.openstack_cloud(**module.params)
+        nova_client = cloud.nova_client
+        changed = False
+
+        secgroup = cloud.get_security_group(module.params['security_group'])
```
omgjlk: Same question as in the other module, what if there are multiple groups with the same name? Does shade bubble up an exception?

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+        if module.params['state'] == 'absent' and secgroup:
+            rule = _get_rule_from_group(module, secgroup)
+            if secgroup and rule:
+                _security_group_rule(module, nova_client, 'delete',
+                                     rule=rule['id'])
+                changed = True
+
+        module.exit_json(changed=changed, result="success")
```
omgjlk: Might be friendly to return the security group rule ID with the return value too, for later attaching to an instance.

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,156 @@
...
+                                     parent_group_id=secgroup.id,
+                                     ip_protocol=module.params['protocol'],
+                                     from_port=module.params['port_range_min'],
+                                     to_port=module.params['port_range_max'],
+                                     cidr=module.params['remote_ip'],
+                                     group_id=module.params['remote_group'],
+                changed = True
+
+
+        if module.params['state'] == 'absent' and secgroup:
```
omgjlk: For some reason, Travis's python doesn't like this line.

omgjlk:
Completed review for this current offering. Some questions in the code, and then Travis can't compile the module even though the syntax looks right.

Shrews:
@j2sol Just pushed a new version of os_security_group that addresses all of your comments. With the latest shade API that it now uses, an exception will be raised on duplicate security group names. And use of the nova client (or neutron client) is now hidden. Also updated the doctoring as you suggested.

cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,145 @@
...
+        cloud = shade.openstack_cloud(**module.params)
+        secgroup = cloud.get_security_group(name)
+
+        if module.check_mode:
+            module.exit_json(changed=_system_state_change(module, secgroup))
+
+        if state == 'present':
+            if not secgroup:
+                secgroup = cloud.create_security_group(name, description)
+                module.exit_json(changed=True, result='created',
```
omgjlk: I don't think it's proper to put a result= here. exit_json is always assumed a success, and the changed is used to indicate change. Adding a result here is confusing.

cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,145 @@
...
+                                 id=secgroup['id'])
+            else:
+                if _needs_update(module, secgroup):
+                    secgroup = cloud.update_security_group(
+                        secgroup['id'], description=description)
+                    module.exit_json(changed=True, result='updated',
```
omgjlk: Again shouldn't have a result=

cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,145 @@
...
+                                     id=secgroup['id'])
+                else:
+                    module.exit_json(changed=False, result='success')
```
omgjlk: success is assumed from exit_json vs fail_json.

cloud/openstack/os_security_group.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if state == 'absent':
+            if not secgroup:
+                module.exit_json(changed=False, result='success')
```
omgjlk: Same about result=

cloud/openstack/os_security_group_rule.py
omgjlk: Also shouldn't have result= here in the exit_json params.

omgjlk:
Travis is still not happy with the syntax, and the group name vs group ID input for os_security_group_rule is not addressed. Other comments in-line.

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,156 @@
...
+                                 module.params['security_group'])
+
+            if not _get_rule_from_group(module, secgroup):
+                _security_group_rule(module, nova_client, 'create',
+                                     parent_group_id=secgroup.id,
+                                     ip_protocol=module.params['protocol'],
+                                     from_port=module.params['port_range_min'],
+                                     to_port=module.params['port_range_max'],
+                                     cidr=module.params['remote_ip'],
+                                     group_id=module.params['remote_group'],
```
omgjlk: This function call is missing a closing paren, which is causing the statement on line 141 to be invalid syntax.

cloud/openstack/os_security_group.py
Shrews: @j2sol I've removed all 'result=' values. These appear to be holdovers from the old module code and I agree that it's redundant/confusing.

Shrews:
@j2sol Agree with all your comments on the secgroup module, and the new commit has the fixes.

There is a fix in the PR pipeline for one of the @emonty branches for the rules module. The rules module is also due to be rewritten very soon anyway to rely more on shade and clean up the old module code.

Shrews:
Removed os_security_group_rule module as it has it's own PR and branch.

omgjlk:
This now looks good to me! :+1:

PR #1054, opened 2015-03-31
Add OpenStack Security Group Rule module

emonty:


emonty:
Closing this because it doesn't make sense for it to be separate from os_security_group

PR #1057, opened 2015-03-31
Add OpenStack Server Volume module

emonty:


cloud/openstack/os_server_volume.py
```diff
@@ -0,0 +1,130 @@
...
+   volume:
+     description:
+      - Name or id of volume you want to attach to a server
+     required: true
+   device:
+     description:
+      - Device you want to attach. Defaults to auto finding a device name.
+     required: false
+     default: None
+   state:
```
ibratoev: Duplicat option description

ibratoev:
I get volume already attached error when the volume is really already attached.

cloud/openstack/os_server_volume.py
Shrews: Thanks. Fixed in an updated version.

Shrews:
The "already attached" problem is fixed in an updated version, soon to be published.

sky-shiny:
I thought new modules went into extras?

emonty:
@sky-shiny for these, since they are replacing the old ones, we decided to put them into core.

cloud/openstack/os_server_volume.py
```diff
@@ -0,0 +1,151 @@
...
+
+EXAMPLES = '''
+# Attaches a volume to a compute host
+- name: attach a volume
+  hosts: localhost
+  tasks:
+  - name: attach volume to host
+    os_server_volume:
+      state: present
+      cloud: mordred
```
dwcramer, line 67: "cloud" isn't mentioned in the options above. The other openstack modules use login_tenant_name, login_username, and login_password as credentials, but I don't see those listed above either.

emonty, line 67: login_\* are only used in the old modules which are being deprecated. The changes are to allow supporting keystone v3 with auth plugins- where username and password may not actually be things.

You'll want to check out: https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/module_docs_fragments/openstack.py which are the doc fragments that explain the common parameters for all of the new modules. There is a tutorial doc coming - I'll try to write that tonight.

dwcramer, line 67: @emonty Ok, good to know, thanks. Oh, but please do add "cloud" to the options in the documentation section above.

detiber:
@bcoca @emonty I have a question regarding the new openstack modules and dynamic inventory (and their inclusion in 2.0). Who will be handling the packaging for shade and os-client-config? If it is going to fall on the distributions to provide the packaging for these libraries I see it as a pretty serious regression in usability.

emonty:
@detiber not sure I understand the question or the concern? It falls on distributions to package software- that is, in fact, the thing that they do. It also falls on the distros to package python-novaclient as well. However, this is one of the reasons we're deprecating and not deleting the old modules ... should someone be in a situation where they are installing ansible v2 not from their distro but do not want to install shade not from their distro (which is weird) - they can wait until their distro has shade to switch modules.

detiber:
@emonty I suppose my main concern is that shade doesn't appear to be packaged for any distributions currently (and os-client-config only appears to be in debian-experimental and arch).

What I probably should have inquired about with my first comment is if there was already someone that was planning to be responsible for packaging the libraries for some or all common distributions or if there was some type of outreach to the distributions to request packaging for the libraries.

emonty:
@detiber AHA! I grok what you say now. So - yes, we will absolutely be making sure that they are upstreamed. os-client-config will get done for sure since it's a dependency of python-openstackclient - so the distro folks doing openstack will take care of that.
I know that wikimedia is working on debian packages for nodepool, which is about to depend on shade, and RedHat are working on rpms for everything openstack-infra related. But, ultimately, that's on me, so I'll sure we either have someone signed up to own it.

detiber:
@emonty great to hear, that definitely clears things up!

mcameron:
@emonty I'm not aware of prior art on an os_server_volume module.  This should be in extras.

mcameron:
Wtf is a 'cloud'?  

```
msg: one of the following is required: cloud,auth.  
```

Do I need a config file?

emonty:
You can either have a config file as per openstack/os-client-config and reference a named cloud from that config with the cloud param, or you can provide auth info in an auth dict parameter. The values in the auth dict vary from cloud to cloud, but you will probably have auth_url, username, password and project_name
On May 7, 2015 4:25 PM, Max Cameron notifications@github.com wrote:Wtf is a 'cloud'?  

msg: one of the following is required: cloud,auth.  

Do I need a config file?

mcameron:
So it breaks the environment variable convention?  :-1:

emonty:
Nope. It also honors all the env vars
On May 8, 2015 4:11 AM, Max Cameron notifications@github.com wrote:So it breaks the environment variable convention?

mcameron:
It honors env vars, however not by default.  You need to add:

```
cloud: envvars
```

To your requests to get the config.  Suboptimal, but at least it works.

mcameron:
Running this module:

```
TASK: [attach the volume] *****************************************************
failed: [127.0.0.1] => {"failed": true}
msg: Error attaching volume 3efed637-def5-43f9-8144-65a264b02078 server d3bed518-475f-4295-9ae6-70214ba37f07: The supplied device path (/dev/vdb) is in use. (HTTP 409) (Request-ID: req-fc0a0b90-d884-4d74-8d6f-ce3e1e98c6d4)
```

When that volume is already attached to that instance on that mountpoint.

```
nova show d3bed518-475f-4295-9ae6-70214ba37f07

| os-extended-volumes:volumes_attached | [{"id": "3efed637-def5-43f9-8144-65a264b02078"}]
```

emonty:
On 05/08/2015 10:40 AM, Max Cameron wrote:

> It honors env vars, however not by default.  You need to add:
> 
> ```
> cloud: envvars
> ```
> 
> To your requests to get the config.  Suboptimal, but at least it works.

Thanks - that's a bug. It should honor them without that.

Issue #994, opened 2015-03-25
NameError: name 'sys' is not defined


greenscar:
osx 10.10.2
Python 2.7.9
ansible 1.8.4

I create a script & call it exit.py

```
#!/usr/bin/python
import sys
print "test"
sys.exit(0)
```

and I execute in ansible via:

```
- debug: msg="test"

- name: shell
  shell: python --version

- name: test script
  script: ../../../exit.py

- name: launch ansible cloudformation c
  cloudformation:
    stack_name: "ansible-cloudformation"
    state: "present"
    region: "us-east-1"
    disable_rollback: true
    template: "files/cloudformation-example.json"
    template_parameters:
      KeyName: "jmartin"
      DiskType: "ephemeral"
      InstanceType: "m1.small"
      ClusterSize: 3
    tags:
      Stack: "ansible-cloudformation"

```

Exit works fine but cloudformation barfs with sys not defined.

```
__
| ~/workspaces/github/ansible @ macmini (james) 
| => ansible-playbook -i envs/localhost ./tasks/aws_test.yml -vvvv 

PLAY [localhost] ************************************************************** 

TASK: [debug msg="test"] ****************************************************** 
ok: [localhost] => {
    "msg": "test"
}

TASK: [shell] ***************************************************************** 
<localhost> 
<localhost> 
<localhost> 
<localhost> u'LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8 /usr/bin/python /Users/james/.ansible/tmp/ansible-tmp-1427249652.83-278423677659105/command; rm -rf /Users/james/.ansible/tmp/ansible-tmp-1427249652.83-278423677659105/ >/dev/null 2>&1']
changed: [localhost] => {"changed": true, "cmd": "python --version", "delta": "0:00:00.012900", "end": "2015-03-24 19:14:12.953713", "rc": 0, "start": "2015-03-24 19:14:12.940813", "stderr": "Python 2.7.9", "stdout": "", "warnings": []}

TASK: [test script] *********************************************************** 
<localhost> 
<localhost> 
<localhost> 
<localhost> LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8
<localhost> 
changed: [localhost] => {"changed": true, "rc": 0, "stderr": "", "stdout": "test\n"}

TASK: [launch ansible cloudformation c] *************************************** 
<localhost> stack_name=ansible-cloudformation state=present region=us-east-1 template=files/cloudformation-example.json
[...]

PR #1062, opened 2015-04-01
issue #994: use HAS_BOTO to determine if import was successful

steenzout:
- removed import of sys module where it wasn't necessary.
- HAS_BOTO constant to check if import was successful.
- trigger a failure when import fails.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -487,7 +487,6 @@ EXAMPLES = '''
 
 '''
 
-import sys
 import time
 from ast import literal_eval
 
@@ -496,9 +495,10 @@ try:
     from boto.ec2.blockdevicemapping import BlockDeviceType, BlockDeviceMapping
     from boto.exception import EC2ResponseError
     from boto.vpc import VPCConnection
+    HAS_BOTO = True
 except ImportError:
-    print "failed=True msg='boto required for this module'"
-    sys.exit(1)
+    HAS_BOTO = False
+
 
 def find_running_instances_by_count_tag(module, ec2, count_tag, zone=None):
 
@@ -782,7 +782,6 @@ def create_instances(module, ec2, vpc, override_count=None):
     # group_id and group_name are exclusive of each other
     if group_id and group_name:
         module.fail_json(msg = str("Use only one type of parameter (group_name) or (group_id)"))
-        sys.exit(1)
 
     vpc_id = None
     if vpc_subnet_id:
@@ -1202,6 +1201,9 @@ def main():
                                 ['exact_count', 'instance_ids']
                              ],
     )
+
+    if not HAS_BOTO:
+        module.fail_json(msg='boto required for this module')
 
     ec2 = ec2_connect(module)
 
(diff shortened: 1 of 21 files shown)


bcoca:
don't change the shebang, it breaks the ansible_python_interpreter feature.

steenzout:
if I don't change it I can't use the EC2 module on a `virtualenv` that has ansible and boto installed.
that's the root cause of the issue I'm having.

if the implementation of `ansible_python_interpreter` feature needs this then it also changes the expected behavior of all other non-ansible python packages that are installed on a `virtualenv`.
does it make sense to break a standard to implement a feature?

I read the doc about this feature and wouldn't `virtualenv -p python2 ~/env` and `source ~/env/bin/activate` and `~/env/bin/deactivate` be a cleaner solution to overcome the problem where systems only have a python3 interpreter (or need/want to use a python2 non-default interpreter) and you would need to switch between interpreters to run ansible?
plus you wouldn't need code to implement it and maintain it and it would be standard, right?

I'll try and see what happens if I set the `ansible_python_interpreter` but even if it works it's not practical since you're making me set a variable on all of my inventories, even if my system has a default python2 interpreter because I use `virtualenv` directly or through `tox` to test my playbooks, roles, etc.

if I install ansible through `macports` I guess I'll also need to set this variable as well on all of my inventories since my python interpreter is `/opt/local/bin/python` and `pip` is `/opt/local/bin/pip` and `boto` will be installed in

```
/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/boto
```

otherwise I won't be able to access the EC2 core module since the default interpreter looks somewhere in:

```
...
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/
/Library/Python/2.7/site-packages
...
```

would people using `brew` have the same problem?

now, imagine I set that variable on all my inventories.
my production environment is say... redhat.
I guess I need to double the inventory files or pass different parameters to the ansible playbook?

I rather `sed` all of shebang lines to put it as they are on the commit when I set the `virtualenv` and install ansible on my mac and completely forget about this feature. :)

conclusion: I can remove the commit (or you can discard the commit) but anyone using `virtualenv` or using a non-default python interpreter to install ansible and external dependencies will have the same issue as I'm having.

your call... :)

sivel:
The correct answer is of course to remove that commit. Changing the interpreter line is the incorrect approach.

steenzout:
I've removed the shebang line change from the commit.
I'll post the issue of using `virtualenv` with ansible modules with an external dependency on the dev mailing list.

abadger:
@steenzout For virtualenv, you want something like this solution: http://stackoverflow.com/questions/20575084/best-way-to-always-run-ansible-inside-a-virtualenv-on-remote-machines

Basically leveraging the ansible_python_interpreter and a short wrapper script to enable the virtualenv.

steenzout:
@abadger thank you for the link.
my case is actually having different local machines where I can run playbooks (in different OS), which use `virtualenv` where ansible and 3rd party dependencies are installed.
I've managed to find a easier workaround with just using bash (for local and remote execution of playbooks tasks) without the need for the `ansible_python_interpreter` feature.

hope the commit is acceptable.
let me know if more tweaks are needed.

abadger:
yep, looks good.  Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

PR #1073, opened 2015-04-03
Add ability to specify using ssh_args in synchronize

sivel:
This PR, coupled with https://github.com/ansible/ansible/pull/10612 for the action module, adds 2 new options:
1. `use_ssh_args`
2. `ssh_args`

`use_ssh_args`, which is a boolean (defaulted to False to keep current behavior), is what is exposed currently via documentation, but not in the `argument_spec`.

`ssh_args`, a string, is not exposed via documentation, and does appear in `argument_spec`

`use_ssh_args` is popped from the options, and `ssh_args` is added.

### DIFF
--- a/files/synchronize.py
+++ b/files/synchronize.py
@@ -138,6 +138,13 @@ options:
       - put user@ for the remote paths. If you have a custom ssh config to define the remote user for a host
         that does not match the inventory user, you should set this parameter to "no".
     default: yes
+  use_ssh_args:
+    description:
+      - Use the ssh_args specified in ansible.cfg
+    default: "yes"
+    choices:
+      - "yes"
+      - "no"
   rsync_opts:
     description:
       - Specify additional rsync options by passing in an array.
@@ -227,7 +234,8 @@ def main():
             group = dict(type='bool'),
             set_remote_user = dict(default='yes', type='bool'),
             rsync_timeout = dict(type='int', default=0),
-            rsync_opts = dict(type='list')
+            rsync_opts = dict(type='list'),
+            ssh_args = dict(type='str'),
         ),
         supports_check_mode = True
     )
@@ -254,6 +262,7 @@ def main():
     owner = module.params['owner']
     group = module.params['group']
     rsync_opts = module.params['rsync_opts']
+    ssh_args = module.params['ssh_args']
 
     cmd = '%s --delay-updates -F' % rsync
     if compress:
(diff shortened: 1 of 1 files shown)


files/synchronize.py
```diff
@@ -306,7 +316,7 @@ def main():
     else:
         private_key = '-i '+ private_key 
 
-    ssh_opts = '-S none -o StrictHostKeyChecking=no'
+    ssh_opts = '-S none -o StrictHostKeyChecking=no %s' % ssh_args
```
bcoca, line 319: this does not seem to do anything but add yes/no to the ssh_opts

tima:
I agree something seems off here @bcoca, but line 319 won't add yes/no. More likely a blank string or None. The intent stated in the PR/ticket is not what I'm seeing in the commit. Is there a missing file? Perhaps in the associated action module? That's the only way I can make sense of this PR.

tima:
I'm digging around on this one a bit more and I think I see why I'm confused and what you meant @sivel. 

I see the action plugin PR this one needs to make it complete. Specifically commit d5eb4df23ee7fd8086eae988a85c42204832777d in the core. I tripped over the core modules-core split and missed that this one. Seems I wasn't the only one since it seems @bcoca didn't see it either.

My only nit it that the concatenation of ssh_args should be conditional rather than tack that extra whitespace on if ssh args is not defined.

files/synchronize.py
sivel, line 319: `ssh_args` and `use_ssh_args` are 2 different things.  I think you are missing the other piece which is the action plugin which has a PR at https://github.com/ansible/ansible/pull/10612

sivel:
This functionality requires 2 PRs.  This one and the one in the core ansible repo at https://github.com/ansible/ansible/pull/10612

> I'm digging around on this one a bit more and I think I see why I'm confused and what you meant @sivel.
> 
> I see the action plugin PR this one needs to make it complete. Specifically commit d5eb4df23ee7fd8086eae988a85c42204832777d in the core. I tripped over the core modules-core split and missed that this one. Seems I wasn't the only one since it seems @bcoca didn't see it either.
> 
> My only nit it that the concatenation of ssh_args should be conditional rather than tack that extra whitespace on if ssh args is not defined.

Also, who cares if an extra whitespace is added, what harm does that cause?

sivel:
BTW, `needs_revision` should probably be removed.

PR #1039, opened 2015-03-31
[docker] Record facts containing each changed container.

smashwilson:
This restores the registration of `ansible_facts` containing each changed container, until a more featureful `docker_facts` module is implemented.

Fixes #1005.

/cc @thoughtjigs, @abadger, @bcoca

Issue #1005, opened 2015-03-26
docker_containers


thoughtjigs:
the core docker module no longer publishes ansible facts using the docker_containers variable.

abadger:
@smashwilson: I see where we accidentally removed this here: 80aca4b9369073d7b58ff168cf4eef8bb126c2d0

@thoughtjigs: Do you know what was in the docker_containers fact before?  It doesn't look like it was the complete list of containers.  Was it the same value that's now returned in containers (containers.changed inside of the module code) it just needs to be automatically stored in a fact?

smashwilson:
Oh, I see, interesting. Yeah, I can take a look at adding that back in.

It looks like it was only the changed containers before the refactor, but because it's a fact, I would personally expect it to be the "inspect" results for _all_ of the running containers that match, not just the ones that were changed -- the output of `get_deployed_containers()` after the state has been applied. Does that make more sense?

It'll be easy enough to put back either way, though.

abadger:
That's what I expected too but it seems like it's not what was there before.   If get_deployed_containers() turns out to be a superset of what we would have used before then I think it's fairly safe to do that (especially if we are returning just the changed containers in a different variable now).  If it's not a superset then it's a tougher call but I'd like to hear what people using the docker module want -- I don't use it enough to form an opinion on my own.

thoughtjigs:
@abadger @smashwilson It was only the changed containers. As that is what my playbooks are expecting returning all containers would break my playbooks. However, it is a slightly under-documented feature ... having ansible facts returned for different collections of containers would be useful. Perhaps there could be several dictionaries of facts: docker_changed, docker_running, docker_stopped. docker_containers could continue to contain only the changed docker_container facts and be deprecated in favor of the above three. 

Additionally, providing switches for container fact gathering might be useful for large container inventories ... though I have seen no performance hit when using my dynamic docker container inventory script that gathers facts on all containers. 

I could take a look at creating a patch for this as this is critical to our playbooks (I had to revert to 1.8.4 for the time being) if that would be useful. 

Thanks

bcoca:
there is a docker_facts module in the queue, I would not overload this one with too many features (changed containers seems fine).

gimoh:
ehm, in 1.8.4 it was actually returning inspect data for all containers that matched or matched and were running (depending on value of state option), whether their state was changed or not.  It was actually returning output of `get_deployed_containers()` or `get_running_containers()`.

[...]

PR #1078, opened 2015-04-03
ansible_facts reflects 1.8 output

andrewgdavis:


bcoca:
??

andrewgdavis:
the output of ansible_facts changed between versions.  the one line change provided the data in the same manner 1.8 did.

The relevant history:
https://github.com/ansible/ansible-modules-core/issues/1005

The solution put in place-- but the behavior changed between versions:
https://github.com/ansible/ansible-modules-core/pull/1039

I don't know what the direction is of the future "docker_facts" module so that may supercede my fix.  Until then, the one line change may help someone.

ritterwolf:
I quickly tested this against a playbook that fails under 1.9, and it works for me.

gimoh:
This isn't a complete fix, in 1.8.4 `docker_containers` included inspect data for all matched or running containers, not only the ones where state changed.  See [my comment on #1005](https://github.com/ansible/ansible-modules-core/issues/1005#issuecomment-99393219).

gregdek:
@ansible/docker Please review this PR to make sure it adheres to the following guidelines:

PR #1080, opened 2015-04-04
Allow parsing of a multiline metadata value by escaping newline caracter.

lpereir4:
This is useful for the parsing of a google-container-manifest metadata to succeed.


bcoca:
cc @erjohnso

Issue #1076, opened 2015-04-03
local_action wait_for takes 5 minutes to run


tetherit:
I have this task:

``` yaml
- local_action:
    module: wait_for
      "host={{ ipaddress }} port=22 delay=10 connect_timeout=5 timeout=1800"
  sudo: false
  tags:
    - 'testwaitfor'
```

I can telnet to a running SSH just fine from this machine:

```
$ telnet 10.8.0.97 22
Trying 10.8.0.97...
Connected to 10.8.0.97.
Escape character is '^]'.
SSH-2.0-OpenSSH_6.6.1p1 Ubuntu-8
^]
```

But when I try this from Ansible, I see this:

```
$ time ansible-playbook timeboxes.yml --tags testwaitfor -vvv

PLAY [carpc_rpi2] ************************************************************* 

GATHERING FACTS *************************************************************** 
<vpn.carpc_rpi2.zanview.com> ESTABLISH CONNECTION FOR USER: deployer
<vpn.carpc_rpi2.zanview.com> REMOTE_MODULE setup
<vpn.carpc_rpi2.zanview.com> EXEC ssh -C -v -o ForwardAgent=yes -o ControlPersist=30m -o ControlPath="/Users/Hackeron/.ansible/cp/ansible-ssh-%h-%p-%r" -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=deployer -o ConnectTimeout=10 vpn.carpc_rpi2.zanview.com /bin/sh -c 'sudo -k && sudo -H -S -p "[sudo via ansible, key=xkjuogremxhelybleroxixbnhtktppgt] password: " -u root /bin/sh -c '"'"'echo BECOME-SUCCESS-xkjuogremxhelybleroxixbnhtktppgt; LANG=en_US.UTF-8 LC_CTYPE=en_US.UTF-8 /usr/bin/python'"'"''
ok: [carpc_rpi2]

TASK: [common | wait_for "host=10.8.0.97 port=22 delay=10 connect_timeout=5 timeout=1800"] *** 
<vpn.carpc_rpi2.zanview.com> REMOTE_MODULE wait_for "host=10.8.0.97 port=22 delay=10 connect_timeout=5 timeout=1800"
[...]

PR #1082, opened 2015-04-04
Ensure one of port or path is passed to wait_for.

jder:
If you pass neither path or port, the module will not do anything
useful, so we should give an error for this case.

This would have helped debug this case: https://github.com/ansible/ansible-modules-core/issues/1076


bcoca:
it waits for a certain amount of time, that is also a feature, it is a 'sleep()'.

jder:
@bcoca OK. If that's intentional, I'll close this. Thanks!

Issue #1085, opened 2015-04-05
ansible 1.9.0.1 module: group traceback on osx 


thudry:
Issue Type:

Bug Report 

Ansible Version:
ansible 1.9.0.1
  configured module search path = /usr/share/ansible

Environment:

Ansible is installed on debian wheezy with pip, ansible manage osx 10.10

Summary:

When the gid is not set ansible traceback

Steps To Reproduce:

My playbook:

```

---
-   hosts: all
    user: root
    tasks:
    -   name: test
        group: name=lul state=present
```

Actual Results:

```
TASK: [group name=lol state=present] ******************************************
failed: [192.168.29.130] => {"failed": true, "parsed": false}
Traceback (most recent call last):
  File "/var/root/.ansible/tmp/ansible-tmp-1428202088.65-56239082117602/group", line 2041, in     <module>
    main()
  File "/var/root/.ansible/tmp/ansible-tmp-1428202088.65-56239082117602/group", line 424, in main
    (rc, out, err) = group.group_mod(gid=group.gid)
TypeError: group_mod() got an unexpected keyword argument 'gid'
OpenSSH_6.2p2, OSSLShim 0.9.8r 8 Dec 2011
debug1: Reading configuration data /etc/ssh_config
debug1: /etc/ssh_config line 20: Applying options for *
debug1: auto-mux: Trying existing master
debug1: mux_client_request_session: master session id: 2
Shared connection to 192.168.29.130 closed.


FATAL: all hosts have already failed -- aborting
```

The group isn't modified.
The problem is the same with a new group.

PR #973, opened 2015-03-20
1-line fix Darwin group_mod in system/group module by adding **kwargs

rmgrimm:
This also contains two tiny spacing/indentation changes, contained in a second commit.

PR #1086, opened 2015-04-05
module group: Check if the gid is set

thudry:
Check for Darwin if the gid is set. #1085

### DIFF
--- a/system/group.py
+++ b/system/group.py
@@ -273,7 +273,8 @@ class DarwinGroup(Group):
     def group_add(self, **kwargs):
         cmd = [self.module.get_bin_path('dseditgroup', True)]
         cmd += [ '-o', 'create' ]
-        cmd += [ '-i', self.gid ]
+        if self.gid is not None:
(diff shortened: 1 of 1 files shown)


rmgrimm:
Would it be better to have line 287 read as `def group_mod(self, **kwargs):` instead of `def group_mod(self, gid=None):`? It could then be consistent with the rest of the `group_mod` signatures in the file.

I made a similar and much smaller fix with #973. Unlike this one, mine doesn't allow gid to be optional as is described in the documentation.

bambou42:
In the documentation gid is not required

    gid	no		Optional GID to set for the group.

gregdek:
@bambou42 Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

thudry:
It should be good now

bcoca:
you have a merge commit now

thudry:
Sorry :/ is it good now ?

bcoca:
yes, i merged it already, thanks!

PR #1101, opened 2015-04-07
Update postgresql users only when necessary

willthames:
For read-only databases, users should not change when no changes
are required.

Don't issue ALTER ROLE when role attribute flags, users password
or expiry time is not changing.

In certain cases (hashed passwords in the DB, but the password
argument is not hashed) passlib.hash is required to avoid
running ALTER ROLE.

### DIFF
--- a/database/postgresql/postgresql_user.py
+++ b/database/postgresql/postgresql_user.py
@@ -44,7 +44,7 @@ options:
   password:
     description:
       - set the user's password, before 1.4 this was required.
-      - "When passing an encrypted password, the encrypted parameter must also be true, and it must be generated with the format C('str[\\"md5\\"] + md5[ password + username ]'), resulting in a total of 35 characters.  An easy way to do this is: C(echo \\"md5`echo -n \\"verysecretpasswordJOE\\" | md5`\\")."
+      - "When passing an encrypted password, the encrypted parameter must also be true, and it must be generated with the format C('str[\\"md5\\"] + md5[ password + username ]'), resulting in a total of 35 characters.  An easy way to do this is: C(echo \\"md5`echo -n \\"verysecretpasswordJOE\\" | md5`\\"). Note that if encrypted is set, the stored password will be hashed whether or not it is pre-encrypted."
     required: false
     default: null
   db:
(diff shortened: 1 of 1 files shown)


abadger:
@bcoca, what revisions need to be made here?

abadger:
I wasn't sure about why USAGE was being removed but after talking with willthames on IRC I examined it more closely.  Found all of these facts which seem to indicate that removing USAGE from postgresql_user.py as this PR does is correct:
- As will informed me on IRC, USAGE isn't valid for either database or table objects.  And if it's listed as a valid priv then the expansion of ALL would have to change.
- If the user were to specify USAGE as a priv in postgres_user, then the code would throw an error.
- USAGE was added here at the same time as it was added to postgresql_privs.  Likely it should only have been added to postgresql_privs and not to postgresql_user.

abadger:
This was breaking our integration tests.  On looking at it further, the has_database_privileges (and maybe has_table_privileges) function don't seem to do what we need here.  They find out whether the user in question has the ability to make that command.  This includes permissions assigned to roles that the user belongs to.  For the postgresql_user module it seems to be more sensible to test whether the user itself (not the roles the user belongs to) has the proper privileges.  This is because the changes will be made to the user and not to the roles that the user belongs to.

looking into whether we can make use of the get_*_privileges() functions here.

willthames:
@abadger so what happens now - do my changes get reverted?

Can you point me at the failing test, and I'll see if I can whether I can fix the problem myself, and possibly also improve the testsuite so that running the module against a readonly database works as long as no changes are required.

abadger:
The failing test is here: https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_postgresql/tasks/main.yml#L463
https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_postgresql/tasks/main.yml#L485
https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_postgresql/tasks/main.yml#L504

(Line 504 is the assert that fails.  Line 485 is the code that asks postgres for the current permissions of the user. Line 463 is the postgres_user task that is supposed to be setting the permissions)

What's happening is that the newly created ansible_user_1 is placed in a default role.  The default role on a fresh postgres install has connect and temp privileges.  So the has_database_privileges function sees that the user is able to perform this action and so it does not recommend any changes.  However, when we check the db user account itself, we get this output:

```
 {=Tc/postgres,postgres=CTc/postgres,ansible_db_user1=C/postgres}
```

Which shows that the db user itself only has create privileges.

abadger:
Before resorting to reverting I'm seeing if we can replace the has___privileges calls with something based on get___privileges.

abadger:
Okay, pull request that switches to using get_table_privileges and get_database_privileges: https://github.com/ansible/ansible-modules-core/pull/1368

Fixes the integration test failures.

PR #1102, opened 2015-04-07
Add support for extra_hosts to docker module

heiparta:
extra_hosts parameter (maps to --add-host in 'docker run' syntax) is used
to add host-to-ip mappings to the container.

Example usage:
    - name: start myfoo
      docker:
        name: myfoo
        image: busybox
        state: reloaded
        extra_hosts:
          dockerhost: 127.0.0.1
          foobarhost: 127.0.0.2

After running the playbook you can use docker inspect to verify that the new hosts are defined in the HostConfig section

Support for --add-host was added to docker in 1.31:
http://jasani.org/2014/11/19/docker-now-supports-adding-host-mappings/

extra_hosts was added to docker-py in 0.7.0:
https://github.com/docker/docker-py/commit/29c67c3a0503d2ad8ee1643d3db58962f3e9cfe1

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -246,6 +246,9 @@ options:
         retries.
     default: 0
     version_added: "1.9"
+  extra_hosts:
+    description:
+    - Dict of custom host-to-IP mappings to be defined in the container
   insecure_registry:
     description:
       - Use insecure private registry by HTTP instead of HTTPS. Needed for
@@ -492,6 +495,7 @@ class DockerManager(object):
             'dns': ((0, 3, 0), '1.10'),
             'volumes_from': ((0, 3, 0), '1.10'),
             'restart_policy': ((0, 5, 0), '1.14'),
+            'extra_hosts': ((0, 7, 0), '1.3.1'),
             'pid': ((1, 0, 0), '1.17'),
             # Clientside only
             'insecure_registry': ((0, 5, 0), '0.0')
@@ -1233,7 +1237,7 @@ class DockerManager(object):
 
         optionals = {}
         for optional_param in ('dns', 'volumes_from', 'restart_policy',
-                'restart_policy_retry', 'pid'):
+                'restart_policy_retry', 'pid', 'extra_hosts'):
             optionals[optional_param] = self.module.params.get(optional_param)
 
         if optionals['dns'] is not None:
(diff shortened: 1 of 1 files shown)


sebi-hgdata:
+1

wvlancke:
+1

PavelVanecek:
I fail to use this feature. My setup:

```
# my machine
$ ansible --version
ansible 1.9.2
$ cat playbook.yml
# ...
- name: mushrooms
  docker:
    name: mushrooms
    state: reloaded
    extra_hosts:
      kafka: 54.40.17.122

# remote host
$ docker --version
Docker version 1.5.0, build a8a31ef
$ pip list | grep docker
docker-py (1.1.0)
```

From my point of view, all is ready. But:

```
$ ansible-playbook playbook.yml
TASK: [mushrooms] ************************************************************* 
failed: [54.40.29.137] => {"failed": true}
msg: unsupported parameter for module: extra_hosts

```

What am I doing wrong?

Edit: It seems that even though my ansible --version says 1.9.2, the actual docker.py file included on my filesystem was out of date. So I uninstalled the brew version and installed ansible using pip, but the issue still remains.

Edit2: I managed to get it (almost) working after [installing ansible from source](http://docs.ansible.com/ansible/intro_installation.html#running-from-source). Now it does not throw an error, but it doesn't do anything either.

coreygale:
+1

We are experiencing the same situation PavelVanecek described - using `extra_hosts` doesn't throw any errors upon run-time but also didn't do anything. Any hosts added with `extra_hosts` can not be resolved inside the provisioned container.

Our setup:

OS: Ubuntu Server 14.04.2 LTS 64-bit
Ansible: 2.0.0 (latest devel)
Docker Client: 1.6.0
Docker API: 1.18
docker-py: 1.2.3

I was, however,  able to find a work-around by pinning a specific commit hash (`256a323de56d8259c9cd65ae4c55ab761d432b85`) after cloning the devel branch. With this modification, our Ansible setup looks like this:

``` shell
sudo apt-get install -yqq software-properties-common
sudo apt-get update -qq
sudo apt-get install git python-setuptools -yqq
sudo git clone -b devel git://github.com/ansible/ansible.git --recursive
cd ./ansible
sudo git checkout 256a323de56d8259c9cd65ae4c55ab761d432b85
source ./hacking/env-setup -q
sudo easy_install pip
sudo pip install paramiko PyYAML Jinja2 httplib2
sudo git pull --rebase
sudo git submodule update --init --recursive
// copy your relevant config files here
sudo make install
```

Setting up Ansible in this fashion fixes the `extra_hosts` issue, but should really only be used as a temporary solution.

PR #1103, opened 2015-04-07
remove calling format to make fakes3 work

ghost:
fakes3 only works after this augment removed

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -317,7 +317,6 @@ def main():
                 is_secure=fakes3.scheme == 'fakes3s',
                 host=fakes3.hostname,
                 port=fakes3.port,
-                calling_format=OrdinaryCallingFormat(),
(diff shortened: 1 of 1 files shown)


bcoca:
what happens when you don't use fakes3? should this not be conditional like is_secure?

ch3n2k:
if OrdinaryCallingFormat is specified, the buckets and files created by Ansible are not accessible by other tools (like s3cmd). And files created by other tools are not accessible by Ansible either. If remove this parameter or using the default SubdomainCallingFormat, there will be no such issue.

ch3n2k:
From the fakes3 side, we can see that there will be one more level of directory in the middle if OrdinaryCallingFormat is specified.

fakes3_root
  +---- bucket_name_1
  |
  +---- fakes3 ----  bucket_name_2

ghost:
my fakes3 setup fault, needs to specify -H start parameter

PR #1104, opened 2015-04-07
Replace catching of non-existing Exception with check against error code

tsu:
When creating or modifying an existing AWS autoscaling group fails with an exception, the original exception is masked by an exception stating `AttributeError: 'module' object has no attribute 'InvalidException'`


gregdek:
@tsu Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

tsu:
This PR does no longer apply.

PR #1105, opened 2015-04-07
Adding openstack load balancer management to ansible-modules-core.

sky-shiny:


sivel:
cc @emonty 

Also, I believe new modules should be submitted to the extras repo.

emonty:
Thanks for your contribution! As @sivel mentions, you'll want to submit this to ansible-modules-extras since it's completely new functionality (and also because the current state of neutron load balancers is a bit in flux)

A few comments before you re-submit there:
- Ansible is GPL, not MIT or BSD - so if you could adjust the headers in your submission that would be great
- Please rename these to be:  "os_server_load_balancer.py" and "os_load_balancer.py" 
- Please see #1058 for an example of how to deal with common auth parameters, common documentation and how to get a neutron_client object from shade. We'd like to make sure that these modules possess a common auth interface.

sky-shiny:
Moved to https://github.com/ansible/ansible-modules-extras/pull/379

PR #1107, opened 2015-04-07
Port `force` functionality from `get_url` module

elventear:
Make the Windows `win_get_url` module behave more like `get_url`, with respect of overwriting files.

### DIFF
--- a/windows/win_get_url.ps1
+++ b/windows/win_get_url.ps1
@@ -26,6 +26,15 @@ $result = New-Object psobject @{
     changed = $false
 }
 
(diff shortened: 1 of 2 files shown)


gregdek:
@elventear Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

elventear:
rebased

cchurch:
Thanks for your submission.

It looks like #1485 had also implemented the `force` option and added the additional feature of checking the last modified timestamp of the file.  We've chosen to use those changes instead.

PR #1109, opened 2015-04-08
updates win_msi to allow install/uninstall to wait

schwartzmx:
- Adds additional `wait` param for specifying whether to wait on the install/uninstall to complete.
- Updates to use `Start-Process` (https://technet.microsoft.com/en-us/library/hh849848(v=wps.620).aspx) to allow `-Wait` param
  - `Start-Process` **is** available in PowerShell V2.0 and greater
  - This will allow `msiexec.exe` cmdlet process to wait if specified (which currently isn't supported as an extra arg to `msiexec.exe`)

Questions, concerns, let me know.

Phil

### DIFF
--- a/windows/win_msi.py
+++ b/windows/win_msi.py
@@ -45,12 +45,25 @@ options:
         description:
             - Path to a file created by installing the MSI to prevent from
               attempting to reinstall the package on every run
+    wait:
+        version_added: ""
+        description:
+            - Specify whether to wait for install or uninstall to complete before continuing.
+        choices:
+            - true
+            - yes
+            - false
+            - no
+        default: false
 author: Matt Martz
 '''
 
 EXAMPLES = '''
 # Install an MSI file
 - win_msi: path=C:\\\\7z920-x64.msi
+
+# Install an MSI, and wait for it to complete before continuing
+- win_msi: path=C:\\\\7z920-x64.msi wait=true
 
 # Uninstall an MSI file
 - win_msi: path=C:\\\\7z920-x64.msi state=absent
--- a/windows/win_msi.ps1
+++ b/windows/win_msi.ps1
@@ -25,7 +25,7 @@ $path = Get-Attr $params "path" -failifempty $true
 $state = Get-Attr $params "state" "present"
 $creates = Get-Attr $params "creates" $false
 $extra_args = Get-Attr $params "extra_args" ""
-$wait = $false
+$wait = Get-Attr $params "wait" $false | ConvertTo-Bool
 
 $result = New-Object psobject @{
     changed = $false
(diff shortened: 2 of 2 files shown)


itshanney:
Phil, do you think that this commit will address the issue I found in the win_msi module with extra_args?

https://github.com/ansible/ansible-modules-core/issues/1229

It appears your change in how msiexec is executed may fix it, but I can't really tell due to my PowerShell inexperience.

schwartzmx:
Give it a test try, it may just be that the extra arguments that you are providing to the `.msi` are not supported.  I've ran into some of those issue before trying to do silent installs (not with ansible) but if the `.msi` file doesn't like an argument a pop up will occur with "natively" supported extra arguments and how to run the installer from the command line <- This popup could be causing your install to hang...

I'd try it locally with the arguments you are passing, if you can, and then go from there.  If the arguments are correct, give this update a try.

gregdek:
@sivel Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@sivel friendly ping, time to review this PR?

gregdek:
@schwartzmx Thanks for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

Once you have rebased this PR, please add a comment with the text "ready_for_review" and we will continue with the review process. 

(We may also at that point call out to other Windows reviewers besides @sivel for help here.)

schwartzmx:
@gregdek I've rebased, and fixed conflicts.  Thanks!

schwartzmx:
ready_for_review

gregdek:
Thanks @schwartzmx.

Ping @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch -- I hope one of you can review. :)

cchurch:
A few comments:
- Doesn't work if MSI path contains spaces
- Use `Get-Attr $params "wait" $false | ConvertTo-Bool` instead of explicitly checking for `"yes"` or `"true"`.  (We should probably make a note to update the argument parsing example at http://docs.ansible.com/ansible/developing_modules.html#windows-modules-checklist to provide an example of handling a boolean argument).
- Specify boolean choices in docs as `"yes"` and `"no"` for consistency with (most) other Windows modules.  (These currently show up as True and False on the docsite -- that may be a separate issue).

schwartzmx:
Thanks for the review @cchurch.  I updated the module to use `Get-Attr` and also fixed issues when specifying a path to an **msi** containing spaces by escaping quotes around the path.

As for the boolean values, I do specify `"yes"` and `"no"`  [here](https://github.com/schwartzmx/ansible-modules-core/blob/update/win_msi/windows/win_msi.py#L53-L56), along with `"true"` and `"false"`.  So I'm not sure, it could be another issue like you mentioned.

schwartzmx:
Could also add quotes around `$extra_args` and escape them to possibly fix https://github.com/ansible/ansible-modules-core/issues/1229,  what do you guys think?

``` PowerShell
Start-Process -FilePath msiexec.exe -ArgumentList "/x `"$path`" /qn /l $logfile `"$extra_args`"" -Verb Runas -Wait;
```

Or that could possibly be a separate PR.

h0nIg:
@schwartzmx add please and then ship it

h0nIg:
shipit

gregdek:
@h0nIg sorry, but you're not one of the module owners here... 

re-pinging windows folks for re-review: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

jhawkesworth:
meets_guidelines

cchurch:
I'm now seeing an issue when running this test https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_win_msi/tasks/main.yml#L31 and adding `wait: true` -- I get an error `Cannot convert value to type System.String` from the last `Exit-Json` line, which can be worked around by not adding the log contents to the result object.  Testing against Windows Server 2008 R2.

h0nIg:
@cchurch should be fixed with https://github.com/ansible/ansible-modules-core/pull/2563

gregdek:
OK, this one has been through the wringer. I presume it's now in "ready_for_review" state. 

@cchurch since you were the last reviewer here, could you give us your latest review on it and give us a "shipit" or "needs_revision" depending, so our triage bot can do the right thing?

windows/win_msi.py
```diff
@@ -45,6 +45,16 @@
         description:
             - Path to a file created by installing the MSI to prevent from
               attempting to reinstall the package on every run
+    wait:
+        version_added: "2.0"
+        description:
+            - Specify whether to wait for install or uninstall to complete before continuing.
+        choices:
```
smadam813, line 52: Why so many choices that mean the same thing?  I know ConvertTo-Bool supports them all, but I feel a module shouldn't have more than one option that means the same thing.

smadam813:
@schwartzmx I commented on the code for a potential change to choices.  Instead of 4 choices with 2 meaning identical things, just specify two that can be used.  

This is not a breaking issue, more of a consistency issue.  Other than this, I would say:

shipit

windows/win_msi.py
jhawkesworth, line 52: I think this is just reflecting the rest of ansible - the various ways of specifying booleans are mentioned here, for example: http://docs.ansible.com/ansible/YAMLSyntax.html

PR #1132, opened 2015-04-13
Fixes #1131: Add option to only add user to groups that exist

### LINKED ISSUE #1131: Add option to only add user to groups that exist
I'd like an option in the user module to allow for users to only be added to groups specified in the 'groups' parameter if the groups exist, and just happily skip any groups that do not rather than having the task fail.

This looks pretty trivial to add, so I'll try implementing it myself first and hopefully submit a pull request shortly.

tonyyarusso:
Adding a simple option to the user module to allow additional groups that don't exist to just be skipped, rather than triggering a failure.


tonyyarusso:
I need to replace this with a properly-named branch rather than having it right off devel.

tonyyarusso:
Replaced by #1176.

PR #1016, opened 2015-03-27
Add ec2_vpc_route_table and ec2_vpc_subnet modules.

dopry:
As discussed in #322, this is the ec2_vpc_route_table and and ec2_vpc_subnet module written by @erydo without the modifications to ec2_vpc. 

I also included patch from @herbygillot, https://github.com/ansible/ansible-modules-core/commit/5fd1aa6e646071f0dd6a515a2d6fc0ce809c9b54, that adds support for specifying the gateway_id as 'igw'. I switched from INTERNET_GATEWAY to confirm with how you specify it in the AWS managment console.

PR #322, opened 2014-11-14
Split out route table and subnet functionality from VPC module.

rwe:
This change factors out subnet and route table creation from the creation of the VPC itself. This changes the interface of the `ec2_vpc` module to take in subnet IDs and route table IDs, rather than large inline definitions.

This is a fairly large change. This PR I believe supersedes the functionality of #258, from which it was developed independently. It makes sense to be merged along with #321, which is still okay on its own.

The `ec2_vpc` module now supports check mode, as do the new `ec2_vpc_subnet` and `ec2_vpc_route_table` modules.

The basic premise was: the `ec2_vpc` module does too much monolithically, and is both inflexible and difficult to use. (For example: getting meaningful subnet IDs out of the VPC creation in order to provision EC2 instances on that subnet was nigh impossible). The code was very stateful and could not support check mode.

These refactorings:
1. Make idempotent playbooks easier. Route tables are not always replaced anymore.
2. Make check mode possible for VPC setup.
3. Make it easier to refactor playbooks for VPC setup. (Subnets, route tables, etc.)
4. More easily allow advanced configurations such as instance allocation on specific subnets.
5. Clean up the stateful code considerably.
6. Generally make this module set more "ansible"-like, I believe, by splitting the configuration into finer-grained verifiable tasks.

PR #1133, opened 2015-04-14
New vpc module. does not contain subnet or route table functionality.

juliedavila:
This is a rewrite of the ec2_vpc module this presumes the future merging of https://github.com/ansible/ansible-modules-core/pull/1016 and is a derivative of  https://github.com/ansible/ansible-modules-core/pull/322

Overall the ec2_vpc module had grown too monolithic. This module is intended to reside along the pre-existing module until full deprecation occurs.

### DIFF
--- a/cloud/amazon/ec2_vpc_net.py
+++ b/cloud/amazon/ec2_vpc_net.py
@@ -1,0 +1,344 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_vpc_net
+short_description: configure AWS virtual private clouds
+description:
+    - Create or terminates AWS virtual private clouds.  This module has a dependency on python-boto.
+version_added: "2.0"
+options:
+  name:
+    description:
+      - The name to give your VPC. This is used in combination with the cidr_block paramater to determine if a VPC already exists.
+    required: yes
+  cidr_block:
+    description:
+      - The CIDR of the VPC
+    required: yes
+    aliases: []
+  tenancy:
+    description:
+      - Whether to be default or dedicated tenancy. This cannot be changed after the VPC has been created.
+    required: false
+    default: default
+  dns_support:
+    description:
+      - Whether to enable AWS DNS support.
+    required: false
+    default: true
+  dns_hostnames:
+    description:
+      - Whether to enable AWS hostname support.
+    required: false
+    default: true
+  dhcp_id:
+    description:
+      - the id of the DHCP options to use for this vpc
+    default: null
+    required: false
+  tags:
+    description:
+      - The tags you want attached to the VPC. This is independent of the name value, note if you pass a 'Name' key it would override the Name of the VPC if it's different.
+    default: None
+    required: false
+  state:
+    description:
+      - The state of the VPC. Either absent or present.
+    default: present
+    required: false
+  multi_ok:
+    description:
+      - By default the module will not create another VPC if there is another VPC with the same name and CIDR block. Specify this as true if you want duplicate VPCs created.
+    default: false
+    required: false
+author: Jonathan Davila
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Create a VPC with dedicate tenancy and a couple of tags
+
+- ec2_vpc:
+    name: Module_dev2
+    cidr_block: 170.10.0.0/16
+    region: us-east-1
+    tags:
+      new_vpc: ec2_vpc_module
+      this: works22
+    tenancy: dedicated
+
+'''
+
+
+import time
+import sys
+
+try:
+    import boto
+    import boto.ec2
+    import boto.vpc
+    from boto.exception import EC2ResponseError
+
+    HAS_BOTO=True
+except ImportError:
+    HAS_BOTO=False
+
+def boto_exception(err):
+    '''generic error message handler'''
+    if hasattr(err, 'error_message'):
+        error = err.error_message
+    elif hasattr(err, 'message'):
+        error = err.message
+    else:
+        error = '%s: %s' % (Exception, err)
+
+    return error
+
+def vpc_exists(module, vpc, name, cidr_block, multi):
+    """Returns True or False in regards to the existance of a VPC. When supplied
+    with a CIDR, it will check for matching tags to determine if it is a match
+    otherwise it will assume the VPC does not exist and thus return false.
+    """
+    exists=False
+    matched_vpc=None
+
+    try:
+        matching_vpcs=vpc.get_all_vpcs(filters={'tag:Name' : name, 'cidr-block' : cidr_block})
+    except Exception, e:
+        e_msg=boto_exception(e)
+        module.fail_json(msg=e_msg)
+
+    if len(matching_vpcs) == 1 and not multi:
+        exists=True
+        matched_vpc=str(matching_vpcs).split(':')[1].split(']')[0]
+    elif len(matching_vpcs) > 1 and not multi:
+        module.fail_json(msg='Currently there are %d VPCs that have the same name and '
+                             'CIDR block you specified. If you would like to create '
+                             'the VPC anyways please pass True to the multi_ok param.' % len(matching_vpcs))
+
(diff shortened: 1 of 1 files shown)


bcoca:
so im thinking we can rename this ec2_vpc_net and the other 2 will be ec2_vpc_subnet and ec2_vpc_routes

juliedavila:
done

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,338 @@
...
+short_description: configure AWS virtual private clouds
+description:
+    - Create or terminates AWS virtual private clouds.  This module has a dependency on python-boto.
+version_added: "2.0"
+options:
+  name:
+    description:
+      - The name to give your VPC. This is used in combination with the cidr_block paramater to determine if a VPC already exists.
+    required: yes
+    default: null
```
bcoca: when required, there is no default

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,338 @@
...
+    aliases: []
```
bcoca: empty aliases should be omitted

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,338 @@
...
+      - Whether the VPC should have an IGW. If set to true, the module will try to check if any idle IGWs are availble otherwiseit will attempt to create a new igw and attach it to the VPC.
+    default: present
+    required: false
+    aliases: []
+  tags:
+    description:
+      - The tags you want attached to the VPC. This is independent of the name value, note if you pass a 'Name' key it would override the Name of the VPC if it's different.
+    default: null
+    required: false
+    aliases: null
```
bcoca: aliases null?

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,338 @@
...
+
+try:
+    import boto
+    import boto.ec2
+    import boto.vpc
+    from boto.exception import EC2ResponseError
+
+    HAS_BOTO = True
+except ImportError:
+    sys.exit(msg='Boto is required')
```
bcoca: don't do sys.exit here, set HAS_BOTO to false and use module's fail_json in main

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,338 @@
...
+    elif len(matching_vpcs) > 1 and not multi:
+        module.fail_json(msg='Currently there are %d VPCs that have the same name and '
+                             'CIDR block you specified. If you would like to create '
+                             'the VPC anyways please pass True to the multi_ok param.' % len(matching_vpcs))
+
+    return exists, matched_vpc
+
+def vpc_needs_update(module, vpc, vpc_id, dns_support, dns_hostnames, igw_state, dhcp_id, tags):
+
+    # This returns True or False. Intended to run after vpc_exists.
```
bcoca: use docstrings """ for multiline comments on a function/method's description

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,338 @@
...
+    sys.exit(msg='Boto is required')
+
+
+def vpc_exists(module, vpc, name, cidr_block, multi):
+    """Returns True or False in regards to the existance of a VPC. When supplied
+    with a CIDR, it will check for matching tags to determine if it is a match
+    otherwise it will assume the VPC does not exist and thus return false. """
+    exists=False
+    matched_vpc=None
+    matching_vpcs=vpc.get_all_vpcs(filters={'tag:Name' : name, 'cidr-block' : cidr_block})
```
bcoca: could this raise exceptions? if so they should be trapped and dealt with.
idem for other calls below

juliedavila: If there is an exception here it would have been raised earlier before this call. When nothing matches it returns an empty list. Def need to do some other captures though.

juliedavila:
Updated. Still need to add the deletion logic. Via boto all IGWs, RTs and subnets have to be deleted before the VPC can go.

bcoca:
considering they will each have their own modules, I think it is OK to just return a user friendly error that asks them to do so using the other modules.

juliedavila:
Ok it's been updated, should be good to for deletion. The only thing needed for deletion was to add logic to remove the internet gateway. The other stuff should/will be functions of the other vpc modules.

wimnat:
Is this, and the other vpc modules mentioned, aimed for v2 release?

juliedavila:
@wimnat  yes.

juliedavila:
Refactored out IGW functionality to leave it in place for https://github.com/ansible/ansible-modules-extras/issues/607

gregdek:
@defionscode is this basically done, then? All the refactoring you wanted to do is complete with this latest patch, or no?

juliedavila:
@gregdek yes, good to go AFAIK, but I don't like to +1 my own stuff :wink:

bcoca:
@defionscode let me know when we can deprecate the old module.

wimnat:
@bcoca @defionscode I know this has already been merged but I only got around to my review today and i have some questions.  I've commented inline.

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,344 @@
...
+    exists=False
+    matched_vpc=None
+
+    try:
+        matching_vpcs=vpc.get_all_vpcs(filters={'tag:Name' : name, 'cidr-block' : cidr_block})
+    except Exception, e:
+        e_msg=boto_exception(e)
+        module.fail_json(msg=e_msg)
+
+    if len(matching_vpcs) == 1 and not multi:
```
wimnat, line 126: What happens if len == 1 and multi is specified?  Surely, if len == 1 then we don't care about multi so this should just be "if len(matching_vpcs) == 1:"

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,344 @@
...
+        exists=True
+        matched_vpc=str(matching_vpcs).split(':')[1].split(']')[0]
+    elif len(matching_vpcs) > 1 and not multi:
+        module.fail_json(msg='Currently there are %d VPCs that have the same name and '
+                             'CIDR block you specified. If you would like to create '
+                             'the VPC anyways please pass True to the multi_ok param.' % len(matching_vpcs))
+
```
wimnat, line 133: What happens if len > 1 and multi is true?  How do you decide which vpc you are going to use for comparison?

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,344 @@
...
+    description:
+      - Whether to enable AWS DNS support.
+    required: false
+    default: true
+  dns_hostnames:
+    description:
+      - Whether to enable AWS hostname support.
+    required: false
+    default: true
+  dhcp_id:
```
wimnat, line 49: This is defined as 'dhcp_opts_id' in module params

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,344 @@
...
+                e_msg=boto_exception(e)
+                module.fail_json(msg=e_msg)
+
+            update_dhcp, update_tags=vpc_needs_update(module, vpc, vpc_id, dns_support, dns_hostnames, dhcp_id, tags)
+
+            if update_dhcp:
+                new_dhcp_opts=update_dhcp_opts(module, vpc, vpc_id, dhcp_id)
+            if update_tags:
+                new_tags=update_vpc_tags(module, vpc, vpc_id, tags, name)
+                module.exit_json(changed=True, name=name, vpc_id=vpc_id, dhcp_options=new_dhcp_opts, tags=new_tags)
```
wimnat, line 327: If update_tags is not set then exit_json is never called

cloud/amazon/ec2_vpc_net.py
```diff
@@ -0,0 +1,344 @@
...
+    elif state == 'absent':
+        if already_exists:
+            changed=True
+            try:
+                vpc.delete_vpc(vpc_id)
+                module.exit_json(changed=changed, vpc_id=vpc_id)
+            except Exception, e:
+                e_msg=boto_exception(e)
+                module.fail_json(msg="%s. You may want to use the ec2_vpc_subnet, ec2_vpc_igw, "
+                "and/or ec2_vpc_rt modules to ensure the other components are absent." % e_msg)
```
wimnat, line 337: ec2_vpc_rt should be ec2_vpc_route_table

wimnat:
@defionscode I made a few changes to address the above as well as some other issues including returning all vpc object parameters such as instance_tenancy and state.  Can you take a look and let me know what you think? https://github.com/wimnat/ansible-modules-core/commit/f2f7e10f3d3d2bf5c5bac49486c3ceec08c0abba

gregdek:
Since this is already merged, perhaps these should be a separate PR?

wimnat:
@gregdek as its quite a big change, I'd like @bcoca and @defionscode to have a look see here first :)

bcoca:
@winmat, submit pr that is how we examine these things, we normally filter out comments on closed tickets or we would never have time to look at open ones.

juliedavila:
@winmat, a very brief glance take, looks fine, but def open a PR and mention me so i can do some testing.

PR #1137, opened 2015-04-14
Elasticache subnet group management

timmahoney:
This module allows creation of an Elasticache Subnet Group, which is necessary for creation of Elasticache Clusters inside a VPC.


worldofchris:
Great work.  Just used this to create an Elasticache node in a VPC to use with Redis.

bennojoy:
@timmahoney  can you please sqash the commits to a single one, and change the version added to 2.0

timmahoney:
All set - thanks @bennojoy

PR #1140, opened 2015-04-15
Add support for specifying unique hosted zone identifier

Pallokala:
## Add support for defining unique hosted zone identifier when modifying route53 hosted zone

With this change, the users get the possibility for defining with Hosted Zone ID which hosted zone they want to modify. This is important when there exists multiple identically named private hosted zones in different VPCs.

The change is (should be) fully backwards compatible.

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -41,7 +41,6 @@ options:
     required: false
     version_added: 2.0
     default: null
-    aliases: []
   record:
     description:
       - The full DNS record to create or delete
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -39,6 +39,7 @@ options:
     description:
       - The Hosted Zone ID of the DNS zone to modify
     required: false
+    version_added: 2.0
     default: null
     aliases: []
   record:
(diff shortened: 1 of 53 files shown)


Pallokala:
Are there any news for this? It would really benefit me if this was merged soonish.

gregdek:
@bpennypacker can you review this PR? If you give it +1 we will tag as "P1_reviewed" for quicker inclusion.

bpennypacker:
+1 looks good!

gregdek:
Thanks @bpennypacker

cloud/amazon/route53.py
```diff
@@ -35,6 +35,12 @@
     required: true
     default: null
     aliases: []
+  hosted_zone_id:
```
bcoca, line 38: needs version added and you can drop empty aliases

gregdek:
@Pallokala back to you -- if you can add version (2.0) and drop empty aliases, we should be good to go.

gregdek:
Oh, and apparently this also needs to be rebased now... quite a bit going on in this module...

Pallokala:
I hopefully rebased this correctly and modified the patch to fit with the latest version.

PR #1143, opened 2015-04-15
Hkariti ec2 snapshot idempotent

robjailall:
Via pycon sprint. Modified https://github.com/ansible/ansible-modules-core/pull/76 to be more testable and wrote tests here:

https://github.com/ansible/ansible/pull/10730


abadger:
Gave some feedback about naming of the parameter and style cleanups at pycon.

@robjailall is also going to look into the speed hit that was reported in #76.

@defionscode   How is this looking to you?  Is speed the only thing that you see?

robjailall:
@defionscode

Regarding the speed, since wait is on by default, it will wait until snapshot.status is 'completed'

I ran this using test-module, and it simply took long for snapshot.status to go from 'pending' to 'completed'. In the EC2 console, the snapshot was created immediately and status set to 'completed'.

The internet suggests that it might take long for the latest snapshot status to become available in the API.

I confirmed that if you use the aws command line tool, it takes a long time to update the status there, too, even though the aws console says it is completed:

aws ec2 describe-snapshots --snapshot-ids [snapid]

hkariti:
Hey thanks for that. Haven't had time to poke this.  I'll close my orig PR

tzz:
Sweet, this would be a nice addition.

Unfortunately it does not handle daily, weekly, etc. snapshots.  That functionality is built right into boto so I just do this at the end of my snapshot creation:

```
- shell:
    cmd: python -c 'import boto; ec2_conn = boto.connect_ec2(); ec2_conn.trim_snapshots(8, 7, 4, True)'
```

I hope something like it can be added to the module as well.  Maybe call it `retain_policy` and just have four keys underneath representing each of the `trim_snapshots()` parameters.  It seems like a trivial feature that would benefit many.

gregdek:
@willthames Please review this PR to make sure it adheres to the following guidelines:

cloud/amazon/ec2_snapshot.py
```diff
@@ -124,13 +258,19 @@ def main():
...
             snapshot_tags = dict(type='dict', default=dict()),
             state = dict(choices=['absent','present'], default='present'),
         )
     )
     module = AnsibleModule(argument_spec=argument_spec)
+    return module
+
+
+def main():
+    module = create_snapshot_ansible_module()
```
willthames, line 273: This seems unnecessary.

cloud/amazon/ec2_snapshot.py
```diff
@@ -115,7 +128,128 @@
     HAS_BOTO = False
 
 
-def main():
+# Find the most recent snapshot
+def _get_snapshot_starttime(snap):
```
willthames, line 132: I'm not convinced about prefixing methods with underscores in modules. I know it means they're private methods, but nothing really imports modules anyway.

I'd also argue that given how specific create_snapshot is (it takes a module argument that must implement fail_json and exit_json) there's no reason for the difference in privacy of the methods.

gregdek:
@willthames are these blockers in your opinion?

juliedavila:
@gregdek I don't think those are blockers, but I agree with @willthames. I give it a +1 should @robjailall not have the time to be able to make the updates.

@tzz I think it would be good to get this through as is, then if you could, open a GH issue for that feature added.

willthames:
@gregdek probably more just an indication that we don't have a one true style guide for modules. Actually CODING_GUIDELINES.md says that functions should be prefixed with an underscore, but on that basis I'm not sure how function is being defined (surely we don't prefix everything that returns a value with an underscore)

Anyway, not a blocker, just an expression of preference.

abadger:
Rebased and merged here:

a91db23b564df8a7d3024558b7f9799e4bc91494
1f447ad5ba15f841473ada14cfbbac3ba6a06a11
d10f3f7a7e337059665a6357d73a3a4872f5bac7

github is still displaying the merge button but as far as I can tell, the code has been committed now.  Let me know if I've missed something by accident.

@hkariti and @robjailall thanks for the code, @willthames and @defionscode thanks for the review.

# Merged

Hi!

issue is resolved.

Thank you!

robjailall:
Thanks!

PR #1147, opened 2015-04-16
Support encrypted volumes in ec2_lc

bobby:
I didn't do any checking for whether boto supports encrypted volumes, as is done in [ec2_vol.py](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/ec2_vol.py#L292).  Is that necessary now, or is that check a vestige of a long-forgotten past?

If we do need the check, we should probably pull that check out into a little util file, and load it into both modules, but I don't know the Ansible idiom for doing so.


bobby:
Turns out AWS doesn't support encrypted volumes (other than via Snapshot ID) in Launch Configurations (cf. http://stackoverflow.com/questions/28909764/aws-auto-scaling-launch-configuration-encrypted-ebs-cloud-formation-example).  Closing this request.

PR #1149, opened 2015-04-16
Add support for 'update' parameter to hg module

catlee:
This adds support for an 'update' parameter similar to git's. Set it to false to prevent ansible from trying to update an existing checkout.

### DIFF
--- a/source_control/hg.py
+++ b/source_control/hg.py
@@ -65,6 +65,13 @@ options:
         required: false
         default: "no"
         choices: [ "yes", "no" ]
+    update:
+        required: false
+        default: "yes"
+        choices: [ "yes", "no" ]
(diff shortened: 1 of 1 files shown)


gregdek:
@yeukhon Please review this PR to make sure it adheres to the following guidelines:

yeukhon:
@gregdek  this looks good so far. I will run a unit test before EoD. Only thing to add is probably an example (but I think this module is simple enough doesn't really warrant that many examples).

cc: @catlee

gregdek:
Thanks @yeukhon!

yeukhon:
@gregdek  sorry I was too busy these days. Looks good to me. We can strip community review off.

PR #1153, opened 2015-04-17
make migrate command idempotent with django built-in migrations

Ernest0x:
Django since version 1.7 has built-in migrations, so no need to have south installed with recent django versions. The 'migrate' command works with built-in migrations without any change, but the output is different than the output produced by south, which breaks idempotence. This commit fixes this.


gregdek:
@tastychutney Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@tasychutney a friendly ping to review this PR if you have time. Thanks.

cchurch:
:+1: Thanks!

PR #1157, opened 2015-04-18
Remove mention of custom error message

jeffwidman:
Although mentioned in the description, the assert module doesn't appear to support custom error messages. So removed description from docs.


jeffwidman:
Actually, my preferred solution would probably be for `assert` module to support custom error messages,  and deprecate the `fail` module... no need for two modules that do almost the same thing, just add a `state` switch that can be set to 'always_fail'. Maybe it's possible to alias `fail` to `assert`?

But that's a fairly big change, and there may be distinct use cases for both that I'm not understanding...

bcoca:
assert adds it's own list of conditionals on top of the when: clause, this and the expressed syntax are the only real differences. I would consider it a more complex/nuanced version of fail not a simpler one, which is why I'm closing this ticket.

jeffwidman:
Does the assert module support a custom message on failure?

PR #130, opened 2014-10-03
Mysql password hash support

trevoro:
As referenced in ansible/ansible#6049 and ansible/ansible#5633, this feature branch adds the optional `pasword_hash` setting to the mysql database ansible module.

PR #1174, opened 2015-04-20
allow for passing pass_hash instead of password in mysql_user module

kaHaleMaKai:
For reasons of security, I'd like to use password hashes instead of storing plain-text passwords for mysql user management.


bcoca:
I was thinking it is better to use something like the postgresql_user module does, have a encrypted: true|false(default) parameter that signifies if the password field is encrypted or not

kaHaleMaKai:
yes, makes perfectly sense
I will make the changes and file a new pull request

bcoca:
you can just update this one, this is one of the few instances where git push --force is actually nicer than the alternatives

kaHaleMaKai:
sorry, didn't realize that

kaHaleMaKai:
I haven't yet changed anything

kaHaleMaKai:
sorry for my clumsiness, first time pull request ;-)

clok:
+1 Any updates on when this feature will be merged?

bcoca:
@clok code still needs to be updated, as it stands now it cannot be merged

gregdek:
@the-lars Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

kaHaleMaKai:
I updated to devel branch.
I couldn't test locally, as the test-module script threw errors (https://github.com/ansible/ansible/issues/11180). I hope the CI tests are enough.
Back in April before the first pull request, all tests passed and my changes don't seem to affect any other code.

gregdek:
Thanks @the-lars. @banyek can you review?

clok:
@gregdek doesn't this collide with the work in https://github.com/ansible/ansible-modules-core/pull/1581 ?

bcoca:
hmm @abadger you reviewed 1581, can you take a look here and get the best of both worlds?

database/mysql/mysql_user.py
```diff
@@ -211,14 +221,17 @@ def user_exists(cursor, user, host):
     count = cursor.fetchone()
     return count[0] > 0
 
-def user_add(cursor, user, host, password, new_priv):
-    cursor.execute("CREATE USER %s@%s IDENTIFIED BY %s", (user,host,password))
+def user_add(cursor, user, host, password, encrypted, new_priv):
+    if not encrypted:
+        cursor.execute("CREATE USER %s@%s IDENTIFIED BY %s", (user,host,password))
+    else:
+        cursor.execute("CREATE USER %s@%s IDENTIFIED BY PASSWORD %s", (user,host,password))
```
clok, line 228: Some of the core feedback on #1581 and #130 given by @morgo was concerning MySQL 5.7 compatibility. The `PASSWORD` function will be deprecated and the methods for updating hashed passwords have changed. Also, Line 252 is affected by this. This has been accounted for in #1581 

I have tested thoroughly on VMs runnig MySQL 5.5, 5.6 and 5.7 Beta.

The hold up at the moment for #1581 is putting together Integration Tests. The current test suite in not functioning on a clean `devel` build. Looks like @the-lars ran into this as well.

gregdek:
Ah, yes, I guess it does collide with #1581. 

Putting into core_review for the team to sort out.

abadger:
yeah, @morgo, do you see anything here that should be applied on #1581 ?  My brief look (and not necessarily knowing all the intricacies of mysql password handling ;-) is that this code doesn't handles a subset of the cornercases that #1581 does now.

morgo:
Yes, this patch will not work for 5.7. It uses `SET PASSWORD FOR %s@%s = %s` expecting it to be treated as a hash, but 5.7 will treat this as a password.

Also, from a feature point of view I would prefer that this be referred to more clearly as "supporting the `mysql_native_password` hash", since MySQL already supports a stronger `sha256` hash, and it is reasonable to anticipate may become the default at some stage.

(Officially the MySQL team does not like it's internal hash format being copied around as an 'format' so to speak, but I acknowledge it is useful, particularly in the case of pre-5.7 where there is not a full `ALTER USER` command.)

bcoca:
so to make this fully compatible would we need to try 'ALTER USER' and if it fails fallback to this method?

morgo:
@bcoca kind of. The new command is:
`ALTER USER %s@%s IDENTIFIED WITH mysql_native_password AS %s`

(The other patch has a version detection system rather than try and fall-back.)

I have a [blog post here](http://www.tocker.ca/2014/09/05/proposed-changes-to-user-management-in-mysql-5-7.html) explaining "why" the previous password commands are a problem.  The short version is that the `PASSWORD()` function can not clearly determine the authentication plugin for the user.

clok:
#1581 includes the features that this PR is adding plus the added 5.7 support for passing in a `mysql_native_password` hash.

morgo:
BTW, @bcoca your comment https://github.com/ansible/ansible-modules-core/pull/1174#issuecomment-94485064 is what I think makes sense to _officially_ recommend:
- Store passwords (not hashes) in ansible
- Turn on encryption

This gets out of the business of copying around a `hash`, which is really just an internal state representation for MySQL.  Copying around hashes is not really secure.  We are supposed to assume if the attacker gets the hash, they can eventually break the password.

bcoca:
hashes are only more secure if you change passwords often, hopefully within the margin of the brute force being able to get the clear text version. Also if we never reuse passwords for more than one system.

yes, I know, HUGE IFs.

morgo:
@bcoca Perhaps, but lets not get into that.  What we do know is that the `mysql_native_password` hash is `sha1` based, which is no longer considered 'strong'.

abadger:
Okay,  If no one objects, I'll close this in favor of https://github.com/ansible/ansible-modules-core/pull/1581  as I think that that pull request handles a superset of the functionality of this one.

kaHaleMaKai:
true that!

PR #1177, opened 2015-04-20
Remove check on template and template_url

### LINKED ISSUE #10792: CloudFormation module requires "template" parameter even when doing state absent.
Ansible version = 1.9.0.1

Very simple task to delete a CloudFormation stack is throwing up unless I provide a valid path to a file for 'template' parameter. The documentation says it is not required and the example shows it as missing from the 'absent' example.

Example task:

``` yaml
- name: delete cloudformation stack
  cloudformation:
    region: "{{ stack_region }}"
    stack_name: "{{ stack_name }}"
    state: absent
#    template: "{{ stack_templates_root }}/fake.json"
```

Generates the following:

> TASK: [delete_cloudformation | delete cloudformation stack] ******************\*  
> failed: [localhost] => {"failed": true}  
> msg: missing required arguments: template                                                                                                                                                                 
> 
> FATAL: all hosts have already failed -- aborting

If you uncomment the template parameter and point it to any file (not even a valid CF template) it works fine.

mscherer:
This part is already checked later, and prevent from erasing a
stack, fixes https://github.com/ansible/ansible/issues/10792

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -230,9 +230,6 @@ def main():
     if not HAS_BOTO:
         module.fail_json(msg='boto required for this module')
 
-    if module.params['template'] is None and module.params['template_url'] is None:
-        module.fail_json(msg='Either template or template_url expected')
-
     state = module.params['state']
     stack_name = module.params['stack_name']
 


gregdek:
@jsmartin Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@jsmartin friendly ping if you have time...

gregdek:
@jsmartin ping for review?

jsmartin:
This patch needs a bit of work -- it should only ignore the requirements of those files if `state==absent`. A better implementation of it is here: https://github.com/ansible/ansible-modules-core/pull/334/files.  It used to be in the code, but somehow got reverted.  See https://github.com/ansible/ansible/issues/10792 for more details.

gregdek:
@mscherer Please make the suggested revisions, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

gregdek:
@mscherer A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

mscherer:
So I did look quite hard at the code, but I am not sure to understand what @jsmartin mean. 

The current code is doing the same test twice:

Once here: https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/cloudformation.py#L261

And then 2 lines after again, but this time, taking in account that the state is "present":
https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/cloudformation.py#L267

And there is only 2 assignations between the 2, so unless I am missing something subtle, the first test is a old duplicate and should be removed.

Given that the state is either present or absent, I think the current removal is what is needed, because the code is present twice, which is unneeded.

jsmartin:
This appears to be a duplicate of #2826 (which has been merged), should probably close this.

mscherer:
Indeed, already merged, closing it

PR #1179, opened 2015-04-21
Allow custom methods in uri module

hver:
I want to use Ansible in a task to refresh Varnish caches. To do so, I have to send a _REFRESH_ request. Unfortunately it is currently not possible, because the _uri_ module restricts the methods to `[GET,POST,PUT,HEAD,DELETE,OPTIONS,PATCH]`.
I would like to propose to remove this restriction and allow any string to be use as a method. I looked into the [RFC2616](https://datatracker.ietf.org/doc/rfc2616/) if there is a list of restricted methods to allow, but it tells only that 

> The methods GET and HEAD MUST be supported by all general-purpose servers. 

So I think there is no reason to restrict the methods to this list. (Otherwise we should rename the _uri_ module to something like _rest_api_)

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -73,7 +73,6 @@ options:
     description:
       - The HTTP method of the request or response.
     required: false
-    choices: [ "GET", "POST", "PUT", "HEAD", "DELETE", "OPTIONS", "PATCH" ]
     default: "GET"
   return_content:
     description:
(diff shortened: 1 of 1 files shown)


gregdek:
@romeotheriault Please review this PR to make sure it adheres to the following guidelines:

bcoca:
the choices are there more to guide users than to restrict them, if the list is incomplete we should add to it, not remove it.

rrtj3:
When I added the list it was mostly to protect people from misspelling methods. I would say adding REFRESH (and any other method you need, like bcoca suggests) to the choices is safer than just removing them all. As far as I know this is the first I've heard of a method that was needed not being available, so I don't think it's a super common thing that warrants removing the choices.

hver:
I'm thinking about WebDAV, it uses a lot of own HTTP methods. Varnish uses REFRESH, PURGE and BAN. But if no one requests it, maybe it is not so important to be able to use any word as method name. 

I have updated the commit and added 'TRACE' and 'CONNECT' from RFC2616 and 'REFRESH'

PR #1180, opened 2015-04-21
Adding openstack multiregion support for floating ip module.

sky-shiny:


sky-shiny:
Imo: not a feature.  If you have > 1 region and request a floating ip with this module, requests fail with an "ambiguous endpoint" error.

emonty:
I agree - this is a bug, so I'm fine with it going in to the soon-to-be deprecated modules.

PR #1182, opened 2015-04-21
Validate openstack floating ip against provided network name.

sky-shiny:
Currently this module assumes that floating ips for a host are on the correct network.  
This patch validates the network against the network provided by the user.

### DIFF
--- a/cloud/openstack/quantum_floating_ip.py
+++ b/cloud/openstack/quantum_floating_ip.py
@@ -171,7 +171,7 @@ def _get_port_info(neutron, module, instance_id, internal_network_name=None):
         return None, None
     return fixed_ip_address, port_id
 
-def _get_floating_ip(module, neutron, fixed_ip_address):
+def _get_floating_ip(module, neutron, fixed_ip_address, network_name):
     kwargs = {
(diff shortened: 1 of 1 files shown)


emonty:
I'm also inclined to view this one as a bugfix - but it's a bit more so than the other - so I'd like to think about it for a bit.

@dguerri could you check and make sure that the logic @sky-shiny has added here is in the shade floating-ip support so that we'll have this for the new modules as well?

dguerri:
It is still a WiP, but yes, `_get_floating_ip` with a network name will be supported by shade when https://review.openstack.org/177036 is merged.
Nevertheless, it would be better to use `add_ip_from_pool(self, server, pools, fixed_address=None)` from shade (where `pool` is a list of neutron external network names or nova floatingIP pools) in order to support both neutron and nova-network while hiding all the machinery about that.

sky-shiny:
Discussions around work of greater scope than is handled here, is better served by a discussion in its own pull request.

Is there anything preventing us from moving forward with _this_ pull request?  I am happy to make any necessary modifications, however I'd like to avoid scope creep, and am happy with the possibility that this will be deprecated in favor of a more general solution in the near future.

Issue #386, opened 2014-11-24
Set stickiness policy using ec2_elb_lb?


bobobox:
##### ISSUE TYPE

Feature Idea
##### COMPONENT NAME

ec2_elb_lb module
##### ANSIBLE VERSION

N/A
##### SUMMARY

I notice that it's not possible to set a stickiness policy via the ec2_elb_lb module, while this functionality is supported by boto. I was wondering if there was a particular reason for this (is this somehow problematic?) or if it's just waiting for someone to add it in?

This is kind of an important parameter for a lot of load balancer uses.

haad:
+1 if someone is not working on this I will have a look at it.

olvesh:
Was just looking for this feature.. :+1:  here as well (not up to ansible module hacking yet though, but will gladly test).

tomfotherby:
I think this issue can be closed now because `stickiness` is in version 2 of Ansible - see http://docs.ansible.com/ansible/ec2_elb_lb_module.html

ansibot:
@bobobox we are unable to validate that "


##### " is an existing ansible module.

We are expecting a component name that matches something that can be found in:
* http://docs.ansible.com/ansible/list_of_all_modules.html

Please revise the component name in the description.

[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

bobobox:
Closing as this feature has been implemented in Ansible 2.0

PR #1190, opened 2015-04-23
Add ability to define policies for a given ELB.

haad:
This commit will add ability for ec2_elb_lb module to create policies for an ELB instance.


haad:
This fixes following issues: #447, #386

PR #1196, opened 2015-04-24
avoid exception when boto is not available

bcoca:
shoudl fix #10831


abadger:
+1

I've also been thinking it would be nice to come up with a best practices
doc that expands on the writing modules doc.  Things like this could be
part of that. (Someday, when more time appears....)

-Toshio

Issue #1197, opened 2015-04-24
S3 module  overwrite=true should download files only if they are changed


jblackburn22:
Issue Type:

Bug Report

Ansible Version:

ansible 1.9 

Environment:

Mac OSX (Yosemite) -> Amazon Linux 2014.09 or Fedora 20

Summary:

We are using the s3 module to download artifacts for installation. Once we have installed something, we want to re-download the file only if it has been changed.  

Steps To Reproduce:
1. Upload a file to s3. 
2. Use the s3 module to download the file. 
3. Run the task again.  
4. Change the file in s3. 
5. Run the task again.

Expected Results:

If the overwrite=True, then the expected results are

Run 1:
The first time the task runs, the file is downloaded and returns with a "changed" status since it doesn't exist.

Run 2:
The second time the task runs, the file is NOT downloaded and returns with an "ok" status since the check sums for the local and remote files matches.

Run 3:
The 3rd time the task runs, the file is downloaded and returns with a "changed" status since the file has been changed.

Actual Results:

The task downloads the file every time.

bennojoy:
Hi!

Thanks very much for your submission to Ansible. It sincerely means a lot to us.

We're not sure this is a bug, and we don't mean for this to be confrontational. Let's explain what we're thinking:

The default behaviour of the module is to keep the destination file intact even if the checksum's dont match, the only way to overrride this behaviour is to have the "overwrite" variable set to true and this ensures that the file is downloaded and overwritten.

As such, we're going to close this ticket. However, we're open to being corrected, should you wish to discuss. You can stop by one of our two mailing lists to talk about this and we might be persuaded otherwise.

https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code
Comments on closed tickets aren't something we monitor, so if you do disagree with this, a mailing list thread is probably appropriate.

Thank you once again for this and your interest in Ansible!

bennojoy:
In further discussion with the team, we think we should change the behaviour of the module a bit by adding the following options to the 'overwrite ' parameter 
['always' (true), 'never' (false), different (overwrite only when cheksums are different)]
this would help keep the backward compatibly yet add the feature where the module wont download the file if the source and the dest checksums match and overwrite is set to true

bennojoy:
@jblackburn22  can you please give #1326 a try

PR #1198, opened 2015-04-24
issue #1197: Only download s3 object if changed

jblackburn22:


bcoca:
this breaks the overwrite feature, if you don't want the file overwritten set overwrite to false

PR #1205, opened 2015-04-25
Updated documentation

camilonova:
So examples work with ansible > 1.9

Otherwise you will get this error:

```
ERROR: module is not a legal parameter of an Ansible Play
```

I fixed this from:

http://stackoverflow.com/questions/29271960/error-local-action-is-not-a-legal-parameter-at-this-level-in-an-ansible-playboo

### DIFF
--- a/cloud/linode/linode.py
+++ b/cloud/linode/linode.py
@@ -95,60 +95,70 @@ notes:
 '''
 
 EXAMPLES = '''
-# Create a server
-- local_action:
-     module: linode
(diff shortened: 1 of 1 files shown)


camilonova:
cc @bcoca

bcoca:
Eamples are just tasks, modules do not normally show a full playbook, it is not for modules to give you a 'full implementation example' just one about the usage of the module itself. For full examples you should read the full docs.

PR #1212, opened 2015-04-27
ec2_group: select own group if the name matches

zimbatm:
This fixes an issue where multiple VPC have the same group name and
the one from the other VPC is selected.

Related to this issue, it would be nice if the vpc_id groups where prioritized as well but it would require more changes to the module.

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -223,7 +223,12 @@ def main():
     groups = {}
     for curGroup in ec2.get_all_security_groups():
         groups[curGroup.id] = curGroup
-        groups[curGroup.name] = curGroup
+        if curGroup.name in groups:
+            # Prioritise groups from the current VPC
+            if vpc_id is None or curGroup.vpc_id == vpc_id:
+                groups[curGroup.name] = curGroup
+        else:
+            groups[curGroup.name] = curGroup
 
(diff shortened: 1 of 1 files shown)


bcoca:
i might not be reading this correctly, but won't there always be a single group_name? it seems that if we need this partitioned by  vpc before we look at group_name

zimbatm:
The issue I am facing is that the following config:

``` yaml
ec2_group:
  state: present
  name: default
  region: "{{ aws_region }}"
  description: "Default group for {{ cluster }}"
  vpc_id: "{{ vpc.vpc_id }}"
  rules:
    # Breaking with 
    - proto: all
      group_name: default
```

Yields the following error:

```
<Response><Errors><Error><Code>InvalidGroup.NotFound</Code><Message>You have specified two resources that belong to different networks.</Message></Error></Errors><RequestID>32358200-233f-4aed-b437-f555431f8691</RequestID></Response>
```

The issue is that each VPC has a security-group with the same "default" name and that ec2_group is trying to associate the wrong one in the rules.

bennojoy:
Hi!

https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code
Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #1223, opened 2015-04-29
Added ec2 support for AWS access token

tjarvstrand:
This is the same pattern as the cloudformation module


tjarvstrand:
Sorry, this is a duplicate of #1206

PR #1225, opened 2015-04-29
use the right way to unescape string in lineinfile module

jaypei:
Reference https://github.com/ansible/ansible/issues/10864

### DIFF
--- a/files/lineinfile.py
+++ b/files/lineinfile.py
@@ -370,25 +370,13 @@ def main():
 
         line = params['line']
 
-        # The safe_eval call will remove some quoting, but not others,
-        # so we need to know if we should specifically unquote it.
-        should_unquote = not is_quoted(line)
-
-        # always add one layer of quotes
-        line = "'%s'" % line
-
         # Replace escape sequences like '\n' while being sure 
         # not to replace octal escape sequences (\ooo) since they
         # match the backref syntax.
         if backrefs:
             line = re.sub(r'(\\[0-9]{1,3})', r'\\\1', line)
-        line = module.safe_eval(line)
 
-        # Now remove quotes around the string, if needed after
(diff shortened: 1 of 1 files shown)


abadger:
This doesn't look at all right to me.  string.decode() is for changing between byte str and unicode type, not for escaping and unescaping quotes.  pretty sure that this code will throw unicodeerrors given the right set of input.

abadger:
Hmm... reading docs it looks like python2 does have codecs that are for str-str transformation and string_escape is one of those.  so maybe this could work.  we'd have to make sure that the input is a str before running .decode() on it otherwise we might throw a unicodeerror.

In v2, we can fix this by removing all of this logic as v2 does not escape the newline before sending it to the modules. (v1 sometimes does and sometimes doesn't)

abadger:
Okay, I think this is probably the best we can do in the v1 tree:

``` python
import codecs
[...]
if isinstance(line, unicode):
    line = line.encode('utf-8')
line = codecs.escape_decode(line)
```

It's the same logic as you have but slightly more future-proof. (Not much... string_escape is gone from python3 so that's known un-future-proof code.  codecs.escape_(de|en)code() is undocumented so it works for now but might not work in the future).

I'll merge this into the v1 tree, merge it into the v2_command branch, and then remove all of this escaping logic in v2.  (Since v2 won't have the problem with escape/unescape that lead to this.)

abadger:
merged, fixed up, and removed altogether on the v2 branch.  Thanks!

Note that this won't go into 1.9.x as it is a potentially big code change if we've missed any cornercases.  It could go into a 1.10 but we'll probably release 2.0 before/instead of 1.10.

# Merged

Hi!

issue is resolved.

Thank you!

Issue #1156, opened 2015-04-17
route53: incorrectly handles string TTL


moorereason:
##### Issue Type:

Bug Report (with proposed fix below)
##### Ansible Version:

ansible 1.9.0.1
  configured module search path = None
##### Environment:

CentOS Linux release 7.1.1503 (Core)
ansible-1.9.0.1-1.el7.noarch
python-boto-2.34.0-4.el7.noarch
##### Summary:

Setting the `ttl` value to a string causes the task to always return `changed=True`.  I ran across the issue file doing a `item.ttl | default('3600')` in a task.  I'm in the habit of quoting default values as strings for some reason. :innocent: 
##### Steps To Reproduce:

``` yaml
- route53:
    command: create
    zone: abc.com.
    record: host.abc.com.
    type: A
    ttl: '3600'
    value: 127.0.0.1
    overwrite: yes
```
##### Expected Results:

Expect subsequent plays to return `changed=False`.
##### Actual Results:

Always returns `changed=True`.
##### Proposed Fix

Change [this line](https://github.com/ansible/ansible-modules-core/blob/d960dc1584fb17ebdb720c0c65f7ae155ab612fa/cloud/amazon/route53.py#L212) to:

``` python
ttl_in = int(module.params.get('ttl'))
```

moorereason:
I just realized that using a `default()` filter triggers this bug.  I thought the issue only happened when quoting the ttl as the `default` parameter; however, using filters requires that the `value` is quoted, so the `value` always ends up being a string.

Example that triggers this bug:

``` yaml
- route53:
    command: create
    zone: item.zone
    record: item.record
    type: item.type
    ttl: "{{ item.ttl | default(3600) }}"
    value: item.value
    overwrite: yes
  with_items: records
```

abadger:
Thanks to @xiacio for the fix!

# Closing This Ticket

Hi!

We believe recent commits (likely detailed above) should resolve this question or problem for you.

This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular
issue is resolved.

Thank you!

PR #1231, opened 2015-04-30
Fix route53 TTL validation

ghost:
Fix TTL validation for route53 records by casting the provided TTL into an int (similar to how the TTL fetched from the current route53 record is).

Fix as suggested in the issue: https://github.com/ansible/ansible-modules-core/issues/1156

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -192,7 +192,7 @@ def main():
             command              = dict(choices=['get', 'create', 'delete'], required=True),
             zone                 = dict(required=True),
             record               = dict(required=True),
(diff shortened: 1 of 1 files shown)


abadger:
Instead of fixing this here by casting, fix it on https://github.com/ansible/ansible-modules-core/pull/1231/files#diff-044114001b80d58bf544d58a80af3576L195  by setting type='int'.  That allows validation to be run on the value given to the module.

ghost:
Done, thanks for the suggestion.

abadger:
Thanks for the fix!

# Merged

Hi!

issue is resolved.

Thank you!

PR #1235, opened 2015-04-30
Set default for mysql_user config_file

juliakreger:
The default value set by the module was a value of None for the
config_file parameter, which propogates into the connect method
call overriding the stated default in the method.

Instead, the default should be set with-in the parameter
specification so the file check is not requested to check None.

This corrects issue 1215


abadger:
Looking at a related problem caused by https://github.com/ansible/ansible-modules-core/pull/63 Will review this at the same time.

abadger:
@juliakreger Looks good!  Thanks for fixing that!

# Merged

Hi!

issue is resolved.

Thank you!

PR #1236, opened 2015-04-30
Allow playbook specified login_user and login_password to override config file settings

abadger:
..


abadger:
In https://github.com/ansible/ansible-modules-core/pull/63 the config file takes precedence over the login_user and login_password.  We want those to take precedence instead.

PR #1238, opened 2015-05-01
Also show stdout when validation fails

bertvv:
Some validation commands (e.g. `named-checkconf`) show their error messages on standard output, but Ansible will only show the standard error stream. In that case, the output of the playbook run gives insufficient information to fix the error in the config file:

```
TASK: [bind | Main BIND config file] ****************************************** 
failed: [testbind] => {"failed": true}
msg: failed to validate: rc:1 error:

```

With this PR, standard output will also be shown:

```
TASK: [bind | Main BIND config file] ****************************************** 
failed: [testbind] => {"failed": true}
msg: failed to validate: rc:1
error:
output:/home/vagrant/.ansible/tmp/ansible-tmp-1430473488.4-148430499733954/source:6: unknown option 'pa'

```


bcoca:
all validate= should be consistent, this is used across several modules (move into module shared code?)

bertvv:
Thanks for your reaction, @bcoca 

This change works for templates as well, at least. It's how I tested it before submitting the PR. But indeed, in [files/lineinfile.py](https://github.com/ansible/ansible-modules-core/blob/devel/files/lineinfile.py), there's some code duplication.

I don't know exactly where this code should go then, though.

bcoca:
I'm pondering if I should just move it to the module_utils/basic.py so I don't have to copy/paste all around

bertvv:
I had a go at implementing that. For now, I didn't creat a PR, but you can find the code in:
- https://github.com/bertvv/ansible/tree/wip/validation-output (function `validate_config_file` in `module_utils/basic.py`
- https://github.com/bertvv/ansible-modules-core/tree/wip/validation-output (`file/copy.py` and `file/lineinfile.py` now use this function.

I tested the new behaviour in the `file`, `template` and `lineinfile` modules with a test playbook that I put in this gist: https://gist.github.com/bertvv/8883f7c215560763b5c3

gregdek:
@bertvv Thanks for the PR.  Because this modules is very central to Ansible operation, it can take quite a while for PRs to be accepted. Thanks for your patience.

bertvv:
Fair enough. I appreciate the acknowledgment!

files/copy.py
```diff
@@ -297,7 +297,7 @@ def main():
                     module.fail_json(msg="validate must contain %%s: %s" % (validate))
                 (rc,out,err) = module.run_command(validate % src)
                 if rc != 0:
-                    module.fail_json(msg="failed to validate: rc:%s error:%s" % (rc,err))
+                    module.fail_json(msg="failed to validate: rc:%s\nerror:%s\noutput:%s" % (rc,err,out))
```
bcoca, line 300: tempted to rewrite as:

`module.fail_json(msg="failed to validate", rc=rc, stderr=err, stdout=out)`

and put it all in a nice json response

bertvv, line 300: That makes nothing but sense...

gregdek:
Hey @bertvv, feel free to make the changes that @bcoca recommended.

In the meantime, since this is in core_review and we're pushing the 2.0 release out the door, we're gonna put the "next" release tag on this. Just an acknowledgement that we won't get to it before 2.0 goes out the door, but we'll pick it up afterwards. Thanks!

quinot:
I've just stumbled on the exact same issue with named-checkcon, so count me as an additional vote towards this change!

babadofar:
+1
This would make it a lot easier to fix errors in templates. I'm trying to validate Logstash configurations, and the error message is only printed to stdout.

gregdek:
Thanks @bertvv for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

bertvv:
That makes sense, what with the changes in v2.0. Not sure if I'm going to be able to work on this in the short term, though. I'll close this PR, and open a new one if I get to it.

PR #1242, opened 2015-05-02
apache2_site module for managing apache2 sites in debian based distros

ghost:
This is just a copy of the apache2_module module along with the changes to allow an apache instance to specified present in pull request 1194


sivel:
New modules should be submitted to the extras repo.

Also, this is a duplicate of https://github.com/ansible/ansible-modules-extras/pull/104

ghost:
Ah, my bad. I assumed because the other module was in core, this should be too. Apache2_module should probably belong in extras. If it was I would have spotted existing PR

bcoca:
i will probably deprecate that one and create an apache2 module that does both in extras

PR #1243, opened 2015-05-02
Fix for Issue #10636 in Ansible/ansible (and more) - aws variables broken in ec2.py

andynelson:
Fixes an issue where not all credential types were only being used during creation of the vpc connection.
(introduced in commit bc0c169 when vpc connection was added)
That commit meant that only if aws_access_key & aws_secret_key set in a playbook were used. 

Now allows environment variables or boto profile name.
Also fixes related issue #10638 in ansible/ansible


cloud/amazon/ec2.py
```diff
@@ -1207,17 +1207,15 @@ def main():
...
+    region, ec2_url, boto_params = get_aws_connection_info(module)
 
     if region:
         try:
-            vpc = boto.vpc.connect_to_region(
-                region,
-                aws_access_key_id=aws_access_key,
-                aws_secret_access_key=aws_secret_key
-            )
+            vpc = boto.vpc.connect_to_region(region, **boto_params)
```
willthames: `vpc = connect_to_aws(boto.vpc, region, **boto_params)`

fixes a couple of other minor issues (particularly arround security tokens with profiles)

cloud/amazon/ec2.py
```diff
@@ -1207,17 +1207,15 @@ def main():
...
         except boto.exception.NoAuthHandlerFound, e:
             module.fail_json(msg = str(e))
+        if boto_params.get('profile_name'):
```
willthames: and then you don't have to do this.

cloud/amazon/ec2.py
andynelson: got it... just committed that change.

bennojoy:
@andynelson could you please close this PR as it is already fixed upstream.

andynelson:
:-)

PR #1255, opened 2015-05-05
APT: Add multiple packages inline + support Async

eloycoto:
Hi, 

If you want to use Async module in apt at the moment it's difficult. 

Async task are not compatible with loops, so my behaviour is to use multiple packages inline. I used this approach for pip as example. 

So, my idea is to use the following: 

``` task.yml
- name: Install deps
  apt:
    name: "git make ngrep"
    update_cache: yes
  async: 10000
  poll: 0
  register: apt_sleeper
```

So with this patch, the system will accept multiple packages, as apt-get install do. 

I've just check, and debian names can't be with space. So no exception it's expected. 

Thanks
Regards.

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -601,8 +601,12 @@ def main():
             install_deb(module, p['deb'], cache,
                         install_recommends=install_recommends,
                         force=force_yes, dpkg_options=p['dpkg_options'])
+        try:
+            packages = p['package'][0].split(' ')
+        except:
+            module.fail_json(
+                msg="invalid package specs: {0}".format(p['package']))
 
-        packages = p['package']
         latest = p['state'] == 'latest'
         for package in packages:
             if package.count('=') > 1:


jordanell:
:+1:

eloycoto:
Hi, 

I [wrote a post in my blog](http://acalustra.com/acelerate-your-ansible-playbooks-with-async-tasks.html) about Async that it's related with this PR. Hope to help to understand. 

Regards.

bcoca:
you really don't need this patch as a comma separated list would also be taken by the module

bcoca:
also apt/yum and others have a with_ optimization, which bypasses the async limitation, so I'm going to close this as a non-issue.

eloycoto:
Hi @bcoca,

I've just testing with commas and yes it's working, but without spaces. Like that: 

```
---
- hosts: 127.0.0.1
  connection: local
  vars:
    packages:
      - vim
      - zsh

  tasks:
  - name: APT with commas
    apt:
      name: "{{ ','.join(packages) }}"
      state: present
```

Thanks for your time, and sorry for invalid PR ;-)

PR #1260, opened 2015-05-05
Fix ec2_eip ip reuse functionality

elventear:
This is an issue that seems to have been sprung up before in the reverse
way:

> https://github.com/ansible/ansible/pull/8518

I believe that this issue is enviroment dependent, but haven't verified
the root cause. Given that in the wild we should see users seeing
various combinations of all the dependencies that control the issue,
both situations should be handled.

Enviroment where the issue occurs:
- osx 10.10.3
- python 2.7.9 x86_64 (from MacPorts)
- ansible 1.9.1
- boto 2.38.0


elventear:
Seems it has been take care in https://github.com/ansible/ansible-modules-core/pull/300

PR #1262, opened 2015-05-05
mysql_user: Only revoke actually granted permissions, not 'ALL'.

bobobox:
When applied to an existing user to reduce the user's privileges on a database, the 'privileges_revoke' function gets invoked, which attempts to perform a 'REVOKE ALL'. However, if the 'login_user' does not have 'ALL' privileges on the database or on *.*, this results in failure and thus the inability to reduce granted permissions via the module.

This is not just a hypothetical: with Amazon Web Services RDS there is no (accessible) user with 'ALL' permissions on *.*, so any attempt to reduce a user's permissions on *.* with the module results in failure. 

This change makes it such that only the actually granted privileges are revoked, thus achieving the same end result but allowing the module to function in environments with more limited privileges like Amazon RDS.

### DIFF
--- a/database/mysql/mysql_user.py
+++ b/database/mysql/mysql_user.py
@@ -350,7 +350,7 @@ def privileges_revoke(cursor, user,host,db_table,priv,grant_option):
         query.append("FROM %s@%s")
         query = ' '.join(query)
         cursor.execute(query, (user, host))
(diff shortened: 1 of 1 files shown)


abadger:
Thanks for catching this and working on it!

# Merged

Hi!

issue is resolved.

Thank you!

PR #1273, opened 2015-05-07
fixes issue 122, returns changed due to routes being recreated each time

bennojoy:
This PR fixes issue 122, where the module recreates the routing table each time the module is run, here we compare if the subnets are associated with route tables with the routes mentioned in the list and if so nothing gets changed.


cloud/amazon/ec2_vpc.py
```diff
@@ -227,6 +227,81 @@ def find_vpc(module, vpc_conn, vpc_id=None, cidr=None):
...
+    needs_change = False
+    #Check if we have any other extra route tables
+    #We add a one for the main table
+    rtb_len = len(route_tables) + 1
+    remote_rtb_len = len(vpc_conn.get_all_route_tables(filters={'vpc_id': vpc.id}))
+    if remote_rtb_len != rtb_len:
+        needs_change = True
+    for rt in route_tables:
+        if needs_change:
+            break;
```
benbridts: Instead of breaking out of the loop, you could change line 272 to `return True`.

The same thing can be done for line 278, 288, 300 and 301.

If you do that you can also remove line 266 and change 302 to `return False`

bennojoy:
thanks @ikben lkben, changed and tested.

cloud/amazon/ec2_vpc.py
```diff
@@ -227,6 +227,76 @@ def find_vpc(module, vpc_conn, vpc_id=None, cidr=None):
...
+            if set(i.items()).issubset(set(j.items())):
+                change = "true"
+        match.append(change)
+    if 'false' in match:
+        return False
+    else:
+        return True
+
+#Check if the remote routes match the local routes
+def rtb_match(route_tables=None, vpc_conn=None, module=None, vpc=None, igw=None):
```
benbridts, line 265: Sorry for not giving all my remarks at the same time, but you may want to rename this function to reflect that it returns true if the rtb has changed. rtb_match implies the opposite behavior. (you could also swap true/false in the function itself).

The preferred way to add documentation of a function is like this:

```
"""
    Checks if the remote routes match the local routes.

    route_tables : description of the route_tables parameter
    vpc_conn : description of the vpc_conn parameter
    module : description of the module parameter
    vpc : description of the vpc parameter
    igw : description of the igw parameter

    Returns:
        True when ... False when ....
"""
```

bennojoy:
Hi Ben,

I think the function is named right because the function checks if the route table given in the yaml and and the route table returned by ec2 is the same, if same it returns true and if not it returns false.

and down we check it as 

if not routes_match(rt['routes'], nrt, igw):
                      return True (where true implies the routing entries have to be changed)

benbridts:
I was talking about the rtb_match function, not routes_match.

bennojoy:
Sorry yeah. let me close this PR and make the changes and send a new one

PR #1257, opened 2015-05-05
Route53 routing policy

zimbatm:
Adds support for routing policies by using the underlying boto facilities. Fixes #207:

Additional work:
- [x] Improve change-detection to support all the extra attributes (by comparing record.to_xml())
- [x] Improve the record updates to use UPSERT instead of DELETE+CREATE
- ~~[ ] Add a route53_healthcheck module to describe these resources.~~ (see #1276)

PR #1276, opened 2015-05-07
route53_health_check: new module

zimbatm:
Complements #1257 by adding a module that manages the Route53 health checks.

The module is still having a couple of issues that I need to fix:
- [x] Make the `present` state idempotent
- [x] Add support for updating existing health-checks (boto doesn't implement the underlying API call)
- ~~[ ] Tagging~~ (not going to implement that as boto doesn't support it either)


zimbatm:
Alright, this is stable now

bennojoy:
@zimbatm looks good, just that empty alias[] is not required and when required: true default: null is not required

zimbatm:
@bennojoy fixed

Do I also need a version_added on the whole module or something ?

zimbatm:
Just added the `version_added: "2.0"` on the top-level

bennojoy:
@zimbatm  yes please, Also as per current policies all new modules go into the module_extras repo first. let me check with team , since this was added a month ago. will udpate .. thanks

zimbatm:
Ok keep me posted. Thanks for looking into this.

gregdek:
Some new modules can go directly into core, but it's a much easier process if we go through extras; to be approved as a new module in core, the Ansible team must review it much more closely, and that takes longer.  

I would recommend resubmitting as an Extras module, where we have a community of AWS module maintainers who are very good about reviewing new modules in a timely manner.

zimbatm:
Thanks

PR #1279, opened 2015-05-07
Not use "is" to compare strings

no2a:
As "is" tests whether if operands are the same object rather than
they have the same value, potentially causes a wrong result.


gregdek:
@no2a Thanks for your pull request. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

PR #1286, opened 2015-05-08
Use of proper fstab file on FreeBSD

0tmc:
Currently mount command doesn't use custom fstab file, trying to do so and set state=mounted returns error, although the file itself is properly edited.

Fixed that for FreeBSD, which uses -F option for custom fstab in mount command.

It's more complicated for Linux systems, as different releases have different options for that action, some (CentOS 6) have none at all.

### DIFF
--- a/system/mount.py
+++ b/system/mount.py
@@ -206,13 +206,29 @@ def unset_mount(**kwargs):
 
 def mount(module, **kwargs):
     """ mount up a path or remount if needed """
+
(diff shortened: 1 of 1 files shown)


bcoca:
nice, now you can set jails fstab easily!

gregdek:
@0tmc Thanks for your pull request. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

PR #1300, opened 2015-05-09
TLS options for docker_image

fampinheiro:
The same options and configuration of the PR #926 are applied in this one but for `docker_image`.
This solves a problem I was having using ansible with boot2docker.

I'm not very experienced with Python but I'm sure someone can point me in the right direction if the PR is missing something.


michaeljs1990:
I just re implemented this completely today because I wasn't really sure what I was searching for until after I fixed this. Being able to use docker and docker_image on OSX would be awesome.

michaeljs1990:
I tried this patch out and it does not fix the issue on OSX. somehow the environment variables are dropped before it reaches this module. The variables are available is you look for them right at the start of the ansible script but somehow are gone later. I am looking into where they are being dropped and if this is an issue with my environment or the ansible source code.

fampinheiro:
@michaeljs1990 this solved my issue when I created the PR, meanwhile I stopped using ansible with `boot2docker` so I really don't know if this is supposed to work on the current version.

michaeljs1990:
yeah it seems like the issue is not in your patch but may need to be a fix in the core ansible repo I'll have a little more information on this tonight when i get a change to find out where the environment variables are being dropped. If i hard code the proper setting into os.getenv() this does work.

gregdek:
Since @fampinheiro isn't using this, I'm going to close this one. @michaeljs1990 do you have a patch coming?

michaeljs1990:
Yes I will have one coming tonight after a little more tests

gregdek:
Awesome, looking forward to your PR.

michaeljs1990:
https://github.com/ansible/ansible-modules-core/pull/1639 needs a bit of work (documentation mostly) and I won't lie I haven't played with python for about 2 years now.

PR #1304, opened 2015-05-11
get meaningful error message before disconnecting from VSphere

rosowiecki:
I've been constantly getting:
pysphere.resources.vi_exception.VIException: [Not Connected]: Must call 'connect' before invoking this method
instead of actual error message, befre I applied this workaround.


bcoca:
this 'fix' just exists before calling disconnect, might as well remove it. A proper fix would check that there is still a connection there (or probably move this to a finally: ) somewhere.

rosowiecki:
Oh, I wasn't aware fail_json calls exit()...  I'll try to submit something better soon, thanks!

bcoca:
you just did not disconnect explicitly, which might happen anyways when script ends and python exists (I would need to check the library). This would be unrelated to you getting the error message or not.

PR #1323, opened 2015-05-13
Tell me what the igw id is if we created an igw.

feanil:
Right now if you create an internet gateway with your vpc.  There is no easy way to know what the ID of that gateway was.  This adds a new output that has the igw info if the gateway was created.

### DIFF
--- a/cloud/amazon/ec2_vpc.py
+++ b/cloud/amazon/ec2_vpc.py
@@ -182,17 +182,6 @@ def get_vpc_info(vpc):
         'state': vpc.state,
     })
 
-def get_igw_info(igw):
-    """
-    Get info about the internet gateway.
-    """
-    if igw is None:
-        return {}
-
-    return ({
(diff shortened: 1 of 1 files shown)


feanil:
@carsongee

bcoca:
i would avoid the function and just always return {gw_id: gwid} even if it is None it is OK to return that

carsongee:
This looks good to me

feanil:
@bcoca do you mean that it's ok if `id` is none or do you mean that it's ok if `igw` is `None`?  Would you like to always return `json_igw={id: (None or igw.id)}`?

bcoca:
probably cleaner to just return:

`module.exit_json(changed=changed, vpc_id=new_vpc_id, vpc=vpc_dict, igw_id=igw_id, subnets=subnets_changed)`

feanil:
@bcoca Updated as per your suggestion.

PR #1330, opened 2015-05-14
Get specific object version with S3 module.

wtcross:
This PR introduces the capability of specifying an object version id when using the `get` and `getstr` modes of the `S3 module.

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -73,6 +73,13 @@ options:
       - Keyname of the object inside the bucket. Can be used to create "virtual directories", see examples.
     required: false
     default: null
+  version:
+    description:
+      - Version ID of the object inside the bucket. Can be used to get a specific version of a file if versioning is enabled in the target bucket.
+    required: false
+    default: null
+    aliases: []
+    version_added: "2.0"
   overwrite:
     description:
       - Force overwrite either locally on the filesystem or remotely with the object/key. Used with PUT and GET operations.
@@ -114,6 +121,9 @@ EXAMPLES = '''
 # Simple GET operation
 - s3: bucket=mybucket object=/my/desired/key.txt dest=/usr/local/myfile.txt mode=get
 
+# Get a specific version of an object.
+- s3: bucket=mybucket object=/my/desired/key.txt version=48c9ee5131af7a716edc22df9772aa6f dest=/usr/local/myfile.txt mode=get
+
 # PUT/upload with metadata
 - s3: bucket=mybucket object=/my/desired/key.txt src=/usr/local/myfile.txt mode=put metadata='Content-Encoding=gzip,Cache-Control=no-cache'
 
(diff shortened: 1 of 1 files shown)


wtcross:
@bcoca Thanks! This is going to be a useful feature.

wtcross:
I have tested both `get` and `getstr` in buckets that have versioning enabled as well as disabled. To enable versioning (if you would like to test yourself) go to an S3 bucket in the AWS Console, click Properties and then the Versioning dropdown. There's a big button to enable it.

wtcross:
bucket-versioning

wtcross:
For reference: http://docs.aws.amazon.com/AmazonS3/latest/dev/ObjectVersioning.html

wtcross:
@bennojoy I resolved the merge conflict by rebasing `devel`. All tests are passing

wtcross:
rebased `devel` again

Is there a specific order in which PRs are evaluated after being assigned a status? I want to write a blog post about using the `S3` module on my companies blog, but need this feature before doing so.

bcoca:
normally by date, ease of integration and sometimes random pings from users reminding us that this is almost done

cloud/amazon/s3.py
```diff
@@ -73,6 +73,12 @@
       - Keyname of the object inside the bucket. Can be used to create "virtual directories", see examples.
     required: false
     default: null
+  version:
```
bcoca, line 76: still requries version_added

bcoca:
@bennojoy test functionally when you can?

cloud/amazon/s3.py
wtcross, line 76: Good catch! I'll add it now.

bennojoy:
@wtcross  could you please change the version added from 1.9 to 2.0 also the put operation doesn't seem to support version, any reason we have not aded version for put operation ?

wtcross:
@bennojoy  I rewrote my last commit to have the correct version.

Updating a specific version of an object would break the semantics of versioning. In any case, the [`boto.s3.key.Key#set_contents_from_file`](http://boto.readthedocs.org/en/latest/ref/s3.html#boto.s3.key.Key.set_contents_from_file) method does have a `version_id` parameter. Only read or metadata-related operations have the `version_id` parameter. These are the methods of the `boto.s3.key.Key` class that the `s3` module uses:
- [`get_contents_to_filename`](http://boto.readthedocs.org/en/latest/ref/s3.html#boto.s3.key.Key.get_contents_to_filename)
- [`get_contents_as_string`](http://boto.readthedocs.org/en/latest/ref/s3.html#boto.s3.key.Key.get_contents_as_string)

Here's a good image showing what happens for a versioned object:
version diagram

Updates go on top and reads can specify any version.

wtcross:
@bennojoy I made sure that all my changes are in one commit, so that I'm not polluting Ansible commit history. Please let me know if there is anything else you need me to do and I will.

wtcross:
By the way, I apologize for not using the issue template. Future PRs will definitely use that.

wtcross:
@bennojoy Do you have any other questions or concerns?

bennojoy:
@wtcross  all good will just give it a try on my end and merge

bennojoy:
Hi!

https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code
Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

PR #1331, opened 2015-05-14
Fix problem writing binary content to a temporary file in the uri module

### LINKED ISSUE #7606: [Bug Report] uri module fails to create file with certain binary files
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.5.x, ansible 1.6.x, ansible 1.7
##### Environment:

N/A
##### Summary:

Changes in the uri module in 1.5 and higher introduced code to deal with data by converting it to Unicode, but this seems to cause problems with some binary files, like Java KeyStore files.
##### Steps To Reproduce:
- Create an empty Java KeyStore file by running the following:
- keytool -genkey -keystore empty.jks
- enter simple 6 character password
- Hit enter until you get the question where you are asked to verify if CN=Unknown, etc. Enter yes
- keytool -delete -alias mykey -keystore empty.jks
- Try to fetch this empty Java KeyStore with the URI module
##### Expected Results:

A successful retrieval of the file and written to the filesystem of the host that's being deployed to.
##### Actual Results:

msg: failed to create temporary content file: 'ascii' codec can't encode characters in position 0-3: ordinal not in range(128)

The problem lies with the following:

``` python
        try:                                                                                                                                                                                                                           
            return r, unicode(content.decode('unicode_escape')), dest                                                                                                                                                                  
        except:                                                                                                                                                                                                                        
            return r, content, dest                         
```

Restoring this to the 1.4 way of simply having `return r, content, dest` fixes the problem for us.

### LINKED ISSUE #10938: failed to create temporary content file: 'ascii' codec can't encode character
In uri.py file，it always report  error msg when i run ansible-playbook to download a file from internet,the error message is  "failed to create temporary content file: 'ascii' codec can't encode character u'\x89' in position 0: ordinal not in range(128)". 

it seems that the code cant catch UnicodeEncodeError.

 As i change the code **_unicode(content.decode('raw_unicode_escape'))**_  to  **_str(content, encoding='utf-8', errors = 'ignore')**_ in uri.py,the error  disappeared and can be catched.

``` python
        try:
            return r, unicode(content.decode('raw_unicode_escape')), dest
        except:
            return r, content, dest
```

abadger:
Fixes https://github.com/ansible/ansible/issues/10938
Fixes https://github.com/ansible/ansible/issues/7606

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -20,6 +20,7 @@
 #
 # see examples/playbooks/uri.yml
 
+import cgi
 import shutil
 import tempfile
 import base64
@@ -188,7 +189,6 @@ try:
 except ImportError:
     HAS_URLPARSE = False
 
-
 def write_file(module, url, dest, content):
     # create a tempfile with some test content
     fd, tmpsrc = tempfile.mkstemp()
@@ -309,10 +309,7 @@ def uri(module, url, dest, user, password, body, body_format, method, headers, r
         r['redirected'] = redirected
         r.update(resp_redir)
         r.update(resp)
-        try:
-            return r, unicode(content.decode('raw_unicode_escape')), dest
-        except:
-            return r, content, dest
+        return r, content, dest
     except httplib2.RedirectMissingLocation:
         module.fail_json(msg="A 3xx redirect response code was provided but no Location: header was provided to point to the new location.")
     except httplib2.RedirectLimit:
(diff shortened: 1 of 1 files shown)


abadger:
Lots of notes for this one:
- In general code should avoid using unicode_escape and raw_unicode_escape codecs.  Those hardcode latin-1 as the byte representation they are decoding from which is becoming increasingly wrong in the utf-8 oriented world we live in.  This leads to mangled output:

``` python
>>> a = u'Café Eñyei'.encode('utf-8')
>>> print(a)
Café Eñyei
>>> print(a.decode('raw_unicode_escape'))
CafÃ© EÃ±yei
```
- Using the charset from the content-type header may be wrong if the server is wrong.  I'm not sure if the present code is the best fallback or not.  We could try utf-8, if that fails, try the charset from the content-type, and if that fails, then use xmlcharref replace. (I wish there was an error handler that would do python escaping like raw_unicode_escape does but alas, there is none.)
- There are a whole lot of other tickets and issues tied together with this one.  I'm going to reference all of those tickets that I can find on here.  If you are seeing this pull request from one of those tickets, please test this out to make sure that your issue does not regress because of this.
- One of the former PRs has some integration tests.  I'll look at trying to add those to this pull request.  I'm not certain that they test this precise problem but they do look like they test some of the other problem areas that the other tickets address.

abadger:
I have the integration tests from @darkk in https://github.com/ansible/ansible/pull/8027 queued up as well.  I had to modify it a little for some things that have changed.  Most notably removing the tests comparing raw_content to content since this version of the fix does not make that distinction.  I'm not sure what the comparison was testing so I was unable to port that portion of the test over to the new code.  @darkk if you have information about what that section of the tests was intended for I can look at making it work.

abadger:
Integration test PR: https://github.com/ansible/ansible/pull/11014

abadger:
Pushing changes to devel so that they receive wider testing.

abadger:
Pushed to stable-1.9: https://github.com/ansible/ansible-modules-core/commit/6906760d46b8afb884c5340415a17df315afd1b2

PR #1342, opened 2015-05-16
Decompress mysql dumps on the fly during an import to simplify operation

fdupoux:
The mysql_db module is able to import compressed MySQL dump files. The way it currently works is a bit complicated: it first decompresses the dump file, then runs the import, and finally the dump is being recompressed. A lot of things can go wrong during these operations: we may not have the right permissions on the disk, or we may not have enough disk space to decompress the file entirely. The recompressed file may not be exactly identical to the original file and the user probably does not expect changes in what is the input file. These operation also fail on symbolic links (the user may want a symlink such as "latest_backup" to point to a regular file). Also recompressing the dump is wasting resources and it can take a long time on large files.

This change uses a simple pipeline of two programs to decompress the file on the fly using zcat/bzcat/xzcat and import it. Hence a single command has to run rather than three commands. It does not make any change at all to the original dump file. It is also faster as there is no need to recompress the file and there is less disk operations involved. And the operation will work even if there are permissions or space restrictions on the disk where the dump is stored.

I have tested this change with the three types of supported compression modes.

### DIFF
--- a/database/mysql/mysql_db.py
+++ b/database/mysql/mysql_db.py
@@ -166,56 +166,20 @@ def db_import(module, host, user, password, db_name, target, all_databases, port
     if not all_databases:
     	cmd += " -D %s" % pipes.quote(db_name)
     if os.path.splitext(target)[-1] == '.gz':
-        gzip_path = module.get_bin_path('gzip')
-        if not gzip_path:
-            module.fail_json(msg="gzip command not found")
-        #gzip -d file (uncompress)
-        rc, stdout, stderr = module.run_command('%s -d %s' % (gzip_path, target))
-        if rc != 0:
-            return rc, stdout, stderr
-        #Import sql
-        cmd += " < %s" % pipes.quote(os.path.splitext(target)[0])
-        try:
-            rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True)
-            if rc != 0:
-                return rc, stdout, stderr
-        finally:
-            #gzip file back up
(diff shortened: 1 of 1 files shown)


bcoca:
this is how the module functioned before, the problem was being able to tell a user at which point of the pipeline the process had failed, sadly pipefail is not universal.

I agree that the current solution leaves much to be desired, I had on my list to use subproccess pipes to get the best of both worlds, remove the complicated steps and intermediate files in favor of streaming through pipes and still being able to report at which part of the pipeline a failure occurred and why.

fdupoux:
It makes sense, but I hoped the second program of the pipe (mysql command) would fail anyway if the decompression fails in the middle for some reason. I also think using python subprocess pipes would be ideal, and apparently there is also native support for reading gzip and bzip2 files in python. It would be even better as it would not require external processes for the decompression (except for xz).

A simpler short-term change would be to simply execute two commands: one would create an uncompressed copy of the compressed dump, and the second one would restore the uncompressed dump. Then the temporary file would be deleted. This way we would not have to recompress the original file which is wasting time and may fail. Would it not be better than the current approch which involved three steps ?

bcoca:
yes the mysql will fail, but the user won't know that it failed because gzip could not read the compressed file or was out of space, that was the whole point of the refactor to how it is now. This PR takes us back to that situation which is worse for the user.

fdupoux:
I have created PR #1417 where the pipelines have been implemented using python subprocess so we can check the exit status of both programs involved in the operation.

bcoca:
closing this in favor of the new PR

PR #1347, opened 2015-05-18
Support optional dockerfile name

octplane:
- Allow the docker-image module to build Dockerfile which are not named Dockerfile.

### DIFF
--- a/cloud/docker/docker_image.py
+++ b/cloud/docker/docker_image.py
@@ -35,6 +35,12 @@ options:
     required: false
     default: null
     aliases: []
+  dockerfile:
+    description:
+       - Dockerfile to use
+    required: false
+    default: Dockerfile
+    version_added: "2.0"
   name:
     description:
        - Image name to work with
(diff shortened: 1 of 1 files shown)


cloud/docker/docker_image.py
```diff
@@ -35,6 +35,12 @@
     required: false
     default: null
     aliases: []
+  dockerfile:
+    description:
```
bcoca, line 39: version_added?

cloud/docker/docker_image.py
```diff
@@ -35,6 +35,12 @@
...
+       - Dockerfile to use
+    required: false
+    default: Dockerfile
+    aliases: []
```
bcoca: no need for aliases

cloud/docker/docker_image.py
octplane, line 39: I've added 1.9 but I have no clue whether this is correct or not...

bcoca, line 39: no, since 1.9 was already released, normally its the 'next version to release', 2.0 in this case

octplane:
Fixed and rebased.

octplane:
Thanks !

PR #1348, opened 2015-05-18
fix apt to handle virtual packages correctly

### LINKED ISSUE #1230: apt: state=present for virtual packages always shown as changed
If you already have a package installed which _Provides_ a virtual package, then a
`apt: state=present name=virtual-package`
always appears as changed, apparently because the `apt-get install` output is slightly different (this is not very robust).

Eg:
_exim4-daemon-light_ is installed
Then try:
`- apt: name=default-mta state=present`
There apt output includes these 2 lines which seem to disturb the apt module:

> Note, selecting 'exim4-daemon-light' instead of 'default-mta'
> exim4-daemon-light is already the newest version.

devjatkin:
When virtual package providing only one package, look up status of target package

fixes #1230

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -206,8 +206,16 @@ def package_status(m, pkgname, version, cache, state):
     except KeyError:
         if state == 'install':
             try:
-                if cache.get_providing_packages(pkgname):
-                    return False, True, False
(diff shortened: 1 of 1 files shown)


mrwacky42:
Don't you want to merge this to `stable-1.9` branch also?

PR #1354, opened 2015-05-18
Correct syntax in PR #1349

### LINKED ISSUE #1344: sysctl bug - yaml dict format in sysctl task with an int value reports 'changed=true' every run.
Issue Type: Bug Report
Ansible Version:

```
~/vagrant/centos $ ansible --version
ansible 1.9.1
```

Ansible Configuration: default

Environment:
mac osx controller using vagrant centos image

```
[vagrant@localhost ~]$ cat /etc/centos-release 
CentOS release 6.6 (Final)
```

Summary:
Using the yaml dict format in sysctl task with an int value reports 'changed=true' every run.  

Steps To Reproduce:
- use the following vagrantfile

```
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure(2) do |config|
  config.vm.box = "chef/centos-6.6"

  config.vm.provision "ansible" do |ansible|
    ansible.playbook = "play.yml"
  end 

end
```
- run the following playbook

```

---
- hosts: all 
  remote_user: vagrant
  sudo: true
  tasks:

    - yum: name=libselinux-python state=present
    - selinux: state=disabled

    - name: set vm swappiness
      sysctl:
        ignoreerrors: true
        name: vm.swappiness
        value: 10
      register: result

    - debug: var=result
```

e.g.

```
~/vagrant/centos $ ANSIBLE_KEEP_REMOTE_FILES=1 ansible-playbook play.yml -i hosts -vvvv --private-key=.vagrant/machines/default/virtualbox/private_key
```
- re-run the playbook

Expected Results:
After re-running, it should not change anything

```
TASK: [set vm swappiness] ***************************************************** 
ok: [default]
```

Actual Results:
It reports it as changed

```
TASK: [set vm swappiness] ***************************************************** 
changed: [default]
```

Debugging:
The module appears to parse the sysctl.conf file into a json dict, and all values are strings, when the module is passed an int value, it fails on the comparison.

Full debugging is here - https://gist.github.com/halberom/6bd0abcff79a50af529e

halberom:
Fixes #1344


sivel:
Shouldn't you have just recommended that #1349 be updated?  As opposed to pulling in that commit and submitting your own?

halberom:
yes probably, i'm a noob at this stuff.

PR #1355, opened 2015-05-18
Update ec2.py

aioue:
Update instance_type description to feature up-to-date list of instance types. Matches format of 'ebs_optimized' param which also includes a link.

Update region description to feature up-to-date list of ec2 regions. Matches format of 'ebs_optimized' param which also includes a link.

Updated choices for monitoring param to be explicit bool matching default 'yes'.

Updated choices for source_dest_check to be explicit bool matching default 'true'.

Updated tenancy choices to remove redundant phrasing in comments.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -44,7 +44,7 @@ options:
   region:
     version_added: "1.2"
     description:
-      - The AWS region to use.  Must be specified if ec2_url is not used. If not specified then the value of the EC2_REGION environment variable, if any, is used.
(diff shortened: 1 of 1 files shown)


cloud/amazon/ec2.py
```diff
@@ -187,6 +189,7 @@
       - Enable or Disable the Source/Destination checks (for NAT instances and Virtual Routers)
     required: false
     default: true
+    choices: [ "true", "false" ]
```
bcoca: should be yes/no, default yes (just to keep all bools in module the same)

aioue:
File consistency updated as requested.

PR #1363, opened 2015-05-19
restore creating ec2_vol w/o name when more than one is present

### LINKED ISSUE #1362: ec2_vol: Commit 6b27cdc break some use case
Commit 6b27cdc08eba5f2ff817b4e0470dd4b6034012c7 changed the volume creation logic to "find and name" volume at creation time but it breaks some use cases when name is not specified and more than one volumes are present with no name tag.

patlachance:
Commit 1d433cf0628c6cdb4a6bec6364492a3ff88c1a64 changed the behavior of the ec2_vol module to allow finding volume based on the name and tagging them with the provided name at creation time but at the same time basic creating a volume without specifying the name doesn't work anymore when more than one volume are present in AWS.

Proposal to improve filtering used by the 'get_volume' function based on the DescribeVolumes filer possibilities.

fixes #1362

### DIFF
--- a/cloud/amazon/ec2_vol.py
+++ b/cloud/amazon/ec2_vol.py
@@ -201,12 +201,30 @@ def get_volume(module, ec2):
     name = module.params.get('name')
     id = module.params.get('id')
     zone = module.params.get('zone')
+    instance = module.params.get('instance')
+    volume_size = module.params.get('volume_size')
+    device_name = module.params.get('device_name')
+    iops = module.params.get('iops')
+    encrypted = module.params.get('encrypted')
+    snapshot = module.params.get('snapshot')
     filters = {}
     volume_ids = None
     if zone:
         filters['availability_zone'] = zone
     if name:
         filters = {'tag:Name': name}
+    if device_name:
+        filters['attachment.device'] = device_name
+    if volume_size:
+        filters['size'] = volume_size
+    if encrypted:
+        filters['encrypted'] = encrypted
+    if snapshot:
+        filters['snapshot-id'] = snapshot
+    if instance and instance != "None":
+        filters['attachment.instance-id'] = instance
+    if iops:
+        filters['volume-type'] = 'io1'
     if id:
         volume_ids = [id]
     try:


bennojoy:
@patlachance did some checks , seems like if we have a volume with same size and device_name it will try to attach that to the instance so maybe we need to add the filter for instance id to

if instance:
        filters['attachment.instance-id'] = instance

patlachance:
@bennojoy :  More filters could be added to the get_volume function. I just added some related to my use case. The example you mentioned seams to be the behavior wanted by the previous contributor: create the volume if it doesn't exist and connect it if one is found. From that perspective, adding instance the instance ID doesn't seam required. Code was added in the create_volume function to handle the volume connection state.

bennojoy:
@patlachance Just trying to understand this more here, so previous to the commit if the instance id, device_name, volume_size is given the code checks if the instance as an attachment of that device_name if not it creates a new volume and attaches it to the instance. 

commit 1d433cf breaks this by searching for volume first andif  it finds multiple volumes in that AZ then it fails, with your commit we do search for a specific volume of the specified size and device name and if not found create one and attach it, so if there happens to be an instance which is of the same size and device name , it woudl try to attach it and if that volume is already attched to a another instance , it woudl fail.  as far as i can see adding the instance filter would make the behaviour almost similar to the previous verison, do you see any other issues adding the instance filter ?

patlachance:
@bennojoy : My understanding of the code (not tested) is that adding the instance id to the filter would make the EC2 API "DescribeVolumes" call only look for volumes already attached to the instance. The commit (code) will find volume with the specified name even if not attached to an instance so I think adding the instance ID to the filter would break this logic. We could argue on the reason why it should be possible to attach a volume instead of creating a new one when the invoker ask for a new one. Personally I prefer filtering on the instance ID, but since the pull request was approved there must be a good reason!

bennojoy:
@patlachance yeah i think we need instance filter too, would it be possible to also add the instance filter to your pr and squash it to a single commit.

patlachance:
@bennojoy done! BTW I also added other allowed parameters to the create_volume function to the filter. Shouldn't be harmful and it will generate amore selective search!

bennojoy:
@patlachance  sorry for this but i think i overlooked one aspect when i said we should include the instance filter, if we have a volume with the same name tag, size and other properties but it is not attached to the instance specfied in the instance tag, the module would create a new volume wiht same properties and attach it to the instance, which might not be the behaviour user wants .. the user may want that same voulme to be attached to that instance .. 
how about making the following change in get_volume

if len(vols) > 1 and name:
        module.fail_json(msg="Found more than one volume in zone (if specified) with name: %s" % name)
elif len(vols) > 1 and not name:
        return None
return vols[0]

thoughts ?

patlachance:
@bennojoy I don't think that making your suggested change will allow handling of the corner case you mentioned. If the filter contains the instance ID and the volume with same tag, size and other properties exists but is not attached to the instance, "vols" will be empty and the following code will be executed. 
 if not vols:
        if id:
            msg = "Could not find the volume with id: %s" % id
            if name:
                msg += (" and name: %s" % name)
            module.fail_json(msg=msg)
        else:
            return None

To handle this corner case, we should remove the instance ID from the filter and rely on the following code in the create_volume function, which should do the job:
    volume = get_volume(module, ec2)
    if volume:
        if volume.attachment_state() is not None:
            if instance is None:
                return volume
            adata = volume.attach_data
            if adata.instance_id != instance:
                module.fail_json(msg = "Volume %s is already attached to another instance: %s"
                                 % (name or id, adata.instance_id))
            else:
                module.exit_json(msg="Volume %s is already mapped on instance %s: %s" %
                                 (name or id, adata.instance_id, adata.device),
                                 volume_id=id,
                                 device=adata.device,
                                 changed=False)

bennojoy:
@patlachance  sounds good can you please remove the instance filter..  Thanks

patlachance:
@bennojoy I made a test to remove the instance filter and it the following code doesn't work:

ec2_vol:
  region: "region"
  zone: "zone"
  instance: "instane id"
  state: present
  volume_size: "size"
  device_name: "name"

error message is "Volume None is already attached to another instance: instance-id

That previous commit breaks many different simple use cases. The module should not require to specify name. What about rollbacking the commit and ask the contributor to implement the feature using a new flag parameter? It could be something like "find_before_create: yes | no".

So I kept the instance filter and added another check because I notice that detach call failed when searching for volume attached to instance "None"

bennojoy:
@patlachance  can you please give this a try   https://github.com/bennojoy/ansible-modules-core/blob/ec2_vol_fix/cloud/amazon/ec2_vol.py

gregdek:
Hey @bennojoy what's the status of this one?

patlachance:
@bennojoy : provisioning works but delete silently ignore the existing volume. I will try to debug when I have more time.

gregdek:
Hi @patlachance, what's the status of this PR?

gregdek:
@patlachance A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

patlachance:
@gregdek Sorry for delay, I'll have a look this week-end

gregdek:
Closing due to inactivity.

PR #1365, opened 2015-05-19
Spelling. :/

sky-shiny:


bcoca:
actually i would change the if as there is another module using HAS instead oF HAVE, fixed via https://github.com/ansible/ansible-modules-core/commit/5011593ee398895d25c8078f3bf073442f4fa33d

PR #1366, opened 2015-05-19
Changed shebang to use environment python in GCE modules

zakx:
The Google Compute Engine modules point their shebang to `/usr/bin/python`, thereby creating issues on Mac OS X setups with a custom python installation in `/usr/local/bin`.

This PR fixes this by changing the shebang to point to the more considerate `/usr/bin/env python`.


bcoca:
# Possible Misunderstanding

Hi!

We believe the ticket you have filed is being somewhat misunderstood.

The shebang on all modules is /usr/bin/python so the ansible can match and modify to use ansible_python_interpreter per host or host group, allowing it to work in different environments. env is not a good solution as it sometimes points to the 'wrong python'.

In the future, this might be a topic more well suited for the user list, which you can also post here if you'd like some more help with the above.
- https://groups.google.com/forum/#!forum/ansible-project

PR #1367, opened 2015-05-19
Adding directory creation example to file module.

gtaylor:
Our team has found ourselves creating directories pretty often if they don't already exist, so I added an example of this in the file module documentation.


bennojoy:
@gtaylor merged thanks

PR #1368, opened 2015-05-19
Fix a problem introduced with #1101 and optimize privilege handling

abadger:
- If a db user belonged to a role which had a privilege, the user would
  not have the privilege added as the role gave the appearance that the
  user already had it.  Fixed to always check the privileges specific to
  the user.
- Make fewer db queries to determine if privileges need to be changed
  and change them (was four for each privilege.  Now two for each object
  that has a set of privileges changed).

### DIFF
--- a/database/postgresql/postgresql_user.py
+++ b/database/postgresql/postgresql_user.py
@@ -324,12 +324,21 @@ def user_delete(cursor, user):
     cursor.execute("RELEASE SAVEPOINT ansible_pgsql_user_delete")
     return True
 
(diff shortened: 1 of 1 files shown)


abadger:
@willthames Does this look good to you?

abadger:
Actually... willthames, I'm going ahead with merging this so that I get further with the integration tests.  But if you notice any problems feel free to let me know (can find me on IRC #ansible-devel too... I read my history from there if I'm not immediately present).

abadger:
# Merged

Hi!

issue is resolved.

Thank you!

PR #1383, opened 2015-05-20
Add state to easy_install

schlueter:
State may be either `present` or `latest`. If it is latest, the `--upgrade` option is passed to the easy_install executable when it is run.

### DIFF
--- a/packaging/language/easy_install.py
+++ b/packaging/language/easy_install.py
@@ -70,6 +70,12 @@ options:
     version_added: "1.3"
     required: false
     default: null
+  state:
+    description:
+      - The desired state of the gem. C(latest) ensures that the latest version is installed.
+    required: false
+    choices: [present, latest]
+    default: present
 notes:
     - Please note that the M(easy_install) module can only install Python
       libraries. Thus this module is not able to remove libraries. It is
@@ -83,7 +89,7 @@ author: Matt Wright
 
 EXAMPLES = '''
 # Examples from Ansible Playbooks
-- easy_install: name=pip
+- easy_install: name=pip state=latest
 
 # Install Bottle into the specified virtualenv.
 - easy_install: name=bottle virtualenv=/webapps/myapp/venv
--- a/packaging/language/easy_install.py
+++ b/packaging/language/easy_install.py
@@ -143,7 +143,8 @@ def main():
     site_packages = module.params['virtualenv_site_packages']
     virtualenv_command = module.params['virtualenv_command']
     executable_arguments = []
-    module.params['state'] is 'latest' and executable_arguments.append('--upgrade')
+    if module.params['state'] == 'latest':
+        executable_arguments.append('--upgrade')
 
     rc = 0
     err = ''
--- a/packaging/language/easy_install.py
+++ b/packaging/language/easy_install.py
@@ -71,6 +71,7 @@ options:
     required: false
     default: null
   state:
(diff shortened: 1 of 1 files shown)


packaging/language/easy_install.py
```diff
@@ -89,8 +89,9 @@
 - easy_install: name=bottle virtualenv=/webapps/myapp/venv
 '''
 
-def _is_package_installed(module, name, easy_install):
-    cmd = '%s --dry-run %s' % (easy_install, name)
+def _is_package_installed(module, name, easy_install, executable_arguments):
+    executable_arguments.append('--dry-run')
```
abadger: Won't this mean that all further uses of executable_arguments outside of the function will contain '--dry-run'?

packaging/language/easy_install.py
```diff
@@ -137,6 +142,8 @@ def main():
     executable = module.params['executable']
     site_packages = module.params['virtualenv_site_packages']
     virtualenv_command = module.params['virtualenv_command']
+    executable_arguments = []
+    module.params['state'] is 'latest' and executable_arguments.append('--upgrade')
```
abadger: more readable to do this as a regular if-then rather than a one-liner.

When comparing to a string, use == rather than is.

abadger:
Need to have documentation.

packaging/language/easy_install.py
```diff
@@ -70,6 +70,12 @@
     version_added: "1.3"
     required: false
     default: null
+  state:
+    description:
+      - The desired state of the gem. C(latest) ensures that the latest version is installed.
```
abadger: s/gem/package/

schlueter: oops

packaging/language/easy_install.py
```diff
@@ -70,6 +70,12 @@
...
+    required: false
+    choices: [present, latest]
+    default: present
```
abadger, line 79: Also add a version_added: "2.0" field so people know which version of ansible this becomes available in.

abadger:
Once version_added is also added to documentation I think this is ready to merge.

schlueter:
Should that be "2.0.0 0.0.pre" or just "2.0.0"?

schlueter:
or actually, it looks like the postgresql_user resource has '2.0' so I'll do that as well

abadger:
Thanks!

# Merged

Hi!

issue is resolved.

Thank you!

br0ziliy:
@abadger This breaks easy_install completely for those who're using latest devel branch.

The patch below fixes the issue:

``` patch
diff --git a/packaging/language/easy_install.py b/packaging/language/easy_install.py
index 77efeae..e14d566 100644
--- a/packaging/language/easy_install.py
+++ b/packaging/language/easy_install.py
@@ -96,7 +96,7 @@ EXAMPLES = '''
 - easy_install: name=bottle virtualenv=/webapps/myapp/venv
 '''

-def _is_package_installed(module, name, easy_install, executable_arguments):
+def _is_package_installed(module, name, easy_install, executable_arguments=[]):
     executable_arguments = executable_arguments + ['--dry-run']
     cmd = '%s %s %s' % (easy_install, ' '.join(executable_arguments), name)
     rc, status_stdout, status_stderr = module.run_command(cmd)
```

br0ziliy:
After 2nd look I created a pull request with more appropriate fix.

PR #1385, opened 2015-05-20
Adding ability to filter AWS Route 53 private hosted zones by attache…

joelthompson:
…d VPC
##### Issue Type:

Feature Pull Request
##### Ansible Version:

devel
##### Ansible Configuration:

N/A
##### Environment:

N/A
##### Summary:

AWS allows you to configure multiple Route53 private hosted zones with the same zone name. The limitation is that multiple private hosted zones with the same name cannot be attached to the same VPC. This feature allows you to take advantage of this, filtering which private hosted zone you wish to make modifications on by attached VPC.
##### Steps To Reproduce:

``` yaml
- route53:
      command: create
      zone: foo.com
      record: new.foo.com
      type: A
      ttl: 7200
      value: 1.1.1.1
      private_hosted_zone: true
      vpc_id: vpc-xxxxxx
```
##### Expected Results:

Creates new.foo.com in the foo.com private hosted zone attached to vpc-xxxxxx
##### Actual Results:

N/A

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -138,6 +138,15 @@ options:
     required: false
     default: null
     version_added: "2.0"
+  vpc_id:
+    description:
+      - When used in conjunction with private_zone: true, this will only modify
+        records in the private hosted zone attached to this VPC. This allows you
+        to have multiple private hosted zones, all with the same name, attached
+        to different VPCs.
+    required: false
+    default: null
+    version_added: "2.0"
 author: "Bruce Pennypacker (@bpennypacker)"
 extends_documentation_fragment: aws
 '''
@@ -250,14 +259,26 @@ try:
 except ImportError:
     HAS_BOTO = False
 
-def get_zone_by_name(conn, module, zone_name, want_private, zone_id):
+def get_zone_by_name(conn, module, zone_name, want_private, zone_id, want_vpc_id):
     """Finds a zone by name or zone_id"""
     for zone in conn.get_zones():
         # only save this zone id if the private status of the zone matches
         # the private_zone_in boolean specified in the params
         private_zone = module.boolean(zone.config.get('PrivateZone', False))
         if private_zone == want_private and ((zone.name == zone_name and zone_id == None) or zone.id.replace('/hostedzone/', '') == zone_id):
-            return zone
+            if want_vpc_id:
+                # NOTE: These details aren't available in other boto methods, hence the necessary
+                # extra API call
+                zone_details = conn.get_hosted_zone(zone.id)['GetHostedZoneResponse']
+                # this is to deal with this boto bug: https://github.com/boto/boto/pull/2882
+                if isinstance(zone_details['VPCs'], dict):
+                    if zone_details['VPCs']['VPC']['VPCId'] == want_vpc_id:
+                        return zone
+                else: # Forward compatibility for when boto fixes that bug
+                    if want_vpc_id in [v['VPCId'] for v in zone_details['VPCs']]:
+                        return zone
+            else:
+                return zone
     return None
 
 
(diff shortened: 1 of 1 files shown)


gregdek:
@joelthompson Thanks for submitting this pull request. Please accept our apologies for taking so long to review it.

joelthompson:
@gregdek -- rebased.

gregdek:
Thanks!

@bpennypacker please have a look. If you +1 we will promote to priority_reviewed.

cloud/amazon/route53.py
```diff
@@ -344,6 +367,11 @@ def main():
           elif not alias_hosted_zone_id_in:
               module.fail_json(msg = "parameter 'alias_hosted_zone_id' required for alias create/delete")
 
+    if vpc_id_in and not private_zone_in:
```
jsmartin, line 370: Use the "required_together" option if applicable --  see https://github.com/ansible/ansible-modules-core/blob/047d7f0775ea9c1fca0f6beecd468fc2e8280219/cloud/vmware/vsphere_guest.py#L1239-L1250

joelthompson, line 370: Thanks @jsmartin but I don't think that's applicable here. It's valid to specify private_zone True and not specify a vpc_id, but it's not valid to specify a vpc_id but not private_zone True. I think that required_together enforces the second use case but doesn't allow for the first, if I understand it correctly.

I think what I actually want is something similar to required_if, but I can't find any documentation about it, it doesn't seem to be used in any of ansible-modules-core or ansible-modules-extras. I also don't think that required_if quite supports the use case. Looking at https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L979 I think there are two things that required_if doesn't support:
1. I want to specify that the private_zone must have a specific value, not just that is passed in (i.e., somebody could pass in private_zone: False and it would pass the required_if check)
1. I want to enforce the previous condition if and only if vpc_id is specified, while required_if only lets me enforce it if the value of vpc_id matches a specific value. It doesn't seem to support the case of "enforce the condition if my parameter is specified" (it's currently "enforce the condition if my parameter takes on a specific value").

Thoughts? I didn't see anything else in that file that looked like it would do what I want.

--Joel

bennojoy:
@joelthompson  did some testing and looks good, couple of things, if a user sets vpc_id it automatically means that he wants to operate on a hosted zone in private , so can't we just make the private_zone to True so that the user doesnt have to enter that value explicitly to true. seems a but redundant to me.
incase i missed something and private_zone is required to be set explicitlyfor some reason  then may be change this error message "parameter 'private_zone_in' must be true when specifying parameter" " 'vpc_id_in'" to the real module parameter names 'private_zone' and 'vpc_id' as the suffix _in is the variable name that user doesnt know about.

joelthompson:
@bennojoy -- you're right, it's really a question of user experience. In my mind, I could easily see somebody who misunderstands how Route53 private hosted zones works and has a public hosted zone that they happen to be using from a VPC, and thinks, "Oh, I'll just specify my VPC ID here" without realizing that this will only look for a private hosted zone, so being explicit seems safest to me.

Agree about the misleading error message, fixed.

mikejquinn:
Is this PR necessary along with #1140? They both seem to achieve the same goal.

joelthompson:
Hey @mikejquinn -- I think they solve slightly different use cases. #1140 allows you to say, "I want to modify records in this exact zone regardless of which VPC it happens to be attached to" whereas this allows you to say, "I want to modify records in whichever hosted zone happens to be attached to my VPC." That's useful because you no longer have to keep track of the opaque hosted zone ID that AWS generates for you, and I think the semantics are a little more natural, i.e., you don't actually care which hosted zone you're talking about, you just want to ensure that DNS queries in a given VPC return particular values, so this allows you to converge to that more naturally.

Thoughts?

PR #1388, opened 2015-05-21
Make it possible to create ACLs recursively

puiterwijk:


files/acl.py
```diff
@@ -36,6 +36,15 @@
     description:
       - defines whether the ACL should be present or not.  The C(query) state gets the current acl without changing it, for use in 'register' operations.
 
+  recurse:
+    required: false
+    default: "no"
+    choices: [ "yes", "no" ]
+    version_added: ""
```
resmo, line 43: this would be "2.0" most likely

bcoca:
competing implementation with #150

puiterwijk:
The fix in #150 is more complete, so will close this in favor of that one.

PR #1399, opened 2015-05-22
Add SLE12 support to hostname.py

vlewin:
Apply `SystemdStrategy` for SLES12 machines as SLES 12 uses system.d


alxgu:
It works. :+1:

PR #1405, opened 2015-05-22
Add support for --log-driver option that docker released with Docker 1.6

### LINKED ISSUE #1393: Add support for --log-driver option for docker run command to docker module
This will enable features like logging to syslog from docker containers.  It is in Docker version 1.6.0+.

ryanwalls:
This fixes #1393.

docker-py has deprecated passing host config parameters to the client.start method.  This PR moves creating host configs to a separate method `create_host_config` in the docker module and passes the host config into the `create_container` method.  

This will work for docker api > 1.15.  If we need to support older versions, the PR will need to be updated.

As part of the change, added support for --log-driver option introduced with docker 1.6 (and api 1.18).

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1129,7 +1129,7 @@ class DockerManager(object):
             # LOG_DRIVER
 
             expected_log_driver = set(self.module.params.get('log_driver') or [])
-            actual_log_driver = set(container['HostConfig']['LogConfig'] or [])
+            actual_log_driver = set(container['HostConfig']['LogConfig']['Type'] or [])
             if actual_log_driver != expected_log_driver:
                 self.reload_reasons.append('log_driver ({0} => {1})'.format(actual_log_driver, expected_log_driver))
                 differing.append(container)
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -92,6 +92,21 @@ options:
       - 'alias. Use docker CLI-style syntax: C(redis:myredis).'
     default: null
     version_added: "1.5"
+  log_driver:
+    description:
+      - You can specify a different logging driver for the container than for the daemon.
+        "json-file" Default logging driver for Docker. Writes JSON messages to file.
+        docker logs command is available only for this logging driver.
+        "none" disables any logging for the container. docker logs won't be available with this driver.
+        "syslog" Syslog logging driver for Docker. Writes log messages to syslog.
+        docker logs command is not available for this logging driver.
+        Requires docker >= 1.6.0.
+    required: false
+    default: json-file
+    choices:
+      - json-file
+      - none
+      - syslog
   memory_limit:
     description:
       - RAM allocated to the container as a number of bytes or as a human-readable
@@ -506,6 +521,7 @@ class DockerManager(object):
             'restart_policy': ((0, 5, 0), '1.14'),
             'extra_hosts': ((0, 7, 0), '1.3.1'),
             'pid': ((1, 0, 0), '1.17'),
+            'log_driver': ((1, 2, 0), '1.18'),
             # Clientside only
             'insecure_registry': ((0, 5, 0), '0.0')
             }
@@ -1110,6 +1126,15 @@ class DockerManager(object):
                 self.reload_reasons.append('volumes_from ({0} => {1})'.format(actual_volumes_from, expected_volumes_from))
                 differing.append(container)
 
+            # LOG_DRIVER
+
+            expected_log_driver = set(self.module.params.get('log_driver') or [])
+            actual_log_driver = set(container['HostConfig']['LogConfig'] or [])
+            if actual_log_driver != expected_log_driver:
+                self.reload_reasons.append('log_driver ({0} => {1})'.format(actual_log_driver, expected_log_driver))
+                differing.append(container)
+                continue
+
         return differing
 
     def get_deployed_containers(self):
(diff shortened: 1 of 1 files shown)


abadger:
CC: @smashwilson

cloud/docker/docker.py
```diff
@@ -1244,45 +1316,8 @@ def do_create(count, params):
...
-            self.ensure_capability('pid')
-            params['pid_mode'] = optionals['pid']
-
-        if optionals['extra_hosts'] is not None:
-            self.ensure_capability('extra_hosts')
-            params['extra_hosts'] = optionals['extra_hosts']
-
         for i in containers:
-            self.client.start(i['Id'], **params)
+            self.client.start(i)
```
smashwilson, line 1322: Reading the docker-py docs, it looks like `.start()` [still accepts a container ID as a parameter](http://docker-py.readthedocs.org/en/latest/api/#start). Is that not the case?

smashwilson:
:+1: Looks solid to me so far! I'll take it for a spin when I get the chance and make sure my test playbook still works.

cloud/docker/docker.py
ryanwalls, line 1322: Yeah, I think it still does, but I thought it looked cleaner this way.

ryanwalls:
@smashwilson @abadger Just wanted to check in on the PR.  Any updates?

cloud/docker/docker.py
```diff
@@ -1206,6 +1231,52 @@ def pull_image(self):
...
+        if optionals['pid'] is not None:
+            self.ensure_capability('pid')
+            params['pid_mode'] = optionals['pid']
+
+        if optionals['extra_hosts'] is not None:
+            self.ensure_capability('extra_hosts')
+            params['extra_hosts'] = optionals['extra_hosts']
+
+        if optionals['log_driver'] is not None:
+            self.ensure_capability('log_driver')
```
smashwilson, line 1275: Because `log_driver` is defaulting to `"json-file"`, this check is being triggered even if the task doesn't include an explicit `log_driver` parameter, which causes the task to fail.

I think the best way to prevent this would be default the parameter to `None` instead of `json-file`. You may want to update the docs entry to say something like "the default is the Docker daemon's default, json-file" too.

smashwilson:
Sorry, I got sidetracked by a deadline this week :smile: I just gave it a quick look and found that snag :point_up:. I'll update `docker-py` in my venv and keep poking throughout the day.

ryanwalls:
@smashwilson good catch.  Updated it per your suggestion.

smashwilson:
:+1:

I hit a bunch of docker TLS issues on my work laptop today (all of which is unrelated to your patch, of course) so I wasn't able to get much further I'm afraid. I'll let you know if I find anything else. In the meantime, if someone else beats me to testing, feel free to chime in and pre-empt me :grin:

cloud/docker/docker.py
```diff
@@ -1110,6 +1127,15 @@ def get_differing_containers(self):
                 self.reload_reasons.append('volumes_from ({0} => {1})'.format(actual_volumes_from, expected_volumes_from))
                 differing.append(container)
 
+            # LOG_DRIVER
+
+            expected_log_driver = set(self.module.params.get('log_driver') or [])
+            actual_log_driver = set(container['HostConfig']['LogConfig']['Type'] or [])
+            if actual_log_driver != expected_log_driver:
```
smashwilson: Two more things that I found:
1. The `log_driver` values here are strings, not lists of strings, so the `set()` constructor should be unnecessary (although it would technically still work, as long as no two options are anagrams :wink:)
2. This appears to trigger a false positive in the reloading logic if no log_driver was set when the container was launched:

```
"msg": "removed 1 container, stopped 1 container, started 1 container, created 1 container.", 
"reload_reasons": "log_driver (set([u'e', u'f', u'i', u'j', u'-', u'l', u'o', u'n', u's']) => set([]))", 
```

Finally -- is it possible for a dict to be missing on the path `['HostConfig']['LogConfig']['Type']`, if you're running with a older docker daemon that doesn't support log_driver? I don't have an older docker handy to test against -- I wish I had a build matrix of docker and docker-py versions available.

smashwilson: On point 2 :point_up:, there are actually a _lot_ of false positives like this for options, especially if you use `state=reloaded` against docker containers that were launched with something other than Ansible. I hit a bunch of other ones this morning (I remember `net` was one).

cloud/docker/docker.py
smashwilson, line 1322: :+1:

smashwilson:
No other regressions in my testing playbook, and I was able to use it to spin up a container that logs to syslog. :sparkles: :heart:

cloud/docker/docker.py
ryanwalls: Ok, I've fixed both 1 and 2.  I think I've probably fixed the false positives for things launched with something other than ansible as well by checking the api version.  That also takes care of worrying about whether `['HostConfig']['LogConfig']['Type']` exists.

smashwilson:
@ryanwalls There was one tiny typo left - I just sent you ryanwalls/ansible-modules-core#1 to fix it. With that change, everything passes cleanly :+1:

smashwilson:
@ryanwalls Excellent. Thanks for being patient!

@abadger LGTM :+1:

ryanwalls:
@abadger @smashwilson what are the next steps to getting this merged?

ryanwalls:
@abadger @smashwilson next steps?

gregdek:
Hi @ryanwalls -- since it's been successfully reviewed by one of the module's owners, we've promoted it to "P1_reviewed". This means it's now at the top of our queue for merge candidates. If all goes well, it should be merged soon-ish.

abadger:
Thanks @ryanwalls and @smashwilson !  Sorry for losing track of this recently.

# Merged

Hi!

issue is resolved.

Thank you!

abadger:
Added a version_added to the log_driver documentation: 964d73172207df628cc7e8cfa9c9782d28f93fa9

Issue #1116, opened 2015-04-09
Error if htpasswd file includes a blank line 


xrow:
Hi, 
I would like to reopen the following ticket https://github.com/ansible/ansible/issues/5328.

The problem is that if a htpasswd file has na empty line in the end, ansible will fail with:

TASK: [nagios | htpasswd path=/etc/nagios/passwd name=admin password=admin owner=root group=apache mode=0640] **\* 
failed: [monitoring.xrow.lan] => {"failed": true}
msg: need more than 1 value to unpack

FATAL: all hosts have already failed -- aborting

```
- htpasswd: path=/etc/nagios/passwd name=admin password=admin owner=root group=apache mode=0640
```

Contents of /etc/nagios/passwd 

```
nagiosadmin:RbdO4ou4PNyMg

```

My ansible verison is  1.9.0.1

mscherer:
Yep, that's a issue in the underlying library IIRC. I am not sure exactly on how/what would be the right answer. Either reimplement it, or fix upstream ( https://code.google.com/p/passlib/ )

Jmainguy:
Suppose we could have ansible sed the file first and remove any blank lines, ^$, and then run the normal python stuff. @mscherer @xrow @bcoca, what do you guys think? Worth coding in?

xrow:
This would have worked for my case.

Jmainguy:
@xrow give #1406 a shot

xrow:
I thanks I will try this for sure once it is available in my version from EPEL.

jimi-c:
# Closing This Ticket

Hi!

The commit a1181b490b7e00953a954878f3694a32378deca4 was recently merged in regarding this, which should resolve this problem for you. This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular issue is resolved.

Thank you!

rbu:
I opened https://bitbucket.org/ecollins/passlib/issues/73/htpasswdfile-does-not-like-empty-lines

PR #1406, opened 2015-05-22
remove blank lines from htpasswd file

Jmainguy:


Jmainguy:
This is designed to fix #1116 . Note this only removes blank lines, the module can still fail if the file contains invalid entries, like a username without a password.

Jmainguy:
This also creates a temporary file if check_mode, and removes the temp file once its finished. This is so the module does not fail if check_mode and the file contains a blank line.

Jmainguy:
This will now return change = True if there are blank lines in the file, since it will then need to edit the file and remove blank lines

bcoca:
soo, use mktemp functions to create a tmp file, no need for using uuid and /tmp yourself.

jimi-c:
Merged, thanks!

Issue #826, opened 2015-02-20
Cannot use urls with yum and rpm_key modules on CentOS 5


JorisVanEijden:
Yum on RHEL 5 and CentOS 5 does not support installing from urls.
It returns "nothing to do" with a code of 0, making ansible think the package is already installed.

Please either document this at the relevant module pages or make ansible detect and error on this case.

I believe this affects both the yum and the rpm_key modules.

This is a reopening of the previous ansible issue https://github.com/ansible/ansible/issues/5327

Jmainguy:
@JorisVanEijden give #1408 a shot. It fails on CentOS 5 now.

JorisVanEijden:
I checked and tested and it works.
Hope to see this in a future release.

I also noticed that I was wrong about the rpm_key. That does work with urls.

abadger:
Merged to devel and also to the stable-1.9 branch.

# Closing This Ticket

Hi!

We believe recent commits (likely detailed above) should resolve this question or problem for you.

This will also be included in the next major release.

If you continue seeing any problems related to this issue, or if you have any further questions, please let us know by stopping by one of the two mailing lists, as appropriate:
- https://groups.google.com/forum/#!forum/ansible-project - for user questions, tips, and tricks
- https://groups.google.com/forum/#!forum/ansible-devel - for strategy, future planning, and questions about writing code

Because this project is very active, we're unlikely to see comments made on closed tickets, but the mailing list is a great way to ask questions, or post if you don't think this particular
issue is resolved.

Thank you!

PR #1408, opened 2015-05-23
forces EL5 to fail if name=:// url style path to rpm

Jmainguy:


Jmainguy:
Intended to fix https://github.com/ansible/ansible/issues/5327 and #826

Jmainguy:
```
[root@phy01 ansible]# ansible -i hosts all -m yum -a "name=http://soh.re/soh.rpm"
centos5.soh.re | FAILED >> {
    "failed": true,
    "msg": "Enterprise Linux 5 cannot install rpms via url like http or https://example.com/example.rpm"
}

centos6.soh.re | FAILED >> {
    "changed": false,
    "failed": true,
    "msg": "Package at http://soh.re/soh.rpm could not be installed",
    "rc": 1
}
```

JorisVanEijden:
Works like a charm.

Jmainguy:
Discussed on live chat today, code in the ability to install rpm's from url's, download it to tmp location, install it, delete rpm

Jmainguy:
@JorisVanEijden added the ability for url's to actually work in EL5, works by downloading the file to /tmp, installing it, then deleting it after to clean up space. give it a shot now

JorisVanEijden:
Interesting idea.
But I feel this is better placed in install() than in ensure(). Then you wouldn't need to pass around the deletelist either, right?
And you should probably let urlretrieve() handle the temp file or use tempfile.mkstemp().

abadger:
We should probably check that the distribution is redhat in addition to the version.  Maybe something like this:

``` python
import platform
[...]
-            distribution_version = get_distribution_version()
-            if LooseVersion(distribution_version) < LooseVersion("6"):
+            distribution = platform.dist()
+            if distribution[0] = "redhat" and LooseVersion(distribution_version) < LooseVersion("6"):
[...]
```

I agree we should use tempfile here but we make use of the package name for things like error messages.  So tempfile.mkstemp() probably isn't the best thing to use.  `tempfile.mkdtemp` is probably better.  Something like this:

``` python
import shutil
[...]
        tempdir = tempfile.mkdtemp()
[...]
                            package = os.path.join(tempdir, str(item.rsplit('/', 1)[1]))
                            try:
                                urllib.urlretrieve (str(item), package)
                                pkg[index] = package
                            except Exception, e:
                                shutil.rmtree(tempdir)
                                module.fail_json(msg="Failure downloading %s, %s" % (item, e))
                            # For cleaning up we can also use shutil.rmtree(tempdir) instead of looping through each downloaded file.
```

Pushing the logic into install as @JorisVanEijden mentions does seem cleaner.  The previous coding of the yum module is somewhat to blame here. It makes sense to retrieve the packages and delete the packages at the same level.  Since the yum module doesn't return its result to the toplevel the cleanup has to happen at the level that we call module.fail_json(). So we end up wanting to push the retrieving of packages into the same level.

Jmainguy:
@JorisVanEijden Great suggestion, went ahead and did that
@abadger good suggestions as well, they have now been incorporated.

Jmainguy:
@JorisVanEijden added your suggestions on comment clean up, and it now removes dir correctly on check mode

Jmainguy:
talked with abadger, going to change out urllib for module_utils.urls.fetch_url()

Jmainguy:
@abadger updated to use fetch_url

abadger:
Thanks for all of your work on this @Jmainguy !

# Merged

Hi!

issue is resolved.

Thank you!

PR #1417, opened 2015-05-25
Decompress mysql dumps on the fly using python subprocess …

fdupoux:
…during an import to simplify operation

The mysql_db module is able to import compressed MySQL dump files. The way it currently works is a bit complicated: it first decompresses the dump file, then runs the import, and finally the dump is being recompressed. A lot of things can go wrong during these operations: we may not have the right permissions on the disk, or we may not have enough disk space to decompress the file entirely. The recompressed file may not be exactly identical to the original file and the user probably does not expect changes in what is the input file. These operation also fail on symbolic links (the user may want a symlink such as "latest_backup" to point to a regular file). Also recompressing the dump is wasting resources and it can take a long time on large files.

This change uses a simple pipeline of two programs to decompress the file on the fly using zcat/bzcat/xzcat and import it. Hence a single command has to run rather than three commands. It does not make any change at all to the original dump file. It is also faster as there is no need to recompress the file and there is less disk operations involved. And the operation will work even if there are permissions or space restrictions on the disk where the dump is stored.

The pipelines have been implemented using python subprocess so we can verify the exit status of both programs and make sure none of them failed. I have tested this change with the three types of supported compression modes.

### DIFF
--- a/database/mysql/mysql_db.py
+++ b/database/mysql/mysql_db.py
@@ -111,6 +111,7 @@ import ConfigParser
 import os
 import pipes
 import stat
+import subprocess
 try:
     import MySQLdb
 except ImportError:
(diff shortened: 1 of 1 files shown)


abadger:
I like the general concept here.  Some changes to make to the implementation:
- Instead of using `zcat, bzcat, xzcat` use `gzip -dc, bzip2 -dc, xz -dc``.  For gzip this is important as zcat may be compress (and using the LZW algorithm) on some *nix systems.  For all of them, it means that we can just depend on one binary existing instead of one for compression and one for decompression.
- You don't need to check the return value of `get_bin_path()`.  Instead do `get_bin_path('gzip', required=True)`.  That will allow `get_bin_path()` to check and throw the error for you.
- It's probably time to refactor the code here so that every mysql restore using any compressor shares the code that's the same for them.  Could do something like:

```
compressor_path = None
if os.path.splitext(target)[-1] == '.gz':
    compressor_path = module.get_bin_path('gzip', required=True)
elif [...]:

if compressor_path:
    # Common decompress and restore code for all compressed dump files
else:
    # Restore code when the dump file is not compressed
```

fdupoux:
Thanks for the review, I have implemented these recommendations in the new commits.

abadger:
Thanks for the enhancement.  I'm sure that will help a lot of people out!

# Merged

Hi!

issue is resolved.

Thank you!

PR #1422, opened 2015-05-26
sleep when only doing a time delay to avoid cpu churn

bcoca:


abadger:
Probably should make this apply to all values of state.  Perhaps something like:

``` python

if not(port or path):
    state = 'timeout'
[....]
elif state == 'timeout':
    time.sleep(timeout)


```

bcoca:
there is no timeout state, it is just a 'default' that happens when no path/port is specified

bcoca:
but added it to drained/absent now

abadger:
bcoca: Yep, I know there's no "timeout" state that the user can specify.  So setting it to "timeout" inside of our code is just assigning a name to the implicit default.  We could also set `state = 'default'` or `state = None` and adjust the conditional check for state appropriately to achieve the same result.

Doing it that way eliminates the redundancy of specifying `time.sleep()` in multiple places for the same implicit state.  that eliminates bugs... For instance, the latest patch still doesn't take care of drained ;-).

bcoca:
default state is 'started' which I think we want to keep that, but I see what you mean, check if no path/socket is passed and just create a time delay, saves on sleeps

bcoca:
drained didn't need one as it either succeeds or sleeps

abadger:
Looks good.  +1.

PR #1423, opened 2015-05-26
Improved output of mysql_db plugin

ercpe:
Show error number and error description on connect error to ease debugging.


Jmainguy:
looks solid

PR #1429, opened 2015-05-28
[gce] [gce_net] fix docs; only delete network if fwname is not provided

flyinprogrammer:
Updated the documentation so that we're not talking about persistent disks. :smile: 

Also, when managing a fw rule we specify the attributes fwname and name indicating the name of the firewall rule and name of the network it should belong too. Currently without this fix, if you go to delete a fw rule, it will also try and delete the network once it is done deleting the fw because a fw rule will always have a name attribute.

With this fix the network will only be deleted if fwname is not present, indicating our state was about the network and not a fw rule.

### DIFF
--- a/cloud/google/gce_net.py
+++ b/cloud/google/gce_net.py
@@ -75,7 +75,7 @@ options:
     aliases: []
   state:
     description:
-      - desired state of the persistent disk
+      - desired state of the network or firewall
     required: false
     default: "present"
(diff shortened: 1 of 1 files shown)


gregdek:
@erjohnso Please review this PR to make sure it adheres to the following guidelines:

erjohnso:
This LGTM, :+1:  - thanks for the fix @flyinprogrammer!

@gregdek, feel free to merge.

PR #1435, opened 2015-05-28
cloudformation: accept local templates in yaml format

billwanjohi:
Since the YAML data format is a subset of JSON, it is trivial to convert
the former to the latter. This means that we can use YAML templates to
build cloudformation stacks, as long as we translate them before passing
them to the AWS API. This could potentially be quite popular in
the Ansible world, since we already use so much YAML for our playbooks.

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -81,6 +81,12 @@ options:
       - Location of file containing the template body. The URL must point to a template (max size 307,200 bytes) located in an S3 bucket in the same region as the stack. This parameter is mutually exclusive with 'template'. Either one of them is required if "state" parameter is "present"
     required: false
     version_added: "2.0"
+  template_format:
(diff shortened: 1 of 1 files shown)


cloud/amazon/cloudformation.py
```diff
@@ -81,6 +81,11 @@
       - Location of file containing the template body. The URL must point to a template (max size 307,200 bytes) located in an S3 bucket in the same region as the stack. This parameter is mutually exclusive with 'template'. Either one of them is required if "state" parameter is "present"
     required: false
     version_added: "2.0"
+  template_format:
```
bcoca, line 84: needs version added (2.0)

tedder:
I have a pull request open for a module called cloudformation_assemble, which does some nice things like concatenation.

cloud/amazon/cloudformation.py
billwanjohi, line 84: OK, fixed

PR #1436, opened 2015-05-28
Add OpenStack Client Config module

emonty:
All of the ansible OpenStack modules are driven by a clouds.yaml config
file which is processed by os-client-config. Expose the data returned by
that library to enable playbooks to iterate over available clouds.

### DIFF
--- a/cloud/openstack/os_client_config.py
+++ b/cloud/openstack/os_client_config.py
@@ -38,7 +38,7 @@ EXAMPLES = '''
 
 
 def main():
-    module = AnsibleModule()
+    module = AnsibleModule({})
     p = module.params
 
     try:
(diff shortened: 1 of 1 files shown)


emonty:
@j2sol @Shrews check me on this?

omgjlk:
If the purpose of this is to generate facts, perhaps this should just be a facts module and inject that data directly into ansible facts rather than force the user to register the return and process it.

http://docs.ansible.com/developing_modules.html#module-provided-facts

cloud/openstack/os_client_config.py
```diff
@@ -0,0 +1,70 @@
...
+     required: false
+     default: 'yes'
+version_added: "2.0"
+requirements: [ os-client-config ]
+author: Monty Taylor
+'''
+
+EXAMPLES = '''
+# Inject facts about OpenStack clouds
+- os-client-config
```
omgjlk: This example seems to be ... missing some content.

omgjlk:
I think this looks good now. :+1:

Shrews:
I'm having issues running it. Instantiating the AnsibleModule object needs an argspec.

Shrews:
Works for me now with the latest commit.

PR #1455, opened 2015-06-02
Fix typos and make docs consistent re: File modules' validate option

jeffwidman:


bcoca:
I'm thinking we should just move it to doc fragments and make it shared so there is no need to maintain N copies.

jeffwidman:
Good idea. I'll update the PR

gregdek:
@jeffwidman any luck on that docs reordering?

jeffwidman:
Sorry, I've been distracted, I'll try to get it handled tomorrow.

jeffwidman:
Hey @gregdek I've just been looking this up here, and little confused on how doc fragments work.

1) it appears that implementing this is a simple matter of adding a 'validate' option to https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/module_docs_fragments/files.py

2) Additionally, I assume I need to remove the current listing of the 'validate' option from the `copy`, `lineinfile`, `template`, and `replace` modules in this repo. They will automagically inherit the validate item from the doc fragment, so if I don't remove the manual listing, I assume it will be double-listed.

3) What happens to the other modules listed in the [files section](http://docs.ansible.com/list_of_files_modules.html) that inherit from files doc fragments? 

This is where I'm confused. For example, the docs for the `assemble` module don't currently list a  'validate' option, but the module code has `extends_documentation_fragment: files`. 

If I add `validate` to the doc fragment for files, won't the docs for `assemble` inherit the 'validate' option from the doc fragment? Which doesn't make sense because that functionality wasn't actually added to the `assemble` module, just to the doc fragment... 

Any clarification is appreciated.

bcoca:
I already added:

```
utils/module_docs_fragments/validate.py 
```

jeffwidman:
Fixed and rebased.

I think this is ready to be merged?

PR #1458, opened 2015-06-03
Explain that state=present should be used with yum and url

Tahvok:
Document that url or file install should be used with 'state=present'.
Some get confused when using 'state=latest' and seeing it failing with no obvious reason.

### DIFF
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -49,7 +49,7 @@ description:
 options:
   name:
     description:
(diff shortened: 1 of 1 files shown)


gregdek:
Tossing over to @abadger who is handling all yum PRs currently.

gregdek:
Since yum is a very central module, we're not going to merging any yum PRs until after 2.0 is released. Thanks for your continued patience, @Tahvok.

h-michael:
:+1:

abadger:
Thanks @Tahvok !  Sorry 2.0 development took so long.  This is now merged to devel and should be in the 2.1.0 release.

PR #1467, opened 2015-06-04
Add missing state parameter

emonty:


omgjlk:
:+1: Yeah, these should get in.

PR #1468, opened 2015-06-04
Make sure that we compare int to int. In case that ttl is provided with jinja template ttl_in is string.

haad:
Fix problem with route53 where is always marked as changed. Becase we compare int to str.

### DIFF
--- a/cloud/amazon/iam.py
+++ b/cloud/amazon/iam.py
@@ -28,48 +28,42 @@ options:
     required: true
     default: null
     choices: [ "user", "group", "role"]
-    aliases: []
   name:
     description:
       - Name of IAM resource to create or identify
     required: true
-    aliases: []
   new_name:
     description:
       - When state is update, will replace name with new_name on IAM resource
     required: false
-    aliases: []
+    default: null
   new_path:
     description:
       - When state is update, will replace the path with new_path on the IAM resource
     required: false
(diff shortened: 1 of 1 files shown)


bcoca:
would this not be covered already?

```
   ttl                  = dict(required=False, type='int', default=3600),
```

...

```
ttl_in                  = module.params.get('ttl')

```

haad:
I thought so, but have a look below to see my config+logs from debug. And to me I got correct bahaviour only when I user int(ttl_in)

``` console
caracal:rsd_aws_setup (master*) $ python /Users/haad/.ansible/tmp/ansible-tmp-1433414932.8-218720879400053/route53
300 -> print int(record['ttl'])
300 -> print ttl_in
<type 'str'> print type(ttl_in)
{"msg": [], "changed": true} ->             if value_list == sorted(rset.resource_records) and int(record['ttl']) == int(ttl_in) and command_in == 'create': is False
```

``` yaml
- name: AWS Route53 | Add DNS A record for a gw instance
  route53: >
    command=create
    zone=rsd-cloud.com
    record={{ vpc_dns_gw_name }}
    type=A
    ttl=300
    overwrite=True
    value={{ item.public_ip }}
  with_items: ec2_gw.results[0].tagged_instances
```

PR #1471, opened 2015-06-05
Initial implementation of s3_lifecycle module

echiu64:
Per issue https://github.com/ansible/ansible-modules-core/issues/1420
See main project for unit tests and README (https://github.com/echiu64/ansible-s3-lifecycle)
Welcome comments/review as I work to fully complete this module

### DIFF
--- a/cloud/amazon/s3_lifecycle.py
+++ b/cloud/amazon/s3_lifecycle.py
@@ -1,0 +1,419 @@
+#!/usr/bin/python
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


bcoca:
The module manages lifecycle to s3 relationships more than the lifecycles themselves which are fully defined inline.  It expects deep knowledge of the lifecycle definitions and does not give an easy (guided by options) way to create one, usable by someone very familiar with the aws internals/API but requires a big learning curve for others.

Sadly we already have some aws modules this way which makes it hard for people to get started, I would like to move away from this mode of doing things.

Tests can be added to ansible/ansible under test/integration with the other cloud tests

echiu64:
Any suggestions how to make it easier? Only thing i can think of is maybe we auto generate rule ids on behalf of the user. And maybe get rid of status and just use state (absent/present) vs combination of status and state. When I do the glacier transition, i could always do a transition_to_glacier option, vs using a "storage_class" option?

gregdek:
@ansible/core is this better as an extras module? I'm inclined to think so.

wimnat:
Working with @echiu64, this has been addressed with a PR in extras - https://github.com/ansible/ansible-modules-extras/pull/718.  This PR can probably be closed.

bcoca:
closing as per above

PR #1472, opened 2015-06-05
fix for 11177 where module readds rule even if the rules exists

### LINKED ISSUE #11177: ec2_group failure When proto is either 50 or 51
Hi

I am having the following issue when applying the variable set below:

```
ec2_security_groups_test:
    - sg_name: test-grup
      sg_description: test-group
      vpc_id: "vpc-XXXXXX"
      region: "{{ region }}"
      sg_rules:
       - proto: 50
         from_port: -1
         to_port: -1
         cidr_ip: 12.12.12.12/32
       - proto: 50
         from_port: -1
         to_port: -1
         cidr_ip: 13.13.13.13/32
       - proto: 50
         from_port: -1
         to_port: -1
         group_name: test-group

```

it will work on the 1st run, but on the second I will have:

```
<Response><Errors><Error><Code>InvalidPermission.Duplicate</Code><Message>the specified rule "peer: 12.12.12.12/32, protocol: 50, ALLOW" already exists</Message></Error></Errors><RequestID>e890595b-9609-4c31-a611-87da7b5de7ae</RequestID></Response>


```

I have tried specifying Ports, also "-1".. Nothing seems to work.
If I say TCP or UDP it works fine, but when I specify a numeric port as above I got the issue:

Any possible workaround?

bennojoy:
This should fix ansible/ansible#11177 where a rule is not matched by the module since the api responds with from_port and to_port as 'none' when the protocol is not tcp/udp/icmp and the port is set to 'all/-1'

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -131,6 +131,11 @@ def make_rule_key(prefix, rule, group_id, cidr_ip):
     """Creates a unique key for an individual group rule"""
     if isinstance(rule, dict):
         proto, from_port, to_port = [rule.get(x, None) for x in ('proto', 'from_port', 'to_port')]
+        #fix for 11177 
+        if proto not in ['icmp', 'tcp', 'udp'] and from_port == -1 and to_port == -1:
+            from_port = 'none'
+            to_port   = 'none'
+
     else:  # isinstance boto.ec2.securitygroup.IPPermissions
         proto, from_port, to_port = [getattr(rule, x, None) for x in ('ip_protocol', 'from_port', 'to_port')]
 


bcoca:
won't this prevent settings for custom protocols?

bennojoy:
@bcoca  this wont block the module to add a custom protocol, the aws api return a 'None' for port if the port is 'any/all' for custom protocols and the api takes -1 for addding the rule, so when the module is run the second time, there is mismatch between the existing rule and the given rule, the existing rule's port show up as None , while in the task args it is given as -1, the code jsut changes -1 to none for custom protocols when checking against existing rules.

PR #1478, opened 2015-06-06
added innitial daemon-reloaded support to service module

bcoca:
WIP


gregdek:
@bcoca looks like you need to rebase :)

bcoca:
I might be dropping this in favor of #1616

gregdek:
ping @bcoca -- have you decided that #1616 is better yet?

bcoca:
yes, #1616 is better route, but will still take time

PR #1485, opened 2015-06-08
win_get_url re-download file only if modified

talcloudshare:
Avoid downloading the same file again if the file on the server wasn't modified, this is based on the file modified date

### DIFF
--- a/windows/win_get_url.ps1
+++ b/windows/win_get_url.ps1
@@ -62,7 +62,7 @@ Else {
         
         $stream = New-Object System.IO.StreamReader($response.GetResponseStream())
         
-        $stream.ReadToEnd() | Set-Content -Path $dest -Force
+        $stream.ReadToEnd() | Set-Content -Path $dest -Force -ErrorAction Stop
         
         $result.changed = $true
     }
--- a/windows/win_get_url.ps1
+++ b/windows/win_get_url.ps1
@@ -1,7 +1,7 @@
 #!powershell
 # This file is part of Ansible.
 #
-# Copyright 2014, Paul Durivage <paul.durivage@rackspace.com>
+# (c)) 2015, Paul Durivage <paul.durivage@rackspace.com>, Tal Auslander <tal@cloudshare.com>
 #
 # Ansible is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


bcoca:
I would not add this w/o adding an option to 'force' the download

talcloudshare:
Makes sense, will try to address that tomorrow

talcloudshare:
Added force parameter, it accepts yes/no and defaults to no

trondhindenes:
On line 65, Set-content won't throw an error since it proably will be a non-terminating one. You should probably add an `-erroraction "stop"` to the set-content cmdlet. Non-terminating errors are not picked up by try/catch blocks, so there's a real chance your catch block won't be executed even if something goes wrong with the code as is.

talcloudshare:
Thanks, added `-ErrorAction Stop`. Let me know if you find any other issues.

gregdek:
@angstwad Please review this PR to make sure it adheres to the following guidelines:

trondhindenes:
@talcloudshare I'm picky here, but the "force" param accepts any string value without failing. Could you make sure you fail if the specified param value is outside of the accepted values? We usually use the Convertto-Bool function to auto-accept any boolean-able values, such as "true", "yes" and "1".

This should fail:
`win_get_url -a "url=http://download.microsoft.com/download/1/6/B/16B06F60-3B20-4FF2-B699-5E9B7962F9AE/VSU_4/vcredist_x64.exe dest='C:\vcredist' force=something"`

talcloudshare:
Sorry, have been quite busy. Not sure if this resolves the issue you mentioned as it still doesn't fail when a string value doesn't match the accepted boolean vaules, but it does defaults to false on unexpected inputs.

trondhindenes:
Yep, I just noticed on another PR that that's the behavior of convertto-bool. That's fine.

trondhindenes:
Looks good now :+1: from me. Nice addition to the module.

gregdek:
Thanks @talcloudshare for the PR, and thanks @trondhindenes for the review, as always :)

gregdek:
Oops... thought this was an Extras module, merged prematurely. @bcoca @jimi-c @toshio please double-check to make sure we're good here (and revert if needed)

PR #1487, opened 2015-06-08
Fixed more issues with the delete_group and paramater mismatch

toninog:
Fixed more issues with the delete_group and paramater mismatch

### DIFF
--- a/cloud/amazon/iam.py
+++ b/cloud/amazon/iam.py
@@ -133,7 +133,7 @@ task:
     iam_type: user
     name: jdavila
     state: update
-    groups: "{{ item.created_group.group_name }}"
+    group: "{{ item.created_group.group_name }}"
   with_items: new_groups.results
 
 '''
@@ -388,7 +388,7 @@ def create_group(module=None, iam=None, name=None, path=None):
     return name, changed
 
 
-def delete_group(module, iam, name):
+def delete_group(module=None, iam=None, name=None):
     changed = False
     try:
         iam.delete_group(name)
@@ -662,7 +662,7 @@ def main():
         group_exists = name in orig_group_list
 
         if state == 'present' and not group_exists:
-            new_group, changed = create_group(iam, name, path)
+            new_group, changed = create_group(iam=iam, name=name, path=path)
             module.exit_json(changed=changed, group_name=new_group)
         elif state in ['present', 'update'] and group_exists:
             changed, updated_name, updated_path, cur_path = update_group(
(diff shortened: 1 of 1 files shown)


cloud/amazon/iam.py
```diff
@@ -133,7 +133,7 @@
     iam_type: user
     name: jdavila
     state: update
-    groups: "{{ item.created_group.group_name }}"
+    group: "{{ item.created_group.group_name }}"
```
bcoca: the correct one here is groups, not group

bcoca:
cc @defionscode

juliedavila:
@toninog is this just a style fix pr? or is there an issue with the module itself?

cloud/amazon/iam.py
toninog: The original commit I had had the groups and not group... it's weird it got reverted...

The other changes are to fix a bug that caused the delete of a group to work.  Below are the fixes for that

toninog:
It was a module fix itself.  The delete_group function fails to delete unless my submitted fix is there.

juliedavila:
@toninog can you fix/squash your commit so it's `groups` and not `group`

toninog:
Sure.  Shall I do a new commit or try and merge the 3 commits ?

If merge : how do I do that ?

juliedavila:
Ah, no. Make a commit to the same branch (no merging).

Then do `git rebase -i -p HEAD~2`

Then you should get a screen with two lines at the top with your commit messages.

The second line, on the far left should be changed to say `fixup` or `squash`.

The do a `git push -f`

toninog:
ok - if you delete the PR I will resubmit a new PR with all the changes in

toninog:
when running : git rebase -i -p HEAD-2 I see 

pick fccc925 fixes to code to enable updates of user to groups and delete groups.  Fixed example yaml to use groups
pick 9f3658d Fixed bug in example where updating a user to a group Fixed bug in create_group and update_group whereby paramaters were not being set correctly.
pick 68cfbca Fixed more issues with the delete_group and paramater mismatch
pick e49a90a Merge branch 'devel' of https://github.com/toninog/ansible-modules-core into devel

# Rebase ccfc9d2..e49a90a onto ccfc9d2

at the top

juliedavila:
So normally you'd just be able to change all but the top most `pick` to `fixup` but in this case there is a merge commit which we can't accept AFAIK. Kind of a pain to do but you'll need to revert to remove the merge commit.

The simplest way would be to copy your patched up `iam.py` to say your home directory, rebranch (maybe even clone if you really want to start clean) off from devel and move over the iam.py again

toninog:
ok - well then we can close this PR and I'll resubmit from a clean fork.

juliedavila:
Ok, go ahead and close this then mention me in your new PR (sorry for the drag)

toninog:
hey no worries.  I need a working iam module :)

PR #1491, opened 2015-06-08
git: Add option to verify GPG signature of commit

Erethon:
This patch adds a new option in the git module called `verify_commit`, that
uses the `git verify-commit` command introduced in git 2.1.0, to check the
signature of a GPG signed commit. The public key used to verify the signature
MUST be trusted in the trustdb GPG file on the target host.

This feature will come in handy when using `ansible-pull` and cloning a git
repo over a clear text protocol.

Refs ansible/ansible#4503

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -160,6 +160,18 @@ options:
               main project. This is equivalent to specifying the --remote flag
               to git submodule update.
 
+    verify_commit:
+        required: false
+        default: "no"
+        choices: ["yes", "no"]
+        version_added: "2.0"
+        description:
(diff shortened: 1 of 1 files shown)


source_control/git.py
```diff
@@ -160,6 +160,17 @@
               main project. This is equivalent to specifying the --remote flag
               to git submodule update.
 
+    verify_commit:
+        required: false
```
bcoca, line 164: missing version added

Erethon, line 164: Didn't know when/if it will be merged, so I left it blank. Adding it now.

PR #1495, opened 2015-06-10
Docs Pull Request: update mysql_user example to remove outdated information

louim:
The example about the `.my.cnf`was still containing outdated information that was fixed in ansible/ansible@c9990b8


bennojoy:
merged thanks.

Issue #69, opened 2014-09-30
s3: Unnecessary bucket_check condition while getting an object from s3


gmpuran:
Hi There,

I was advised here: https://github.com/ansible/ansible/issues/8123 to open a separate issue on newly separated repos.

Summary:

AWS IAM policy enables us to restrict users to specific file or folder in a bucket. While S3 get object operation is successful from other tools, ansible s3 module fails and returns an error Target bucket not found.

The operation succeeds when list bucket rights are provided to that IAM user.

Thanks!

Ladret:
To work around this issue you'll need to add a bucket policy to allow the IAM user to list the bucket.

This is in addition to the policy to allow get acccess on the contents of the bucket

Example:

```
{
            "Sid": "Stmt1412808555554",
            "Action": [
                "s3:ListBucket"
            ],
            "Effect": "Allow",
            "Resource": "arn:aws:s3:::some-bucket",
            "Principal": {
                "AWS": [
                    "arn:aws:iam::1234567890123:user/ansible-iam"
                ]
            }
        }
```

gmpuran:
Thanks @Ladret , yes I'm aware of it.

By "The operation succeeds when list bucket rights are provided to that IAM user.", I meant the same.

juliedavila:
@gmpuran could you do me a favor and format this issue like seen here: https://raw.githubusercontent.com/ansible/ansible/devel/ISSUE_TEMPLATE.md

I think I understand your issue but want to be 100% clear before working on a patch

gmpuran:
Here you go. As the issue quite old, haven't got a chance to grab version details, but have provided detailed description of the issue.

##### Issue Type:

Bug Report

##### Ansible Version:

##### Environment:

N/A

##### Summary:

While getting an s3 object using ansible s3 module, explicit S3:ListBucket permission is required in IAM, for GET operation to be successful. This additional listing permission should not be required as one can restrict a particular IAM user to specific file or folder in S3 bucket.

##### Steps To Reproduce:
1. Create S3 bucket - ansible-s3
2. Create folder test/ and a file inside that folder test.txt
3. GET operation command will be:
   
   #### Simple GET operation
   - s3: bucket=ansible-s3 object=/test/test.txt dest=/usr/local/test.txt mode=get
4. If IAM policy is kept as follows:

``` json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Stmt1426510138000",
            "Effect": "Allow",
[...]

PR #1498, opened 2015-06-10
Added validate_bucket option for any S3 GET requests.

JoelGilliland:
Only added for GET's as it didn't necessarily make sense for PUTS, but maybe that should be added as well. I'm not sure what kind of error a PUT request made to a nonexistent bucket would cause.


JoelGilliland:
FYI, this is related to:
https://github.com/ansible/ansible-modules-core/issues/69

JoelGilliland:
I've found a bug in this pull request when checking md5(overwrite: different). Also I'm not a huge fan of how this is implemented, or really, how the S3 module flows as a whole(ie. forced checks for bucket existing when lookup() is already doing that). I've started a bit of a refactor, but maybe a discussion is needed around this. This is definitely a need for many I think.

Issue #38, opened 2014-09-29
(zabbix package install) yum state=present with with_items fails with conflict, while 'yum install ' works


kustodian:
##### Issue Type:

Bug Report
##### Ansible Version:

1.7.2
##### Environment:

Centos 6
##### Summary:

If you try to install `zabbix-server` and `zabbix-server-pgsql` package from the [official zabbix repository](http://repo.zabbix.com/zabbix/2.4/rhel/6/x86_64/zabbix-release-2.4-1.el6.noarch.rpm) with yum:

```
yum -y install zabbix-server zabbix-server-pgsql
```

It works without a problem, but if you do the same thing with Ansible:

```
- hosts: zabbix
  tasks:
    - yum: name=http://repo.zabbix.com/zabbix/2.4/rhel/6/x86_64/zabbix-release-2.4-1.el6.noarch.rpm state=present

    - yum: name={{ item }} state=present
      with_items:
        - zabbix-server
        - zabbix-server-pgsql
```

The task fails and reports:

```
Error: zabbix-server-mysql conflicts with zabbix-server-pgsql-2.4.0-1.el6.x86_64
Error: zabbix-server-pgsql conflicts with zabbix-server-mysql-2.4.0-1.el6.x86_64
```

I guess the reason for this is because zabbix-server depends on either `zabbix-server-mysql` or `zabbix-server-pgsql` and yum module tries to install one package at a time, so if it first installs `zabbix-server`, that package automatically installs `zabbix-server-mysql`, which later conflicts with `zabbix-server-pgsql`.

The workaround is very easy, you just have to only pass `zabbix-server-pgsql` to the yum module, since it will install `zabbix-server` as a dependency. The problem is that running a yum command with a list of packages doesn't work the same as the Ansible `yum` module with a list of packages passed using `with_items`. This means you will never know what will the module install, compared to the yum command, which is bad in my opinion.
##### Steps To Reproduce:

You can reproduce the bug, by just running the above playbook.
##### Expected Results:

It should install both packages.
##### Actual Results:

Only installs `zabbix-server` and the second package reports a conflict with `zabbix-server-mysql`, which was installed as a dependency for `zabbix-server-mysql`.

Jmainguy:
This happens via command line as well, this is an issue with Zabbix, and not with Ansible. I suggest closing this bug report out.

```
yum install zabbix-server zabbix-server-pgsql
Loaded plugins: fastestmirror
Determining fastest mirrors
[...]

PR #1501, opened 2015-06-11
Execute yum once for all supplied packages

EdTorbett:
Rather than executing yum once per package, execute yum once for all supplied packages. This is necessary when performing a yum upgrade involving multiple dependent packages installed from RPM, for example when upgrading from PostgreSQL 9.0.11 to 9.0.21 on a Red Hat server.

### DIFF
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -663,7 +663,7 @@ def remove(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
 
     if pkgs:
         # run an actual yum transaction
-        cmd = yum_basecmd + ["remove"] + pkg
+        cmd = yum_basecmd + ["remove"] + pkgs
 
         if module.check_mode:
             module.exit_json(changed=True)
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -485,6 +485,7 @@ def list_stuff(module, conf_file, stuff):
 
 def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
 
+    pkgs = []
     res = {}
     res['results'] = []
     res['msg'] = ''
@@ -586,7 +587,10 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
             # the error we're catching here
             pkg = spec
 
-        cmd = yum_basecmd + ['install', pkg]
+        pkgs.append(pkg)
+
+    if pkgs:
+        cmd = yum_basecmd + ['install'] + pkgs
 
         if module.check_mode:
             # Remove rpms downloaded for EL5 via url
@@ -596,15 +600,15 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
                 module.fail_json(msg="Failure deleting temp directory %s, %s" % (tempdir, e))
             module.exit_json(changed=True)
 
-        changed = True
-
         rc, out, err = module.run_command(cmd)
 
-        # Fail on invalid urls:
-        if (rc == 1 and '://' in spec and ('No package %s available.' % spec in out or 'Cannot open: %s. Skipping.' % spec in err)):
-            err = 'Package at %s could not be installed' % spec
-            module.fail_json(changed=False,msg=err,rc=1)
-        elif (rc != 0 and 'Nothing to do' in err) or 'Nothing to do' in out:
+        if (rc == 1):
+            for spec in items:
+                # Fail on invalid urls:
+                if ('://' in spec and ('No package %s available.' % spec in out or 'Cannot open: %s. Skipping.' % spec in err)):
+                    err = 'Package at %s could not be installed' % spec
+                    module.fail_json(changed=False,msg=err,rc=1)
+        if (rc != 0 and 'Nothing to do' in err) or 'Nothing to do' in out:
             # avoid failing in the 'Nothing To Do' case
             # this may happen with an URL spec.
             # for an already installed group,
@@ -614,16 +618,16 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
             out = '%s: Nothing to do' % spec
             changed = False
 
-        res['rc'] += rc
+        res['rc'] = rc
         res['results'].append(out)
         res['msg'] += err
 
         # FIXME - if we did an install - go and check the rpmdb to see if it actually installed
-        # look for the pkg in rpmdb
-        # look for the pkg via obsoletes
+        # look for each pkg in rpmdb
+        # look for each pkg via obsoletes
 
-        # accumulate any changes
-        res['changed'] |= changed
+        # Record change
+        res['changed'] = True
 
     # Remove rpms downloaded for EL5 via url
     try:
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -541,9 +541,9 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
             # short circuit all the bs - and search for it as a pkg in is_installed
             # if you find it then we're done
             if not set(['*','?']).intersection(set(spec)):
-                pkgs = is_installed(module, repoq, spec, conf_file, en_repos=en_repos, dis_repos=dis_repos, is_pkg=True)
-                if pkgs:
-                    res['results'].append('%s providing %s is already installed' % (pkgs[0], spec))
+                installed_pkgs = is_installed(module, repoq, spec, conf_file, en_repos=en_repos, dis_repos=dis_repos, is_pkg=True)
+                if installed_pkgs:
+                    res['results'].append('%s providing %s is already installed' % (installed_pkgs[0], spec))
                     continue
             
             # look up what pkgs provide this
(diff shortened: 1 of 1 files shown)


EdTorbett:
This should resolve issue #38 as well.

packaging/os/yum.py
```diff
@@ -596,15 +600,15 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
                 module.fail_json(msg="Failure deleting temp directory %s, %s" % (tempdir, e))
             module.exit_json(changed=True)
 
-        changed = True
-
```
abadger, line 600: This should remain and then set res['changed'] = changed at the bottom.  The reason is that yum could still have returned "Nothing to do" which will set changed = False.

packaging/os/yum.py
```diff
@@ -614,16 +618,16 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
...
         # FIXME - if we did an install - go and check the rpmdb to see if it actually installed
-        # look for the pkg in rpmdb
-        # look for the pkg via obsoletes
+        # look for each pkg in rpmdb
+        # look for each pkg via obsoletes
 
-        # accumulate any changes
-        res['changed'] |= changed
+        # Record change
+        res['changed'] = True
```
abadger, line 633: Here we should do re['changed'] = changed.  changed will be False if yum output was "Nothing to do".  Otherwise it should be True.

packaging/os/yum.py
EdTorbett, line 600: Completely correct. I'll update.

abadger:
Other than those two changes for returning changed or not, this looks like a good change.

Could probably do the same thing to latest and remove as well as install but those would just be speed optimizations... they aren't causing the yum module to fail because it can't find all of the packages that it needs to complete the transaction.

EdTorbett:
I've been able to implement the remove change as suggested, but latest is more complex due to its support for '*'. As I'm going on holiday for a week - starting tomorrow - I'd best leave implementing that for when I get back.

kustodian:
This looks like a really easy, but an important optimization.

Since this optimization should also increase the speed of the `yum` module, which is one of the slower Ansible modules, would it be possible to increase the speed even more. For example if the packages are all passed to yum in one go, why does it need to call `is_installed()` and `what_provides()` to check if the packages are installed or not? Why not pass all the packages every time? It would make the execution a lot faster, than running `repoquery` for each package separately.

I guess in `check_mode` it would still need to do this, so that it knows if anything is going to be changed, but it would be great if `check_mode` could be faster as well.

packaging/os/yum.py
```diff
@@ -652,17 +659,20 @@ def remove(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
                 res['results'].append('%s is not installed' % pkg)
                 continue
 
+        pkgs.append(pkg)
+
+    if pkgs:
         # run an actual yum transaction
-        cmd = yum_basecmd + ["remove", pkg]
+        cmd = yum_basecmd + ["remove"] + pkg
```
abadger: This should be pkgs rather than pkg

abadger:
@EdTorbett Sounds good -- If you'd like to fix that one thing I found now, I'd be willing to merge this PR and you can submit another one for latest when you get back.  that way we'll close off the buggy behaviour now :-)

@kustodian yeah, there's room for more improvements along the same lines.  However, calling repoquery repeatedly shouldn't be as major a win as this one.  The reason is that is_installed and what_provides() should just be querying the metadata.  So the first run of repoquery might be slow (to download the metadata if it's out of date and also to load the python into the disk caches) but subsequent runs should be much[1]_ faster.  Passing multiple packages to repoquery would also mean that we have to parse the output to figure out which line goes with which requested package so there's some more code that needs to be written and it may be a bit more fragile than the current code.  I also note that most of the code that calls repoquery has alternate code that only uses the yum API, not repoquery.  I'm not sure why we prefer repoquery in these cases.  My naive assumption would be that using the yum API would be quicker.  If someone wants to explore what happens if we make it use the yum API (and only fallback to repoquery?  Not sure when we'd have /usr/bin/yum installed but not have access to the yum API) that might improve speed the same amount as calling is_installed() fewer times.

[1]: Getting it to a single run of repoquery will save noticable time... just not as much as optimizing install/remove/update will.

bcoca:
agreed, this is a significant speedup on it's own and the individual metadata queries are useful for fine grained reporting the state of the packages

kustodian:
Check mode always reports as changed and filtering of packages currently doesn't work, it always passes all packages to yum, because there is an old `pkgs` variable used on line 543:

``` python
             if not set(['*','?']).intersection(set(spec)):
                 pkgs = is_installed(module, repoq, spec, conf_file, en_repos=en_repos, dis_repos=dis_repos, is_pkg=True)
                 if pkgs:
                     res['results'].append('%s providing %s is already installed' % (pkgs[0], spec))
                     continue
```

You should probably rename this variable to something else.

kustodian:
~~@abadger I tried disabling `repoquery` and I noticed that it runs faster with the yum API. I tried querying 12 packages on a VM and here are the resutls:~~
- ~~repoquery ~ 3.7s~~
- ~~yum API ~ 1.1s~~

~~What I also noticed is that Ansible 1.9 runs faster using `repoquery`, than 2.0, but slower with yum API. Here are the average timings for 1.9:~~
- ~~repoquery ~ 2.7s~~
- ~~yum API ~ 3.3s~~

Edit: Ignore this comment, my test was bad.

EdTorbett:
@kustodian Last commit message was meant to reference you, not @strahinja - oops!

kustodian:
@abadger I ran the tests with `repoqury` and with yum API with 100+ packages (with `-C`) to see how fast it will query the packages and something strange happened using Ansible 1.9 and 2.0:
- 1.9 (repoquery) ~ 2.2s
- 1.9 (yum API) ~ 0.8s
- 2.0 (repoquery) ~ **3+ minutes**???
- 2.0 (yum API) ~ **35s***???

Why is 2.0 so slow compared to 1.9?

abadger:
@EdTorbett Great!  I'll try to do a final review and merge this today.

@kustodian Ouch.  We probably need a new bug for that.  I suspect that the problem has something to do with some change in the way ansible forks off worker processes in v2 versus 1.x.  yum (via the API or via repoquery) will acquire a lock so that multiple running instances won't download the metadata concurrently.  I'm guessing that something about the way ansible is handling its worker processes is running into that lock.  @jimi-c would probably have to take a look at that if that's the case.

Alternately, it could be that there's another process on the system that happened to start and acquired the lock during your testing.  I only think of that because yesterday I was testing the dnf module and I managed to run it when a cron job fired off a dnf makecache.  I couldn't figure out what change I'd made that caused it to go so slowly until I saw the cron job in ps's output.

kustodian:
@abadger I did some more investigating and it's not the problem with 2.0 (nor with another app holding a lock), since I copied the module of 1.9.1 and ran it with Ansible 2.0 and it was as fast as with 1.9.1, so it's something introduced with the module. I saved the scripts Ansible compiles and I'm doing some testing on them to see what exactly is the reason. For now when I ran the scripts with `python -m cProfile yum.py` I noticed that the slow script has 1056 calls to `select.select`, while the fast one has only 11 and in the slow one those calls take 177 seconds in total.

kustodian:
I found what the problem is. In the old module for every package we checked if check_mode is set, so when the first package which isn't installed is found, the module would just exit reporting as changed. In this pull request it first goes over all the packages and than it checks if check_mode is set. In my tests, the second package was not installed, so the old version needed to check only 2 packages to exit, while the new version checked all 100+ packages.

I'm not sure if we should leave the old functionality regarding the check mode, since it was faster, but it also wasn't 100% correct. Since for example if you passed a package which doesn't exist after the package which not installed, in check mode yum wouldn't return an error, because it would exit before it reached that package, but in a regular run it would return an error.

The other thing I would like to ask is do we want the yum module to return an error if a package doesn't exist. The reason why I'm asking is because regular yum command doesn't exit with an error status if a package doesn't exist, it only reports which package is it.

Also what is important is that the yum API is about 6 times faster than repoquery.

abadger:
@EdTorbett Thanks for this fix and optimization!

@kustodian I'll look at history to see if I can figure out why the yum module uses repoquery by default and try to change it to use yum api by default instead.  Of course, if I get side tracked with other issues, anyone else is free to beat me to it ;-) 

For your questions -- I think that check mode should be faithful to what non-check more does.  Since an error stops execution of a playbook it makes sense that check_mode should detect that condition as well.  Whether we should exit with an error if the package does not exist, I'm not sure.  We should probably take that to the list.  Throwing an error handles typos so that's good.  On the other hand, replicating what the commandline does could be good.  On the third hand we should compare to what the other os/packaging modules do when they encounter a missing package.  If 90% of them throw an error then we probably should to.  If it's a mixed bag then we're more free to make the decision solely as what makes sense for the yum module.

# Merged

Hi!

issue is resolved.

Thank you!

kustodian:
`apt-get` and `zypper` return a none-zero status if a package doesn't exist (and I'm pretty sure `pacman` does that as well), so It's better to leave it as is, since yum is in the minority.

PR #1504, opened 2015-06-11
documentation: remove include of files options only relevant to unix from win_file.py

jhawkesworth:
spotted that I'd accidentally left in group etc options which aren't relevant to windows.


jhawkesworth:
@bcoca Thanks for picking up this and the other windows stuff recently.

Jon

PR #1507, opened 2015-06-12
Disable warnings in command module

willthames:
Convert `warn` to a boolean so that truthiness tests work correctly


bcoca:
This should not be needed as the spec setting the type to bool should already do this.

willthames:
It wasn't working but I was using an older version of ansible. 
It does work now, although no warnings are printed any longer. I'll close this as that's a separate issue.

PR #1402, opened 2015-05-22
Make the vsphere_guest module really reconfigure the network interfaces

vtsingaras:
Code was missing for reconfiguring devices in state=reconfigured handling code. This commit adds initial support for reconfiguring devices and adds complete code for handling the network devices.

Signed-off-by: Vyronas Tsingaras vtsingaras@it.auth.gr

PR #1511, opened 2015-06-12
Make the vsphere_guest module really reconfigure the network interfaces

aioue:
Small patch to update https://github.com/ansible/ansible-modules-core/pull/1402.


aioue:
Code brought into referenced PR.

PR #1525, opened 2015-06-16
Removed check whether destination directory is writable from files/un…

mjuenema:
Removed check whether destination directory is writable from `files/unarchive.py`. This check will prevent extraction of an archive even if the archive does not actually write to the destination directory but only writes to any writable sub-directories of it. The underlying tar command will report errors should it try to write to read-only directories.

### DIFF
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -277,8 +277,6 @@ def main():
     # is dest OK to receive tar file?
     if not os.path.isdir(dest):
         module.fail_json(msg="Destination '%s' is not a directory" % dest)
-    if not os.access(dest, os.W_OK):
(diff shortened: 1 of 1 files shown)


gregdek:
@pileofrogs Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@pileofrogs are you available to review this PR?

pileofrogs:
Yes!  Thanks!

pileofrogs:
Looks fine.  Good point.  Leave it to the underlying command to decide if there is a problem.

+1

(this is my first time doing a review, so if I need to do anything besides leave this message, please let me know)

Thanks!
-Dylan

PR #1538, opened 2015-06-17
Add OpenStack Security Group Rule Module

emonty:


cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,214 @@
...
+    if state == 'absent' and rule_exists:
+        return True
+    return False
+
+
+def main():
+    argument_spec = openstack_full_argument_spec(
+        security_group   = dict(required=True),
+        protocol         = dict(default='tcp',
+                                choices=['tcp', 'udp', 'icmp']),
```
omgjlk: At least when dealing with a Neutron end point, a protocol is not required. The default, given that this is optional according to the API spec, should be None, or Null. It's how one can create a rule that spans all IP protocols. This module should allow for protocol to be not specified and thus default to None.

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,214 @@
...
+        port_range_min   = dict(required=True, type='int'),
+        port_range_max   = dict(required=True, type='int'),
```
omgjlk: With Neutron port ranges are optional, with a default of Null, which would span all ports. As such this module should not require port_range_min/max and default to None.

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,214 @@
...
+        remote_ip_prefix = dict(required=False, default=None),
+        # TODO(mordred): Make remote_group handle name and id
+        remote_group     = dict(required=False, default=None),
+        ethertype        = dict(default='IPv4',
+                                choices=['IPv4', 'IPv6']),
```
omgjlk, line 258: Technically the API doesn't require IPv4 or IPv6 as a ethertype, the whole argument is listed as Optional, but behavior is unknown as to what it'll do. Food for thought.

cloud/openstack/os_security_group_rule.py
Shrews: Fixed! Thanks for pointing this out. Shade has been changed to toss an exception if Nova is being used and no protocol is supplied.

cloud/openstack/os_security_group_rule.py
Shrews: Fixed.

Shrews:
Let's keep the needs_revision label on this until a newer version of shade is released.

cloud/openstack/os_security_group_rule.py
```diff
@@ -0,0 +1,257 @@
...
+    port_range_min = module.params['port_range_min']
+    port_range_max = module.params['port_range_max']
+    remote_ip_prefix = module.params['remote_ip_prefix']
+    ethertype = module.params['ethertype']
+    direction = module.params['direction']
+
+    for rule in secgroup['security_group_rules']:
+        # No port, or -1, will be returned from shade as None
+        if rule['port_range_min'] is None:
+            rule_port_range_min = -1
```
omgjlk: This is breaking comparison on rules that do not have port ranges. Rules without port ranges are coming back from shade as None and this code is setting it to -1, however rules being passed to the module are also coming through as None and not being converted here. Can we do away with the conversion to -1, or match on both -1 AND None?

omgjlk:
Other than the -1/None issue this module is working as expected for our environment.

cloud/openstack/os_security_group_rule.py
Shrews: I believe you need a shade patch I created for this. Will verify tomorrow.

gregdek:
Looks like this one is bitten by our keypair docs failure from a few days back -- @emonty can you rebase?

cloud/openstack/os_security_group_rule.py
Shrews: @j2sol Yep, you were correct. It was easier to just convert incoming -1 values to None so that they match what is returned from shade. I added a lot of inline documentation to describe what the various actors are doing and how we need to do comparisons. I also created a playbook to test _several_ various scenarios, and tested against both Nova and Neutron, and it looks pretty solid now. Hopefully another set of eyes will help catch anything I may have missed.

cloud/openstack/os_security_group_rule.py
emonty, line 258: Let's none of us spend too much time thinking about why it would be optional or we might do harm to someone.

omgjlk:
This works for me now, and passes tests. Are we still waiting on a shade version?

Also the doc string needs to be updated indicating that the port min/max is not required.

emonty:
@j2sol nope- shade is current and in good shape. @gregdek rebased and ready to fly

omgjlk:
:+1:

PR #1539, opened 2015-06-17
Mysql vars port definition

acaveroc:
Add "login_port" variable to allow connection to a port other than default MySQL installations

### DIFF
--- a/database/mysql/mysql_variables.py
+++ b/database/mysql/mysql_variables.py
@@ -53,6 +53,7 @@ options:
             - mysql host to connect
         required: False
     login_port:
+        version_added: "1.9"
         description:
             - mysql port to connect
         required: False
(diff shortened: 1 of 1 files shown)


bennojoy:
@acaveroc  can you please add version_added  to the new paramter and also we can specifty the type of the parameter in argument spec so that we dont have to cast it .
login_port=dict(default="3306",type='int')

acaveroc:
@bennojoy I've just added this modifications, thank you for your advice

gregdek:
@banyek Please review this PR to make sure it adheres to the following guidelines:

banyek:
LGTM
+1

gregdek:
Thanks @banyek. Promoting to priority_reviewed.

PR #1559, opened 2015-06-19
This enable one to find a datastore with no config_target supplied

vtsingaras:
Signed-off-by: Vyronas Tsingaras vtsingaras@it.auth.gr


bennojoy:
Merged Thanks

PR #1562, opened 2015-06-19
Add the option to pass a string as policy

micxer:
Currently the `iam_policy` module only allows to pass a static file as policy. To make it more flexible and allow the usage of the template lookup, this adds a `policy_json` option that accepts the policy as string. This is mutually exclusive to `policy_document`.


gregdek:
@defionscode Please review this PR to make sure it adheres to the following guidelines:

PR #60, opened 2014-09-30
added `include_doc` option

az7arul:
Use case: If one uses ansible to create their dev setup, this will allow them to install gems with/without documentations

PR #1566, opened 2015-06-19
Fix syntax error in gem.py

sivel:
The following PR introduced a syntax error into gem.py that is causing travis to fail: https://github.com/ansible/ansible-modules-core/pull/60

This PR corrects the syntax error by changing a `-` to a `=`.


bcoca:
sorry, i beat you to it 7a3383260620078fe4e797c3aad0588de71d3005

PR #1567, opened 2015-06-19
Use dict not object notation in OpenStack Security Group

emonty:


cloud/openstack/os_security_group.py
```diff
@@ -48,6 +48,8 @@
        - Should the resource be present or absent.
      choices: [present, absent]
      default: present
+
+requirements: ["shade"]
```
bcoca: if using docfragments that already have requirements, you don't need to add them again.

emonty:
@gregdek updated from @bcoca comment

PR #1579, opened 2015-06-23
[core|pip] Add setuptools option to virtualenv creation

G2fx:
These changes allow user to specify if he want to use --setuptools (with virtualenv_setuptools = yes)
Some modules need setuptools to be installed into a virtualenv (flask-restful-swagger for example)

### DIFF
--- a/packaging/language/pip.py
+++ b/packaging/language/pip.py
@@ -62,6 +62,14 @@ options:
     required: false
     default: "no"
     choices: [ "yes", "no" ]
+  virtualenv_setuptools:
+    version_added: "1.X"
+    description:
+      - Use Setuptools instead of Distribute.
+        This option add --setuptools to virtualenv creation command.
+    required: false
+    default: "no"
+    choices: [ "yes", "no" ]
   virtualenv_command:
     version_aded: "1.1"
     description:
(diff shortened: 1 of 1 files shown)


gregdek:
@mattupstate please review. If you give +1 we will promote to P1_reviewed.

abadger:
Are you sure this is needed?  Perhaps you just need to update your version of virtualenv?

From virtualenv --help:

```
  --setuptools          DEPRECATED. Retained only for backward compatibility.
                        This option has no effect.
  --distribute          DEPRECATED. Retained only for backward compatibility.
                        This option has no effect.
```

setuptools and distribute merged two years ago I believe.

If you have constraints that do make this necessary, please let us know what those are :-)

G2fx:
Hello,

You're right, my bad.
I had an old version of virtualenv (provided by ubuntu 12.04 LTS), with the new version of virtualenv (installed with a pip install --upgrade virtualenv) it's ok, and the patch is useless.

Really sorry for the noise :/

gregdek:
No worries @G2fx :)

mattupstate:
Agreed, this is not necessary. However, whats the general consensus on supporting an older version of pip? Is it necessary? What about people who must rely on distro packages?

bcoca:
@mattupstate to support the older versions we can merge this and make a note in the options that this should not be needed in newer versions of pip. But we are in no hurry to support something unless there are people out there wanting to use it, I'll keep this in mind in case I see a request for it.

Issue #1515, opened 2015-06-13
authorized_key fails in check mode if user doesn't exist


gregsymons:
**Issue Type:**

Bug Report

**Ansible Version:**

1.9.1

**Ansible Configuration:**

No changes from defaults

**Environment:**

Both manager and managed host are Ubuntu 14.04

**Summary:**

I have a set of tasks that create local users and manage their authorized_keys file using the authorized_key module. If I run a play containing these tasks in check mode against a brand-new managed host, the user creation is skipped as expected, but the authorized_key module fails because the user doesn't exist:

**Steps to reproduce**
1. Create a play with the following contents:

``` yaml

---
- hosts: localhost
  become: yes
- tasks
    - name: Manage nonexistent user's ssh keys
      authorized_key:
        user: some_user_that_dont_exist
        key: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDSxrCsbXg6bXZglxShkbSQoOHAU6/1M+K+jRgv8qvQ3sEYDD1s1Rsjj+M5iBS94ZIGEujkMdKm/DjdymMbyQMYm0z8EfMBF6BCuHCwqP89TmXNd01Su9YajMB7C6E8RsbUh7MsMRuLd7EoXqHwCqUVNo8Pywdgcfnfmxy+O/KdyQNyhct8gjxcESlSys4FJiey5saRW3HOvuYcwlC4YGHD1ikW18p09NAhRR1pZf1AKp4gVsBduzONfj3t4rLi/9WW1SoMJAQ5IynK/856GBRy4HX1BtoHxEQfORCVFtPLprHLTfS7nFMoNMyFvYgJsFh90DQxAtYWXFEaRm7zh/Qt
```
1. Execute the playbook with the following command:

```
> ansible-playbook <play> -vvv --check
```

**Expected Results**

I'm not 100% sure, but they should be successful.

**Actual Results:**

```
<localhost> ESTABLISH CONNECTION FOR USER: <user>
[...]

PR #1580, opened 2015-06-23
fix authorized_keys in check_mode

verm666:
This change is in response to issue #1515.

The original problem is: in authorized_key module you have no idea about users
which will be created by Ansible at first run. I can propose next two ways to
solve this problem:
1. Combine modules system/user.py and system/authorized_key.py in one module
    (so you will know everything about users in that module)
2. Use small workaround: add my commit and always provide 'path' parameter
    for authorized_key module during runs with --check option (in case of simplicity of Ansible it's looks not really bad).


gregdek:
Thanks @verm666!

@bradobro can you review this patch? If you give it a +1 we will tag as "P1_reviewed" for quicker inclusion.

abadger:
I think this is probably the best workaround for now.  @verm666 could you add a documentation note in the modules' description  that path is necessary if check mode is used and the user doesn't exist yet?

Other than that, the code itself looks fine.

bcoca:
seems like an extra commit to apt_repository snuck in here, please fix

verm666:
@bcoca, who should remove extra commit? If it's I - what is the best way to do this?

gregdek:
It should just be a rebase, right? If you rebase your repo and then push, the PR should automatically update.

abadger:
Yeah, rebase interactive should let you do this:

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages

verm666:
Can I just recreate pull request from separate branch (issue-1515, e.g.)? This pull request (and another one with commit to apt_repository.py) was created from my devel branch, so it was a mistake. This PR should be closed / rejected.

abadger:
@verm666 Yep, that would be fine.  Just link to this PR in the new one so that we don't forget the conversation here :-)

PR #1581, opened 2015-06-23
redux: Mysql password hash support

### LINKED ISSUE #1662: mysql_user doesn't accept encryption
I normally use Passlib encryption with the prompt module in order to set passwords. This works well with SSH passwords, for example.

However, when trying to use prompt to set the password variable with mysql_user it just doesn't work. I've tried with sha512_crypt and md5_crypt but mysql wouldn't let me log in after changing passwords. 

Using prompt without encryption works fine. 

Is this because mysql_user is not compatible with passlib?

Just mentioning @marktheunissen as it appears as author of the module

clok:
Taking the work from #130 and applying the feedback there. This is a feature we really need. I am sorry to break some of the rules here and create yet another PR for this, but I want resolve the conflicts and get this feature implemented. GREAT WORK by @trevoro here. I will happily take on the work from here if we do not hear from @trevoro 

This will resolve #130 and fixes #1662


gregdek:
@clok you have the ball.

@marktheunissen can you review? If you can give +1 here we will tag "P1_reviewed" to fast-track merge.

clok:
Roger that @gregdek. Thanks!

trevoro:
Thanks guys - I never had a chance to go and update everything against the latest version, after this thing sat stale for so long.

Best!

Jmainguy:
ty @trevoro for writing the original code. Looking forward to this feature

clok:
As @morgo points out in #130 there is a change coming in MySQL 5.7 that will change how existing user passwords will be managed. Since MySQL will be deprecating the `SET PASSWORD` method and merging it with `ALTER USER` we will have a situation where we may need to know the mysql version that is being accessed by Ansible.

Since 5.7 has not officially released yet, I would like to suggest we move forward with merging this feature and open a separate issue for MySQL 5.7 support. Thoughts?

Further reading:
http://mysqlblog.fivefarmers.com/2015/04/10/improved-alter-user-syntax-support-in-5-7/
http://www.tocker.ca/2014/09/05/proposed-changes-to-user-management-in-mysql-5-7.html
http://dev.mysql.com/doc/refman/5.7/en/alter-user.html
http://dev.mysql.com/doc/refman/5.7/en/grant.html
http://dev.mysql.com/doc/refman/5.7/en/set-password.html

morgo:
I noticed another problem with the patch and MySQL 5.7 support:  You can no longer `SELECT password FROM mysql.user`, because the column is renamed.  It's possible to workaround this using conditional version syntax.  Copied from my comment on https://github.com/ansible/ansible/issues/10482:

```
select /*!50706 authentication_string as */ password from mysql.user;
```

morgo:
Sorry to scope creep the feature, but it might be worth denoting in the usage that adding hash support is really "adding `mysql_native_password` hash support", which is based on `sha1`.  In MySQL 5.6 and above there is a much stronger `sha256` hash, which will likely become the default in the future.  This patch makes some assumptions which are not true of `sha256`.

(I don't specifically recommend copying password hashes around, but I acknowledge that it is a useful convenience feature.)

abadger:
Probably want no_log in the argument_spec.

In #130 @bcoca mentioned that using encrypted=yes|no and passing the hashed password string via  password to match with other db modules would be better.

Regarding 5.7 (and you mean 5.7.8, correct?)... I think it would be a very good idea to support it now.  If @morgo can give us the information and testing to get that support in now I'd like it better.  If the interface is still changing or that is too hard to coordinate then we could push the feature in if it doesn't support 5.7.8 but it's always better to be proactive about these things if possible.

(One of the main reasons is that otherwise we're playing catch up... MySQL releases 5.7.8 in January.  Someone discovers that ansible doesn't work with 5.7.8 and reports it.  Someone proposes a PR in February.  We get it reviewed and merged by the end of February.  The next ansible point release isn't ends up happening in May... lots of lag that it would be good eliminate if we can.)

morgo:
@abadger by 5.7, I mean the current release candidate of 5.7.7.  The MySQL release model is to explicitly name the GA (for 5.6 it was 5.6.10).  It's coming _soon_.

Happy to test any code on 5.7 and provide results.

 (I work on the MySQL team, and of course would be really excited to state that Ansible supports 5.7 features on launch :+1: )

clok:
Thanks for the feedback @abadger 

I will change the `true/false` to `yes/no` for the `encrypted` parameter. 

As for support for MySQL 5.7, @morgo please correct me if I am wrong, since 5.7 changes how `ALTER USER` functions we will need to know what the remote mysql server version is to send the correct user modification command. Is there a backwards compatible query that we can use or will we need to know/detect the MySQL server version to send the correct query?

Cheers!

morgo:
@clok `SELECT VERSION()` will return the server version.  The protocol also reveals the version on initial connection (which is exposed through some drivers), but a query is probably the best way.

abadger:
@clok --- oops, sorry, I was looking at an old version (or misunderstood the code when I looked at it before) where encrypted had the password hash instead of a boolean.  true/false vs yes/no doesn't make any difference to me.  No need to change anything for that.

gregdek:
So I guess we're just waiting on @morgo's suggested revisions here?

clok:
Right on. Here is what I am thinking of putting in place. @morgo I would love your feedback on the query. This will work for detecting subversions greater than 6.

``` sql
# Plain Text
SET @subversion := (SELECT SUBSTRING_INDEX(SUBSTRING_INDEX(VERSION(), '.', 2), '.', -1));
SELECT CASE
  WHEN @subversion > 6 THEN
    ALTER USER %s@%s IDENTIFIED BY %s;
  ELSE
    SET PASSWORD FOR %s@%s = PASSWORD(%s);
END;

# Encrypted
SET @subversion := (SELECT SUBSTRING_INDEX(SUBSTRING_INDEX(VERSION(), '.', 2), '.', -1));
SELECT CASE
  WHEN @subversion > 6 THEN
    ALTER USER %s@%s IDENTIFIED WITH mysql_native_password AS %s;
  ELSE
    SET PASSWORD FOR %s@%s = %s;
END;
```

morgo:
@clok I recommend taking an approach similar to rails has and write a method called `supports_alter_user`, then checking first and second number in python code.  This patch would fail if the version was 6.0, 7.0 etc - which can be assumed will be the superset of 5.7.  It also requires something called "anonymous code blocks" - so doesn't run on 5.7.

Here is the rails patch I'm talking about, where they branched away from inline version detection due to bugs:  https://github.com/rails/rails/commit/11338183265f759b07c8dffd526b92fb550ba79d

I'm not any good at Python code, or understanding ansible conventions.. but I'm happy to review all patches and comment on how they will work cross-version in MySQL :)

clok:
Thank you for the feedback @morgo. Great reference in the rails code. I understood that the patch would fail with major version change, but my thought process is that with a major version release there would be significant review and potential refactor of the code anyways.

I am a little concerned about your "anonymous code block" comment though. I have been referring to the 5.7 release notes and I do not see reference to the way embedded queries like the ones suggested above. Have I missed something?

morgo:
Anonymous code blocks is described here: http://docstore.mik.ua/orelly/oracle/prog2/ch15_03.htm - basically using IF/WHEN/ELSE outside of a procedure is not a mysql feature ( https://bugs.mysql.com/bug.php?id=70948 )

Rails chose to patch the way they did so they have a clean implementation as MySQL forks will differ on future features more over time (MariaDB is based on MySQL 5.5).  I can only assume ansible will want to do the same :)

clok:
Right on. Thanks again for the awesome insight into MySQL best practices. As an update for this PR, I am working through resolving some 5.7 related bugs that have popped up while testing the changes. Will push soon.

clok:
@morgo I noticed that the `PASSWORD` function is tagged for deprecation in the dox. https://dev.mysql.com/doc/refman/5.7/en/encryption-functions.html#function_password

We use `SELECT PASSWORD('passed in password string')` to generate the native mysql password hash to compare with the current password hash to determine if a password should be updated. Is there a way, one this is deprecated, to generate that hash? Or will we need to start enforcing other hashing methods? (like SHA256)

gregdek:
@clok ready for re-review?

morgo:
@clok RE: `sha256`, that was my comment on https://github.com/ansible/ansible-modules-core/pull/1581#issuecomment-114701432 - this patch is catering for `mysql_native_password` (default).  Here is another example:

```
if len(password) is 41 and password[0] is '*':
```

A `sha256` password will be longer than 41, and does not start with `*`:

```
mysql> select password('test')\G
*************************** 1. row ***************************
password('test'): $5$n#TP]yP(>;-Ekx1$FwBQaPLPcyoV2fuTNsJYEIUASkNHAzFhtSo4iixn/N7
1 row in set, 1 warning (0.01 sec)
```

The usage and deprecation of `PASSSWORD()` and `old_passwords` is described here:
http://www.tocker.ca/2014/09/05/proposed-changes-to-user-management-in-mysql-5-7.html

To compare in a `sha256`-safe way, you need to look at the `plugin` being used for the user and then set `old_passwords` to the correct value before calling `password()`:

```
mysql> select user, host,plugin from mysql.user;
+-------------+-----------+-----------------------+
| user        | host      | plugin                |
+-------------+-----------+-----------------------+
| root        | localhost | mysql_native_password |
| msandbox    | 127.%     | mysql_native_password |
| msandbox    | localhost | mysql_native_password |
| msandbox_rw | localhost | mysql_native_password |
| msandbox_rw | 127.%     | mysql_native_password |
| msandbox_ro | 127.%     | mysql_native_password |
| msandbox_ro | localhost | mysql_native_password |
| rsandbox    | 127.%     | mysql_native_password |
+-------------+-----------+-----------------------+
8 rows in set (0.01 sec)
```

5.7 of course has specific command for `ALTER USER` and `SHOW CREATE USER` which include the auth plugin, hence why `PASSWORD()` is deprecated.  5.6 supports the query above ^^.  5.5 will not (the columns exist but the values are blank + sha256 is not supported):

```
mysql> select version();
+-----------+
| version() |
+-----------+
| 5.5.44    |
+-----------+
1 row in set (0.00 sec)

mysql> select user,host,authentication_string, plugin from mysql.user;
+-------------+-----------+-----------------------+--------+
| user        | host      | authentication_string | plugin |
+-------------+-----------+-----------------------+--------+
| root        | localhost |                       |        |
| msandbox    | 127.%     | NULL                  |        |
| msandbox    | localhost | NULL                  |        |
| msandbox_rw | localhost | NULL                  |        |
| msandbox_rw | 127.%     | NULL                  |        |
| msandbox_ro | 127.%     | NULL                  |        |
| msandbox_ro | localhost | NULL                  |        |
| rsandbox    | 127.%     | NULL                  |        |
+-------------+-----------+-----------------------+--------+
8 rows in set (0.00 sec)
```

It is a valid FR to have `sha256` support.  Maybe for now just keep that door open, and make it clearer that this is `mysql_native_password` (sha1) hash support, and relies on the deprecated `password()` usage?  I'm not sure, maybe because the commands are cleaner, maybe you will only want to implement `sha256` on 5.7+, and skip 5.6?

clok:
@morgo So, the issue here is that in order for Ansible to know whether or not to update a user's password (keeping in mind the goal of idempotent) we would like to generate the `authentication_string` for a given plain text password to compare against what is stored on the server. This is where `SELECT PASSWORD('string')` was used. With the deprecation forthcoming, I am looking to determine another way to validate these strings.

Here is what I have for `mysql_native_password` generated passwords:

``` sql
mysql> SELECT CONCAT('*', UCASE(SHA1(UNHEX(SHA1('random_string'))))) `mysql_native_password`;
+-------------------------------------------+
| mysql_native_password                     |
+-------------------------------------------+
| *588D5985F677B55B16AC43AB8633C92D530BA9F1 |
+-------------------------------------------+
1 row in set (0.00 sec)

mysql> SELECT PASSWORD('random_string');
+-------------------------------------------+
| PASSWORD('random_string')                 |
+-------------------------------------------+
| *588D5985F677B55B16AC43AB8633C92D530BA9F1 |
+-------------------------------------------+
1 row in set, 1 warning (0.00 sec)
```

I like your idea to call out that Ansible supports `mysql_native_password` for the time being and not sha256 as of yet. I will update the code and the documentation.

morgo:
Deprecation != removal, so you will have until at least 5.8 until you can not use `password()`.  By that time it might make sense to plug-in `sha256` support using the new commands.

(We are still much better off than where we started this discussion, where 5.7 would not work at all.)

clok:
@morgo Yeah, understood. It's one of those things that since I was in there working on it anyways I wanted to make it not fail the next time MySQL bumps a subversion.

@gregdek Okay, ready for review :feelsgood:

database/mysql/mysql_user.py
```diff
@@ -206,31 +218,81 @@ def connect(module, login_user=None, login_password=None, config_file=''):
...
 
+# User Authentication Management was change in MySQL 5.7
+# This is a generic check for if the server version is grater than 5.6
+def server_version_check(cursor):
+    cursor.execute("SELECT VERSION()");
+    result = cursor.fetchone()
+    version_str = result[0]
+    version = version_str.split('.')
+
+    if (int(version[0]) <= 5 and int(version[1]) < 7):
```
clok, line 231: @morgo I believe this is the line of code you are referring to. This logic does not include 5.7 as a `True` result. It will return `False` which is the expected behavior. I will fix the typo in the comment. Also, the comment is inverted since it returns `True` when the version is 5.6 or lower. Very sorry for the confusion.

gregdek:
So! Now that everyone in the Ansible+MySQL world is looking at this module...

@marktheunissen, the author of this module, is no longer actively using it and would like to step down as reviewer of this module.  

Can we use this opportunity to have someone else step up as author/reviewer? I would recommend @clok since he's clearly motivated to keep the module in good repair, with @morgo as backup.  Basically, it means that whenever anyone recommends changes to this (and potentially other) MySQL modules, you'd be pinged for review.  Are you willing?

clok:
@gregdek thank you for the consideration. I am willing and happy to take that on. I use Ansible everyday in my work and want to see it continue to improve. Ready to rock! :facepunch:

gregdek:
@clok then add yourself to the "author:" field with name and Github ID as an addendum to this PR and let's make it official. :)  Once you've done that, I'll go ahead and promote this to priority_reviewed for inclusion.

clok:
It appears the build failure is unrelated to commit 7f9451fed67cdb4c8e5652bd440cc0b902e7ad2f. Looking at the [details](https://travis-ci.org/ansible/ansible-modules-core/builds/69037899) for the build failure it is in `../lib/ansible/modules/core/cloud/openstack/os_keypair.py`

clok:
@gregdek Added myself as author to the `mysql_user.py` module. Is anything else needed?

gregdek:
Sorry for the spurious build failure earlier. I think this is ready. Flagging priority_reviewed. Thanks @clok!

database/mysql/mysql_user.py
```diff
@@ -206,31 +220,81 @@ def connect(module, login_user=None, login_password=None, config_file=''):
...
 
-def user_mod(cursor, user, host, password, new_priv, append_privs):
+def is_hash(password):
+    ishash = False
+    if len(password) is 41 and password[0] is '*':
+        ishash = True
+        for i in password[1:]:
+            if i not in string.hexdigits:
+                ishash = False
+                break
```
abadger: This could be refactored as:

``` python
ishash = False
if len(password) == 41 and password[0] == '*':
    if frozenset(password[1:]).issubset(string.hexdigits):
        ishash = True
```

Should be faster as it avoids looping in python, avoids setting ishash prematurely, doesn't use "is" here as we're really wanting to compare values rather than that the objects are the same.

database/mysql/mysql_user.py
```diff
@@ -206,31 +220,81 @@ def connect(module, login_user=None, login_password=None, config_file=''):
...
+        else:
+            cursor.execute("SELECT authentication_string FROM user WHERE user = %s AND host = %s", (user,host))
         current_pass_hash = cursor.fetchone()
-        cursor.execute("SELECT PASSWORD(%s)", (password,))
-        new_pass_hash = cursor.fetchone()
-        if current_pass_hash[0] != new_pass_hash[0]:
-            cursor.execute("SET PASSWORD FOR %s@%s = PASSWORD(%s)", (user,host,password))
-            changed = True
+
+        if encrypted and password:
```
abadger: In this condtional and the one on line 286, do we want to check password?  We already checked whether password was None on line 266.  If someone wants to set password to the empty string, it seems like checking if bool(password) was True here would prevent that.

abadger:
Okay, one code change to make and one question that may need a code change.

We probably also should have at least one test added here that exercises the encrypted parameter: https://github.com/ansible/ansible/tree/devel/test/integration/roles/test_mysql_user/tasks

database/mysql/mysql_user.py
clok: I like that refactor. Will put that in place.

database/mysql/mysql_user.py
clok: Using `bool(password)` does feel cleaner and more direct. When `password` is `None` then the conditional will be `False` as desired. We can then simplify the if/else block to only consider `encrypted`. Thanks @abadger

clok:
@abadger I agree we should have a test. It's a bummer that the integration tests for the core modules live within a different repository like that. What is the process for adding tests to the ansible integration test suite? This is my first rodeo with contributing to ansible.

abadger:
@clok The process is pretty similar for changes to modules.  Just checkout the main repository https://github.com/ansible/ansible/ make changes and then submit a pull request there.

You can link the new pull request here so that we know that they're related.

Most of the integration tests for modules are a few simple tasks inside of an ansible role devoted to that module.

database/mysql/mysql_user.py
```diff
@@ -206,31 +220,78 @@ def connect(module, login_user=None, login_password=None, config_file=''):
...
+def user_mod(cursor, user, host, password, encrypted, new_priv, append_privs):
     changed = False
     grant_option = False
-
-    # Handle passwords
-    if password is not None:
-        cursor.execute("SELECT password FROM user WHERE user = %s AND host = %s", (user,host))
+    
+    # Handle clear text and hashed passwords.
+    if bool(password):
```
abadger, line 263: So I was wondering earlier whether doing a boolean check is desirable or if we want to just do a is None check.

if password == "" then bool(password) would evaluate False and we'd skip all of the password handling.  Is that what we want?

clok:
@abadger I am working through adding Integration tests now. I am having to build another test VM with the older versions of MySQL because so far it looks like the current Integration tests were failing. I'll see if I can get an update soon.

clok:
Yeah looks like the `test_mysql_user.yml` integretation test has failures even on a completely green pull of Ansible and clean install of dependancies.

```
TASK [test_mysql_user : test_mysql_user : assert output user2 was not updated] **
fatal: [testhost]: FAILED! => {"failed": true, "changed": false, "assertion": "result.changed == false", "evaluated_to": false}

PLAY RECAP **********************************************************************
testhost                   : ok=94   changed=56   unreachable=0    failed=1
```

In order to put new tests in, I will first need to fix this test suite. Unfortunately, the destructive test suite doesn't clean up after a failure, so this may take a while.

clok:
Got tired of building new VMs. Put together a test VM with Vagrant for building and testing Ansible. https://github.com/clok/ansible-test-env-vagrant

abadger:
@jimi-c  I know you were talking about the mysql_user integration test yesterday -- did you get that working?

clok:
@abadger @jimi-c is there a PR open for the mysql integration tests updates that you can point me too? I would love to fix the tests and add more for this feature.

jimi-c:
@clok / @abadger I thought I had it run successfully for me on 7/10, but it did require a complete wipe of mysql (mariadb) on my system, so not sure if there's some cruft leftover from previous runs that's breaking things on certain DB versions. I would definitely try it again now though on a clean system and see if it works for you.

clok:
Hmmm... @jimi-c I am using the Ubuntu vagrant image built by https://github.com/clok/ansible-test-env-vagrant

It pulls ansible `devel` and builds from there. There is a completely raw mysql instance, not mariadb, that is being tested against. I will try with mariadb and see what happens. Mind you, the destructive integration test install mysql from apt-get.

morgo:
@clok You can get MySQL 5.7 very easily from the apt repos too: http://dev.mysql.com/downloads/repo/apt/
Example to change to 5.7:
http://unix.stackexchange.com/questions/158052/how-to-configure-the-mysql-apt-repo-on-ubuntu-on-a-non-interactive-shell

clok:
@morgo Thanks, I have been testing against 5.7 for a bit now. The comment above is regarding regression testing. :+1:

tsalisbury-zz:
Hi,
Great feature.  Any idea when it will get merged/released?
It will really help me a lot.
Thanks!

tsalisbury-zz:
Hi there,
Would there be an eta on this one?
Thanks

bitmand:
Also very interested in ETA for release on this one? :)

( And great job on this one! )

Jmainguy:
Hey, are we waiting for anything on this PR, @bcoca @abadger anything yall need to see before this is merged?

Oneiroi:
Hey guys, this was looking positive; are we likely to see a merge before this request hits the 2 year mark? 

(https://github.com/ansible/ansible/pull/5633 -> https://github.com/ansible/ansible/pull/6049 -> https://github.com/ansible/ansible-modules-core/pull/130 -> https://github.com/ansible/ansible-modules-core/pull/1174 -> https://github.com/ansible/ansible-modules-core/pull/1581)

abadger:
@Jmainguy yes! this just needs an integration test or two added to get this through the door.  Link to the PR in the main repo here and ping me on IRC and I'll get both of them merged ASAP (we're closing in on the deadline for features in modules so sooner is better.)

Jmainguy:
@clok @abadger @Oneiroi Did some testing, doesn't appear to be working as I would expect. When using encrypted=yes, it says 'changed' everytime, regardless if it needs to run or not.

```
[root@phy01 mysql_hash]# cat site.yml 
---
- name: test
  hosts: all
  gather_facts: false
  tasks:
    - name: Set bob with password123
      mysql_user: name=bob state=present password=123

    - name: set bob with password123, but hashed (expect ok)
      mysql_user: name=bob state=present password='*A0F874BC7F54EE086FCE60A37CE7887D8B31086B' encrypted=yes

    - name: set bob with password1234, but hashed (expect changed)
      mysql_user: name=bob state=present password='*D65798AAC0E5C6DF3F320F8A30E026E7EBD73A95' encrypted=yes

    - name: Set bob with password1234 (expect ok)
      mysql_user: name=bob state=present password=password1234
[root@phy01 mysql_hash]# ansible-playbook -i hosts  site.yml 

PLAY [test] ********************************************************************

TASK [Set bob with password123] ************************************************
changed: [centos6.soh.re]

TASK [set bob with password123, but hashed (expect ok)] ************************
changed: [centos6.soh.re]

TASK [set bob with password1234, but hashed (expect changed)] ******************
changed: [centos6.soh.re]

TASK [Set bob with password1234 (expect ok)] ***********************************
ok: [centos6.soh.re]

centos6.soh.re             : ok=4    changed=3    unreachable=0    failed=0   
```

gregdek:
Thanks @clok for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

Jmainguy:
@clok @abadger @Oneiroi If yall are ok with it always saying changed, then I will write the unit tests and submit those, just wanted to be sure thats what yall were expecting. It would be cool if this did not say changed when it didn't need to run. Let me know.

Oneiroi:
@Jmainguy indeed it is, the reason for this is it that it is not looking up the original hash and comparing it for the user (at least from my  original code), it could I suppose but then seeing a non change would indicate the hashes match and introduce an albeit small attack vector for matching hashes.

@gregdek getting anyone to rebase given the history of this relatively simple original PR may be difficult, it seems to be forever left to dwindle then citing some merge conflict / new requirement preventing merge, I refer back to my comment here: https://github.com/ansible/ansible-modules-core/pull/1581#issuecomment-155213796 I've personally rebased, and re-submitted a number of times all for naught. I'm grateful to others whom have continued to push for this in subsequent requests however given the history and the the treatment from the project I'm not exactly hopeful.

Jmainguy:
@Oneiroi @gregdek Sounds good, I will write the unit tests to accept that it always shows changed, and then @abadger will merge it.

@Oneiroi I will let you know once unit tests are run so you can rebase this, and then we all party.

Jmainguy:
@Oneiroi mb got confused, I see you are not latest author on this =). 

@clok standby for integration tests and then you can fix conflict.

Oneiroi:
@Jmainguy indeed, seems I've dropped off the commit history entirely for this one. Oh well so long as the functionality gets merged in!

Jmainguy:
@clok Please rebase, integration tests written.

gregdek:
Yep, needs rebase and revision. :)

trevoro:
Can you guys give an ETA on time to merge following the next rebase?
This is the exact same situation that happened almost a year and a half ago on this exact same feature.

Jmainguy:
Well they promised to merge it already, so should be pretty instant.

gregdek:
The biggest outstanding issue is that we don't have an active maintainer for this module; the maintainer of record no longer responds. 

It appears as though @Jmainguy has done more than enough work to justify merging this, so as soon as you can rebase and @Jmainguy gives us the thumbs up, we will merge.

@Jmainguy would you be willing to become a maintainer of record for this module?

Jmainguy:
@gregdek Yeah that would be cool

Jmainguy:
@gregdek @clok @trevoro @Oneiroi @abadger I have rebased this, and added a one line fix to make it only change password if the hash differs from existing hash when encrypted=true.

Please see #2673  I believe it keeps the additions by @trevoro and @clok intact and can be merged immediately (Unless 2.0 line should be changed to 2.1, which can be done quickly)

gregdek:
A very long and painful process to get this in. Apologies for the delays. Now closing this one in favor of #2673. Thanks to all who participated in pushing this one to the finish line: @clok @trevoro @Oneiroi @abadger.

PR #1583, opened 2015-06-23
iam: use modern helper to allow sts

billwanjohi:
The previous implementation ignored the session token when present. This change is modeled after #566.


bennojoy:
Merged Thanks

PR #1585, opened 2015-06-23
Improved check_mode output for yum module

queridiculo:
The output of the yum module isn't very informational when called in check_mode. 

I've made some minor changes that provide a bit more information about the anticipated changes as a result of a yum operation.


gregdek:
Thanks @queridiculo. This module is maintained by the core team, so it may take a little while for this to make it through.

I haven't tested, but a quick look gets a tentative +1 for me. Seems quite useful.

abadger:
@queridiculo The changes to install and remove look fine.  The changes to latest seem to have some flaws.  I think part of the reason is that install and remove have been optimized and the changes to check_mode's return that you propose fit well into the flow of the optimized code.

Maybe it would be good to do a minor optimization to latest and implement the difference in check_mode return with that.  You could pull the '*' and check_mode conditional block outside of the loop and make the loop only run if the conditional is not met.  the check_mode enhancements can then be done in that section of code.

``` python
if '*' in items or module.check_mode:
   # Do the '*' handling and the check_mode handling
else:
    for spec in items:
        # the current loop code goes here minus the '*' and check_mode handling
```

queridiculo:
Thanks for pointing that out, took me a few reads of latest to understand what you were talking about.

queridiculo:
A lot to learn about git. Made a mess of this pull request, closing to open a new one.

PR #1588, opened 2015-06-24
Service: optional daemon-reload for systemd

### LINKED ISSUE #191: should systemctl do daemon-reload automatically?
##### Issue Type:

Feature Idea
##### Ansible Version:

ansible 1.7.2, but tested against the github version
##### Environment:

host OS: Gentoo Linux
managed OS: Arch Linux
##### Summary:

On /etc/init.d scripts, the start/stop/enable routines always use the service files on disk, which makes it easy to schedule service restart on redeployment of changed service files.
I feel that systemd should behave similarily from within the service module, that is, when I have a rule like

```
service: name=foo state=started
```

I would want service to go ahead and tell systemd to load the latest service files from disk if necessary. At the moment, I would have to invoke `systemctl daemon-reload` manually, or reboot, which does not feel right.
##### Steps To Reproduce:

update a .service file via a copy rule, then call:

```
service: name=foo state=started
```
##### Expected Results:

the service should start using the new .service file
##### Actual Results:

the service files are not reloaded and the old version is used instead

pieterlexis:
This adds the `daemon_reload` option to the service module. This is
ofcourse a systemd-only option. When this option is set, we issue a
`systemctl daemon-reload` before we try to manipulate the service.

Closes #191

I have not tested this code yet (will do later), but I'd like some early feedback as I have not contributed code to ansible before. Also, looking at the current tests, the `runlevel` option is untested, does this mean this option is also 'esoteric'?

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -70,6 +70,14 @@ options:
         default: 'default'
         description:
         - "For OpenRC init scripts (ex: Gentoo) only.  The runlevel that this service belongs to."
+    daemon_reload:
+        required: false
+        choices: ["yes", "no"]
+        default: "no"
+        version_added: "2.0"
+        description:
+        - For systemd only. Whether we should execute `systemctl daemon-reload`
+          before we issue another `systemctl` command for this service
     arguments:
         description:
         - Additional arguments provided on the command line
@@ -155,6 +163,7 @@ class Service(object):
         self.pattern        = module.params['pattern']
         self.enable         = module.params['enabled']
         self.runlevel       = module.params['runlevel']
+        self.daemon_reload  = module.params['daemon_reload']
         self.changed        = False
         self.running        = None
         self.crashed        = None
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks for the PR, @pieterlexis. Since this is pretty central, it will need to be reviewed closely by the core team, so it might take a while.

bcoca:
I think the proposal in #1616 could cover this and other 'special' options.

pieterlexis:
`daemon-reload` is not a special or custom service command as is the case in #1616. It is a 'feature' of the init-system (systemd) itself to re-read all service files. When you pass daemon-reload a service name, it fails:

```
root@debian:~# systemctl daemon-reload apache2
Invalid number of arguments.
```

Hence, this issue is different from #1616

bcoca:
@pieterlexis I don't disagree with your reasoning, but we are looking at keeping the service module generic, also probably split out into specific systemd/upstart/sysV modules, more or less how the packaging is now. The 'custom' state might not be the optimal solution but I was looking at it as something interim vs adding more permanent specific option to service, if that makes sense.

gregdek:
So @bcoca does this mean we should close this because we're planning to do it differently, or is there some path forward for this PR?

philm:
@bcoca: +1 on the idea of splitting apart service module into specific modules.

gregdek:
@bcoca would like some feedback here. @pieterlexis originally submitted and noted that he hadn't yet tested because he was looking for advice before he proceeded further, and it looks like you favor #1616.

If #1616 is really the "right" way to do it, then should we just close this PR?  Or is this PR right "enough" to continue with? We should probably make a call one way or the other so @pieterlexis can move on with his life. :)

gregdek:
I'm taking the silence here to mean that the core team does not agree with the approach in this PR, for the reasons @bcoca mentioned above. Closing. Apologies to @pieterlexis for the delay.

Issue #1452, opened 2015-06-02
yum module install URL reports changed in --check


kustodian:
##### Issue Type:

Bug Report
##### Ansible Version:

1.9.1
##### Ansible Configuration:

Nothing
##### Environment:

Centos 6
##### Summary:

`yum` module always reports as changed in check mode (`--check`) when installing an RPM from a URL.
##### Steps To Reproduce:

Here is a short play:

```

---
- hosts: all
  gather_facts: no
  tasks:
    - yum: name=http://nginx.org/packages/rhel/6/noarch/RPMS/nginx-release-rhel-6-0.el6.ngx.noarch.rpm state=present
```

If you run it regularly everything is fine (since `nginx-release` was installed before):

```
$ ansible-playbook -i test_hosts yum.yml        

PLAY [all] ******************************************************************** 

TASK: [yum name=http://nginx.org/packages/rhel/6/noarch/RPMS/nginx-release-rhel-6-0.el6.ngx.noarch.rpm state=present] *** 
ok: [pg-01]

PLAY RECAP ******************************************************************** 
pg-01                      : ok=1    changed=0    unreachable=0    failed=0   
```

But if you run it in check mode, it reports as changed:

```
$ ansible-playbook -i test_hosts yum.yml -C

PLAY [all] ******************************************************************** 

TASK: [yum name=http://nginx.org/packages/rhel/6/noarch/RPMS/nginx-release-rhel-6-0.el6.ngx.noarch.rpm state=present] *** 
changed: [pg-01]

PLAY RECAP ******************************************************************** 
pg-01                      : ok=1    changed=1    unreachable=0    failed=0
```
##### Expected Results:

Idempotency should work in check mode as well.
##### Actual Results:

Idempotency doesn't work in check mode.

mntechguy:
+1 I've noticed this issue as well.

PR #1589, opened 2015-06-24
fixes yum always changed on --check

yhekma:
I made a semi quick fix for ansible/ansible-modules-core#1452 that downloads the remote rpm and parses the header to see if the package is already installed when run with --check.


gregdek:
Thanks @yhekma. This is a core module that's maintained by the Ansible team, and it's a pretty important module, so it may take longer to get a full in-depth review of this PR.

yhekma:
@gregdek sure no prob. I just saw the issue and remembered that I actually already fixed (assuming I did't break stuff) it for my own use at the client I am at, so I just decided to make a pr for it.

abadger:
@yhekma Thanks, i think that the yum module may have had a couple changes recently that need to be incorporated into here:
- We're now using the loop through the "items" list to assemble a list of pkgs that need installing and then installing them all with a single yum install.  This check needs to move into the loop that processes items or it needs to loop over the pkgs list separately.  (right now the check is only handling the last package that was processed in the items loop so it won't handle the case where the package-by-url isn't the last package in items.
- There's some code that downloads packages by url for EL5 (where yum doesn't handle the urls itself).  We probably want to be more like that.  The things that I definitely see we could reuse here are the tempdir that we download into and the use of fetch_url instead of urllib2.  Until later versions of python-2.7 (2.7.9 I believe), urllib2 doesn't verify that the host we connect to has a valid ssl certificate.  So we need to use our fetch_url function which does.

yhekma:
Good points. I will take a look at them tomorrow if and when work allows

abadger:
@yhekma Looks like maybe you did a merge instead of a rebase and it pulled a lot of other commits into the pull request :-(  If you're able to correct that it will make it easier to do a final review of this.

yhekma:
Hmz, I thought I did a rebase on the upstream/devel....
Maybe not enough sleep. I just force-pushed the changes in a single commit

yhekma:
@abadger  could you indicate if the code is in a state you find acceptable for merging?

packaging/os/yum.py
```diff
@@ -603,6 +603,33 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
...
+            ts.setVSFlags(yum.rpm._RPMVSF_NOSIGNATURES)
+            yb = yum_base()
+
+            for package in pkgs:
+                if package.startswith('http'):
+                    tmp_file = tempfile.TemporaryFile()
+
+                    try:
+                        tmp_file, info = fetch_url(module, pkg)
+                        tmp_file.write(rsp.read())
```
abadger, line 616: We'll want to chunk reads here like we do at line 532.  rpm files can be rather large and we don't want to read() the whole file into memory at one time.

packaging/os/yum.py
```diff
@@ -603,6 +603,33 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
...
+                    except Exception, e:
+                        module.fail_json(msg="Error downloading %s: %s" % (package, e))
+
+                    tmp_file.seek(0)
+
+                elif os.path.exists(package):
+                    tmp_file = os.open(package, os.O_RDONLY)
+
```
abadger, line 624: I think we need one more else: here to handle the case where package is a package name rather than a file or url.

packaging/os/yum.py
```diff
@@ -603,6 +603,33 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
...
+                hdr = ts.hdrFromFdno(tmp_file)
+
+                if yb.rpmdb.searchNevra(name=hdr[yum.rpm.RPMTAG_NAME]):
+                    module.exit_json(changed=False)
+
+                if yb.rpmdb.searchNevra(name=package):
+                    module.exit_json(changed=False)
+
```
abadger, line 632: Oh I see what you're trying to do here.  You always create a tmp_file.  then you check both the hdr from ts.hdrFromFdno() and searching by the package name.  I don't think this is the best idea.
1. There's several error conditions that you'd have to catch here to make this logic work.  For instance, ts.hdrFromFdno() run on a non-rpm file (the case if tmp_file is empty because package is neither a url or file) then it will raise an exception.
2. This doesn't handle cleanup of the temporary files that we download.

abadger:
A few thoughts:
- there's another PR that's working on yum's check_mode as well: https://github.com/ansible/ansible-modules-core/pull/1657  I believe that we'll still need something along the lines of this PR to handle URLs in the install case though.  So this is just a heads up that you might have to rebase if that gets merged first.
- After looking at the problems with cleaning up the temporary files and then looking at whether it could be merged into the code that downloads on older RHEL, I wondered if there's a better way to do this.  Perhaps we should be checking whether the package needs to be installed in the block that starts at line 523.  If the package is installed, then skip to the next package.  If it's not installed, then we can continue processing to where we add the package to the pkgs list.  This would mirror the processing for package names.  that way when we get down to line 605 (where we do special processing for check_mode) we'll already know whether there's any packages to install or not (if the pkgs list is empty, there are no packages to install).  I'm not sure if you'll still need to download the package in check_mode or if there's some yum API that can get the NEVR of a package at a URL. to compare to what's installed on the system.  If you do need to download , you can piggy back on the RHEL5 download code which would also mean you can piggy back on the cleanup code so that portion should be easier to manage.

queridiculo:
Minimizing the download operations makes sense, something I was thinking about when I read the code in check_mode section.

Assuming that we have a URL based package, check_mode could probably be reduced to a simple urllib2 based check to see whether the package is available for download.

I'd imagine something like this:

``` python
if '://' in spec:
    url = spec  
    pkg = url.rsplit('/', 1)[1]
    base = url.rsplit('/', 1)[0]
    request = urllib2.Request(url)
    request.get_method = lambda : 'HEAD'

    try:
        conn = urllib2.urlopen(request)
        conn.getcode()
        conn.close()
        msg = '{0} will be installed from {1}'.format(pkg, base)
        res['results'].append(msg)
    except Exception, e:
        msg = 'Error downloading {0} ({1})'.format(url, e)
        module.fail_json(msg)
```

abadger:
@kustodian -- what do you think about @queridiculo's comment above?  I know you opened the bug about check-mode prompting this and also care about the speed of the module.  In this case we can't have it both ways.  We either have to download the rpm to determine if it is already installed or we will pretty frequently report that there's something to do when there isn't.  (Actually, there may be a third way but I'm not familiar enough with rpm to attempt to code it.  Using http range requests we may be able to download just the rpm header.  And then there is a way to use just the header to check what package is at that URL and decide whether an update is necessary.)

queridiculo:
Something I've also noticed in latest, module.fail_json is called for a single failed yum operation.

Not an issue if we're dealing with a single package, but I'm wondering if this is really the best way of approaching operations on a number of packages. 

Perhaps it may be more desirable to record the failures, and the provide more comprehensive information with a module.fail_json once all the items are processed when the error count is > 0.

abadger:
@queridiculo I would be okay with that.  I think that there's currently inconsistent behaviour in this area so making it all conform to that model would be good.  note that as the changes land to make updates, installs, and removals happen as a single yum transaction, we'll be able to be less fine grained in how we fail, though.  So I wouldn't want to get people to get used to us being able to install all but one package and then be asked to implement that behaviour with the single-transaction code.

kustodian:
@abadger I don't like @queridiculo solution, since it only checks if the URL is valid, if it's not we still get the same result with check-mode as we have now (it would report as changed). In my case I would be up for downloading a whole package and checking if it's installed, since when we install packages from an URL they are usually repository packages and they are small, but I can see this being a problem if the package is big. On the other hand if check-mode is misleading like it is now and it reports that an RPM isn't installed, I would run that task again without check-mode so that it can install it, so I would have to wait for package to download again anyway.

I'm not sure if it's possible just to download part of the RPM to get it's name (I think it's not), but I have a different suggestion. Why not introduce an additional parameter for installing packages from URLs so that we can define their name. I'm not sure what is the best approach to this, because of backward compatibility, but for the sake of it, it could be something like this:

```
yum: name=https://somedomain.com/package2-something.rpm pkg_name=package2
```

In this case yum would check if the package is installed by looking up `pkg_name`. Not sure what exactly to do if it's set to latest, in that case I would probably download it and check versions, since if you want to do latest all the time, you probably want to wait in check mode as well.

Or we could leave the old way of doing things just because of backward compatibility and add a new parameter for installing packages from a URL:

```
yum: url=https://somedomain.com/package2-something.rpm name=package2
```

So this works exactly like the one above, but it's a new syntax for installing packages from URLs. If `url` is defined it is used as a source for installing the package and if `name` is defined as well, that name is used to check if the package is installed or not. If `name` is not defined, we download the package in check-mode to see if we need to install it.

So to sum it all up, if we are not going to add new parameters, I'm all for downloading packages in check-mode, if we are going to add a new parameter for URL downloading I would optimize things as described above for it, and in all other cases download in check-mode.

queridiculo:
Not sure if this distinction makes a difference, but HEAD actually checks to see if the file is available.

In my example an exception is triggered when the return code isn't 200 (file doesn't exist).

kustodian:
But it still doesn't tell you what is the package name, so it still cannot tell you in check-mode if the package is installed or not. The only improvement is that it would throw an error if the URL is invalid.

queridiculo:
I think we're on the same page here, just looking at it from different angles.

As @abadger mentioned in his comments, the first step would be to identify whether a package is installed.

Once that determination has been made we can move onto the next step which is either a download and install, or if in check_mode, a test to see if the package is available along with a result that includes the anticipated operation.

My example exclusively dealt with the latter aspect.

A more comprehensive example would look like this:

``` python
if module.check_mode:
    for spec in items:
        if '://' in spec:
            url = spec
            pkg = url.rsplit('/', 1)[1]
            base = url.rsplit('/', 1)[0]

            if is_installed(pkg):
                res['results'].append('%s is already installed' % pkg)
                continue

            request = urllib2.Request(url)
            request.get_method = lambda : 'HEAD'

        try:
            conn = urllib2.urlopen(request)
            conn.getcode()
            conn.close()
            msg = '%s will be installed from %s' % (pkg, base)
            res['results'].append(msg)
        except Exception, e:
            msg = 'Error downloading %s (%s)' % (url, e)
            module.fail_json(msg)
```

abadger:
@queridiculo the problem with only checking the http response code is that we don't know whether the file is a valid rpm, if it needs to be installed or if it's already installed, etc.  So we've traded speed for accuracy.  From @kustodian's description, it seems like accuracy is more important in this case.

yhekma:
I just came back from vacation and it looks like this is somewhat more work than I initially expected:)
I will look into this later today or tomorrow if this is still relevant

kustodian:
I would like to repeat myself and suggest how to solve this problem. We add a new parameter to `yum` called `url` which would work like this:

```
yum: url=https://somedomain.com/package2-something.rpm name=package2
```

If `url` is defined it is used as a source for installing the package, and if `name` is defined, that `name` is used in check-mode to see if the package is installed, but it should also check if the URL is valid (fetch HEAD). If `name` is not defined, we download the package in check-mode to see if we need to install it or not. Also because of backward compatibility if we pass an URL to the `name` parameter, we also download the package in check-mode to see if we will need to install it.

So to sum it all up, if we are not going to add new parameters, I'm all for downloading packages in check-mode, if we are going to add a new parameter for URL downloading I would optimize things as described above, and in all other cases download in check-mode.

wernerb:
I'm hitting this issue as well, and I agree with @kustodian to download the whole file in check mode. Because versions may change, naming might be unclear etc. All requires to download the file.

gregdek:
This one has been in needs_revision for 6 months, so I'm closing. @yhekma thanks for the PR, feel free to resubmit as you might see fit.

PR #1590, opened 2015-06-24
Add the ability to search file name based on regex

ritzk:
files/find,py is currently limited to glob search, which is very limited. 
- Adds a new parameter "pattern" which allows for regex search on the file name.
- "patterns" is retained to allow for glob search, by personally I would suggest this to be removed.


gregdek:
@bcoca can you review? Looks like you're the author. :)

bcoca:
i would actually not make this a new parameter and just allow regexs to be used as an item in patterns option.

ritzk:
replaced glob search with regex .

gregdek:
Back to @bcoca for review.

abadger:
@bcoca what did you mean by allowing regexes to be used?  Seems like we'd want glob for sysadmins who are looking for a way to do the equivalent of /bin/find -name.  To enable regexes as well we'd need some way to differentiate between a glob pattern and a regex pattern.

@ritzk Looks like in the newest commit there's some confusion as to whether pattern or patterns is what is being used.  Probably should wait for @bcoca's response to know what is desired before making too many changes, though.

bcoca:
this still adds a very confusing very similar option that now does the exact same thing, @ritzk what I meant the change to do was overload the existing 'patterns' and work with both shell globs and regexes. 

We should be able to detect or use both within the module w/o adding new options nor changing the default.

ritzk:
@bcoca  It can be hard to distinguish between regex and glob at times. Which is the reason for using pattern and patterns. e.g.,  "Test*.c" is a valid glob and regex.

gregdek:
Hm. So @ritzk are you saying that you basically can't make the changes requested by @bcoca here?

ritzk:
regex and glob can overlap in certain cases, and translate differently. 

glob -> .*.txt -> hidden text files 
regex -> .*.txt -> text files 

We would need to add an option such as "regex:<expression>"  ( assuming no one names there file with "regex:" which is valid on linux ).

Personally, use pattern to define regex, and patterns to define glob.

ritzk:
Failure unrelated to by patch .

Compiling ./system/cron.py ...
  File "./system/cron.py", line 241
    os.chmod(path, 0o644)
                       ^
SyntaxError: invalid syntax

Not sure, how to mark this in travis.

ritzk:
Fixed.

bcoca:
could you clean up the commit history? a bunch of merges are reverts in there

PR #1591, opened 2015-06-24
FIx issue with template when nothing is changed and the destination i…

dago:
…s a hardlink. This is a partial fix for ansible/ansible#9806

### DIFF
--- a/files/file.py
+++ b/files/file.py
@@ -247,7 +247,7 @@ def main():
     # src is None only if this is called from template. The state may also be
     # 'link' or 'hard' which defaults to the previous state if not explicitly set
     # in the above code section.
-    elif state == 'file' or (src is None and state == prev_state):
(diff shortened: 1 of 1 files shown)


gregdek:
@dago Thanks for your pull request. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

dago:
@gregdek Thanks for your heads up. I am also not confident that my patch is really a good idea as it basically works around some conditions in variable state instead of making the hardlink transition explicit. I would much favor a more detailed rewrite of the specific areas affected by hardlinks.

gregdek:
Hmm. Having reread the reference issue in ansible/ansible, it seems like pursuing a fix there will be preferable, and it's set at P2 so I believe that rewrite will happen at some point.  Any objection if we close this PR?

PR #1604, opened 2015-06-25
apt_repository: correct mishandling of PPAs.

verm666:
The original problem is: apt_repository.py connect to launchpad on
every playbook run. In this patch apt_repository.py checks if required
repository already exists or not. If no - paa will be added, if yes -
just skip actions.


gregdek:
Thanks @verm666!

@sashka can you review? If you give +1 we will promote to P1_reviewed for faster inclusion.

sashka:
:+1: 

Thanks @verm666.

gregdek:
Thanks all! Promoting to P1_reviewed.

PR #1605, opened 2015-06-25
Add "force" description

vladimirmartsul:
Add "force" option description

### DIFF
--- a/files/template.py
+++ b/files/template.py
@@ -47,6 +47,14 @@ options:
     required: false
     default: ""
     version_added: "1.2"
+  force:
+    description:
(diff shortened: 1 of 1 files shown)


gregdek:
I presume this is documentation of previously undocumented behavior? Since it's core functionality, and since it seems straightforward enough to verify, promoting to P1_reviewed.

vladimirmartsul:
Yes, it's core base functionality.

bcoca:
this option was actually not added to the module directly, its a side effect of it being part of copy, which template actually uses to push the file to the target server.

PR #1606, opened 2015-06-25
Correct awkward (perhaps accidentally copied into) description of validate option in template module

schlueter:


gregdek:
Thanks @schlueter.

bcoca:
We don't take strictly style updates to module as they break attribution, we do strive to be PEP8ish but we specifically ignore the 80 column limit as we find it to detract from redability, not increase it.

PR #1607, opened 2015-06-25
Adds basic authentication & skip certificate validation to win_get_ur…

abriening:
Adds basic auth and SSL cert skipping to win_get_url module.

### DIFF
--- a/windows/win_get_url.ps1
+++ b/windows/win_get_url.ps1
@@ -62,7 +62,7 @@ If ($force -or -not (Test-Path $dest)) {
         $result.changed = $true
     }
     Catch {
-        Fail-Json $result "Error downloading $url to $dest"
+        Fail-Json $result "Error downloading $url to $dest $($_.Exception.Message)"
     }
 }
 Else {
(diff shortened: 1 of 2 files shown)


gregdek:
@angstwad please review; if you give +1, we will promote to P1_reviewed.

angstwad:
@gregdek Doing just a visual inspection, this looks okay to me.  I'd prefer a Windows/PowerShell expert (I am not an expert :smile:) give their thumbs up before confirming.

gregdek:
Well then let's call in the cavalry. @trondhindenes @jhawkesworth?

trondhindenes:
Looks good. One thing (and call me picky):
`if($params.skip_certificate_validation) {`
Rememer that the object coming in via json params is a string. So whatever the skip_certificate_validation is set to here (even "false") will result in this expression getting validated to true. I recommend using the builtin convertto-bool function to convert the param into a proper boolean object and then perform your test on that. Something like:
`if(($params.skip_certificate_validation) -and (ConvertTo-bool $params.skip_certificate_validation))
{
#execute code
}`

gregdek:
I like picky. :)  Thanks @trondhindenes.

abriening:
@trondhindenes I'm using the Parse-Args from lib/ansible/module_utils/powershell.ps1; this should be a psobject parsed with  ConvertFrom-Json. The values shouldn't need to be converted. Or are you trying to handle quoted "true" and "false" in the config?

jhawkesworth:
Not looked at the code but I think the point is that even if
$params.skip_certificate_validation is set to "false" the if() test will
consider that true (because only an empty string would be false).

abriening:
Rebased and added the ConvertTo-bool; will now support boolean, 1/0, "on", "yes", "true", "1"

jhawkesworth:
Did a little testing on this this morning and is at least backwardly compatible
I confirm the requested revision have been made so I am happy to upvote this
@gregdek / @bcoca please
shipit

gregdek:
Thanks @jhawkesworth and other reviews, and thanks @abriening for the PR. Marking for inclusion.

PR #1609, opened 2015-06-26
[docker] Prevent unnecessary reloads due to the net parameter

smashwilson:
I'm seeing a failure in my test repository right now because the `net` parameter is inconsistently defaulted:

```
"reload_reasons": "net (bridge => )"
```

This triggers an unnecessary reload when the `net` parameter is unspecified.


gregdek:
Promoting to P1_reviewed since @smashwilson is one of the module's authors

PR #1613, opened 2015-06-26
unarchive: fix work with 0 bytes archives

### LINKED ISSUE #1575: unarchive: don't try to work on 0 byte files
When you call unarchive on an archive that's 0 bytes it will report "Failed to find handler to unarchive. Make sure the required command to extract the file is installed.", which is rather misleading.

A solution forward might be to check for 0 byte size before trying to work on the file.

verm666:
This change is to fix #1575

### DIFF
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -303,10 +303,7 @@ def main():
     # skip working with 0 size archives
     try:
         if os.path.getsize(src) == 0:
-            res_args = {
-                'changed': False
-            }
-            module.exit_json(**res_args)
+            module.fail_json(msg="Invalid archive '%s', the file is 0 bytes" % src)
     except Exception, e:
         module.fail_json(msg="Source '%s' not readable" % src)
 
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -300,6 +300,16 @@ def main():
     if not os.access(src, os.R_OK):
         module.fail_json(msg="Source '%s' not readable" % src)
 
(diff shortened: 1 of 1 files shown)


gregdek:
@pileofrogs please review; if you give +1 we will promote to P1_reviewed for faster inclusion.

files/unarchive.py
```diff
@@ -300,6 +300,16 @@ def main():
     if not os.access(src, os.R_OK):
         module.fail_json(msg="Source '%s' not readable" % src)
 
+    # skip working with 0 size archives
+    try:
+        if os.path.getsize(src) == 0:
+            res_args = {
+                'changed': False
```
bcoca: ~~I would add msg="0 byte archive found, nothing to do", so users are not puzzled by the lack of change.~~

actually this should be an error: invalid archive, the file is 0 bytes.

files/unarchive.py
```diff
@@ -300,6 +300,16 @@ def main():
...
+            }
+            module.exit_json(**res_args)
+    except Exception, e:
+        module.fail_json(msg="Source '%s' not readable" % src)
```
bcoca, line 308: append exception message to failure message

ashemedai:
I am not fully versed in current Ansible coding practices, but this looks like a very clean fix for the problem. Thanks! Hadn't had time yet so far to work on it myself, much appreciated.

gregdek:
@pileofrogs can you review?

PR #1615, opened 2015-06-26
Change formatting of description of name parameter

rndmh3ro:
Added punctuation and formatted --all-databases.


gregdek:
Thanks @rndmh3ro.

bcoca:
we don't take pure style updates as they break attribution, these options get read and transformed by a formatter and the final output is not exactly what you see here.

PR #1618, opened 2015-06-26
apache2_module documetation update

vroetman:
requires a2enmod and a2dismod


gregdek:
So these are requirements that were simply not documented? And is this just for 1.9 or for devel as well?

vroetman:
Yes, they were not documented. The python code checks for these binaries and complains if they do not exist.

So this apache2_module will not work on RHEL-6, but only on debian style apache setups.

vroetman:
Both for 1.9 and for devel. I put in two pull requests.

PR #1623, opened 2015-06-28
Add ability to chose a specific private hosted zone to modify

### LINKED ISSUE #1619: Route53 module cannot differentiate between multiple private zones for the same doman
Issue Type:

Bug Report

Ansible Version:

ansible 1.9.2 (stable-1.9 db5ea10ed2) last updated 2015/06/26 14:53:24 (GMT -700)
  lib/ansible/modules/core: (stable-1.9 ea913bada2) last updated 2015/06/26 14:06:49 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD f291e9fcc3) last updated 2015/06/26 14:53:29 (GMT -700)
  v2/ansible/modules/core: (detached HEAD 85c8a892c8) last updated 2015/06/26 14:53:29 (GMT -700)
  v2/ansible/modules/extras: (detached HEAD 70ea058563) last updated 2015/06/26 14:53:29 (GMT -700)

Environment:

N/A

Summary:

route53 module cannot differentiate between multiple private zones with the same domain name.

Steps To Reproduce:

Create a public zone with two private zones (attached to different VPCs). Attempt to update one of the two private zones.

Expected Results:

I'd expect to be able to specify which private zone to modify - instead there is a "private_zone" boolean which incorrectly assumes a domain would only ever have a single private zone.

Actual Results:

You will only be able to update one of the two private zones. The last private zone encountered in the AWS route53 response wins.

mikejquinn:
This PR adds the ability to specify a specific hosted zone ID when making modifications. With the previous code, if you had multiple private zones with the same name, only the final zone returned in the AWS request would be changed. The code still supports the `private_zone` parameter, but will now throw an error if multiple private zones exist.

fixes #1619

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -34,7 +34,7 @@ options:
       - The DNS zone to modify
     required: false
     default: null
-    aliases: []
+    version_added: 2.0
   zone_id:
     description:
       - The hosted zone ID of the zone to modify. Required if you have multiple private zones and want to modify one of them.
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -34,13 +34,12 @@ options:
       - The DNS zone to modify
     required: false
     default: null
-    version_added: 2.0
   zone_id:
     description:
       - The hosted zone ID of the zone to modify. Required if you have multiple private zones and want to modify one of them.
     required: false
     default: null
-    aliases: []
+    version_added: 2.0
   record:
     description:
       - The full DNS record to create or delete
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -32,7 +32,7 @@ options:
   zone:
     description:
       - The DNS zone to modify
-    required: true
+    required: false
     default: null
     aliases: []
   zone_id:
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -24,7 +24,7 @@ description:
 options:
   command:
     description:
-      - Specifies the action to take.  
+      - Specifies the action to take.
     required: true
     default: null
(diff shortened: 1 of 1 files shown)


cloud/amazon/route53.py
```diff
@@ -32,7 +32,13 @@
   zone:
     description:
       - The DNS zone to modify
-    required: true
+    required: false
+    default: null
+    aliases: []
+  zone_id:
```
bcoca, line 37: need version_added for new options
empty aliases can be left out

mikejquinn, line 37: What version would this be included in? 2.0?

bcoca, line 37: if merged soon, yes

gregdek:
Hi @mikejquinn -- I see that you made some revisions, thanks for those. In the meantime, it looks like we've developed some merge conflicts; can you please rebase so we can get it back into review?

mikejquinn:
Hi @gregdek - it looks like there were numerous pull requests dealing with this same issue. Both of these seem to achieve the same goal through different module options (and have already been merged):

https://github.com/ansible/ansible-modules-core/pull/1140
https://github.com/ansible/ansible-modules-core/pull/1385

IMO the addition of both a `vpc_id` _and_ a `hosted_zone_id` option is confusing and unnecessarily complicates the code. #1140 takes the same approach I did - explicitly specifying your hosted zone ID. I believe my PR handles some invalid/ambiguous input cases that #1140 does not, though.

gregdek:
@mikejquinn thanks for the feedback. We'll go ahead and close this one since #1140 has been merged. If you want to factor out your patch for the invalid/ambiguous cases and make another PR, that would be great.

PR #1624, opened 2015-06-28
Add ability to chose a specific private hosted zone to modify

mikejquinn:
This PR adds the ability to specify a specific hosted zone ID when making modifications. It is a port of https://github.com/ansible/ansible-modules-core/pull/1623 for the 1.9 branch.


bcoca:
we don't add features to the stable branch, only fixes, which we normally cherry pick so there is no need to open additional PRs

PR #1626, opened 2015-06-28
Fixed bug, when MX records pointed to ORIGIN(@) caused unintentional …

kmadac:
If we have domain which has only one A Origin record , everything works fine. 
But , when domain managed by Digital Ocean has MX records which points to Origin (@), calling task digital_ocean_domain will cause unintentional change of one of the A records. For example if we have following records in domain testdom.com:

```
A @  1.1.1.1
A   mail   2.2.2.2
A  webmail 2.2.2.2
MX @ "v=spf1 mx ~all" 
```

and we use task which should finished in OK state because domain testdom.com exists with corrct origin record:

```
- name: "ensure that domain testdom.com exists in DO DNS"
  digital_ocean_domain: >
    client_id=xxx
    api_key=xxx
    state=present
    ip=1.1.1.1
    name=testdom.com
```

webmail A record is changed to 1.1.1.1, which is not correct and can negatively affect production systems. 

Root cause for this bug is that algorithm searches for Origin record in all records and not only in A records.

### DIFF
--- a/cloud/digital_ocean/digital_ocean_domain.py
+++ b/cloud/digital_ocean/digital_ocean_domain.py
@@ -195,7 +195,7 @@ def core(module):
             records = domain.records()
             at_record = None
             for record in records:
-                if record.name == "@":
+                if record.name == "@" and record.record_type == 'A':
                     at_record = record
 
             if not at_record.data == getkeyordie("ip"):


gregdek:
@mgregson Please review this PR to make sure it adheres to the following guidelines:

If it passes these guidelines, and if you believe it’s a good PR otherwise, please give it a +1; at that point, we will add the “P1_reviewed” flag to this PR, which will move it to the top of our queue for potential inclusion.

kmadac:
Any update on this issue? Is it going to be merged?

bcoca:
ping @mgregson, quick review?

gregdek:
ping @mgregson, can you review?

mgregson:
I'll take a look. Sorry for the delay.

mgregson:
:+1: Looks good to me.

gregdek:
Thanks @kmadac for the PR, and thanks @mgregson for the review. 

Marking for potential inclusion. cc: @ansible/core team.

PR #1628, opened 2015-06-28
ec2_vpc_net refactor

wimnat:
- doc updates
  - added choices for bool options
  - removed empty alias
  - corrected examples to use correct module name
- moved attribute check to start of module so it fails before making any changes (if dns_hostnames and not dns_support:)
- passed vpc object instead of vpc id from vpc_exists method.  ID can be accessed as value of object and allows other attributes to be included in exit_json statement
- removed vpc_needs_update method in favor of checking each attribute individually.  This allows exit_json to be called with the same attributes every time
  - misc spelling corrections

### DIFF
--- a/cloud/amazon/ec2_vpc_net.py
+++ b/cloud/amazon/ec2_vpc_net.py
@@ -17,10 +17,11 @@
 DOCUMENTATION = '''
 ---
 module: ec2_vpc_net
-short_description: configure AWS virtual private clouds
+short_description: Configure AWS virtual private clouds
 description:
-    - Create or terminates AWS virtual private clouds.  This module has a dependency on python-boto.
+    - Create or terminate AWS virtual private clouds.  This module has a dependency on python-boto.
 version_added: "2.0"
+author: Jonathan Davila (@defionscode)
 options:
   name:
     description:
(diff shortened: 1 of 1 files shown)


gregdek:
@carsongee Please review this PR to make sure it adheres to the following guidelines:

wimnat:
@defionscode PR for your review :)

juliedavila:
@wimnat excellent! Testing now.

juliedavila:
Just one finding.

Tags cannot be changed on a VPC that already exists.

Create a vpc with 

``` yml
     - name: VPC Module Test
       ec2_vpc_net:
         name: Module Testing
         cidr_block: 170.10.0.0/16
         region: us-east-1
         tags:
           new_vpc: ec2_vpc_module
           this: works232
         state: present
       register: vpc_test

     - debug: var=vpc_test
```

Then just modify one of the tags, rerun, it will return 'ok' without any change.

wimnat:
@defionscode sorry, couldn't reproduce this.  When i set a new tag i get a changed and the new is present.

wimnat:
@defionscode should be ok now

halberom:
Bit of inconsistency - All the other vpc modules use `resource_tags` as the param, this one uses `tags`.  Cheers.

wimnat:
Only ec2_vpc uses resource_tags but all other ec2 modules just use tags.

As ec2_vpc is likely to be deprecated now this module has arrived, i think keeping tags is the right move.  I could add an alias?

halberom:
Sounds good to me, but should do the same in ec2_vpc_subnet, ec2_vpc_route_table etc for consistency - they currently use resource_tags with no alias, and are the ones I was referring to when I said 'all other vpc modules'.  I can make a note on those PRs if you'd like me to?

wimnat:
@halberom alias added

@defionscode can you please +1 so we can merge this?

juliedavila:
Yep looks good now, +1

PR #1630, opened 2015-06-29
fixed a bug with calculation md5 sum

UvaK:
the bug appears for files > 1KiB

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -318,7 +318,7 @@ def is_walrus(s3_url):
 def get_md5_digest(local_file):
     md5 = hashlib.md5()
     with open(local_file, 'rb') as f:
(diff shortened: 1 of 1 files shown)


gregdek:
@lwade @ralph-tice please review; if you +1 we will promote to priority_reviewed

bcoca:
it might be better to just use the common md5 function in the shared basic.py.

gregdek:
@bcoca is that recommendation firm enough to make it a requirement?

bcoca:
it is not what this PR means to do, but the PR would not be needed as the shared function already does buffered reads.

gregdek:
Thanks @UvaK for submitting this PR, but it appears as though this code has been entirely rewritten and this PR no longer applies. Closing.

PR #1632, opened 2015-06-29
Add OpenStack Floating IP module

emonty:


gregdek:
@emonty putting this into community_review for bookkeeping purposes.

abadger:
@j2sol sol I've reviewed this one and the code looks good.  Let me/us know when you've had a chance to take a look and we can merge.

abadger:
@emonty Found two documentation problems: there's a `default false` and `default 60`.  Both of these need colons.

omgjlk:
I think this looks fine, good enough to merge. As a close follow up feature I'd like to see an option to delete the floating IP when detaching it, essentially removing it from the allocation pool. We see people all the time consume all the floating IPs into one project not realizing that they need to delete as well as detach. It could be a simple option that's acted on in the state=absent case. But again, not before this initially gets merged.

omgjlk:
This could document the return value too, like we did with previous os_\* modules recently.

gregdek:
Putting in needs_revision until I see that the docs problems have been fixed.

emonty:
@gregdek chatted with @j2sol - this is ready to go now

PR #1634, opened 2015-06-29
Change uri debug example

jmeickle:
As per the httplib2 docs, the right attribute to set is 'debuglevel', not 'debug'. https://code.google.com/p/httplib2/wiki/Examples

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -269,7 +269,7 @@ def url_filename(url):
 
 def uri(module, url, dest, user, password, body, body_format, method, headers, redirects, socket_timeout, validate_certs):
     # To debug
-    #httplib2.debug = 4
+    #httplib2.debuglevel = 4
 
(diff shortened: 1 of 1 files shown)


jmeickle:
(But obligatory, "why isn't this using requests yet?" - I needed to look up this value because httplib2 is choking on a proxy that every other client I've tried handles properly.)

abadger:
My understanding is that we didn't want to add extra dependencies to modules that we considered to implement basic features.  However (unlike get_url) the uri module does use a third-party library.  So it's probably a question of whether this could be ported to ansible.module_utils.urls.fetch_url().  If not, requests probably is a better third party library to standardize on than httplib2.  Would need some thought as to which way to go and then for someone to make the patch that implements the decision.

Thanks for fixing the comment.

# Merged

Hi!

issue is resolved.

Thank you!

Issue #1460, opened 2015-06-03
cloudformation: cloudformation template file is not found when using the module in a role, relative to the role


viper233:
Issue Type: Bug Report
Ansible Version: latest devel branch
Ansible Configuration: as per the github file
Environment: Ubuntu 15.04 GNU/Linux x86_64
Summary:
When using the cloudformation module as a task in a role it's unable to locate the cloudformation template file relative to the role directory i.e. role_name/files/cf_template.json

roles/cf_role/tasks/main.yml

```
- name: launch cloudformation stack
  cloudformation:
    region: "{{ aws_region }}"
    state: "{{ stack_state }}"
    stack_name: "{{ stack_name }}"
    disable_rollback: true
    template: cf_template.json
    template_parameters:
        AmiId: "{{ AmiId }}"

template_body = open(module.params['template'], 'r').read()\nIOError: [Errno 2] No such file or 
directory: 
```

I was expecting the file to made available from either the templates/ or files/ sub-directory of the role. files would make sense though I could see a case for a template being used also.

bennojoy:
Module wont know if it being called from a role, so the way to do it woudl be teach the module to accept the json content and then use lookup('file', foo) to read the data, since file lookup knows about roles. this woudl be a feature request.

bcoca:
actually modules cannot access data from the 'master' or local host unless it is the target of the task.
that is what lookups are for

bcoca:
changing this to a doc ticket as this should be made clear in the module docs.

ansibot:
@jsmartin, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@jsmartin, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@jsmartin, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@tedder, @ryansb, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

tedder:
[...]

PR #1635, opened 2015-06-29
update docs for cloudformation

smiller171:
documentation update for #1460


bcoca:
actually, this is a bit misleading, the path is from the current working directory on the target host, you just happen to be using a 'local' connection.

smiller171:
@bcoca does that look more accurate?

bcoca:
yes, though I'm tempted to just create an action module to allow for good relative paths

smiller171:
@bcoca I think that this would be ideal

PR #1640, opened 2015-06-30
Added a new state 'recreated' for docker module

ArtemChekunov:
"recreated" it is sequence of states "absent" and "started",  you can to use it if you need upgrade version of container.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -204,7 +204,8 @@ options:
         configuration out of date. "restarted" unconditionally restarts (or
         starts) the matching containers. "stopped" and '"killed" stop and kill
         all matching containers. "absent" stops and then' removes any matching
-        containers.
+        containers. "recreated" sequence of states "absent" and "started", 
+        you can to use it if you need upgrade version of container.
     required: false
     default: started
     choices:
@@ -215,6 +216,7 @@ options:
       - stopped
       - killed
       - absent
+      - recreated
   privileged:
     description:
       - Whether the container should run in privileged mode or not.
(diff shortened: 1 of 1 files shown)


gregdek:

smashwilson:
Hmm. What's the actual difference between `state: recreated` and `state: restarted`? We already have seven possible states for this module, so I'd like to be careful about adding additional ones.

I _suspect_ that `state: restarted` is restarting an existing container with the exact configuration that was used to start it in the first place, including a possibly-obsolete configuration and image. If that's the case, I could see an argument for just outright replacing `state: restarted` with the behavior you've written as `restarted`. On the other hand, while you _can_ replicate `state: recreated` by composing existing states...

``` yaml
- name: recreate a docker container
  docker:
    name: ohno
    image: someuser/oldandbusted
    state: "{{ item }}"
  with_items:
  - absent
  - started
```

... you _can't_ replicate the existing `state: restarted` the same way. I suppose it comes down to whether or not `state: restarted` is useful to anyone as it is.

Do any of the other `docker` module users have opinions about this?

ArtemChekunov:
State restarted just restart a container.
But i need to restart container with latest image.
I can to use nest construction, it is true: 
with_items: 
- absent 
- started

but this construction don't declarative

gregdek:
@smashwilson any further opinion on how to proceed here? Leaving in needs_info for now.

joshuaconner:
I think I am with @smashwilson on this. While

```
with_items:
- absent
- started
```

isn't declarative, it's a fundamentally non-declarative operation, so I think makes sense for this set of actions to not be first-class supported in the Docker module.

gregdek:
Thanks @sc0rp1us for the PR, but two -1s from maintainers justifies us closing this PR without merging. Apologies.

PR #1644, opened 2015-06-30
Fix indent in os_keypair docs

omgjlk:


sivel:
Partial dupe of https://github.com/ansible/ansible-modules-core/pull/1642

omgjlk:
So it is.

PR #1651, opened 2015-07-01
fix authorized_keys in check_mode

verm666:
This change is in response to issue #1515.
Original pull request #1580.

The original problem is: in authorized_key module you have no idea about users
which will be created by Ansible at first run. I can propose next two ways to
solve this problem:
1. Combine modules system/user.py and system/authorized_key.py in one module
    (so you will know everything about users in that module)
2. Use small workaround: add my commit and always provide 'path' parameter
    for authorized_key module during runs with --check option.

### DIFF
--- a/system/authorized_key.py
+++ b/system/authorized_key.py
@@ -138,7 +138,7 @@ import shlex
 class keydict(dict):
 
     """ a dictionary that maintains the order of keys as they are added """
-    
+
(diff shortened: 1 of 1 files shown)


abadger:
Thanks @verm666 !

# Merged

Hi!

issue is resolved.

Thank you!

Issue #133, opened 2014-10-05
apt_repository performance issue


johnthedebs:
This is still an issue (ansible 1.7.2 on OS X), original conversation here: https://github.com/ansible/ansible/issues/8779

ansibot:
# Can You Help Us Out?

Thanks for filing a ticket! I am the friendly GitHub Ansibot. 

It looks like you might not have filled out the issue description based on our standard issue template.  You might not have known about that, and that's ok too, we'll tell you how to do it.

We have a standard template because Ansible is a really busy project and it helps to have some standard information in each ticket, and GitHub doesn't yet provide a standard facility to do this like some other bug trackers.  We hope you understand as this is really valuable to us!. 

Solving this is simple: please copy the contents of this [template](https://raw.githubusercontent.com/ansible/ansible/devel/ISSUE_TEMPLATE.md) and **paste it into the description** of your ticket.  That's it!

# If You Had A Question To Ask Instead

If you happened to have a "how do I do this in Ansible" type of question, that's probably more of a user-list question than a bug report, and you should probably ask this question on the project mailing list instead.  

However, if you think you have a bug, the report is the way to go!  We definitely want all the bugs filed :)  Just trying to help!

# About Priority Tags

Since you're here, we'll also share some useful information at this time.

In general tickets will be assigned a priority between P1 (highest) and P5, and then worked in priority order. We may also have some follow up questions along the way, so keeping up with follow up comments via GitHub notifications is a good idea.

Due to large interest in Ansible, humans may not comment on your ticket immediately.  

# Mailing Lists

If you have concerns or questions, you're welcome to stop by the ansible-project or ansible-development mailing lists, as appropriate.  Here are the links:
- https://groups.google.com/forum/#!forum/ansible-project - for discussion of bugs and how-to type questions
- https://groups.google.com/forum/#!forum/ansible-devel - for discussion on how to implement a code change, or feature brainstorming among developers

Thanks again for the interest in Ansible!

Jmainguy:
10 seconds seem pretty accurate, same speed to add as it is to remove.
I tested from a centos7 host against a ubuntu14.04 target, using latest ansible 1.7.2

time ansible-playbook -i hosts apt.yml 

PLAY [apt performance] *******************************************************\* 

GATHERING FACTS **************************************************************\* 
ok: [ubuntu1404.soh.re]

TASK: [apt | MongoDB | Install ppa] ******************************************\* 
[...]

PR #1671, opened 2015-07-04
Backport issue #133 fix to stable-1.9 branch

msabramo:
This takes d435d5ce0ae2597fdde4600dd07edbb8c9c4fdfe from the devel branch and backports it to the `stable-1.9` branch, so the issue can be fixed in a future 1.9 release.

I had deploys fail the other day on apt_repository because it was attempting to hit launchpad.net to validate a cert for a PPA that already existed on the system. So this would be a nice fix to have. Original commit message from d435d5ce0ae2597fdde4600dd07edbb8c9c4fdfe follows below...

---

This change is in response to issue #133.

The original problem is: apt_repository.py connect to launchpad on
every playbook run. In this patch apt_repository.py checks if required
repository already exists or not. If no - paa will be added, if yes -
just skip actions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ addons:
       - python2.4
       - python2.6
 script:
-  - python2.4 -m compileall -fq -x 'cloud/|/accelerate.py' .
   - python2.4 -m compileall -fq cloud/amazon/_ec2_ami_search.py cloud/amazon/ec2_facts.py
   - python2.6 -m compileall -fq .
   - python2.7 -m compileall -fq .
--- a/packaging/os/apt_repository.py
+++ b/packaging/os/apt_repository.py
@@ -267,14 +267,18 @@ class SourcesList(object):
     def dump(self):
         return '\n'.join([str(i) for i in self])
 
+    def _choice(self, new, old):
+        if new is None:
+            return old
+        return new
+
     def modify(self, file, n, enabled=None, source=None, comment=None):
         '''
         This function to be used with iterator, so we don't care of invalid sources.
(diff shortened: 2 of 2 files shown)


bcoca:
seems you got an extra commit added

msabramo:
Oh I also added e63e846 so that the Travis CI tests would pass, but I forgot to mention it. Want me to remove it?

abadger:
I'm fixing this but the changes to .travis.yml are wrong.  It looks like you might have had a conflict there when you cherry-picked and merged it incorrectly.

jstasiak:
My original comment: https://github.com/ansible/ansible-modules-core/issues/133#issuecomment-119273206

Copied content for visibility:

This change may have introduced a regression.

I'm running ansible/ansible@d198b18
Host OS: Debian Wheezy

Task:

```
apt_repository: repo='deb http://httpredir.debian.org/debian wheezy-backports main'
```

Result:

```
  File "/root/.ansible/tmp/ansible-tmp-1436287935.27-182925899891693/apt_repository", line 2836, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1436287935.27-182925899891693/apt_repository", line 454, in main
    if state == 'present' and expanded_repo not in sourceslist.repos_urls:
AttributeError: 'SourcesList' object has no attribute 'repos_urls'
```

It seems like this only works on Ubuntu.

abadger:
Opened a new bug report for the problem as this report is closed: https://github.com/ansible/ansible-modules-core/issues/1684

abadger:
Opened a new bug report for the problem as this report is closed: #1684

PR #1672, opened 2015-07-04
New module for placement group in AWS EC2

szibis:
- Addi/Delete/List placement group in EC2 region
- List all placement groups in EC2 region

### DIFF
--- a/cloud/amazon/ec2_placement.py
+++ b/cloud/amazon/ec2_placement.py
@@ -1,0 +1,157 @@
+#!/usr/bin/python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: ec2_placement
+short_description: create, remove or list placement group(s) in EC2.
+description:
+    - Creates, removes and lists placement groups in EC2. This module has a dependency on python-boto.
+version_added: "2.0"
+options:
+  name:
+    description:
+      - The EC2 placement group name.
+    required: False
+    default: null
+    aliases: []
+  strategy:
+    description:
+      - The placement strategy of the new placement group. Currently, the only acceptable value is cluster
+      required: false
+    default: cluster
+    choices: ['cluster']
+    aliases: []
+  state:
+    description:
+      - Whether the placement group should be present or absent. Use list to show all or search specific placement group.
+    required: false
+    default: present
+    choices: ['present', 'absent', 'list']
+    aliases: []
+  region:
+    description:
+      - region in which the resource exists.
+    required: false
+    default: null
+    aliases: ['aws_region', 'ec2_region']
+
+author: "Slawomir Skowron (@szibis)"
+extends_documentation_fragment: aws
+'''
+
+EXAMPLES = '''
+# Basic example of adding placement group
+tasks:
+- name: add placement group
+  ec2_placement:
+      name: "example_group"
+      state: "present"
+      region: "us-east-1"
+
+# Basic example for delete placement group
+tasks:
+- name: remove placement group
+  ec2_placement:
+      name: "example_group"
+      state: "absent"
+      region: "us-east-1"
+
+# Basic example for listing all placement groups
+tasks:
+- name: list all placement group
+  ec2_placement:
+      state: "list"
+      region: "us-east-1"
+
+# Basic example for listing specific placement groups
+tasks:
+- name: list placement group
+  ec2_placement:
+      name: "example_group"
+      state: "list"
+      region: "us-east-1"
+'''
+
+import sys
+
+try:
+    import boto.ec2
+    HAS_BOTO = True
+except ImportError:
+    HAS_BOTO = False
+
+
+def main():
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(dict(
+            name = dict(required=False),
+            strategy = dict(default='cluster', choices=['cluster']),
+            state = dict(default='present', choices=['present', 'absent', 'list']),
+        )
+    )
+    module = AnsibleModule(argument_spec=argument_spec)
+
+    if not HAS_BOTO:
+        module.fail_json(msg='boto required for this module')
+
+    name = module.params.get('name')
+    strategy = module.params.get('strategy')
+    state = module.params.get('state')
+
+    if not isinstance(name, basestring) and name:
+       module.fail_json(msg="name argument must be string")
+
+    ec2 = ec2_connect(module)
+
+    filters = {'group-name' : name}
+    if name:
+       getplacements = ec2.get_all_placement_groups(filters=filters)
+    else:
+       getplacements = ec2.get_all_placement_groups()
+
+    placelist = []
+
+    for placement in getplacements:
+        placelist.append(placement.name)
+
+    if state == 'present':
+        if not name:
+            module.fail_json(msg="name argument is required when state is present")
+        if name in placelist:
+            module.exit_json(msg="Placement group already exists with name %s." %name, changed=False)
+        else:
+             placer = ec2.create_placement_group(name)
+        module.exit_json(msg="Placement group %s created." % (name), changed=True)
+
+    if state == 'absent':
+        if not name:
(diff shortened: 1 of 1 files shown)


gregdek:
Hmm, are we sure we want this to start life in core and not extras?

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,157 @@
...
+      required: false
+    default: cluster
+    choices: ['cluster']
+    aliases: []
+  state:
+    description:
+      - Whether the placement group should be present or absent. Use list to show all or search specific placement group.
+    required: false
+    default: present
+    choices: ['present', 'absent', 'list']
```
bcoca, line 44: list is a bad pattern that some of the amazon modules already have, I would prefer to move this to the ec2_facts modules or alternatively always return the info

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,157 @@
...
+description:
+    - Creates, removes and lists placement groups in EC2. This module has a dependency on python-boto.
+version_added: "2.0"
+options:
+  name:
+    description:
+      - The EC2 placement group name.
+    required: False
+    default: null
+    aliases: []
```
bcoca: no need to list empty aliases

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,157 @@
...
+  strategy:
+    description:
+      - The placement strategy of the new placement group. Currently, the only acceptable value is cluster
+      required: false
+    default: cluster
```
bcoca, line 36: required is not indented correctly

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,157 @@
...
+version_added: "2.0"
+options:
+  name:
+    description:
+      - The EC2 placement group name.
+    required: False
+    default: null
+    aliases: []
+  strategy:
+    description:
```
bcoca, line 32: is this also a placeholder in the aws side?

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,157 @@
...
+            module.fail_json(msg="name argument is required when state is absent")
+        if name not in placelist:
+            module.exit_json(msg="Placement group with name %s not exist. Nothing to remove." %name, changed=False)
+        else:
+             placer = ec2.delete_placement_group(name)
+        module.exit_json(msg="Placement group %s removed." % name, changed=True)
+
+    if state == 'list':
+        module.exit_json(changed=False, placement_groups=placelist)
+    sys.exit(0)
```
bcoca: modules should never use exit, always exit_json or fail_json

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,157 @@
...
+    if not HAS_BOTO:
+        module.fail_json(msg='boto required for this module')
+
+    name = module.params.get('name')
+    strategy = module.params.get('strategy')
+    state = module.params.get('state')
+
+    if not isinstance(name, basestring) and name:
+       module.fail_json(msg="name argument must be string")
+
```
bcoca, line 128: you should handle exceptions thrown by boto and then give appropriate error messages on top.

cloud/amazon/ec2_placement.py
szibis, line 44: I understand, but in ec2_facts needs to have more background code. What info i should return for example ?

cloud/amazon/ec2_placement.py
szibis, line 32: Just like in boto doc: http://boto.readthedocs.org/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.create_placement_group

cloud/amazon/ec2_placement.py
szibis, line 128: Yes sure, but from my tests i just add here a list or anything else and boto converting this into string and use as name. This should not be allowed, thats why i allow only values as string in name field from ansible side. Maybe i should allow more freedom in this case ?

gregdek:
@szibis thanks for all the revisions. 

Pinging AWS reviewers for further review: @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode @tastychutney @bpennypacker @loia

juliedavila:
initial glance over this looks good, will do functional testing now.

Also, it would be _very_ ideal if you could add a `RETURN:` section to the documentation. You can see an example here: https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/apt.py#L141-L162

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,166 @@
...
+    import boto.ec2
+    HAS_BOTO = True
+except ImportError:
+    HAS_BOTO = False
+
+
+def main():
+    argument_spec = ec2_argument_spec()
+    argument_spec.update(dict(
+                         name=dict(required=False),
```
willthames: This indentation isn't right. Even if it were right, use similar indentation to all the other ec2 modules and it'll be a lot easier to read.

Even if the lines exceed standard PEP-8, it'll still be a lot nicer.

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,166 @@
...
+    if state == 'present':
+        if not name:
+            module.fail_json(msg="name argument is required when state is \
+                             present")
+        if name in placelist:
+            module.exit_json(msg="Placement group already exists with name \
+                             %s." % name, changed=False)
+        else:
+            placer = ec2.create_placement_group(name)
+        module.exit_json(msg="Placement group %s created." % name,
```
willthames: The exit should return something useful. In this case, that might be name and strategy. 

Whatever the results of `state: list` return, the module should return similar results when creating a placement group.

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,166 @@
...
+    module = AnsibleModule(argument_spec=argument_spec)
+
+    if not HAS_BOTO:
+        module.fail_json(msg='boto required for this module')
+
+    name = module.params.get('name')
+    strategy = module.params.get('strategy')
+    state = module.params.get('state')
+
+    if not isinstance(name, basestring) and name:
```
willthames, line 126: Use `type='str'` in the `argument_spec` for `name` if you want this validation.

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,166 @@
...
+        module.fail_json(msg="name argument must be string")
+
+    ec2 = ec2_connect(module)
+
+    filters = {'group-name': name}
+    if name:
+        getplacements = ec2.get_all_placement_groups(filters=filters)
```
willthames: You could just use `filters={'group-name': name}` rather than define `filters` when it won't be used.

cloud/amazon/ec2_placement.py
```diff
@@ -0,0 +1,166 @@
...
+    else:
+        getplacements = ec2.get_all_placement_groups()
+
+    placelist = []
+
+    for placement in getplacements:
```
willthames: A list comprehension seems appropriate here:

```
placelist = [placement.name for placement in getplacements]
```

cloud/amazon/ec2_placement.py
juliedavila: Agreed

juliedavila:
@szibis friendly ping

szibis:
Sorry for late response, but i was very busy with current projects.

cloud/amazon/ec2_placement.py
szibis, line 126: fixed

gregdek:
@szibis thanks again for submitting this new module, and sorry it's taken so long for it to be reviewed.

The issue here is that new modules really should be coming through the Extras process, and the reason this module isn't receiving more attention is because no one knows to review it here.

I'm sorry for the imposition, but please resubmit this module as a new PR to the Ansible Extras repository, and reference this pull request.  Since you already have approval from a couple of AWS committers, we should be able to merge it fairly quickly. 

Ansible Extras repo: https://github.com/ansible/ansible-modules-extras

vanga:
Hello,

I am wondering what happened to this PR, I don't see it merged either in `core` or `extras`.

Thanks.

willthames:
Hi @vanga, it looks like it got closed back in 2015. This repo has since been closed.

This should be resubmitted to the main ansible/ansible repo, preferably maintaining @szibis's original commits and then bringing them up to recent standards (ANSIBLE_METADATA, RETURNS, new imports style, etc.) - if you're not able to do this, let me know, and I will try to find time to do it myself, as I'm really not sure how it's not already a module!

vanga:
Hi @willthames 

Thanks for responding, I have actually mistaken Placement Groups with Target Groups in AWS. I needed a module for later. 

Considering that I don't need this I will skip this for now.

sar009:
@willthames I was not able to find this in main repo either

willthames:
@sar009 I will look into what happened with that, you're right.

Edit: I was thinking this was merged, it has never been merged. It needs resubmitting on ansible/ansible as a new module PR.

PR #1683, opened 2015-07-07
Fix group mod and group add for FreeBSD

sean-:
Fix a regression from previous versions of ansible.  `pw(1)` expects two different arguments to argv, not a single argument separated by a space.


bcoca:
backported to 1.9 tree so it should go out with 1.9.3 also

sean-:
Much appreciated.

PR #1689, opened 2015-07-09
Added support for instance types.

### LINKED ISSUE #1704: Support for type and version parameters in rax_cdb doesn't to create instance types besides mysql 5.6
##### Issue Type:

“Feature Pull Request”.
##### Ansible Version:

$ ansible --version
ansible 1.9.2
##### Environment:

N/A
##### Summary:

Currently rax_cdb module doesn't allow type and version parameters, which are supported by pyrax. Without these parameters, one can only create a default instance of type MySQL 5.6. Other available options are MySQL 5.1, MariaDB 10, Percona 5.6. It is not possible to create the other instance types without these parameters.
I have submitted a PR to add support for these parameters at #1689.

mehulved:
Added type and version parameters to be passed to rax_cdb module
so users can create Percona and MariaDB instance types.

fixes #1704

### DIFF
--- a/cloud/rackspace/rax_cdb.py
+++ b/cloud/rackspace/rax_cdb.py
@@ -38,6 +38,14 @@ options:
     description:
       - Volume size of the database 1-150GB
     default: 2
+  type:
+    description:
+      - type of instance (i.e. MySQL, MariaDB, Percona)
+    default: MySQL
+  version:
+    description:
+      - version of database (MySQL supports 5.1 and 5.6, MariaDB supports 10, Percona supports 5.6)
+    choices: ['5.1', '5.6', '10']
   state:
     description:
       - Indicate desired state of the resource
@@ -68,6 +76,8 @@ EXAMPLES = '''
         name: db-server1
         flavor: 1
         volume: 2
+        type: MySQL
+        version: 5.6
         wait: yes
         state: present
       register: rax_db_server
@@ -91,10 +101,12 @@ def find_instance(name):
     return False
 
 
-def save_instance(module, name, flavor, volume, wait, wait_timeout):
+def save_instance(module, name, flavor, volume, type, version, wait,
+                  wait_timeout):
 
     for arg, value in dict(name=name, flavor=flavor,
-                           volume=volume).iteritems():
+                           volume=volume, type=type, version=version
+                           ).iteritems():
         if not value:
             module.fail_json(msg='%s is required for the "rax_cdb"'
                                  ' module' % arg)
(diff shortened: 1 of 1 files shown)


angstwad:
Hey Ansible peeps and @mehulved.  

Just at first glance, I'm concerned with the use of `type` as a kwarg, shadowing the built-in name. 

 Also, I would prefer to limit the list of DB instance types to the currently support instance types at Rackspace.   The current supported instances are [here](http://docs.rackspace.com/cdb/api/v1.0/cdb-devguide/content/Datastore_Types_and_Versions-d1e9263.html), so your list is complete; but still, I'd rather see it as a restrictive list of choices.

mehulved:
@angstwad your concern for `type` kwarg is justified and I will send another PR with that corrected.
As to restricting DB instance types. The problem is, there is no way to query the list of types and versions which are supported. At best, it can be hardcoded, similar to volume limits. If that is a good enough practice, I can add that.

angstwad:
After thinking about it, a curated list should not be a requirement as long as we document WHERE to find the supported instances.

gregdek:
@jails Please review this PR to make sure it adheres to the following guidelines:

jails:
According to your checklist @gregdek I only saw some minor issues:

`default: null` for the name parameter should be replaced by `required: true` imo
https://github.com/mehulved/ansible-modules-core/blob/devel/cloud/rackspace/rax_cdb.py#L32

`main()` should be called like in other modules (https://github.com/ansible/ansible-modules-core/blob/devel/cloud/docker/docker.py) with:

``` python
if __name__ == '__main__':
    main()
```

And at the bottom, `rax_cdb(module, state, name, flavor, volume, cdb_type, cdb_version, wait, wait_timeout)` doesn't strictly follow the pep8 column limitation of 80 character max.

Otherwise :+1: !

bcoca:
@jails, you dont need the if for main, the modules are not importable due to the ansible.modules.basic import is not a real python import

jails:
@bcoca thanks for the head up !

gregdek:
@jails thanks for the review! Sorry it's taken a while to get back to your review. :/

Do you think this issue:

> default: null for the name parameter should be replaced by required: true imo
> https://github.com/mehulved/ansible-modules-core/blob/devel/cloud/rackspace/rax_cdb.py#L32

...is a blocker, or no?

If it is a blocker, please add a comment with "needs_revision" in the text, along with the needed changes, and we will flag it for revision.

If it's not a blocker, please add a comment with "shipit" in the text, and we will flag it for inclusion.

Thanks!

jails:
Ha yeah you're right @gregdek `default: null` should be `required: true` but it's not related to @mehulved changes so it's ok to shipit imo.
Indeed if it was a blocker my initial PR wouldn't have been accepted in the first place.

gregdek:
Thanks @mehulved for the PR, and thanks @jails for the review. Marking for potential inclusion. cc: @ansible/core.

PR #1692, opened 2015-07-09
Add backport of parameters from ansible 2.0

emonty:
The modules have been developed against devel of ansible, but there are
folks using them with the current 1.9 release. That's fine - they mostly
work. However, there have been a few changes in the openstack
module_utils that seemed fine becasue the modules weren't released in
1.9.

Backport/forwardport these to the modules temporarily with a bunch of
copypasta so that folks using the modules against 1.9 still get working
modules. Once 2.0 is released, remove the function from everything.

### DIFF
--- a/cloud/openstack/os_server.py
+++ b/cloud/openstack/os_server.py
@@ -412,7 +412,7 @@ def main():
             ['image', 'root_volume'],
         ],
     )
-    module = AnsibleModule(argument_spec, **module_kwargs)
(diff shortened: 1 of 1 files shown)


abadger:
(There's a typo in this specific patch -- augment_spec on line 455)

How does this align with backporting the module_utils/openstack.py argument_spec for the 1.9.3 release?

emonty:
AH! if we're doing a 1.9.3 then this is a waste of time.

PR #1700, opened 2015-07-09
Report change status on django_manage collectstatic.

faulkner:


cchurch:
That check won't work for `collectstatic link=yes`.  The output in that case would be `0 static files symlinked`.

faulkner:
@cchurch good catch!  I've updated the pull request to handle that case as well.

PR #1712, opened 2015-07-10
win_get_url: add force to module docs, default to force=yes

cchurch:
Additional changes following https://github.com/ansible/ansible-modules-core/pull/1485 as suggested by @bcoca.  Default `force` to `yes` preserves the existing behavior of the module if `force` isn't specified.

### DIFF
--- a/windows/win_get_url.ps1
+++ b/windows/win_get_url.ps1
@@ -40,7 +40,7 @@ Else {
     Fail-Json $result "missing required argument: dest"
 }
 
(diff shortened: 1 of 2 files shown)


talcloudshare:
Please see my comment here regarding setting "yes" as default and using ConvertTo-Bool, any value not recognized by ConvertTo-Bool will be converted to "false":

https://github.com/ansible/ansible-modules-core/pull/1485#issuecomment-118773132

PR #1722, opened 2015-07-13
Remove default for engine_version

fvant:
Redis and memcached have different engine version numbering, there can not be a shared default value.

### DIFF
--- a/cloud/amazon/elasticache.py
+++ b/cloud/amazon/elasticache.py
@@ -42,6 +42,7 @@ options:
     description:
       - The version number of the cache engine
     required: false
+    default: none
   node_type:
     description:
       - The compute and memory capacity of the nodes in the cache cluster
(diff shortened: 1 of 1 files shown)


bcoca:
the default doc is still needed, it would bye null/None in this case

bcoca:
also, how will the api deal with a None in this setting? would a default depending on db type be a better option?

fvant:
Boto API has it optional too, you get the default engine version from AWS.

create_cache_cluster(cache_cluster_id, num_cache_nodes=None, cache_node_type=None, engine=None, replication_group_id=None, engine_version=None, cache_parameter_group_name=None, cache_subnet_group_name=None, cache_security_group_names=None, security_group_ids=None, snapshot_arns=None, preferred_availability_zone=None, preferred_maintenance_window=None, port=None, notification_topic_arn=None, auto_minor_version_upgrade=None)

Issue #1655, opened 2015-07-01
ec2_lc fails with TypeError: LaunchConfiguration:test-lc is not JSON serializable (devel branch)


joshm:
Running a lc create as simple as:

```
- name: s launch configuration
  ec2_lc:
    name: test-lc
    instance_type: m3.medium
    region: us-west-2
    state: present
```

Returns:

`TASK [s_setup : s_setup : s launch configuration] *******************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: LaunchConfiguration:test-lc is not JSON serializable
fatal: [localhost]: FAILED! => {"invocation": {"module_name": "ec2_lc", "module_args": {"instance_type": "m3.medium", "region": "us-west-2", "name": "test-lc", "state": "present"}}, "failed": true, "changed": false, "parsed": false}`

joshm:
I believe it is this: https://github.com/ansible/ansible-modules-core/commit/5e674ddcfc3ab317d2aa05fc84dfe206768cfdab#diff-74356b4e26d3b0a785194e2aad7fe7f4

Removing that, fixes the issue

evanccnyc:
Yup just ran into this bug as welll..probably should back this out since it breaks Launch Config.

bcoca:
reverted the commit, closing this issue

evanccnyc:
Hmmm i just pulled down ansible and reran with this patch backed out and still seeing it...I am going to dig a bit deeper.

bcoca:
run with -vvvv so we get the full traceback

evanccnyc:
That pull request I just posted should fix this problem.

evanccnyc:
Traceback is posted in the pull request. If you want to double check.

vuorse:
The latest version of ec2_lc is indeed broken. Haven't isolated the actual issue yet, but by backing out to a version b91a232 it works so definately something committed since then has caused the issue.

abadger:
@robbwagoner So in https://github.com/ansible/ansible-modules-core/pull/497 we added a LaunchConfig  object into the return value for ec2_lc.  Unfortunately, that object can only be turned into json some of the time.  Can you tell us what fields of LaunchConfig we care about?  Or which ones might not be python basic types?  One of the other return values is also apparently a non-basic type sometimes... we'll have to track that one down to.

abadger:
[...]

PR #1728, opened 2015-07-14
Revert "ec2_lc - include all launch config properties in the return"

bcoca:
Reverts ansible/ansible-modules-core#497

due to issues reported in #1655


evanccnyc:
Okay this revert didnt fix the problem I am going to open an issue to fix the real issue.

PR #1734, opened 2015-07-15
assemble: add file validation support.

### LINKED ISSUE #577: Feature Request: config file validation in assemble module
##### Ansible Version:

```
ansible 1.8.2
```
##### Environment:

up-to-date Archlinux (as of 2014-12-27)
##### Summary:

Ansible module _assemble_ does not currently support _validate_ attribute .
I think this might be nice feature to have, because if someone decides to split configuration file generation into multiple places, validation may become impossible.

 And having _validate_ would allow to perform this kind of check without putting separate task  after _assemble_ entry.
##### Steps To Reproduce:

N/A
##### Expected Results:

Possibility to validate file generated by assemble module
##### Actual Results:

Module currently does not support this operation

freym:
This PR adds an validate option to the assemble module. I works like the copy or template module.

Fixes #577

### DIFF
--- a/files/assemble.py
+++ b/files/assemble.py
@@ -85,6 +85,13 @@ options:
     required: false
     default: false
     version_added: "2.0"
+  validate:
+    description:
+      - The validation command to run before copying into place.  The path to the file to
+        validate is passed in via '%s' which must be present as in the sshd example below.
+        The command is passed securely so shell features like expansion and pipes won't work.
+    required: false
+    default: ""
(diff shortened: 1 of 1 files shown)


jody-frankowski:
I tested this on ansible 1.9.2-1 and it works fine for me. I tested with an ssh config. The code looks fine too. However maybe you could capitalize the messages in order to be consistent with the rest of the module.

freym:
The messages and the other things came from the copy module. The only thing i have changed was the example.

PR #1738, opened 2015-07-16
apt_repositories: fixes traceback about missing repo_urls

### LINKED ISSUE #1684: apt_repository bugs in repos_urls()
@verm666 could you take a look at some problems in the repos_urls code?
- @jstasiak noted here: https://github.com/ansible/ansible-modules-core/issues/133#issuecomment-119273206 that the repos_url() property was only implemented for Ubuntu, not for Debian.
- I was looking at how to adapt it and found `source, ppa_owner, ppa_name = self._expand_ppa(i[3])
  `.  I couldn't figure out where `i` was getting set.  Is that a bug?

We'll need to either push the fix to stable-1.9 or revert it from there as well: https://github.com/ansible/ansible-modules-core/pull/1671#issuecomment-119273824

verm666:
fixes #1684

### DIFF
--- a/packaging/os/apt_repository.py
+++ b/packaging/os/apt_repository.py
@@ -321,6 +321,21 @@ class SourcesList(object):
         source = self._parse(line, raise_if_invalid_or_disabled=True)[2]
         self._remove_valid_source(source)
 
+    @property
+    def repos_urls(self):
+        repositories = []
+        for parsed_repos in self.files.values():
+            for parsed_repo in parsed_repos:
+                enabled = parsed_repo[1]
+                source_line = parsed_repo[3]
+
+                if not enabled:
+                    continue
+
+                repositories.append(source_line)
+
+        return repositories
+
 
 class UbuntuSourcesList(SourcesList):
 
@@ -377,26 +392,6 @@ class UbuntuSourcesList(SourcesList):
         else:
             source = self._parse(line, raise_if_invalid_or_disabled=True)[2]
         self._remove_valid_source(source)
-
-    @property
-    def repos_urls(self):
-        _repositories = []
-        for parsed_repos in self.files.values():
-            for parsed_repo in parsed_repos:
-                enabled = parsed_repo[1]
-                source_line = parsed_repo[3]
-
-                if not enabled:
-                    continue
-
-                if source_line.startswith('ppa:'):
-                    source, ppa_owner, ppa_name = self._expand_ppa(i[3])
-                    _repositories.append(source)
-                else:
-                    _repositories.append(source_line)
-
-        return _repositories
-
 
 def get_add_ppa_signing_key_callback(module):
     def _run_command(command):


gregdek:
@sashka Please review this PR to make sure it adheres to the following guidelines:

abadger:
Is this logic still correct with the changes to repos_urls()? https://github.com/ansible/ansible-modules-core/pull/1738/files#diff-75ef1e4afb3855581133076bd9581144L447

dalyc:
This still doesn't make it work for Debian. If it's debian, a SourcesList is created (not a UbuntuSourcesList) and SourcesList does not have repos_urls so it dies with 

`AttributeError: 'SourcesList' object has no attribute 'repos_urls'`

sashka:
@gregdek I'll have some time for the code review in two days. Sorry for such a long delay.

abadger:
@dalyc This PR appears to move the repos_urls() method from UbuntuSourcesList to SourcesList.  How did you apply this change to test it out?

jcoombe:
oh you're right - my bad - i did a super hacky test by pasting the the changed code into my version of the file which was open in an editor - didn't realise it had moved aswell.

dalyc:
@abadger that last comment was me  - i was logged in as organization repo owner...

abadger:
@dalyc thanks.

So looking at this more closely I think it's still broken in two ways:
- On Ubuntu this is now broken because it won't detect ppas properly.  In the main() function the ppa repos are expanded but in the repos_url() function they are not so we won't get the containment test right.
- On Debian, if a user is trying to use a ppa repo on Debian then they'll get a traceback because the code in main() still calls _expand_ppa() but the SourcesList() won't have that method.

packaging/os/apt_repository.py
```diff
@@ -321,6 +321,21 @@ def remove_source(self, line):
...
+    def repos_urls(self):
+        repositories = []
+        for parsed_repos in self.files.values():
+            for parsed_repo in parsed_repos:
+                enabled = parsed_repo[1]
+                source_line = parsed_repo[3]
+
+                if not enabled:
+                    continue
+
```
abadger, line 334: Since we are comparing against the expanded ppa in the main() function, we still need to expand ppas here.  Maybe adding back the old code and fixing the reference to "i" is what's needed:

``` python
+                if source_line.startswith('ppa:'):
+                    source, ppa_owner, ppa_name = self._expand_ppa(source_line)
+                    source_line = source
```

abadger, line 334: And in addition to that, need to move the _expand_ppa method from UbuntuSourcesList to SourcesList.

abadger:
@verm666 Okay, I think those two comments I made on the pull request would fix the issues with this and we could merge.

abadger:
Note -- I don't use Ubuntu regularly so I didn't test those code snippets.  Be sure to test that they don't do the wrong thing when enabling a PPA :-)

abadger:
@amenonsen saw that add_source is also using expand_ppa().   On closer look we started wondering if there's any reason for a separate SourcesList for Ubuntu at all.  Perhaps we should just replace the existing SourcesList with UbuntuSourcesList and remove the detection for the two distros.  It seems like the functionality is the same -- it's just that for most ppas, the packages would only run on Ubuntu.  but a knowledgable Debian user could enable specific PPAs that they know have compatible dependencies with their system. 

@sashka It looks like you changed the code to not install PPAs on Debian when you rewrote the module.  Is there a reason for that that we're missing?

sashka:
@abadger PPAs are not guarantied to work on Debian, i.e. there is no way to expand `ppa:nginx/stable` on a Debian machine because there is no Ubuntu release name like `lucid` or `maverick` on a Debian machine — automatic expansion will generate wrong link.

The only way is to pretend you're running Ubuntu and go with this:

```
deb http://ppa.launchpad.net/nginx/stable/ubuntu YOUR_UBUNTU_VERSION_HERE main 
deb-src http://ppa.launchpad.net/nginx/stable/ubuntu YOUR_UBUNTU_VERSION_HERE main 
```

For sure you have to replace `YOUR_UBUNTU_VERSION_HERE` with one of Ubuntu release name, and there is no way to guess which release would work.

That was the reason to separate UbuntuSourceList and SourceList.

sashka:
BTW there shouldn't be any `expand_ppa` in SourceList, it should be only in UbuntuSourceList because it's specific to Ubuntu only.

amenonsen:
I don't really buy the "ppas should not be expanded on Debian" argument. If I run "add-apt-repository ppa:nginx/stable" on my Debian wheezy laptop, it adds the following repository entries:

```
# cat /etc/apt/sources.list.d/nginx-stable-wheezy.list 
deb http://ppa.launchpad.net/nginx/stable/ubuntu wheezy main
deb-src http://ppa.launchpad.net/nginx/stable/ubuntu wheezy main
```

Sure, it doesn't work as a practical matter because Launchpad builds and hosts PPA packages only for Ubuntu. But the _mechanism_ is the same, and in theory someone could publish such packages for wheezy, though I have no idea what hoops they would have to jump through. If `add-apt-repository` on Debian refused to add a PPA, then I would be OK with `apt_repository` also refusing to do so.

But anyway, this is a pretty academic distinction. In practice, I agree that ~nobody on Debian is going to want to install directly from a PPA, so I don't _really_ care about any of the above. I just want it to not die when I try to use it, so I submitted PR #1842 that adds a Debian-specific version of `repos_urls` without ppa handling. That also fixes the `i[3]` reference. I tested those changes on Debian, and they work fine. The code hasn't changed on Ubuntu except for calling `_expand_ppa` with the right arguments.

amenonsen:
BTW, the ppa handling seems broken anyway. The module calls ``_expand_ppa`` in ``main`` and then again on the result in e.g. ``add_source``.

amenonsen:
(Sorry, I was mistaken about the last comment I made about running `_expand_ppa` twice, and mistakenly deleted it when I was trying to edit it.)

sashka:
> If add-apt-repository on Debian refused to add a PPA, then I would be OK with apt_repository also refusing to do so.

I'm terribly sorry, but I can't see any connection between add-apt-repository originated in Ubunty and apt_repository. Continuing your point, apt_repository should add source repository by default because add-apt-repository does so.

Speaking more serious my initial intention was to keep behaviour as explicit as possible. PPA is an apt-compatible service which provides us with `ppa:` shortcut for Launchpad service that doesn't host packages for anything but Ubuntu distributions. As you pointed before, theoretically speaking, mechanisms are the same, but practical results are the opposite: PPA shortcut doesn't work correct on Debian hosts out of box.

amenonsen:
@sashka I really don't care enough about the handling of PPAs on Debian to argue for or against any behaviour. I accept whatever logic/organisation you and @abadger can agree on. The current situation is that `apt_repository` is completely broken on Debian since commit d435d5ce (and because of the `i[3]` thing, also kinda broken on Ubuntu). I just want it to work somehow. My PR #1842 fixes the problem along the lines you describe (i.e. make `ppa:` notation break cleanly on Debian). Could you please review it?

sashka:
@amenonsen I will in 2 hours.

sashka:
This PR doesn't work, but https://github.com/ansible/ansible-modules-core/pull/1842 does.

dalyc:
Will the fix be backported to 1.9-stable?

abadger:
I cherrypicked it to the stable-1.9 branch yesterday.  You can retrieve it
from there and test that it works if you like.  I always worry when I
cherry pick something that I've missed something else that the code depends
on.

amenonsen:
I submitted #1847 to fix the thing you missed. ;-)

PR #1742, opened 2015-07-16
Add new os_nova_flavor OpenStack module.

Shrews:
The os_nova_flavor module allows a user with administrative privileges
to create and delete nova flavors.

### DIFF
--- a/cloud/openstack/os_nova_flavor.py
+++ b/cloud/openstack/os_nova_flavor.py
@@ -1,0 +1,237 @@
+#!/usr/bin/python
+
+# Copyright (c) 2015 Hewlett-Packard Development Company, L.P.
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import shade
+    HAS_SHADE = True
+except ImportError:
+    HAS_SHADE = False
+
+DOCUMENTATION = '''
+---
+module: os_nova_flavor
+short_description: Manage OpenStack compute flavors
+extends_documentation_fragment: openstack
+version_added: "2.0"
+author: "David Shrewsbury (@Shrews)"
+description:
+   - Add or remove flavors from OpenStack.
+options:
+   state:
+     description:
+        - Indicate desired state of the resource. When I(state) is 'present',
+          then I(ram), I(vcpus), and I(disk) are all required. There are no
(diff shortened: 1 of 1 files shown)


Shrews:
Ping @emonty and @j2sol

Shrews:
This will require a new release of shade.

omgjlk:
This looks good to me.

emonty:
me too

cloud/openstack/os_nova_flavor.py
```diff
@@ -0,0 +1,232 @@
...
+    if state == 'present' and not flavor:
+        return True
+    if state == 'absent' and flavor:
+        return True
+    return False
+
+
+def main():
+    argument_spec = openstack_full_argument_spec(
+        state        = dict(required=True, default='present',
```
bcoca: required=true and default don't make sense at the same time

cloud/openstack/os_nova_flavor.py
```diff
@@ -0,0 +1,232 @@
...
+version_added: "2.0"
+author: "David Shrewsbury (@Shrews)"
+description:
+   - Add or remove flavors from OpenStack.
+options:
+   state:
+     description:
+        - Indicate desired state of the resource.
+     choices: ['present', 'absent']
+     required: false
```
bcoca, line 40: spec has required: true

cloud/openstack/os_nova_flavor.py
```diff
@@ -0,0 +1,232 @@
...
+     default: present
+   name:
+     description:
+        - Flavor name.
+     required: true
+   ram:
+     description:
+        - Amount of memory, in MB.
+     required: false (true if I(state) is 'present')
```
bcoca: when required is false document default

it might be easier to document all 3 are needed in state: description for when state is present

cloud/openstack/os_nova_flavor.py
```diff
@@ -0,0 +1,232 @@
...
+- os_nova_flavor:
+    cloud=mycloud
+    state=absent
+    name=tiny
+'''
+
+RETURN = '''
+flavor:
+    description: Dictionary describing the flavor.
+    returned: success
```
bcoca: flavor is not returned on 'absent', success or not

cloud/openstack/os_nova_flavor.py
Shrews: There is no default value because they are required when state is 'present'. I can definitely move the description to 'state' though.

cloud/openstack/os_nova_flavor.py
Shrews: How do I document this correctly? There are few examples of this new documentation.  :(

cloud/openstack/os_nova_flavor.py
Shrews: Good catch. Supposed to be false.

cloud/openstack/os_nova_flavor.py
bcoca: default: null, in that case, even when required with other conditions, the base requried: false stays true, minor issue just trying to keep all docs similar

cloud/openstack/os_nova_flavor.py
bcoca: new feature means few corner cases, probably something like:
   returned: on success when state is 'present'

Issue #1707, opened 2015-07-10
Update docker module to support newer versions of the server (mem_limit has been moved to host_config in API version 1.19)


anthonygreen:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.2
##### Environment:

Running from: OS X 10.95
Managing: Ubuntu 15.04
##### Summary:

Starting a container results in the error "mem_limit has been moved to host_config in API version 1.19" starting container
##### Steps To Reproduce:

Using Ansible to configure starting a Jenkins container 

```
- name: Jenkins data directory
  command: "docker create -v {{HOME}}/.jenkins:/var/jenkins_home:rw -v {{HOME}}/.ssh:/.ssh:ro -v {{HOME}}/.ssl:/.ssl:ro --name jenkins_data busybox"
  sudo: yes
  when: docker_containers.stdout.find('jenkins_data') == -1


- name: Start Jenkins container
  docker: 
    name: jenkins_master
    image: mediaplayout/jenkins
    volumes_from: jenkins_data
    pull: always
    state: started
    ports: 
      - 8080:8080
      - 50000:50000
  sudo: yes  
```

I'm getting the error

```
BECOME-SUCCESS-nnwoeoumeitqznkijaqrhohdpyhijpbr
Traceback (most recent call last):
  File "/home/mediaplayout/.ansible/tmp/ansible-tmp-1436530356.59-234590896125756/docker", line 3132, in <module>
    main()
  File "/home/mediaplayout/.ansible/tmp/ansible-tmp-1436530356.59-234590896125756/docker", line 1494, in main
    started(manager, containers, count, name)
  File "/home/mediaplayout/.ansible/tmp/ansible-tmp-1436530356.59-234590896125756/docker", line 1354, in started
    created = manager.create_containers(delta)
  File "/home/mediaplayout/.ansible/tmp/ansible-tmp-1436530356.59-234590896125756/docker", line 1227, in create_containers
    containers = do_create(count, params)
  File "/home/mediaplayout/.ansible/tmp/ansible-tmp-1436530356.59-234590896125756/docker", line 1217, in do_create
    result = self.client.create_container(**params)
  File "/usr/local/lib/python2.7/dist-packages/docker/client.py", line 237, in create_container
    volume_driver
  File "/usr/local/lib/python2.7/dist-packages/docker/utils/utils.py", line 546, in create_container_config
    'mem_limit has been moved to host_config in API version 1.19'
[...]

PR #1743, opened 2015-07-16
Do not erroneously mask exceptions in docker module

larsks:
There was a catch-all `except` statement in `create_containers`:

```
    try:
        containers = do_create(count, params)
    except:
        self.pull_image()
        containers = do_create(count, params)
```

This would mask a variety of errors that should be exposed, including
API compatability errors (as in #1707) and common Python exceptions (KeyError, ValueError, etc) that could result from errors in the code.

This change makes the `except` statement more specific, and only attempts to pull the image and start a container if the original create attempt failed due to a 404 error from the docker API.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -389,6 +389,7 @@ from urlparse import urlparse
 try:
     import docker.client
     import docker.utils
+    import docker.errors
     from requests.exceptions import RequestException
 except ImportError:
     HAS_DOCKER_PY = False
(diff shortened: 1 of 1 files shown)


cloud/docker/docker.py
```diff
@@ -1322,7 +1323,10 @@ def do_create(count, params):
 
         try:
             containers = do_create(count, params)
-        except:
+        except docker.errors.APIError as e:
```
bcoca, line 1326: not 2.4 compatible, but the module already requires 2.6 cause of docker.py, so not an issue.

PR #1745, opened 2015-07-16
Adding `log_opt` argument for `docker` module.

justnom:
This PR adds in the `log_opt` argument for used during the [`RUN` phase](https://docs.docker.com/reference/run/#logging-drivers-log-driver) of Docker containers.

At the moment Docker only uses this argument for the `syslog` logging driver - and the only possible option is `address`.

I have setup the argument in Ansible as a string, with multiple options split by a comma and key value pairs represented with an equal sign.

Example: `address=tcp://127.0.0.1:123,somethingelse=test`


abadger:
Seems like this would be better specified as either a list or dict.  Then ansible can do the parsing of the value entered in the playbook rather than the module.

the SyntaxError that travis is reporting is because the docker module is compatible with python-2.6 and above.  dict comprehensions aren't available on python-2.6 so that would have to be changed as well.

abadger:
CC for review @cove @joshuaconner @softzilla @smashwilson

smashwilson:
:+1: from me. :shipit:

justnom:
Sorry for the wait - rebased.

xbeta:
when will this merge?

xbglowx:
Please merge.

zfil:
This PR should be closed because PR #1991 that covers the same thing is already merged in devel

gregdek:
Hi @justnom. Thanks again for this PR -- looks like another PR was merged that fixes the issue. Please take a look at #1991 to verify that it solves the problem for you in an acceptable way; if not, we're happy to see a new PR against the updated codebase.

PR #1748, opened 2015-07-17
Making rax_facts ignore servers that are in the deleted state [BR]

richbria:


angstwad:
ping @sivel.

cloud/rackspace/rax_facts.py
```diff
@@ -98,6 +98,11 @@ def rax_facts(module, address, name, server_id):
         except Exception, e:
             pass
 
+    #rackspace likes to leave servers in the deleted state for a while.  
+    #If you make a server with the same name it triggers len(servers) > 1 
+    #so just remove servers with a server status of deleted from the list
+    servers = [server for server in servers if server.status != "DELETED"]
```
sivel: I'm not a huge fan of re-assigning already assigned variables.  My preference would be a normal for loop with a del.

Otherwise a new variable.

cloud/rackspace/rax_facts.py
```diff
@@ -98,6 +98,11 @@ def rax_facts(module, address, name, server_id):
         except Exception, e:
             pass
 
+    #rackspace likes to leave servers in the deleted state for a while.
```
sivel: We try to maintain pep8 in the rax modules.  A space following the `#` is required.

cloud/rackspace/rax_facts.py
richbria: Everywhere I am reading suggests using a list comprehension as opposed to iteration and then deleting (in this scenario you are iterating the list twice)

http://stackoverflow.com/questions/1207406/remove-items-from-a-list-while-iterating-in-python

Based on that I would recommend using a slice to avoid duplicating the list.  I could also just change the variable name if you want.

```
servers[:] = [server for server in servers if server.status != "DELETED"]
```

or 

```
valid_servers = [server for server in servers if server.status != "DELETED"]
```

cloud/rackspace/rax_facts.py
richbria: I actually added these comments to be more clear about why I was doing this.  Do you think it is worth even having them?

I could also shorten it to something like:

```
# Preventing deleted servers from being included in results
```

sivel: I think the code itself is probably self explanatory enough

cloud/rackspace/rax_facts.py
sivel: I am fine with:

```
servers[:] = [server for server in servers if server.status != "DELETED"]
```

PR #1750, opened 2015-07-19
synchronize: add flag for verifying target host.

eest:
Add the possibility to verify the target host using a "verify_host"
flag. It is disabled by default to not change the module behaviour.

In my opinion this should be the default behaviour, since we are otherwise possibly sending files to a rogue host, but one step at a time.


eest:
Just thought I would mention @tima (the author) so he gets notified of the change.

PR #1755, opened 2015-07-20
add sshpass support for `synchronize`

elvismacak:
add sshpass support for rsync
work together with https://github.com/ansible/ansible/pull/11657


mgedmin:
For the record (since I didn't know), [sshpass](https://sourceforge.net/projects/sshpass/) is a 3rd-party program that wraps /usr/bin/ssh and accepts a password on the command-line (i.e. visible to anyone running `ps` on the same machine).

elvismacak:
yeah, it's visible in command-line with `ps`.
just like write the readable password directly in the `inventory` file, I think it's a basic procedure should provide, but suggest not use it.

I sometimes need to synchronize some files just in my local-machine, sure nobody can see that, and readly boring with entering password.

gregdek:
@tima let me know which of this or #2053 is preferred, and I will close one of them.

elvismacak:
#2053 is my frirend Bryce ^_^

gregdek:
:)

Well, both of them are out-of-date, so one of them needs to be rebased. Perhaps the two of you could chat and decide which one you'd like to pursue?

tima:
I appreciate the effort and PRs of both @elvis-macak and @hellofreedom here. My preference is this PR because it integrates with the ansible_ssh_password magic inventory variable. 

I guess my only concern with what is here is that the SSH password would appear in the debugging and logging. The temp file used by @hellofreedom in #2053 would better avoid the password from being exposed. 

@elvis-macak: Would you consider modify your work to use that method?

elvismacak:
Yeah, I see.
But I think simple here is much safer than safe. In my PR, my code provide `key` and `password`.  `password` is not safe, while `key` should be promoted, I think.

It's just like the ansible `inventory` file, you can fill with password directly, right? but it's not suggested.
write the password in a file will be hard to maintain, though it's much safer.

tima:
I hear you argument @elvis-macak. Not sure all users would agree. I'd like some of the core commiters like @bcoca, @abadger and @jimi-c to weigh in on something like this that has security implications.

bryceliu:
@elvis-macak you said:

> It's just like the ansible inventory file, you can fill with password directly, right? but it's not suggested.
> write the password in a file will be hard to maintain, though it's much safer.

I'm afraid I disagree with you on that. Because one can change the mode of his or her inventory file to `600` to prevent others from peeking, but the password put directly into a command line can be visible to other users of the same machine (by running `ps aux` or something else, depends on the OS) as @mgedmin said above.

Moreover, there is no need to maintain the password file in #2053 because it's just a temporary file which will be removed when the command is done.

I think @tima is right, this PR integrates with the ansible_ssh_password magic inventory variable (so wonderful) while #2053 has to add `sshpass={{ ansible_ssh_pass }}` explicItly. So @elvis-macak could you consider @tima's advise again?

gregdek:
Thanks to all for getting this PR moving again. Note to @elvis-macak -- this PR also needs to be rebased. When you've rebased and made all necessary revisions, let us know by adding a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

sivel:
sshpass should be used just like it is in the ssh connection plugin.  In the ssh connection plugin sshpass is communicated with via a socket, and not by providing the password in the command.

See:

https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py#L140-L141
https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py#L369-L371

sivel:
Also, I think we should rely on already established connection info, as opposed to supplying the password in the task.

tima:
Thanks for pointing that out @sivel. That is an excellent point and another (best IMO) way of going about this. I agree that a password thru the task is not a good idea from a security  and its not as convenient either. What say you @elvis-macak and @hellofreedom?

elvismacak:
Thx for @sivel's advice, that's a really wonderful method. Also thanks for @hellofreedom @tima for discussing.
I thought there is no need to care too much about the password like others, that would be wrong.

@hellofreedom, yeah, you're right. and as refered by @sivel , using a `pipe` looks better than `TemporaryFile`:
https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/connection/ssh.py#L131-L132
will you change your code and commit again. it's your thought before.

ritzk:
Using SSHPASS env var - 

https://github.com/ansible/ansible-modules-core/pull/2244
https://github.com/ansible/ansible/pull/12704

elvismacak:
@ritzk , not sure reading the password from the enve, can it also be read from `/proc/<pid>/environ`?
except this, I think @hellofreedom's  and @sivel's  thoughts are better, reading from a temperay file or pipe will be safer, I think

elvismacak:
add a new PR: https://github.com/ansible/ansible-modules-core/pull/2247

ritzk:
##### Security consideration
- These are only readable by owner, or root. root can read almost everything based on selinux/security policies (including /proc/kcore)
- ssh client tends to strip out env variables, SSHPASS should not be seen by client process (SendENv bit)

##### reasons to opt for env var
- module.run_command does not support passing an fd to command module. 
- ssh and sshpass do not support reading password from stdin pipe ( data arg of run_command)
- adding support for setting up env was the least invasive and most reliable change

tima:
@elvis-macak: If you submitted a new revision as a separate PR could you close this one and reference the new one in the comments? I'm getting bogged down trying to figure out which PRs in core and mode-modules I need to be reviewing and commenting on. Thanks.

elvismacak:
@tima okay, I'm going to close it, sorry for bothering.  thanks

tima:
Not a bother. You're doing the community a favor!


[image: photo]
Timothy Appnel
Solutions Architect
919.670.0720
Ansible, Inc. *•* www.ansible.com

tima:
Not a bother! You're doing a great service to the community. Thanks.

files/synchronize.py
```diff
@@ -330,20 +337,23 @@ def main():
...
     if dest_port != 22:
-        cmd += " --rsh 'ssh %s %s -o Port=%s'" % (private_key, ssh_opts, dest_port)
-    else:
-        cmd += " --rsh 'ssh %s %s'" % (private_key, ssh_opts)  # need ssh param
+        ssh_opts = '%s -o Port=%d' % (ssh_opts, dest_port)
+
+    if private_key:
+        cmd += " --rsh 'ssh %s %s'" % (private_key, ssh_opts)
+    elif ssh_passwd:
+        cmd += " --rsh 'sshpass -p %s ssh -l ass %s'" % (ssh_passwd, ssh_opts)
```
ritzk, line 356: Why is the ssh username set here ?

elvismacak, line 356: username?

ritzk, line 356: ssh _-l ass_

Also in https://github.com/ansible/ansible-modules-core/pull/2247

PR #1761, opened 2015-07-21
Added support for --read-only Docker containers

### LINKED ISSUE #1760: add readonly flag to docker module
It will be very useful to have ability to launch docker container in readonly mode with readonly flag (as in docker run --read-only)

jhaals:
Adds support for mounting the container's root filesystem as read only.

fixes #1760

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -571,6 +571,7 @@ class DockerManager(object):
             'cpu_set': ((0, 6, 0), '1.14'),
             'cap_add': ((0, 5, 0), '1.14'),
             'cap_drop': ((0, 5, 0), '1.14'),
+            'read_only': ((1, 0, 0), '1.17'),
             # Clientside only
             'insecure_registry': ((0, 5, 0), '0.0')
             }
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -270,7 +270,7 @@ options:
       - Mount the container's root filesystem as read only
     default: false
     aliases: []
-    version_added: "1.9"
+    version_added: "2.0"
   restart_policy:
     description:
       - Container restart policy.
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -265,6 +265,12 @@ options:
     default: DockerHub
     aliases: []
     version_added: "1.8"
+  read_only:
+    description:
+      - Mount the container's root filesystem as read only
+    default: false
+    aliases: []
+    version_added: "1.9"
   restart_policy:
     description:
       - Container restart policy.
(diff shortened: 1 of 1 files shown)


bcoca:
review ping @cove, @joshuaconner, @softzilla, @smashwilson

cloud/docker/docker.py
```diff
@@ -265,6 +265,12 @@
     default: DockerHub
     aliases: []
     version_added: "1.8"
+  read_only:
+    description:
+      - Mount the container's root filesystem as read only
+    default: false
+    aliases: []
+    version_added: "1.9"
```
smashwilson: Er, this should be bumped to "2.0" now.

bcoca: if that is only issue, we normally fix after merge

cloud/docker/docker.py
```diff
@@ -772,6 +778,7 @@ def get_start_params(self):
             'privileged': self.module.params.get('privileged'),
             'links': self.links,
             'network_mode': self.module.params.get('net'),
+            'read_only': self.module.params.get('read_only'),
```
smashwilson, line 796: I feel like this should be gated behind an `ensure_capability` check like some of the other HostConfig parameters added [further down](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/docker/docker.py#L782-L784), because it was added in a recent revision of the Docker API (and docker-py API).

At a glance it looks like `read_only` was added to the Docker API in [1.17](https://docs.docker.com/reference/api/docker_remote_api/#v1-17). I'm checking for the docker-py version presently.

smashwilson, line 796: Ah, docker-py has supported read_only since https://github.com/docker/docker-py/commit/e379e8ae241faf0d8fc244cf4f7228e615149c8b, which looks like it was released in 1.0.0.

cloud/docker/docker.py
smashwilson: > if that is only issue, we normally fix after merge

:+1: Excellent, I'll keep that in mind for future reviews.

jhaals:
Oh I didn't know 1.9 was locked. I've bumped `version_added` to 2.0 and resolved the merge conflict.

cloud/docker/docker.py
bcoca, line 796: i would still have the version check, people seem to be running many diff versions of docker.py ... makes this module 'fun'

jhaals, line 796: version check added

abadger:
@jhaals, just a note in case you need to add another optional feature to the docker module in the future:  version compatibility checks need to use a conditional to check whether they are available as well.  Adding the relevant versions to the data structure isn't enough by itself.  Here's the remaining code that needed to be added: https://github.com/ansible/ansible-modules-core/commit/6e5a832dc28e1de72f296f7fe2b9bda294bc5b50

(And if you see that I did something wrong there, feel free to point it out to me ;-)

jhaals:
Oh I missed that, thanks for the update @abadger! :+1:

PR #1762, opened 2015-07-21
Add repo_vars module to core ansible modules

cedub:


utilities/logic/repo_vars.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, module, **kwargs):
+        self.module = module
+        self.path = kwargs['path']
+
+    def read_vars(self):
+        if not self.path:
+            return {}
+
+        try:
+            f = open('%sansible.json' % self.path, 'r')
```
mgedmin: Use `os.path.join()` maybe?  What if the path doesn't end with a slash?

utilities/logic/repo_vars.py
```diff
@@ -0,0 +1,103 @@
...
+        repo_vars = RepoVars(module, path=path)
+    except RepoVars.InvalidPathError,e:
+        module.fail_json(msg=e.message)
+
+    return_vals = repo_vars.read_vars()
+
+    # Always setting changed to true since the values will be set with each
+    # run of the playbook.  Perhaps this should just always be set to false
+    # since nothing is ever actually changed?
+    return_vals['changed'] = True
```
mgedmin: The set_fact module always returns changed = False.

mgedmin:
Why is this needed?  Why not use `vars_files`?  JSON is a subset of YAML.

cedub:
From what I understand, `vars_files` is on the Ansible side.  This module allows me to define a manifest in my repository that tells Ansible about the repository.  For example, in our company, we have several different types of deployments we need.  Sometimes it's a Node deployment, sometimes it's a Django deployment, sometimes AngularJS with either a Node or a Django backend.  This way we could have a single deployment job template and it will use different roles based on the content of the manifest in the repository.  I haven't seen anything in Ansible core that would allow this.

As far as supporting YAML instead of JSON (or in addition to) this would be a trivial addition and could certainly be added in.

And if there's already another module that does what this PR tries to solve, I'm all for using that one.  I just hadn't seen anything in my digging around and chatting with the support team.

utilities/logic/repo_vars.py
cedub: Since we're just defining variables from the repo to expose to the playbook/roles, I wasn't sure what to return since we're not technically "changing" anything.  Would be open to recommendations here.

But yah, in order to stay consistent with the `set_fact` module, we could certainly just return `changed = False`.

utilities/logic/repo_vars.py
cedub: Yah, good point.  I'll make that tweak.

gregdek:
@cedub apologies, but new modules are no longer being accepted directly into ansible-modules-core. If you would please resubmit this as a PR against ansible-modules-extras and reference this PR, we'll get it into the community review process over there. Thanks!

PR #1763, opened 2015-07-21
restore mem_limit

larsks:
mem_limit got lost in the #1744; this restores it.  Thanks to @dgromov
for the report.

PR #1744, opened 2015-07-16
Use HostConfig object when creating container with Docker Remote API > 1.15

larsks:
This is mlosev's patch (from #1208), rebased against devel as of
2790af2.  It resolves #1707, which was caused by an API incompatibility
between the docker module and server API version 1.19.

Issue #1602, opened 2015-06-25
Memory limit not working docker 1.7.0


enxebre:
Hi there
Using docker Client version: 1.7.0 Client API version: 1.19
Using docker module for running a container and set memory limit like:
memory_limit: "256MB"
or:
memory_limit: 256MB

when I docker inspect the container the memory is not set.

gregdek:
cc: @cove @joshuaconner @softzilla @smashwilson

bcoca:
what version of ansible, os, python? could you fill in the template? https://github.com/ansible/ansible/blob/devel/ISSUE_TEMPLATE.md

enxebre:
Hi,
running from os x mavericks 10.9 to ubuntu1404
ansible 1.9.0.1
docker Client version: 1.7.0 Client API version: 1.19
ansible.cfg

```
[defaults]
host_key_checking = no
record_host_keys = no
jinja2_extensions = jinja2.ext.do
timeout = 15
gathering = smart

[ssh_connection]
ssh_args = -o ControlMaster=auto -o ControlPersist=30m
scp_if_ssh = True
pipelining = True
```

trying to:

```
- name: run container
  docker:
    name: sample
    image: "{{sample_image }}"
    state: started
    restart_policy: "{{ sample_restart_policy }}"
    ports:
    - "{{ sample_port }}:{{ sample_port }}"
    expose:
    - "{{ sample_port }}"
    net: "{{ sample_net }}"
    command: "{{ sample_command }}"
    volumes:
    - "/var/run/docker.sock:/tmp/docker.sock"
    memory_limit: "{{ sample_container_memory_limit }}"
```

when docker inspect got:

```
 "Memory": 0,
```

ecliptik:
Same here, added

```
 memory_limit: "512MB"
```

and playbook runs, but docker inspect shows:

```
 "Memory": 0,
 "MemorySwap": 0,
```

enxebre:
Think must be due to I am using docker-py version=1.1.0 as per ansible docs http://docs.ansible.com/ansible/docker_module.html docker-py >= 0.3.0 is a requirement

ecliptik:
I put in a pull request that fixes #1766 due to the wrong data structure reference which was causing container deployments with api v1.19 to break.

On a whim I added,

```
 memory_limit: "512MB"
```

to the playbook, and before where it would deploy successfully and a _docker inspect_ would show _"Memory": 0,_ - it now gives the error,

```
[...]

PR #1767, opened 2015-07-21
Use proper HostConfig element which contains the proper Memory value …

### LINKED ISSUE #1766: Docker module gives error KeyError: 'Memory' when using  remote API v1.19
**Summary**

When running a playbook with _docker_api_version: "1.19"_, the function _get_differing_containers_ gives the following error:

```
  File "/home/micheal/.ansible/tmp/ansible-tmp-1437502622.78-176169208388238/docker", line 3132, in <modu
le>
    main()
  File "/home/micheal/.ansible/tmp/ansible-tmp-1437502622.78-176169208388238/docker", line 1496, in main
    reloaded(manager, containers, count, name)
  File "/home/micheal/.ansible/tmp/ansible-tmp-1437502622.78-176169208388238/docker", line 1373, in reloa
ded
    for container in manager.get_differing_containers():
  File "/home/micheal/.ansible/tmp/ansible-tmp-1437502622.78-176169208388238/docker", line 908, in get_di
ffering_containers
    actual_mem = container['Config']['Memory']
KeyError: 'Memory'
```

If  _docker_api_version: "1.18"_ is used it does not give an error and deploys the container successfully. It appears it could be related to a change in Remote API v1.19 where,

> The fields Debug, IPv4Forwarding, MemoryLimit, and SwapLimit are now returned as boolean instead of as an int.
> -[Docker Remote API v1.19 Changes](https://docs.docker.com/reference/api/docker_remote_api/#v1-19)

**Ansible Version**

```
 ansible 1.9.2
```

**Ansible Configuration**

```
 host_key_checking = False
 scp_if_ssh = True
```

**Playbook**

```
  - name: Deploy NGINX container
    docker:
      name: nginx
      registry: "private-docker-registry"
      docker_api_version: "1.19"
      pull: always
      tty: yes
      image: private-docker-registry/nginx
      state: reloaded
      net: bridge
      ports:
      - "0.0.0.0:8090:8090"
      volumes:
      - /mnt/nginx/:nginx:ro
```

**Environment**
- Ubuntu 14.04 ansible master
  - ansible v1.9.2
- CentOS 6.6 container host 
  - Docker v1.7.1 
  - docker-py v1.3.0

**Steps to Reproduce**
- Install Ansible v1.9.2 on Ubuntu 14.04
- Install Docker v1.7.1 and docker-py v1.3.0 on CentOS 6.6
- Create a playbook with _docker_api_version: "1.19"_ to use Remote API version 1.19
- Run playbook on Ubuntu ansible host to deploy a container on the CentOS container host

ecliptik:
…- fixes #1766 

Tested locally and containers deployed properly using _docker_api_version: "1.19"_

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1011,10 +1011,10 @@ class DockerManager(object):
             except ValueError as e:
                 self.module.fail_json(msg=str(e))
 
-            actual_mem = container['HostConfig']['Memory']
-
-            #Use v1.18 API and earlier Memory element location
-            if docker_api_version <= 1.18:
+            #For v1.19 API and above use HostConfig, otherwise use Config
+            if docker_api_version >= 1.19:
+                actual_mem = container['HostConfig']['Memory']
+            else:
                 actual_mem = container['Config']['Memory']
 
             if expected_mem and actual_mem != expected_mem:
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1008,7 +1008,7 @@ class DockerManager(object):
             except ValueError as e:
                 self.module.fail_json(msg=str(e))
 
-            actual_mem = container['Config']['Memory']
+            actual_mem = container['HostConfig']['Memory']
 
             if expected_mem and actual_mem != expected_mem:
                 self.reload_reasons.append('memory ({0} => {1})'.format(actual_mem, expected_mem))
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -942,6 +942,9 @@ class DockerManager(object):
         running = self.get_running_containers()
         current = self.get_inspect_containers(running)
 
+        #Get API version
+        api_version = self.client.version()['ApiVersion']
+
         image = self.get_inspect_image()
         if image is None:
             # The image isn't present. Assume that we're about to pull a new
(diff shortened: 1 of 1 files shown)


bcoca:
what happens to those using earlier versions? this should check api version and use one or the other

bcoca:
as these show, this is now a conflict between those running diff versions #1744 #1763 #1602 #1766

ecliptik:
Updated the pull request to include an API check and use the proper data structure depending on the API version,
- v1.18 and earlier - ['Config']['Memory']
- v1.19 -  ['HostConfig']['Memory']

This still gives the error which is due to #1707 and looks like it should work in Ansible v1.9.3 or v2.0.

```
  File "/usr/lib/python2.6/site-packages/docker/utils/utils.py", line 546, in create_container_config
    'mem_limit has been moved to host_config in API version 1.19'
docker.errors.InvalidVersion: mem_limit has been moved to host_config in API version 1.19
```

cloud/docker/docker.py
```diff
@@ -1008,7 +1011,11 @@ def get_differing_containers(self):
             except ValueError as e:
                 self.module.fail_json(msg=str(e))
 
-            actual_mem = container['Config']['Memory']
+            #For v1.19 API and above use HostConfig, otherwise use Config
+            if docker_api_version >= 1.19:
+                actual_mem = container['HostConfig']['Memory']
+            else:
+                actual_mem = container['Config']['Memory']
```
bcoca, line 1018: is it not possible to use ensure_capability as we do with other features?

PR #1775, opened 2015-07-22
Fix typos in route53.py

arcz:


gregdek:
Looks good to me. Thanks @arcz!

PR #1777, opened 2015-07-22
Make apt_repository work on Debian, not only Ubuntu

amenonsen:
It's implemented only for Ubuntu, and resulted in the following
exception elsewhere:

AttributeError: 'SourcesList' object has no attribute 'repos_urls'


amenonsen:
PR updated to move the property to the base class rather than test for its presence in the caller.

abadger:
Note: @verm666 also started working on a fix here: https://github.com/ansible/ansible-modules-core/issues/1684

gregdek:
@amenonsen thanks for this, but we've already got a fix in progress. Please look at #1684 and add your feedback there.

PR #621, opened 2015-01-09
supervisorctl module: Check if a service exists

hostmaster:
[supervisorctl](http://docs.ansible.com/supervisorctl_module.html) manages the state of a program without checking that a supervised program exists. Even if program does not exist it could be checked for `started`,`restarted` or `stopped` state silently. `present` state reports as expected

PR #1780, opened 2015-07-22
Bugfix Pull Request: supervisorctl module: 'present' state is broken

hostmaster:
##### Ansible Version:

2.0 (devel)
##### Summary:

I introduced a logical bug in https://github.com/ansible/ansible-modules-core/pull/621 `state=present` fails always for newly created service because there is no matched process  yet. Check for process existance has no sense for `state='present`


hostmaster:
@inetfuture please have a look

inetfuture:
LGTM

PR #1789, opened 2015-07-24
headers can be passed as dict

jsmartin:
This is a backwards compatible feature PR that allows you to add headers as a dictionary.

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -129,6 +129,13 @@ options:
         "Content-Type" along with your request with a value of "application/json".
     required: false
     default: null
+  headers:
+    description:
+      - A dictionary of headers you wish to pass to the URI.  This can compliment or be
+        used in place of the HEADER_ method.
+    required: false
+    default: null
+    version_added: "2.0"
   others:
     description:
       - all arguments accepted by the M(file) module also work here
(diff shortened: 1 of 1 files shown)


network/basics/uri.py
```diff
@@ -129,6 +129,13 @@
         "Content-Type" along with your request with a value of "application/json".
     required: false
     default: null
+  headers:
+    description:
+      - A dictionary of headers you wish to pass to the URI.  This can compliment or be
+        used in place of the HEADER_ method.
```
mgedmin, line 135: Typo: compliment -> complement.

It might be good to mention that if the same header shows up in both places, the HEADER_... version will override the one from `headers`.

jsmartin, line 135: Hmmm.  I think the HEADER_ version is a bit hacky and not very obvious.  Maybe we should have the dict version override the HEADER_ version in lieu of deprecating the HEADER_ version down the road.  I'll leave that up to the core team to decide before I make any changes.  Thanks!

gregdek:
ping @ansible/core -- looks like @jsmartin is waiting on a decision from you here.

sivel:
@jsmartin this PR no longer applies cleanly do to some changes in the `uri` module.  Fortunately, those new changes introduced a new `headers` argument to accept a dictionary of headers, so I'm going to go ahead and close this out.

PR #856, opened 2015-02-26
[user] Allow '*' as empty password.

renard:
If `password` is defined as `*` `useradd` or `usermod` returns an error:

```
msg: usermod: Invalid password: `*'
```

This works very well on Linux host to not define any password for a
user (mainly useful if your setup is only based on SSH keys for
auth). On OpenBSD this does not work, so we have to ignore the encrypted
password parameter if it defined as `*`.

PR #1790, opened 2015-07-24
[user] Allow '*' as empty password.

renard:
If `password` is defined as `*` `useradd` or `usermod` returns an error:

```
msg: usermod: Invalid password: `*'
```

This works very well on Linux host to not define any password for a
user (mainly useful if your setup is only based on SSH keys for
auth). On OpenBSD this does not work, so we have to ignore the encrypted
password parameter if it defined as `*`.

Should update #856

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -916,7 +916,7 @@ class OpenBSDUser(User):
             cmd.append('-L')
             cmd.append(self.login_class)
 
-        if self.password is not None:
+        if self.password is not None and self.password != '*':
             cmd.append('-p')
             cmd.append(self.password)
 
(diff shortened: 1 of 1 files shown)


bcoca:
just one comment, can we make it '' instead of '*'?

renard:
@bcoca regarding your comment about making it '' instead of '*', the `usermod` command expects an encrypted value for the password. On OpenBSD and Linux a password disabled login is defined using `*` in the master password or shadow files. Thus I thing having it set to `*` would be less confusing for the users. Don't you?

bcoca:
@renard i realized that 10s after, that is why i deleted it and merged.

renard:
ok nice then.

PR #264, opened 2014-11-03
yum module: add support for check_mode /w yum package groups

jfchevrette:
Currently, when using check_mode with a yum package group (@group), ansible will return changed=True all the time. This is inconvenient when running in check_mode to see if your hosts configs are proper. In my use case, I have a playbook that I run in check_mode every morning and an email report is sent if I have any non-compliant hosts/tasks. (changed != 0)

This pull request adds check_mode support for package groups with state=installed or state=latest

Checking for package removal is tricky and I have not found a clean way to implement this. Ideas are welcome.

There is no way to ask yum or repoquery directly if a package group has installable packages or updateable packages. The way this is implemented is by expanding the package group using 'repoquery' and by passing the list of packages to is_installed which will return wether or not it has any packages that would be installed. Since repoquery will list ALL member packages even though they are excluded in yum.conf, there is a need to use both is_installed and is_available to ensure we only check for those packages yum would normally install (taking into account the excluded ones)

PR #1799, opened 2015-07-24
yum module: add support for check_mode /w yum package groups

jfchevrette:
New PR to replace #264 

Currently, when using check_mode with a yum package group (@group), ansible will return changed=True all the time. This is inconvenient when running in check_mode to see if your hosts configs are proper. In my use case, I have a playbook that I run in check_mode every morning and an email report is sent if I have any non-compliant hosts/tasks. (changed != 0)

This pull request adds check_mode support for package groups with state=installed or state=latest

Checking for package removal is tricky and I have not found a clean way to implement this. Ideas are welcome.

There is no way to ask yum or repoquery directly if a package group has installable packages or updateable packages. The way this is implemented is by expanding the package group using 'repoquery' and by passing the list of packages to is_installed which will return wether or not it has any packages that would be installed. Since repoquery will list ALL member packages even though they are excluded in yum.conf, there is a need to use both is_installed and is_available to ensure we only check for those packages yum would normally install (taking into account the excluded ones)

### DIFF
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -209,11 +209,12 @@ def po_to_nevra(po):
     else:
         return '%s-%s-%s.%s' % (po.name, po.version, po.release, po.arch)
 
-def is_installed(module, repoq, pkgspec, conf_file, qf=def_qf, en_repos=None, dis_repos=None, is_pkg=False):
+def is_installed(module, repoq, pkgspec, conf_file, qf=def_qf, en_repos=None, dis_repos=None, is_pkg=False, group=False):
     if en_repos is None:
         en_repos = []
     if dis_repos is None:
         dis_repos = []
+
     if not repoq:
 
         pkgs = []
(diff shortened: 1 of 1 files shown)


abadger:
I worry a little that this might make things a lot slower.  @kustodian do you have any thoughts on this one?

kustodian:
I agree, this would slow things down, but currently it doesn't work at all. We are yet again in the same situation as before, accuracy vs speed. Regarding the `yum` module I'm more for accuracy, since I think we made it pretty fast anyway in the past few weeks with passing all packages at once and using the yum API.

Looking at the code from a glance as I can see this will only work if `repoquery` is used, and if the yum API is used (which is now default) it won't work. You should make the changes so it works also with yum API, or these changes won't have any impact on most (if not all systems).

With the yum API used, it would be cool if you could run some tests, how long it takes to check different sized groups.

gregdek:
@jfchevrette Thanks for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

gregdek:
@jfchevrette A friendly reminder: this pull request has been marked as needing your action, and it hasn't been updated in the last month. 

robynbergeron:
@jfchevrette Another friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open. If we don't hear from you within another 14 days, we will close this pull request.

gregdek:
Closed due to contributor inactivity.

PR #1802, opened 2015-07-25
ec2 fixes

bcoca:
clarified error messages to actually give back ansible module option instead of internal boto field name

removed redundant checks on type (already done in spec) and just that there is actual content in field


bcoca:
cc @defionscode

juliedavila:
+1

PR #1803, opened 2015-07-25
fix for yum to use localinstall when using an rpm

mscherer:
Try to detect if a rpm is local ( or on http ), and use
localinstall instead of using install, for RHEL system
without subscription.

fixes #11727

### DIFF
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -512,6 +512,7 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
     res['rc'] = 0
     res['changed'] = False
     tempdir = tempfile.mkdtemp()
+    is_local = True
 
     for spec in items:
         pkg = None
@@ -556,6 +557,7 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
         elif  spec.startswith('@'):
             # complete wild ass guess b/c it's a group
             pkg = spec
+            is_local = False
 
         # range requires or file-requires or pkgname :(
         else:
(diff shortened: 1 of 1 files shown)


abadger:
I'm a little afraid of the issue you raise in: https://github.com/ansible/ansible/issues/11727#issuecomment-124822160  With current yum, I would expect that to succeed (if there's no subscription error).

Also, if the rpm specified as a local file needs to have a dependency downloaded, I'm guessing that the user will still get a subscription error?

So maybe the right answer is to have the user fix the subscription error (by registering the machine? disabling the plugin that's causing it? Changing repositories?)

rothgar:
This is also a problem for older versions of `yum install` which do not work the same way with local .rpm packages.

@abadger I don't think the question of if a package has a dependency in a repo should matter. Dependencies can be specified with multiple packages or the rpm may not have any. The problem is older version of `yum install` will always check for subscription state while `yum localinstall` only will if package dependencies are needed.

mscherer:
When you say "older version", can you be a bit more explicit, and what you mean by "do not work the same way", something like "not checking signature", or something else ?

mscherer:
Looking at the code, I do not see a obvious reason why it would fail, since it should just print a warning. I am gonna try with my own laptop, as i suspect the issue is maybe not related to the lack of subscription, but to a side effect of printing the warning or something.

rothgar:
You may be right. It may have just been a failure because of the notice. I'll have to test various versions next week.

gregdek:
This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed -- and since we're trying to get 2.0 out the door, it's unlikely we'll be able to dig into this until then. Thanks for everyone's patience here.

yfried:
We need this feature as well.
ansible 1.9.4
yum 3.4.3
RHEL 7

abadger:
I was asked today what would be needed to move this forward.  I've reviewed what's here and in the referenced bug and I think what we're lacking is a reproducible error case.  Or lacking that, a clear explanation of what's happening in the error condition.

It looks like there's a problem with subscription-manager, not with yum.  There's some statements that this is only happening with old versions of yum (but @yfried-redhat talks about a problem with the yum on rhel7).  

There's also speculation that the problem isn't in what yum does but in our screenscraping of yum output which would mean we just need to adjust our parsing of the output to account for that.

We need answers to those questions in order to figure out how to fix this.

PR #1804, opened 2015-07-25
Add Solaris support in hostname module

saito-hideki:
This patch can be managed hostname for Solaris version 11 or later.

### DIFF
--- a/system/hostname.py
+++ b/system/hostname.py
@@ -21,7 +21,9 @@
 DOCUMENTATION = '''
 ---
 module: hostname
-author: "Hiroaki Nakamura (@hnakamur)"
+author:
+    - "Hiroaki Nakamura (@hnakamur)"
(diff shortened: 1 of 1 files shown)


system/hostname.py
```diff
@@ -363,6 +363,41 @@ def set_permanent_hostname(self, name):
 
 # ===========================================
 
+class SolarisStrategy(GenericStrategy):
+    """
+    This is a Solaris11 or later Hostname manipulation strategy class - it
+    execute hostname command.
+    """
+
+    HOSTNAME_CMD = '/usr/bin/hostname'
```
bcoca: change to use module's get_bin_path function, it future proofs the feature

bcoca:
code looks good aside from comment above

saito-hideki:
@bcoca thank you for your advice.
I modified to use the get_bin_path() function.

PR #1806, opened 2015-07-25
get_url temporary destination option

sysadmin75:
Adds option for temporary destination directory for download file to address ansible/ansible#9512


muffl0n:
Sounds like a nice idea. But you should not put "/tmp" as the default. `tempfile.mkstemp()` does not use /tmp in alle cases. See: https://docs.python.org/2/library/tempfile.html#tempfile.tempdir

For example, on Windows it might be something like c:\tmp or c:\temp.

Edit: Windows seems to be a bad example because there is an own module for that: `win_get_url`. But there is still a possibility that /tmp is not the system's tmpdir.

sysadmin75:
Yep, very valid point.

bcoca:
it defaults to TMPDIR, TEMP or TMP env variables or a platform specific

a default of 'temp dir' and a note in description pointing to the function for details is good enough

sysadmin75:
@bcoca I think the problem @muffl0n is pointing out is that I hard code '/tmp' as the default destination dir.

bcoca:
@sysadmin75 i got that, i was just making a note of what the docs should be when you change it to the default.

sysadmin75:
Closing in preference to https://github.com/ansible/ansible-modules-core/pull/1887

PR #1807, opened 2015-07-26
Support cloning VMs (in addition to templates)

brannon:
Support creating new VMs by cloning an existing VM.
- Existing behavior only supported cloning 'templates'. VMware allows you to clone VMs as well.

Support setting power state of cloned VMs.
- Default behavior of pysphere is to power on VMs after cloning. I want to be able to reconfigure the VMs after cloning, so would prefer that they were powered off.


gregdek:

rhoop:
LGTM

PR #1815, opened 2015-07-27
Allow yum module to handle lists of packages in name:

EdTorbett:
The yum module in 1.9 is not able to handle lists of packages in the name: parameter of the yum module, though the devel branch is and the yum module can benefit from it. This is because the pkgspec is always interpreted as a string. Allowing a list to be provided is a trivial and backward-compatible change with a substantial benefit in forward-compatibility.


EdTorbett:
@kustodian and @abadger this is probably in your area of interest.

bcoca:
so this should not be needed as name is already expected to be a list:

```
 name=dict(aliases=['pkg'], type="list")
```

EdTorbett:
That's from the devel branch - stable-1.9 does not have type="list". I will check if changing that fixes it instead as thst looks like a better solution.

On Monday, 27 Jul 2015 at 18:11, Brian Coca notifications@github.com, wrote:

so this should not be needed as name is already expected to be a list:

 name=dict(aliases=['pkg'], type="list")

kustodian:
1.9 release shouldn't contain any new features, only bug fixes if I understand correctly. Only major releases add new features. 

2.0 will be released in a few weeks and it will contain this feature. I don't see a reason why backport it into 1.9, since after 2.0 nobody will use it in a few months anyway.

PR #1818, opened 2015-07-27
Handle non-updatable openstack subnet details

omgjlk:
Some things cannot be updated via the API, so check for those and fail
if the user is wanting to update them. Also don't try to update ipv6
stuff, as that doesn't work and will cause a traceback.


omgjlk:
Ping @emonty @Shrews

Shrews:
Good patch.  :+1:

emonty:
lgtm

PR #1301, opened 2015-05-10
exclude_hosts does not work as expected with state=drained

vaupelt:
There are established connections for a service. The service is bound to a ipv4-mapped ipv6 address. Wait_for wrongly waits for clients listed in exclude_hosts.

PR #1822, opened 2015-07-28
wait_for: Fix false positives with state=drained

ahill00:
While manually testing wait_for checking for drained connections, I found the
module was wrongly returning drained.

To reproduce the current state:

```
source #] nc -l -p 9797
destination #] nc destination 9797
```

Then, invoke the module:

```
ansible hosts -m "wait_for" -a "port=9797 delay=1 state=drained"
```

The module instantly returns 'drained' even though the connection is open.

This patch has two fixes:
- _get_exclude_ips for Linux returned a list of tuples, not a list of IP strings
- A logic issue detecting the active connection count.


vaupelt:
Eventually you have only forgotten to specify a host. This attribute defaults to 127.0.0.1. I think this fix (the logic issue) will work for you if you want to wait for a connection on any ip different to 127.0.0.1 and haven't specified it properly. Otherwise I encourage you to have a look at PR1301.

ahill00:
@vaupelt - Shouldn't using state='drained' and specifying a host be impossible? There would be no way for a remote host's TCP connections to be inspected.

vaupelt:
@andyhky - the host attribute is compared with local addresses.

_ansible-doc wait_for_ lists some examples (item 2 and 3) with state=drained.

ahill00:
@vaupelt - Thanks for the clarification. I tested #1301 with my scenario and it is more complete since it offers IPv6 support. Closing this PR.

PR #1826, opened 2015-07-28
fix(system/user.py): Python3 try/except compatible

pahaz:


bcoca:
sadly we need to maintain this 2.4 compatible which makes this change a deal breaker.

PR #1832, opened 2015-07-29
Add the option to specify "Canned ACL" for bucket and objects.

j-carl:
Hey,

I added the option "canned_acl" to let you specify "Canned ACL" for objects and buckets. The ACLs for buckets is only set during the CREATE command and for the objects within the PUT command.

If the PUT command creates an bucket no ACL is set for the bucket.


jcowley:
Alternative approach here: https://github.com/ansible/ansible-modules-core/pull/1940 that's consistent with the interface established by the Google gc_storage.py module: https://github.com/ansible/ansible-modules-core/blob/devel/cloud/google/gc_storage.py

bennojoy:
closing since #1940 has been merged..thanks @j-carl

PR #1839, opened 2015-07-30
Windows implementation of lineinfile and related documentation

brianlloyd:
Feature-compatible Windows version of the standard lineinfile module.

Additionally, this module adds support for automatic detection and preservation of file encodings, ability to use in conjunction with create="yes" to create files with specific encodings, and ability to control line separator (windows or unix) used in output files.

Full integration tests translated from the original lineinfile integration tests, plus coverage for added encoding and line sep features are included in a separate pull request to the main ansible repo.

### DIFF
--- a/windows/win_lineinfile.ps1
+++ b/windows/win_lineinfile.ps1
@@ -1,0 +1,452 @@
+#!powershell
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


gregdek:
Marking this as "core_review" since it's a new module being submitted through core.

bennojoy:
@brianlloyd  this looks good, but since it is a bit complex module, we would prefer the module to be added in extras repo first. could you please make a PR against the extras repo.

bennojoy:
merged Thanks..

PR #1840, opened 2015-07-30
Added sha256sum check to files downloaded by unarchive module.

lukehoersten:
This is similar functionality to get_url module and makes sense to be in unarchive with the new download functionally. Really simple addition here. Same code as get_url.

### DIFF
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -59,6 +59,14 @@ options:
     choices: [ "yes", "no" ]
     default: "no"
     version_added: "2.0"
+  sha256sum:
+    description:
+      - If a SHA-256 checksum is passed to this parameter and src contains ://, the digest of the
+        destination file will be calculated after it is downloaded to ensure
+        its integrity and verify that the transfer completed successfully.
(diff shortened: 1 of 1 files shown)


abadger:
get_url in 2.0 will have a new checksum parameter and the sha256sum parameter will be deprecated.  Perhaps we want to use that functionality instead?

lukehoersten:
Yeah we should make it match. I didn't see that in the code though. Is it on a branch other than devel?

abadger:
Ah, I made a mistake - the checksum stuff isn't quite merged yet.  The pr
is here: https://github.com/ansible/ansible-modules-core/pull/141. Just a
few things left to do before that gets merged.

muffl0n:
ansible/ansible#11840 has been merged, so you can use that functionality provided there.

lukehoersten:
Great. I'll work on that today. Thanks.

lukehoersten:
I doubt I'll get around to this any time soon. This PR is deprecated now anyway.

PR #1835, opened 2015-07-29
fix error occurring with Debian

pilou-:
Error was: AttributeError: 'SourcesList' object has no attribute 'repos_urls'

PR #1842, opened 2015-07-30
another solution to apt_repository/repos_urls traceback on Debian (PR #1738, bug #1684)

amenonsen:
This fixes the `i[3]` reference in one commit, and adds a Debian version of `repos_urls` without ppa handling. This does work on Debian and should work on Ubuntu (since the code hasn't changed there, except to call `_expand_ppa` with the right arguments).

### DIFF
--- a/packaging/os/apt_repository.py
+++ b/packaging/os/apt_repository.py
@@ -438,7 +438,7 @@ def main():
     if isinstance(distro, aptsources_distro.UbuntuDistribution):
         sourceslist = UbuntuSourcesList(module,
             add_ppa_signing_keys_callback=get_add_ppa_signing_key_callback(module))
-    elif isinstance(distro, aptsources_distro.DebianDistribution) or isinstance(distro, aptsources_distro.Distribution):
+    elif isinstance(distro, aptsources_distro.Distribution):
         sourceslist = SourcesList(module)
     else:
         module.fail_json(msg='Module apt_repository supports only Debian and Ubuntu.')
--- a/packaging/os/apt_repository.py
+++ b/packaging/os/apt_repository.py
@@ -390,7 +390,7 @@ class UbuntuSourcesList(SourcesList):
                     continue
 
                 if source_line.startswith('ppa:'):
-                    source, ppa_owner, ppa_name = self._expand_ppa(i[3])
+                    source, ppa_owner, ppa_name = self._expand_ppa(source_line)
                     _repositories.append(source)
                 else:
                     _repositories.append(source_line)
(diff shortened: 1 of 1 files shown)


amenonsen:
I pushed an additional commit to this PR to clean up the unnecessarily convoluted HAVE_PYTHON_APT logic (there were double-tests in multiple places, and an annoyingly vague error message). It's tested on Debian and works fine. This can be merged on its own too, if desired.

amenonsen:
And on further consideration, here's another commit that adds a Debian-specific version of `_expand_ppa` that always fails. Because otherwise, if you actually tried to add a ppa repository on Debian, you would get the following traceback because of the call in `main`:

```
AttributeError: 'SourcesList' object has no attribute '_expand_ppa'
```

After this change, here's what I get:

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "ppa repositories are supported only on Ubuntu
```

This implements the behaviour that @sashka seemed to intend for #1738, i.e. that adding ppas fails cleanly on Debian. @abadger does this look OK to you? It works on Debian for me.

/cc @pilou- @verm666 @dalyc does this work for you?

amenonsen:
See https://github.com/amenonsen/ansible-modules-core/tree/repounsplit for an alternative implementation that removes the `UbuntuSourcesList` class altogether. This works with both normal repo lines and ppas on Debian for me. I think this implementation is actually cleaner and more understandable than what we had earlier.

@abadger, I'll await your verdict, if you prefer to commit the single-class version, let me know and I can submit that as a separate PR and close this one. Or if you want to stick to the intention of the original code as @sashka described it, then this PR is the conservative option that still does a bit of worthwhile cleanup.

pilou-:
A PPA is just an APT repository: adding a PPA repository on Debian should work.

sashka:
PPA is an APT repository, `ppa:` shortcut is not.

pilou-:
@sashka indeed, #1835 was not that bad ;)

amenonsen:
@pilou- in fact, I think #1835 was better than what I did. Since your PR was already closed, I included your commit in this PR instead of my earlier solution. @sashka note that @pilou-'s solution doesn't need to put any `repos_urls` or `expand_ppa` into the base class at all. I look forward to any comments. For the record, trying a `ppa:` repo string on Debian now gives this error:

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Invalid repository string: ppa:nginx/stable"}
```

sashka:
:+1:

I also passed it through simple testing like this: https://gist.github.com/sashka/f4f6704224f4fc1e8a98
It works like a charm.

abadger:
Thanks @amenonsen,  @verm666, and @pilou- for implementations and @sashka for reviewing!  glad to get this merged.

# Merged

Hi!

issue is resolved.

Thank you!

PR #1847, opened 2015-07-31
#1842 backport fixup: module is now self.module

amenonsen:
stable-1.9 contained two more references to `module` than devel did, so the backport of #1842 caused build failures such as https://travis-ci.org/debops/ansible-nodejs/builds/73512508#L582

Reported by drybjed on IRC.


abadger:
Thanks @amenonsen !

PR #1358, opened 2015-05-19
observe ` in privileges_unpack

Jmainguy:

PR #1857, opened 2015-08-03
Revert "escapeds changes"

ghost:
While this change doesn't break the creation, it does break
idempotency. This change will convert '*.*' to '`*`.*' which is
functionally the same, however when the user_mod() function looks up
the current privileges with privileges_get() it will read '*.*'

Since '*.*' != '`*`.*' it will go through the process of updating the
privleges always resulting in a 'changed' result.

This reverts commit db9ab9b2629f00350a743a4eca72fb5ee8dc8c77.

### DIFF
--- a/database/mysql/mysql_user.py
+++ b/database/mysql/mysql_user.py
@@ -320,8 +320,13 @@ def privileges_unpack(priv):
     privs = []
     for item in priv.strip().split('/'):
         pieces = item.strip().split(':')
-        dbpriv = pieces[0].rsplit(".", 1)
-        pieces[0] = "`%s`.%s" % (dbpriv[0].strip('`'), dbpriv[1])
+        if '.' in pieces[0]:
+            pieces[0] = pieces[0].split('.')
+            for idx, piece in enumerate(pieces):
+                if pieces[0][idx] != "*":
+                    pieces[0][idx] = "`" + pieces[0][idx] + "`"
+            pieces[0] = '.'.join(pieces[0])
+
         if '(' in pieces[1]:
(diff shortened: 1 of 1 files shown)


Jmainguy:
Some relevent info - This change is to fix the commit in https://github.com/ansible/ansible-modules-core/pull/1358. That commit was introduced to observe ` in privlege unpack. Which allows user to to adjust privs on DB with dots in name.

ghost:
This revert would reintroduce that bug, but this change could be still implemented in a different way. I will submit a new patch for that

Jmainguy:
Can confirm, the commit I created is not working right.

```
[root@phy01 ansible]# ansible -i hosts all -m mysql_user -a "name=bob password=12345 priv=*.*:ALL state=present"
centos7.soh.re | success >> {
    "changed": true,
    "user": "bob"
}

[root@phy01 ansible]# ansible -i hosts all -m mysql_user -a "name=bob password=12345 priv=*.*:ALL state=present"
centos7.soh.re | success >> {
    "changed": true,
    "user": "bob"
}

[root@phy01 ansible]# ansible -i hosts all -m mysql_user -a "name=bob password=12345 priv=*.*:ALL state=present"
centos7.soh.re | success >> {
    "changed": true,
    "user": "bob"
}
```

abadger:
@SamYaple Do we want something like this?

``` python
dbpriv = pieces[0].rsplit(".", 1)
if dbpriv[0].strip('`') != '*':
    pieces[0] = "`%s`.%s" % (dbpriv[0].strip('`'), dbpriv[1])
```

ghost:
Since this got merged rather quickly, here is the new PR that fixes both the dot issue and the '*' issue. Tested and working.

https://github.com/ansible/ansible-modules-core/pull/1934

PR #1867, opened 2015-08-05
Allow a mac address to be specified with vsphere_guest

rjsalts:
When adding network interfaces on a new virtual machine you're unable to manually specify a mac address. This patch will allow an optional element to set the mac address.


rjsalts:
Looks like this has already been requested in https://github.com/ansible/ansible-modules-core/pull/1402

PR #1874, opened 2015-08-06
Changed download behavior to use destination instead of temp directory.

iAnomaly:
I request that the unarchive module's download behavior uses the destination directory rather than tmp because there are cases where:
1. `become_user: theuser` and `theuser`will not have permissions to access tmp
2. We would want to keep the archive

I am also thinking about some date modified and size comparison logic to avoid downloading the archive again if it is unchanged upstream (the idempotent way to do it). I can work on this next.


bcoca:
so I'm not against adding this as an option, but I don't think we should change the default behavior, specially w/o allowing a way to toggle it back.

just curious, why would a user not have access to /tmp? most programs assume it is world accessible.

iAnomaly:
Yes, I think that's a safe assumption for /tmp @bcoca, but the current behavior targets the download to the Play's working directory (e.g. /tmp/ansible-tmp-1438887671.49-42449560457528) and it's permissions and ownership are:
`drwxrwxr-x 2 ansible ansible 4096 Aug  6 12:01 ansible-tmp-1438887671.49-42449560457528`

Note in this case my Play is running as user `ansible` on the host so the working directory is created as that user. Then any `become_user` on unarchive will fail unless said user belongs to the same group as the user running the Play.

So we could:
1. continue using tmp but at the root (/tmp) so there is no permission issues with the working directory (this feels gross to me) and add an option to use the destination instead
2. Use the destination as the default but delete the archive after extraction (and use a non-default option to keep the archive if desired)

What do you think?

bcoca:
This is already factored in, when using become_user that is non root the temporary dir will be made in /tmp instead of the user's home.

iAnomaly:
Right @bcoca, but I am saying that the Ansible temporary directory that is made within /tmp (e.g. /tmp/ansible-tmp-1438887671.49-42449560457528) is constructed by the user running the Play (which is not root) and not by the become_user user. Then when it switches to become_user and tries to download the archive, it fails because the become_user does not have write permissions for /tmp/ansible-tmp-1438887671.49-42449560457528

Example:
1. ansible-playbook runs as user _ansible_
2. Unarchive in the playbook runs as become_user: _test_
3. _/tmp/ansible-tmp-1438887671.49-42449560457528_ is created as user _ansible_
4. Archive fails to download as user _test_ because it lacks write perms on _/tmp/ansible-tmp-1438887671.49-42449560457528_

Am I missing your point? Maybe this specific issue is actually a bug.

bcoca:
this is a bug, it should not be reusing the directory in this case, the only other option would be making it world writable (bad), when the remote user is different and unprivileged it should make a new tmpdir to download to.

gregdek:
Closing, since it looks like this PR is solving the wrong problem given @bcoca comments above. Thanks for the PR @iAnomaly, and sorry it couldn't be included.

PR #1875, opened 2015-08-06
Allow uri to work on older versions of httplib2

wrouesnel:
Simple hack to allow uri to not fail on old versions of httplib2 such as those found in Debian squeeze running python 2.6 ( issue #2291 )

The problem is that disable_ssl_certificate_verification is not a parameter to versions of httplib2 found on these machines, and as a result becomes a hard fail. 

This works around the problem in the uri module but probably needs some more work to make its operation "obvious". I've had no problems using it in practice though.


gregdek:
Thanks @wrouesnel!

@romeotheriault Please review this PR to make sure it adheres to the following guidelines:

abadger:
Took a look and we can't support httplib2 < 0.7.0.  The reason is that versions older than that couldn't properly verify TLS certificates and are thus insecure.

As a workaround I can document the version of httplib2 that is needed for this module to work and then people with older versions can use pip to install a newer one (or use a different module -- if all else fails, using command with curl or wget should do the trick and be secure.)

abadger:
Okay, documentation of hte requirement added and also checked the version when we import httplib2 so that we don't attempt to operate without a new enough version.  If people come across this via google later, the potential workarounds are to install a newer httplib2 via pip, use the get_url module instead, or use command with curl or wget.

Issue #1865, opened 2015-08-04
Custom Windows facts


davidobrien1985:
Hi everybody,

I would like to create custom facts on Windows machines.
The current implementation of setup.ps1 is a good start but lacks the capability to capture custom facts.

The setup module's parameter fact_path doesn't work on Windows yet. (http://docs.ansible.com/ansible/setup_module.html)

I was thinking about editing the setup.ps1 to parse a default location where admins can drop custom scripts (*.ps1) and the setup.ps1 would execute each script and add each script's output to the output JSON.
I'd like to propose C:\Windows\Ansible\facts as this default location.

Also, I'd like this to be customisable via fact_path.

Is the Ansible team already working on this? If not, please discuss my proposal. I'm happy to do the work and later on do a PR.

Cheers
David

nitzmahone:
Merged for 2.1, closing...

PR #1876, opened 2015-08-06
Custom Windows Facts

davidobrien1985:
As explained here: https://github.com/ansible/ansible-modules-core/issues/1865

Following changes to setup.ps1:
- added Function Get-CustomFacts
- enabled parsing of params and using facts_path
  -- this will enable current users to easily understand what this variable is for
- any ps1 file in facts_path will be executed (obviously a security issue, but it's the same on Unix), make sure regular users can't access this folder!

I strongly rely on the person writing those custom ps1 files. They will have to make sure that their code returns data in a nice manner.

### DIFF
--- a/system/setup.py
+++ b/system/setup.py
@@ -54,12 +54,14 @@ notes:
       install I(facter) and I(ohai) means you can avoid Ruby-dependencies on your
       remote systems. (See also M(facter) and M(ohai).)
     - The filter option filters only the first level subkey below ansible_facts.
-    - If the target host is Windows, you will not currently have the ability to use
-      C(fact_path) or C(filter) as this is provided by a simpler implementation of the module.
-      Different facts are returned for Windows hosts.
+    - If the target host is Windows you can now use C(fact_path). Make sure that this path 
+      exists on the target host. Files in this path MUST be PowerShell scripts (*.ps1) and 
+      their output must be formattable in JSON (Ansible will take care of this). Test the 
+      output of your scripts.
 author:
     - "Ansible Core Team"
     - "Michael DeHaan"
+    - "David O'Brien @david_obrien davidobrien1985"
 '''
 
 EXAMPLES = """
@@ -74,6 +76,9 @@ ansible all -m setup -a 'filter=facter_*'
 
 # Display only facts about certain interfaces.
 ansible all -m setup -a 'filter=ansible_eth[0-2]'
+
+# Display facts from Windows hosts with custom facts stored in C(C:\\custom_facts).
+ansible windows -m setup -a "fact_path='c:\\custom_facts'"
 """
 
 
--- a/system/setup.py
+++ b/system/setup.py
@@ -35,7 +35,7 @@ options:
         description:
             - path used for local ansible facts (*.fact) - files in this dir
               will be run (if executable) and their results be added to ansible_local facts
-              if a file is not executable it is read.
+              if a file is not executable it is read. Check notes for Windows options. (from 2.1 on)
               File/results format can be json or ini-format
         required: false
         default: '/etc/ansible/facts.d'
@@ -54,10 +54,13 @@ notes:
       install I(facter) and I(ohai) means you can avoid Ruby-dependencies on your
       remote systems. (See also M(facter) and M(ohai).)
     - The filter option filters only the first level subkey below ansible_facts.
+    - If the target host is Windows, you will not currently have the ability to use
+      C(filter) as this is provided by a simpler implementation of the module.
     - If the target host is Windows you can now use C(fact_path). Make sure that this path 
       exists on the target host. Files in this path MUST be PowerShell scripts (*.ps1) and 
       their output must be formattable in JSON (Ansible will take care of this). Test the 
       output of your scripts.
+      This option was added in Ansible 2.1.
 author:
     - "Ansible Core Team"
     - "Michael DeHaan"
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -35,7 +35,7 @@ Function Get-CustomFacts {
   $FactsFiles = Get-ChildItem -Path $factpath | Where-Object -FilterScript {($PSItem.PSIsContainer -eq $false) -and ($PSItem.Extension -eq '.ps1')}
 
   foreach ($FactsFile in $FactsFiles) {
-      $out = . $($FactsFile.FullName)
+      $out = & $($FactsFile.FullName)
       Set-Attr $result.ansible_facts "ansible_$(($FactsFile.Name).Split('.')[0])" $out
   }
 }
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -17,13 +17,40 @@
 # WANT_JSON
 # POWERSHELL_COMMON
 
-# $params is not currently used in this module
-# $params = Parse-Args $args;
+# enabled $params (David O'Brien, 06/08/2015)
+$params = Parse-Args $args;
+
+
+Function Get-CustomFacts {
+  [cmdletBinding()]
+  param (
+    [Parameter(mandatory=$false)]
+    $factpath = $null
+  )
+
+  if (-not (Test-Path -Path $factpath)) {
+    Fail-Json $result "The path $factpath does not exist. Typo?"
+  }
+
+  $FactsFiles = Get-ChildItem -Path $factpath | Where-Object -FilterScript {($PSItem.PSIsContainer -eq $false) -and ($PSItem.Extension -eq '.ps1')}
+
+  foreach ($FactsFile in $FactsFiles) {
+      $out = . $($FactsFile.FullName)
+      Set-Attr $result.ansible_facts "ansible_$(($FactsFile.Name).Split('.')[0])" $out
+  }
+}
 
 $result = New-Object psobject @{
     ansible_facts = New-Object psobject
     changed = $false
 };
+
+# failifempty = $false is default and thus implied
+$factpath = Get-AnsibleParam -obj $params -name fact_path
+if ($factpath -ne $null) {
+  # Get any custom facts
+  Get-CustomFacts -factpath $factpath
+}
 
 $win32_os = Get-CimInstance Win32_OperatingSystem
 $win32_cs = Get-CimInstance Win32_ComputerSystem


bcoca:
cc @jhawkesworth @cchurch @trondhindenes

jhawkesworth:
I like the idea of being able to have custom facts on windows, but I'm not sure about this approach.

When I wanted custom facts, I took the existing setup.ps1 module and copied it into my /etc/ansible/library folder, then added the extra fact collecting logic in there.  The /etc/ansible/library version takes precedence over core modules so I didn't have to modify my playbooks to use it.  Because I was tweaking a copy of the existing module, it meant I didn't have to create my own logic to return the necessary json.

I'll think about this a bit more as it would definitely be good to have a location within the facts json structure where custom facts are expected.

davidobrien1985:
After reading your comment twice I now think I get what you mean :D 
So, before you usually went and just added the logic you needed into setup.ps1? Okay, but then everybody goes and changes setup.ps1 to their liking.
Also, in larger environments, you have constant changes to setup.ps1, because every project might rely on different facts.

This is why I like the idea of giving users the ability to just drop a custom ps1 file into a default or custom location and that gets automatically picked up by Ansible. This way setup.ps1 is generic for an environment and projects can do whatever they want on their own servers and be sure to get their facts into Ansible.
My plan is to get a next version of this out soon that also supports other file types, not just ps1.

trondhindenes:
I like it. We should maybe look at executing the custom facts scripts natively instead of spawning a new powershell.exe process (just for the sake of execution time).

davidobrien1985:
Any further input needed here? What's the verdict? ;)
Happy to work some further on this if necessary.

lugoues:
Should this not follow the same convention as the core setup module (or as close as it can get)?
- ansible_facts
  - ansible_local
    - [facts_file_name]
      - [file contents]

jhawkesworth:
Seems like a good idea to make it a similar as possible to the existing
convention.  I'd be inclined to drop the .ps1 or other extension off the
[facts_file_name] of course so you could arrange things so that you could
have local facts across windows and linux boxes, but retrieve and process
them in the same way.

davidobrien1985:
So, if you want to drop the ps1 extension, how do you want to execute the file / collect the system information without having anything like facter or ohai installed on windows?

Or, do you mean you expect something to create a local file on Windows which already has all the information and the ansible setup module just reads that file? Like in ini format?
I can add that to the code, but I also don't want to drop the ps1 functionality. This will enable users to add custom code on their Windows servers that will dynamically create the facts every time the setup module is invoked.

jhawkesworth:
No sorry I only meant drop the .ps1 from the facts json response as
returned by the module (following on from Peter B's point).

It would be good to be able to either read the file contents or execute the
file and have its output added to the facts though.

trondhindenes:
@davidobrien1985 , you should be able to just execute the facts script directly (instead of starting a powershell.exe process). Executionpolicy is already handled higher up so we shouldn't have to worry about that. Would you care to rewrite it slightly? It's super-important to have setup run as quickly as possible, that's why I'd like us to be a bit picky about execution time :-)

davidobrien1985:
Hi,

changed the code a bit. I will now execute the scripts directly and have gotten back nice JSON output.

Is this more what you're after?

trondhindenes:
yup, looking good! I'm gonna get this tested thoroughly and then report back. Thanks for the contribution David!

trondhindenes:
One more comment David, looking more closely at the code. Instead of throwing an error if the facts_path does not contain a valid file/folder, you should return a json object using `Fail-Json $result "someexceptionmessage"`.

davidobrien1985:
I am only checking if the facts_path directory exists. If it's empty, it doesn't matter, I will just return nothing. I don't want to fail just because of that. 
So Test-Path only checks if the actual folder exists, not if anything is in there.

Reason why I'm doing it that way is because it might be that someone provides that param facts_path on the ansible call, but the folder either doesn’t exist on all nodes or it's empty on some. That shouldn't make the setup module fail, because I still want the other facts from those machines.

Does that make more sense?

From:  trondhindenes
Reply-To:  ansible/ansible-modules-core
Date:  Friday, 18 September 2015 6:47 am
To:  ansible/ansible-modules-core
Cc:  David O'Brien
Subject:  Re: [ansible-modules-core] Custom Windows Facts (#1876)

One more comment David, looking more closely at the code. Instead of throwing an error if the facts_path does not contain a valid file, you should return a json object using Fail-Json $result "someexceptionmessage". Also make sure to validate that the path is a file (leaf) (test-path also returns true on folders)

trondhindenes:
I've done some testing now, and as far as I can see, this works like a charm. Only thing I'd change before getting it merged is to remove the "throw" statement and instead return the failed status using the return object if for example the path is invalid (look at the win_feature module for instance, if yoy're unsure how to implement the failure response)

davidobrien1985:
Alright, I can see why you want it to fail, so I changed the "break" for a "Fail-Json".

trondhindenes:
Cool. I don't know how the linux version handles non-existing paths, but in Ansible we should never use `throw` as a way of existing script execution.

davidobrien1985:
Nice, let's merge ;)

trondhindenes:
yup, +1 from me.

davidobrien1985:
@jhawkesworth any further comments on this?

windows/setup.ps1
```diff
@@ -17,15 +17,40 @@
...
+      Set-Attr $result.ansible_facts "ansible_$(($FactsFile.Name).Split('.')[0])" $out
+  }
+}
 
 $result = New-Object psobject @{
     ansible_facts = New-Object psobject
     changed = $false
 };
 
+If ($params.facts_path) {
```
blakfeld: I think this will fail on Ansible 2.0 since it forces Powershell strict. I've modify this in a few modules. Check out Get-AnsibleParam in the powershell.ps1 file in module_utils

trondhindenes: Agreed. Although I absolutely hate strictmode it's there now, so we have to support it. Should be something like

``` powershell
$factspath = Get-AnsibleParam -obj $params -name facts_path
if ($factspath -ne $null)
{}
```

davidobrien1985: I agree, won't run on >2.0, but I don't know if I only want this to work in 2.0.

https://groups.google.com/forum/#!topic/ansible-devel/xLqzpl66yeg

I'll ask this here as well: Should "fixes" and new things be written only having 2.0 in mind?

jhawkesworth:
Sorry for delay replying about this - a million things on at the moment.

I would say definitely new things should be written to work with latest development version of ansible as  I don't believe modules are backported unless there is a specific, probably security-related reason.

My main concern with this is not increasing the time taken to gather facts.  Hopefully work described here https://groups.google.com/forum/#!topic/ansible-devel/aAFdMPvYiFs will speed up getting modules onto windows machines sufficiently that this is less of a concern, but I haven't had a chance to test that yet.

Please can you revise for >2.0 support so that this can be tested on dev 2.0 versions.

davidobrien1985:
Ok, tested with version 2.

davidobrien1985:
Anything missing to get this merged?
@jhawkesworth @trondhindenes @bcoca @cchurch

trondhindenes:
Looks like there is a merge conflict in the PR? (line 57)

jhawkesworth:
Ok, finally had some time to catch up with PRs.  
Could the parameter be named 'fact_path' to match the equivalent in the non-windows version of setup?
Could the documentation be updated http://docs.ansible.com/ansible/setup_module.html  (currently it says fact_path not available on windows).  Also I think it should be documented that there is no default fact path on windows, and the scripts need to be ps1 files etc.

gregdek:
Thanks @davidobrien1985 for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

davidobrien1985:
Hi guys,
updated the documentation and changed the parameter name.

All good? ;)

nitzmahone:
@davidobrien1985 - I'm good to merge this for devel, but the included merge commits cause problems (we usually prefer rebases to devel). Sorry for the final hassle- can you re-push this branch as squashed commits (or a single patch commit)? Alternatively, I can build and commit the patch for you, but then it'll have my name on it as far as Git's concerned.

davidobrien1985:
@nitzmahone this should look better now. Proper rebase.
@trondhindenes thanks for reminding me of this :)

davidobrien1985:
Any further comments on this?

windows/setup.ps1
```diff
@@ -17,14 +17,41 @@
...
+  )
+
+  if (-not (Test-Path -Path $factpath)) {
+    Fail-Json $result "The path $factpath does not exist. Typo?"
+  }
+
+  $FactsFiles = Get-ChildItem -Path $factpath | Where-Object -FilterScript {($PSItem.PSIsContainer -eq $false) -and ($PSItem.Extension -eq '.ps1')}
+
+  foreach ($FactsFile in $FactsFiles) {
+      $out = . $($FactsFile.FullName)
```
nitzmahone: Do you have a specific reason for dot-sourcing instead of call-ing (&)? Seems like the definition persistence behavior of dot-sourcing would be a negative here when we're likely only interested in the result (potential to collide with built-in funcs, etc). (sorry I didn't catch this before with all the merge noise)

nitzmahone:
Sorry for the delay, had a few weeks in a row of craziness without a proper PR triage pass.

davidobrien1985:
@nitzmahone Matt, I changed the invocation. Good catch! Tested it, still good. ;)

system/setup.py
```diff
@@ -54,12 +54,14 @@
       install I(facter) and I(ohai) means you can avoid Ruby-dependencies on your
       remote systems. (See also M(facter) and M(ohai).)
     - The filter option filters only the first level subkey below ansible_facts.
-    - If the target host is Windows, you will not currently have the ability to use
-      C(fact_path) or C(filter) as this is provided by a simpler implementation of the module.
-      Different facts are returned for Windows hosts.
+    - If the target host is Windows you can now use C(fact_path). Make sure that this path 
+      exists on the target host. Files in this path MUST be PowerShell scripts (*.ps1) and 
+      their output must be formattable in JSON (Ansible will take care of this). Test the
```
jhawkesworth, line 61: Would it be worth adding that for windows hosts fact_path was added in ansible 2.1 and restoring some of the text about filter not being available on windows hosts?

davidobrien1985:
@jhawkesworth added the notes to `/system/setup.py`

jhawkesworth:
Thanks @davidobrien1985 - I know it was only a tiny thing but might as well be clear about it.

PR #1887, opened 2015-08-06
Adds tmp_dest option to get_url module.

sysadmin75:
Addresses the issue in ansible/ansible#9512

### DIFF
--- a/network/basics/get_url.py
+++ b/network/basics/get_url.py
@@ -55,6 +55,14 @@ options:
         If C(dest) is a directory, the file will always be
         downloaded (regardless of the force option), but replaced only if the contents changed.
     required: true
+  tmp_dest:
+    description:
+      - absolute path of where temporary file is downloaded to.
+      - Defaults to TMPDIR, TEMP or TMP env variables or a platform specific value
+      - https://docs.python.org/2/library/tempfile.html#tempfile.tempdir
+    required: false
+    default: ''
+    version_added: '2.0'
   force:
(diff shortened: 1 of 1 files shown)


manicminer:
I've tested this change with the `tmp_dest` option set for a 14GB file download and it works as expected.

manicminer:
One suggestion from what I've observed, it would be great if the temporary file could be moved, rather than copied, to the final destination. This should probably only happen if the temporary path and destination path are on the same volume, which might be hard to determine. Maybe it could be moved in the event that `tmp_dest` is specified?

muffl0n:
Maybe shutil.copyfile() in [Line 347](https://github.com/sysadmin75/ansible-modules-core/blob/9512_get_url_tmp_option2/network/basics/get_url.py#L347) should be replaced with [shutil.move()](https://docs.python.org/2/library/shutil.html#shutil.move):
`If the destination is on the current filesystem, then os.rename() is used. Otherwise, src is copied (using shutil.copy2()) to dst and then removed.`

So that's basicly the same behaviour as we have now (shutil.copy() then os.remove) but with the benefit of faster operation if tmp_src and dst are on the same fs.

robynbergeron:
@sysadmin75 -- thanks for this PR, and thanks to @tombamford and @muffl0n for your comments. :)

@jpmens -- Please review this PR to make sure it adheres to the following guidelines:

jpmens:
shipit

PR #726, opened 2015-01-30
jinja2 support for cloudformation module

vorlock:
Hi all,
I posted this (https://groups.google.com/d/msg/ansible-devel/j3QiAu7EZmE/UGdYFqWy6x8J) message on ansible-devel. Do you think it's feasible to use something like it or are there philosophical constrains against it?

PR #1890, opened 2015-08-07
Cloudformation jinja templating

eslng:
This will add ability to use jinja2 templates to render cloudformation template whenever it ends with j2.
In addition there are few arguments added:
- j2_vars_file to store variables for jinja template
- j2_dump_json to help debugging which will dump rendered jinja2 template to stdout

This is  rebase of #726

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -92,6 +92,12 @@ options:
     required: false
     default: null
     aliases: []
+  j2_dump_json:
+    description:
+      - if true after jinja2 run will dump whole json template to stdout
+    required: false
+    default: null
+    aliases: []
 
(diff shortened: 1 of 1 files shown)


gregdek:
@jsmartin Please review this PR to make sure it adheres to the following guidelines:

jsmartin:
I don't feel this is the correct approach.  Templating JSON seems wrong.  Troposhpere is much better way to generate proper Cloudformation templates that could then be consumed by this module.

eslng:
I don't think troposhere provides same level of visibility

gregdek:
Interpreting @jsmartin comment here as -1, which means we're not going to accept this pull request. Nevertheless, a sincere thanks to @eslng for making the effort.

PR #1894, opened 2015-08-10
Task to copy a file remotely

aseigneurin:
I would like to submit this task that copies a file from the remote server to the remote server. This task differs from the "copy" task that copies a file from the local host to the remote server.

The copy is only performed if the MD5 sum of the file has changed, making this task idempotent, as opposed to executing the "cp" shell command.

Let me know if this makes sense or if there is a better alternative.

### DIFF
--- a/files/copy_remotely.py
+++ b/files/copy_remotely.py
@@ -1,0 +1,87 @@
+#!/usr/bin/python
+
+# -*- coding: utf-8 -*-
+
+# (c) 2015, Alexis Seigneurin <alexis@seigneurin.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


alex-deref:
Hi, I had a similar requirement just the other day, and put together a draft with the same idea, but also supporting recursive copy. You can see discussion on the mailing list (subject: New module (draft) for copying files locally.)

https://groups.google.com/forum/#!topic/ansible-devel/97ppWbwLp5w

And you can see my code here:

https://gist.github.com/alex-deref/dc66a0ca7e1904c30508

aseigneurin:
Looks very nice! I shall give this a try.

aseigneurin:
I've just tried your module and it's great!

It is a superset of the module I have submitted as you can both copy a whole directory or a single file, in both cases while being idempotent.

I would suggest doing a pull request with your code instead of mine.

gregdek:
Hi @aseigneurin -- thanks for the PR! Unfortunately, new modules aren't being accepted into the Core modules repo. Could you please file this PR against the ansible/ansible-modules-extras repo and reference this PR in the new module? That way we can get the needed reviewers to take a look at it. Thanks, and sorry for the delay in responding here.

bcoca:
actually, this overlaps with existing functionality in the copy module  (remote_src=true)

PR #1902, opened 2015-08-11
Added mutex file locking (lock) option to lineinfile module.

RusDavies:
When using the lineinfile module for delegated tasks, it can happen that multiple processes interfere with each other's changes.  An example may be as follows, where the 'example' group contains multiple machines.

``` yaml

---
name: Example of lineinfile borking a file
  - hosts: example
  - tasks:
    - name: Modify a local file
      lineinfile:
        dest: /example.txt
        regexp: "^{{ inventory_hostname }}"
        line: "{{ inventory_hostname }} {{ 100000 | random }}"
        state: present 
        insertafter: EOF 
      delegate_to: 127.0.0.1 
```

In the above example, data will often be lost, as each forked process attempts to overwrite the same file.  

It is of course possible to set serial=1 on the task, thereby serializing all items.  Alternately, one could split the task, separating those parts that demand serialization  from those that benefit from parallelization.  However, both approaches have potential drawbacks, in terms of performance and ease of implementation.  It would instead be preferable for lineinfile to be tolerant of multiple competing processes. 

To that end, this pull request adds a lock option, using fcntl.flock().   Modified usage is then as follows, (notice the new 'lock: yes' option):

``` yaml

---
name: Example of lineinfile used with 'lock' to prevent borking a file
  - hosts: example
  - tasks:
    - name: Modify a local file
      lineinfile:
        lock: yes
        dest: /example.txt
        regexp: "^{{ inventory_hostname }}"
        line: "{{ inventory_hostname }} {{ 100000 | random }}"
        state: present 
        insertafter: EOF 
      delegate_to: 127.0.0.1 
```

With the optional lock enabled, competing forked processes do not trample on each other's changes.

### DIFF
--- a/files/lineinfile.py
+++ b/files/lineinfile.py
@@ -23,6 +23,7 @@ import re
 import os
 import pipes
 import tempfile
+import fcntl
 
 DOCUMENTATION = """
 ---
@@ -118,6 +119,15 @@ options:
      description:
        - Create a backup file including the timestamp information so you can
          get the original file back if you somehow clobbered it incorrectly.
+  lock:
+     required: false
+     default: "no"
+     choices: [ "yes", "no" ]
+     description:
+       - Require a mutex lock on the destination file.  Useful when multiple 
+         forked processes act on the same file, as may happen when a task
+         is delegated to a central machine.
+
   others:
      description:
        - All arguments accepted by the M(file) module also work here.
@@ -182,7 +192,12 @@ def check_file_attrs(module, changed, message):
 
 def present(module, dest, regexp, line, insertafter, insertbefore, create,
             backup, backrefs):
-
+    
+    lock_file = None
+    if module.params['lock'] == True:
+        lock_file = open(dest + ".ansible_lock", 'w')
+        fcntl.flock(lock_file.fileno(), fcntl.LOCK_EX)
+    
     if not os.path.exists(dest):
         if not create:
             module.fail_json(rc=257, msg='Destination %s does not exist !' % dest)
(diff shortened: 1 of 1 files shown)


gregdek:
@dhozac Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@dhozac @ahtik This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
@dhozac @ahtik still waiting on your review.  Please comment with text 'shipit' or 'needs_revision' as appropriate. If we don't hear from you within 14 days, we will start to look for additional maintainers for this module.

ahtik:
Looks good, sorry about the delay! shipit

bcoca:
This is not a problem with the module itself, this is an issue with any module that is delegated against a single resource for multiple machines.

The solution should be generic so we can apply to multiple modules. Currently the workaround is pushing the task to a play with `serial: 1`.

RusDavies:
@bcoca, yes, mutex file access should be generic and available to any module.  I was surprised it's not already so. 

As a counter point, in my experience, this module happens to one of the most important w.r.t. mutex file access, and as such would benefit the most from the solution.  The performance penalty of "serial: 1" may be less important on other less heavily used modules.  Therefore, the value proposition of being DRY may be based on a false equivalence of the importance that mutex file access represents to each module. 

For this specific module, and ignoring any test code, the solution amounts to 20 extra lines, as compared to perhaps between 11 to 13 that may be necessary even if the solution were moved out as a utility. So, we're talking about saving perhaps 7 to 9 lines for each module that may require such a feature.  Moreover, moving the solution elsewhere will add many more lines of code for as-yet unknown demand in other modules. 

Ultimately, whereas this particular module may be where most of the value is gained, the benefit of being DRY may not be worth the dollar-hours of the effort of implementing a more generic solution.  

I suggest the code be shipped as is, and, if there proves to be demand in other modules, then we investigate the effort required to make it generic.  

Thoughts?

bcoca:
we already have this demand for replace, copy, assemble and template, basically most modules that do file operations.

RusDavies:
OK, so I'll close this pull request as dead, and instead await the generic solution that results from those other feature requests.

Issue #1387, opened 2015-05-21
service module enable/disable does not work properly with sysv init scripts


MikeN123:
##### Issue Type:

Bug Report
##### Ansible Version:

ansible 1.9.1
  configured module search path = None
##### Ansible Configuration:

Nothing.
##### Environment:

Mac OS X client running against a Debian 8.0 (jessie) host.
##### Summary:

The service module uses `systemctl is-enabled` to check service state on a systemd enabled system, which does not work for sysv services (at least not on Debian). There is also a bug report at Debian for this: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=751638

I believe the actual bug is in Debian, but hopefully a workaround can be provided by Ansible.
##### Steps To Reproduce:

Try to enable some sysv-only init script on a Debian jessie system:

```
    - name: auto-start my-service
      service: name=my-service enabled=yes
```
##### Expected Results:

The first time it should report `changed = True`, and it should report `changed = False` on further invocations.
##### Actual Results:

Always reports `changed = True`.

bcoca:
closing this as duplicate of #915

PR #1905, opened 2015-08-12
Fix service.enabled on: Debian 8, Ubuntu 15.04

evverx:
**Issue Type:** Bugfix Pull Request

**Ansible Version:** ansible 2.0.0 (devel e0b074000e) last updated 2015/08/12 10:02:40 (GMT +000)

**Environment:**

``` bash
$ lsb_release -rd
Description:    Ubuntu 15.04
Release:    15.04
```

**Summary:**
There are a "legacy" services (for example, `ntp`): they have a `/etc/init.d/name`script but don't have a `/lib/systemd/system/name.service`.
`get_systemd_service_enabled` always returns `False` for those "legacy" services.

See also: https://github.com/systemd/systemd/issues/937

**Steps To Reproduce:**

```
# apt-get install ntp
# systemctl enable ntp
# ansible all -m service -a 'name=ntp enabled=false' --sudo
```

**Expected Results:**

```
127.0.0.1 | SUCCESS => {
    "changed": true,
    "enabled": false,
    "name": "ntp"
}
```

**Actual Results:**

```
127.0.0.1 | SUCCESS => {
    "changed": false,
    "enabled": false,
    "name": "ntp"
}
```

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -503,10 +503,21 @@ class LinuxService(Service):
             self.svc_initctl = location['initctl']
 
     def get_systemd_service_enabled(self):
-        (rc, out, err) = self.execute_command("%s is-enabled %s" % (self.enable_cmd, self.__systemd_unit,))
+        def sysv_exists(name):
+            script = '/etc/init.d/' + name
+            return os.access(script, os.X_OK)
+
+        def sysv_is_enabled(name):
+            return bool(glob.glob('/etc/rc?.d/S??' + name))
+
+        service_name = self.__systemd_unit
+        (rc, out, err) = self.execute_command("%s is-enabled %s" % (self.enable_cmd, service_name,))
         if rc == 0:
             return True
-        return False
+        elif sysv_exists(service_name):
+            return sysv_is_enabled(service_name)
+        else:
+            return False
 
     def get_systemd_status_dict(self):
         (rc, out, err) = self.execute_command("%s show %s" % (self.enable_cmd, self.__systemd_unit,))


bcoca:
This has been a known bug with debian/ubuntu, but this patch seems to work around it nicely, defaulting to checking systemd itself first.

evverx:
@bcoca , thanks for the review. Do you mean this bug: #1387 ?

system/service.py
```diff
@@ -503,10 +503,21 @@ def check_systemd():
             self.svc_initctl = location['initctl']
 
     def get_systemd_service_enabled(self):
-        (rc, out, err) = self.execute_command("%s is-enabled %s" % (self.enable_cmd, self.__systemd_unit,))
+        def sysv_exists(name):
+            script = '/etc/init.d/{0}'.format(self.name)
```
bcoca: format breaks 2.4 compatibility

evverx: thanks. fixed.

bcoca:
kindof, mean this https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=751638, though this seems to have been fixed and the bug you are addressing is a follow up that appeared when this was fixed.

system/service.py
```diff
@@ -503,10 +503,21 @@ def check_systemd():
             self.svc_initctl = location['initctl']
 
     def get_systemd_service_enabled(self):
-        (rc, out, err) = self.execute_command("%s is-enabled %s" % (self.enable_cmd, self.__systemd_unit,))
+        def sysv_exists(name):
+            script = '/etc/init.d/' + name
+            return os.access(script, os.X_OK)
+
+        def sysv_is_enabled(name):
+            return bool(glob.glob('/etc/rc?.d/S??' + self.name))
```
bcoca: you pass name, but use self.name?

evverx: nice catch! fixed.

alfredocambera:
In which version of ansible is available the fix?? I'm using 1.9.3 on Debian Jessie 8.2 but I'm getting **'changed'** every time I enable the service (jenkins using initd):

```
vagrant@mgmt:~$ ansible jenkins -m service -a "name=jenkins enabled=true" -v -s                                                                  
jenkins | success >> {
    "changed": true, 
    "enabled": true, 
    "name": "jenkins"
}

vagrant@mgmt:~$ ansible jenkins -m service -a "name=jenkins enabled=true" -v -s
jenkins | success >> {
    "changed": true, 
    "enabled": true, 
    "name": "jenkins"
}
```

**OS and ansible versions:**

```
vagrant@mgmt:~$ ansible --version
ansible 1.9.3
vagrant@mgmt:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 8.2 (jessie)
Release:        8.2
Codename:       jessie
vagrant@mgmt:~$ ansible jenkins -m shell -a "lsb_release -a" -v -s
jenkins | success | rc=0 >>
Distributor ID: Debian
Description:    Debian GNU/Linux 8.2 (jessie)
Release:        8.2
Codename:       jessieNo LSB modules are available.
```

Let me know if I can help with the testing of this issue.

evverx:
This fix available in the development version only.
Try it.

See also:
- [What Version To Pick?](http://docs.ansible.com/ansible/intro_installation.html#what-version-to-pick)
- [Running From Source](http://docs.ansible.com/ansible/intro_installation.html#running-from-source)

alfredocambera:
@evverx thanks

evverx:
@alfredocambera , there is a problem with Jenkins: [[JENKINS-29965] Debian Package, Ubuntu 15.04 and systemd results in a non working installation](https://issues.jenkins-ci.org/browse/JENKINS-29965).

You can vote for that issue.
Or you could add the file `jenkins.service` to the [package](https://github.com/jenkinsci/packaging/tree/master/deb/build/debian) :)

alfredocambera:
@evverx I tried adding the jenkins.service file on Debian but I didn't work.

I had come across that bug. Indeed I had on add a restart after the installation of jenkins on Debian Jessie, Ubuntu 14.04 and Centos 7 on my ansible role (https://github.com/alfredocambera/ansible-jenkins/blob/master/tasks/main.yml) to make jenkins work. So I think I'll give that issue a vote.

Thanks for the information!

PR #1913, opened 2015-08-14
Turn off passwword authentication for rsync

alex-deref:
In working with the rsync module I was having trouble with hanging processes, due to password prompts hanging in the background.

### DIFF
--- a/files/synchronize.py
+++ b/files/synchronize.py
@@ -173,6 +173,8 @@ notes:
    - Expect that dest=~/x will be ~<remote_user>/x even if using sudo.
    - To exclude files and directories from being synchronized, you may add 
      C(.rsync-filter) files to the source directory.
+   - Does not use SSH agent, authentication must be done using keys using a
+     key without a passphrase.
 
 
 author: "Timothy Appnel (@tima)"
@@ -334,7 +336,9 @@ def main():
     else:
         private_key = '-i '+ private_key 
 
-    ssh_opts = '-S none'
+    # PasswordAuthentication is disabled to prevent the background SSH process
+    # from blocking on input.
+    ssh_opts = '-S none -o PasswordAuthentication=no'
 
     if not verify_host:
       ssh_opts = '%s -o StrictHostKeyChecking=no' % ssh_opts


files/synchronize.py
```diff
@@ -173,6 +173,8 @@
    - Expect that dest=~/x will be ~<remote_user>/x even if using sudo.
    - To exclude files and directories from being synchronized, you may add 
      C(.rsync-filter) files to the source directory.
+   - Does not use SSH agent, authentication must be done using keys using a
```
jody-frankowski, line 176: Could you elaborate about the ssh agent? `PasswordAuthentication=no` doesn't prevent the use of the agent.
In any case, ansible should be able to work with an ssh agent. Removing key files passphrases is a security risk for a lot of users.

alex-deref, line 176: I had ssh-agent running, and when I ran ssh, Ansible, or rsync directly on the command-line, all three of them would use it to perform the login.

However, when I used the Ansible synchronize task it would hang, presumably waiting for terminal input. If I added a passphraseless default key, id_rsa, then it was able to successfully login. If I added the PasswordAuthentication=No option (and removed id_rsa key), then it would fail to log in and report an error.

The PasswordAuthentication=no option instructs SSH to not use password authentication, and only use cryptographic key authentication. This is fine because using passwords requires a prompt which caused the hang I observed.

My reason for the note "Does not use SSH agent" was due to the fact that the synchronize module did not discover and use my SSH agent. I hypothesise that it couldn't find the SSH agent because the environment had been modified and connection details weren't available to the subprocess.

gregdek:
@tima Please review this PR to make sure it adheres to the following guidelines:

tima:
This is something I was talking to @abadger and @defionscode about a couple of weeks ago. Using a passphrase does complicate things, but I think it's reasonable enough request to support passphrases in some way. IMO committing this will really irk those people so I'm opposed to it.

juliedavila:
What if we made this an optional parameter?

alex-deref:
How would the password prompt work? Ansible does not relinquish control of the terminal, so that can't be used, and it seems the environment isn't passed through either because it couldn't find my agent, which makes accessing X difficult. It might require changes to Ansible itself.

I'd be happy if it were optional, and would favour the default to be not prompting because I consider a hang to be a poor user experience.

files/synchronize.py
mgedmin, line 176: For the record, I've been using Ansible's synchronize module to rsync some files keys from server A to server B, relying on my ssh-agent for authentication.  It worked fine.

(Often what breaks ssh agent auth is using 'sudo' to become root -- sudo clears SSH_AUTH_SOCK from the environment, with the default sudo configuration on Ubuntu.)

abadger:
tl;dr;  The ssh-agent note doesn't seem right but I think we could take the "PasswordAuthentication=no" change for now.

Full explanation:

ssh-agent seems to be working for me.  There could be situations where it works and situations where it  doesn't, though (sudo/su/other become methods might change the environment so that the ssh-agent isn't available.. could be other things that are either limitations or bugs)...  Maybe we need to add the ForwardAgent option explicitly here?

I agree that I can't see a way that password auth for ssh would work easily... might be able to give the module a password parameter and then forward that on to the rsync command.  But I'm not sure that will work without using an expect-like library.  Maybe ssh-pass could be used here in the same way that ssh-pas is used for the main ansible connection.... That's all somewhat complex to add though and I'm pretty sure that password authentication won't work without that being added.  So the disabling of password prompting might be the best method until/unless someone submits a PR to do that.

gregdek:
Just a clarification, sorry @abadger -- what action are you proposing here, exactly?

tima:
If we're just talking about a ssh arg of `PasswordAuthentication=no` couldn't the ssh_args parameter of the synchronize module be used now if this is happening to an end user? Do we see this as a wide-spread problem to embed this setting rather leaving it to the end user? Just thinking outloud.

gregdek:
@tima -- can you help with the status of this versus #2244? 

Just unsure what we're asking of @alex-deref -- need a way to proceed, one way or the other.

bcoca:
closing in favor of feature request https://github.com/ansible/ansible/issues/13099, this will also centralize discussions we've been having across many tickets

PR #1916, opened 2015-08-15
Change behavior of apt.py around installing recommended packages.

### LINKED ISSUE #1189: apt: install_recommends should relies upon -o APT::Install-Recommends=yes
If, on a system: `APT::Install-Recommends "false";`
is set (eg _/etc/apt/apt.conf.d/_), then such an Ansible directive:
`apt: pkg=bla state=present install_recommends=yes`
will **not** install the packages recommended for _bla_.

The effective way to set recommends is to pass `-o APT::Install-Recommends=(yes|no)` to the apt-get run.
That will make _install_recommends_ more reliable.

hlieberman:
Closes #1189.

This will cause the settings in Ansible to override the system settings.
That will have no effect except on systems that have an out-of-Ansible
configuration that disables automatic installation of recommended
packages.  Previously, ansible would use the OS default whenever
install_recommends wasn't part of the playbook.  This change will cause
the Ansible default configuration setting of installing recommended
packages to override the configuration files set on the OS for things
installed through ansible, even when there is no install_recommends
specified in the playbook.  Because the OS default matches the Ansible
default, this shouldn't have wide impact.


bcoca:
i don't think we can accept this in it's current form, this goes against the previous default of the plugin and users would expect it to work with the default of the OS.

If you make this an option (force_recommends_install=yes|no) defaulting to no, then that would be something we could add.

hlieberman:
Hi @bcoca.

I think I might not have done a good enough job explaining what it does.  This isn't changing the default of the plugin, nor the default of the OS.  It changes no behavior, except in two circumstances.
1.  The bug circumstance:  A user has placed configuration files to override the default option to not installing recommendations at the OS level, and then installed something through Ansible with "install_recommends = yes".  Currently, this will result in the user not installing recommended packages.  After this patch, the behavior will change so that Ansible will install the recommended packages.
2.  A user has placed configuration files to override the default option to not installing recommendations at the OS level, and then installed something through Ansible /without/ specifying install_recommends in the playbook.  Currently, this will result in the package being installed without recommended packages, despite the ansible default of "install_recommends" being true.  After the patch, Ansible will install recommended packages under this condition.

Debian policy (which is the important thing here, as the canonical source for apt defaults) has strong things to say about disabling recommended package installation.

Per section 7.2, "[Recommends] declares a strong, but not absolute, dependency.  The Recommends field should list packages that would be found together with this one in all but unusual installations."  If the circumstances are so unusual, specifying install_recommends=no seems a logical choice.

I could extend the patch from a binary yes/no to a trinary: yes, no, and default, where default does whatever the OS wants.  Would that be more suitable?

bcoca:
Sorry about the misunderstanding, still I would argue that he OS default should be what we follow by norm and only override when a user specifically states it should be overridden (a force option?). 

Also I'm thinking the community needs to weigh in here also.

hlieberman:
Alright.  Let me rework the patch to try and handle the default case better.  Will ping you when a new one is pushed up.

hlieberman:
@bcoca, the branch should now be a no-op for existing cases, except for fixing the bug about it not working correctly in above circumstance 1.

bcoca:
cc @mgwilliams for review

mgwilliams:
@bcoca I'm not involved in Ansible development at this point, so if it works for you -- go for it!

packaging/os/apt.py
```diff
@@ -62,7 +62,7 @@
     default: null
   install_recommends:
     description:
-      - Corresponds to the C(--no-install-recommends) option for I(apt). Default behavior (C(yes)) replicates apt's default behavior; C(no) does not install recommended packages. Suggested packages are never installed.
+      - Corresponds to the C(--no-install-recommends) option for I(apt). C(yes) installs recommended packages.  C(no) does not install recommended packages. By default, Ansible will use the same defaults as the operating system. Suggested packages are never installed.
     required: false
     default: yes
```
bcoca, line 67: i would change the default to null

packaging/os/apt.py
```diff
@@ -546,7 +548,7 @@ def main():
             package = dict(default=None, aliases=['pkg', 'name'], type='list'),
             deb = dict(default=None),
             default_release = dict(default=None, aliases=['default-release']),
-            install_recommends = dict(default='yes', aliases=['install-recommends'], type='bool'),
+            install_recommends = dict(default='default', aliases=['install-recommends'], choices=['default', 'yes', 'no']),
```
bcoca, line 551: make the option type='bool' and the default to None, otherwise plays that already use true/false or Yes/No will start failing.

gregdek:
Hi @hlieberman -- looks like this also needs rebase.

Thanks for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

Once you have rebased this PR and made the appropriate revisions per @bcoca, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

abadger:
Okay, I made the requested changes in https://github.com/ansible/ansible-modules-core/pull/2399 bcoca reviewed and approved so I merged.  That should pull in the meaning of the last change here (just with implementation changes here).  Thanks @hlieberman for doing most of the work to get this fixed!

# Merged

Hi!

issue is resolved.

Thank you!

PR #1926, opened 2015-08-17
support creating role only

cjchang:
Currently, for role, keystone_user supports only add role with user. In the heat, to support "trusts" for deferred_auth_method, some roles will need to be created just role without users. This request is just to extend its function to support create role without users using the existing code with one more use case.

### DIFF
--- a/cloud/openstack/keystone_user.py
+++ b/cloud/openstack/keystone_user.py
@@ -252,8 +252,17 @@ def ensure_user_exists(keystone, user_name, password, email, tenant_name,
                                  email=email, tenant_id=tenant.id)
     return (True, user.id)
 
+def ensure_role_exists(keystone, role_name):
+    # Get the role if it exists
+    try:
+        role = get_role(keystone, role_name)
+    except KeyError:
+        # Role doesn't exist yet
+        role = keystone.roles.create(role_name)
+    return (True, role.id)
 
-def ensure_role_exists(keystone, user_name, tenant_name, role_name,
+
+def ensure_user_role_exists(keystone, user_name, tenant_name, role_name,
                        check_mode):
     """ Check if role exists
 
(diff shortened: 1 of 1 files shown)


bcoca:
cc @lorin

cloud/openstack/keystone_user.py
```diff
@@ -252,8 +252,17 @@ def ensure_user_exists(keystone, user_name, password, email, tenant_name,
...
     return (True, user.id)
 
+def ensure_role_exists(keystone, role_name):
+    # Get the role if it exists
+    try:
+        role = get_role(keystone, role_name)
+    except KeyError:
+        # Role doesn't exist yet
+        role = keystone.roles.create(role_name)
+    return (True, role.id)
```
lorin: I find it a little confusing that this method returns `(True, role.id)` if the role already exists. I would expect the boolean in the tuple to encode whether or not a change has happened.

cloud/openstack/keystone_user.py
```diff
@@ -378,14 +389,18 @@ def dispatch(keystone, user=None, password=None, tenant=None,
...
+          X      X     X     present     ensure_user_role_exists
+          X      X     X     absent      ensure_user_role_absent
+                       X     present     ensure_role_exists
+                       X     absent      ensure_role_absent
         """
     changed = False
     id = None
-    if tenant and not user and not role and state == "present":
+    if not tenant and not user and role and state == "present":
+        ensure_role_exists(keystone, role)
```
lorin: The value of `changed` doesn't get updated to true if the `ensure_role_exists` creates a role.

lorin:
It looks like the `changed` value doesn't get updated, see inline comments.

cjchang:
Sorry, I missed handling the changed. will update it.

cjchang:
Latest code has been pushed to handle "changed" correctly.

lorin:
:+1:

PR #1941, opened 2015-08-19
Add support for selinux user when adding a new user

badpass:
useradd modification, from man:
-Z, --selinux-user SEUSER
  The SELinux user for the user's login. The default is to leave this field blank, which causes the
  system to select the default SELinux user.

Add the param "seuser" to set the seuser-name when adding a new user for a system with support for selinux.

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -52,7 +52,7 @@ options:
     seuser:
         required: false
         description:
-            - Optionally sets the seuser type (user_u).
+            - Optionally sets the seuser type (user_u) on selinux enabled systems.
     group:
         required: false
         description:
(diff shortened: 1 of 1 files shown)


bcoca:
i would extend the description ' on selinux enabled systems'

badpass:
good point, I change the commit message

pwnall:
@badpass I think @bcoca was referring to the user-visible documentation here
https://github.com/ansible/ansible-modules-core/pull/1941/files#diff-364488963ab3238692762924e55c4991R55

Commenting because this looks useful for me :+1:

badpass:
@pwnall  and @bcoca I changed the user-visible documentation.

gregdek:
Thanks @badpass. @sfromm please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

badpass:
shipit

PR #1945, opened 2015-08-20
Add ability to create cache_parameter_group for elasticache

frenchfrywpepper:
The cache_parameter_group_name was not previously exposed in elasticachy.py, I have exposed it, as optional.

Also I have created an elasticache_parameter_group module that enables, creation / deletion / modification of the cache_parameter_group.

### DIFF
--- a/cloud/amazon/elasticache.py
+++ b/cloud/amazon/elasticache.py
@@ -43,6 +43,11 @@ options:
       - The version number of the cache engine
     required: false
     default: none
+  cache_parameter_group_name:
+    description:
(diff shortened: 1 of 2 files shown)


gregdek:
Looks like this is a new module combined with changes to an extant module, correct?

@jsdalton could use some review advice here -- should we split into separate PRs or review as one?

frenchfrywpepper:
I can split into two if that makes things easier (or more correct.)  I would put the module into extras in that case I assume?

jsdalton:
@frenchfrywpepper I think splitting it into two and submitting the new module to extras is definitely a sensible approach. Your changes to the existing elasticache module are straightforward so that can hopefully sail right through.

PR #1947, opened 2015-08-20
Fix the acl module

astorije:
While trying to re-BSD-ize the acl module, I have noticed that #150 shipped a few bugs :-(
- When testing the `setfacl` command to run, the wrong boolean was returned, making the command never run when it should... and run when it shouldn't!
- The check for the well-formedness of `entry` was wrong
- Removing an acl would have unexpected behavior when using `entry`

Other commits are here to support the 3 bug fixes above.

(Look at full commit messages for more info)

### DIFF
--- a/files/acl.py
+++ b/files/acl.py
@@ -172,7 +172,7 @@ def build_command(module, mode, path, follow, default, recursive, entry=''):
         cmd.append('--recursive')
 
     if not follow:
-        cmd.append('-h')
+        cmd.append('--physical')
 
     if default:
         if(mode == 'rm'):
--- a/files/acl.py
+++ b/files/acl.py
@@ -127,12 +127,10 @@ def split_entry(entry):
     ''' splits entry and ensures normalized return'''
 
     a = entry.split(':')
-    a.reverse()
-    try:
-        p, e, t = a
-    except ValueError, e:
-        print "wtf?? %s => %s" % (entry, a)
-        raise e
+    if len(a) == 2:
+        a.append(None)
+
+    t, e, p = a
 
     if t.startswith("u"):
         t = "user"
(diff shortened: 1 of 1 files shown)


astorije:
I notice that supporting both `entity`/`etype`/`permissions` and `entry` make it painful to convert back and forth, make sure both versions are in sync, make sure they don't collide, ...
Therefore, I have 2 questions:
- `entry` was deprecated in Ansible 1.5 while introducing the other form. Without trying to argue, may I know the rationale around this? To me having access to `entry` was ideal and I didn't understand the switch. At first, I thought it was to make devs' lives easier, but when working on #150, I noticed it was actually a big pain to maintain both :-)
- Since #150 (and the current fix) is scheduled for Ansible 2.0, major version, would you allow me to get rid of `entry` altogether here? As much as I'd like to keep `entry` around instead of the other form, I respect that you guys decided to go another route and deprecate this one, so I think it'd be for the best to keep only one version, as I still have doubts both versions work flawlessly.

astorije:
While I was at it, I fixed the `follow` option for Linux. I will of course re-add `-h` support when adding BSD support back :-)

astorije:
@bcoca, @abadger, can this issue get labeled maybe? FYI, as long as this isn't merged, the `acl` module will be broken on `devel`.
I'll try of course to react as lightning fast as possible once you get to review this to make sure the module gets fixed ASAP. Thanks!

gregdek:
Sorry for the delay @astorije. Since you are marked as an author of this module, I'm going to flag for inclusion, with the expectation that the core team will double check.

astorije:
Thanks a lot @gregdek! No worries regarding delay ;-)

> Since you are marked as an author of this module, I'm going to flag for inclusion

Not sure what you mean though.

gregdek:
It means that it's ready to go and should be pulled in by the core team soon, pending final review.

cc: @ansible/core

jimi-c:
Merged, thanks!

astorije:
Great! I'll continue working on it to bring back BSD support soon.

PR #1952, opened 2015-08-21
Don't fail in authorized_key check_mode if user exists

manics:
PR #1651 fixed issue #1515 but the requirement for path to be defined is unnecessarily strict, and breaks existing usage of check_mode where a user and key already exists.

If the user has previously been created a path isn't necessary.


mgedmin:
The diff looks good to me.

I also tested the patch: it fixes the problem for me.  (I've a playbook that works fine with Ansible 1.9.2 in --check mode, but fails with the quoted error with Ansible 2, unless this patch is applied.)

PR #1955, opened 2015-08-22
service: Add `use_service`.

dochang:
The motivation is that systemctl cannot reload udev on Debian sid. When reloading services, the `service` script checks systemd property `CanReload` first and fallbacks to sysvinit if `CanReload=no`. Prefering `service` would be better in that case.

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -82,6 +82,16 @@ options:
         - Avoid a module failure if the named service does not exist. Useful
           for opportunistically starting/stopping/restarting a list of
           potential services.
+    use_service:
+        required: false
+        default: 'fallback'
+        choices: [ 'force', 'prefer', 'fallback' ]
+        version_added: "2.0"
+        description:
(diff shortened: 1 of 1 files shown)


gregdek:
@dochang thanks for the PR. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

bcoca:
This will be overridden by the service  split and rewrite, also the service binary has different implementation depending on system and is also present on 'non linux', FreeBSD >= v8 for example, 

In this case there are several systemd bugs on Debian and instead of service it would probably be more appropriate to use the init script directly (which systemd should do but fails due to the mentioned bugs).  

note: The bugs seem to have been fixed but are not released for the current version (jessie).

gregdek:
Closing because this will soon be obsoleted by upcoming changes, per @bcoca.

Thanks @dochang for the PR!

PR #1960, opened 2015-08-23
Fixes #1959

### LINKED ISSUE #1959: google-cloud modules fail to find the "correct" python
GCE modules seem to be fixed to /usr/bin/python when it shouldn't as sometimes you may want to run more than 1 python (or use brew to install python on mac os x). This creates issues finding the python modules necessary for GCE to work. 

Example: 
https://github.com/ansible/ansible-modules-core/blob/devel/cloud/google/gce.py#L1

Should be: 

```
/usr/bin/env python
```

Regards,
Alex

alex-leonhardt:
Change to `/usr/bin/env python` instead of `/usr/bin/python` to get the correct (active) python.


bcoca:
env does not always produce the correct python, this is done on purpose, use ansible_python_interpreter to point at the correct python.

PR #1978, opened 2015-08-27
fix for start/stop ec2 instances

majidaldo:
- state variable undefined
- instance_ids could be undefined
- starting/stopping an instance should have no business in changing instance attributes
- checking for instances in transition (what to do about them?)

for my purposes, this does the job. but let's see what the community thinks.
#1703

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -1182,6 +1182,8 @@ def startstop_instances(module, ec2, instance_ids, state, instance_tags):
 
     wait = module.params.get('wait')
     wait_timeout = int(module.params.get('wait_timeout'))
+    state = module.params.get('state')
+
     changed = False
     instance_dict_array = []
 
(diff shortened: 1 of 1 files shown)


bennojoy:
@majidaldo  The changes look good thanks, i am not exactly sure if we should remove the instance attributes code (source dest check) , do you find a better place where this could be added ?

majidaldo:
@bennojoy i'm really not an ec2 expert nor an expert in this module. i just believe this start/stop method should keep things as is. i didn't find it straightforward to write my code as it had enough issues to deal with.

i assume you could change the attributes somewhere else in the code. plus you would have to know what is allowed by aws.

bennojoy:
@majidaldo  i think the startstop function is not the best part but it would be most viable part to add this fucntionality as of now due to the complexity of the module, Also from the interface point of view it would be consistent as the user states that the instance state should be 'running' state with the attributes specified in the task.
If you could add back the attributes check part, i can test and merge this in.

gregdek:
Closed due to inactivity.

PR #1989, opened 2015-08-31
authorized_key: fix example in documentation

mgedmin:
'key=' cannot be pointing to a file name; it needs to be the key itself as a string (or a URL).


gregdek:
Thanks @mgedmin.

@bradobro Looks like a simple PR, just want to run it by you as the module owner.

If you believe it’s a good PR, please add a comment with "shipit" in the text, and we will flag it for inclusion.

If not, please add a comment with "needs_revision" in the text, along with the needed changes, and we will flag it for revision.

PR #1994, opened 2015-09-01
[DOCUMENTATION] Remove the sample SHA256

hutchic:
putting a full sha256 made the width of the table in the generated docs not fit on some screens

screen shot 2015-09-01 at 10 07 06 am


abadger:
Maybe we could change that to the md5 hash to make it short enough?  or if even that is too long, I guess change "shagoeshere" to "D98291AC[...]B6DC7B97"

"shagoehere" doesn't have much more information than  `<algorithm>:<checksum>` which we already have just before that.

abadger:
Thanks for taking the time to make things look better!

# Merged

Hi!

issue is resolved.

Thank you!

PR #2007, opened 2015-09-03
add check_mode fact to setup module

### LINKED ISSUE #10696: Knowing whether a playbook is run in check mode
##### Issue Type:

Feature Idea
##### Ansible Version:

To this day, this feature does not exist. I am running `ansible 1.9.0.1`.
##### Environment:

N/A
##### Summary:

It is natively impossible to know from a task, a role, or a playbook, if we are running in [check mode](http://docs.ansible.com/playbooks_checkmode.html) or not.
##### Steps To Reproduce:

It can be very useful though. For example, some module (`ufw`, `git`, ...) require that the underlying executable must be installed. Lots of roles do that as a first task. If the task is run in check mode for the first time, the role will fail. Same applies if you try to do changes to a user that was not created, to create a symlink on a file that was not copied over yet, ...

In some cases, just knowing that we are in check mode is enough (although I agree that some cases require more work, like checking the presence of an executable which `which`, the presence of a user with `id`, ... but in the end, knowing that we are in check mode is still very valuable even in those cases).
##### Expected Results:

A variable? An extra fact? A extra directive at the same level than `always_run`, `when`, ... ? Not sure what is the best way to present this but I'd be happy to have others' opinions.
##### Actual Results:

Since it's not part of Ansible yet, it's still doable but through very [hacky solutions](http://stackoverflow.com/questions/28729567/skip-ansible-task-when-running-in-check-mode). This same page also shows some more use cases for this.

ibotty:
closes ansible/ansible#10696

### DIFF
--- a/system/setup.py
+++ b/system/setup.py
@@ -85,6 +85,8 @@ def run_setup(module):
     for (k, v) in facts.items():
         setup_options["ansible_%s" % k.replace('-', '_')] = v
 
+    setup_options["check_mode"] = module.check_mode
+
     # Look for the path to the facter and ohai binary and set
     # the variable to that path.
     facter_path = module.get_bin_path('facter')


bcoca:
i don't think we should be setting this in the facts module, it does not always get executed, this should be an internal var, it is also not 'per host'.

jimi-c:
I agree with @bcoca, this is something that should be set in VariableManager. Going to close this PR, but please feel free to resubmit with the fix where we recommend.

Thanks!

ibotty:
I am new to ansible codebase, so I might be totally wrong here, but...

In `VariableManager,get_vars`, is there any way to get to the `PlayContext`? AFAICT, it's not possible with that signature. Does it make sense to change the signature to use the `PlayContext` instead of the `Playbook`? If not, how can I get the information, that we are in check mode from within the `VariableManager`?

ghost:
@ibotty check out this pull request, it should resolve this issue https://github.com/ansible/ansible/pull/14028

PR #1208, opened 2015-04-27
docker: use HostConfig object when creating container with api_version > 1.15

ghost:
Fixes #1158 

Trying to use HostConfig if possible. This is recommended way as described in https://github.com/docker/docker-py/issues/546#issuecomment-96032262

PR #2011, opened 2015-09-03
We had two separate methods trying to do the same thing but neither o…

abadger:
…ne was complete.  This merges them so that all of the options get parsed and applied.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -801,7 +801,8 @@ class DockerManager(object):
 
         optionals = {}
         for optional_param in ('dns', 'volumes_from', 'restart_policy',
-                'restart_policy_retry', 'pid'):
+                'restart_policy_retry', 'pid', 'extra_hosts', 'log_driver',
+                'cap_add', 'cap_drop'):
             optionals[optional_param] = self.module.params.get(optional_param)
 
         if optionals['dns'] is not None:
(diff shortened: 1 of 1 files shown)


abadger:
Another PR pointed out that we have both get_host_config() vs create_host_config() which seem to do (almost) the same thing.

After some research into the pull requests that lead to both of those:
-    create_host_config: #1405
-    get_host_config: #1208 and #1744

I think we had two PRs that implemented very similar changes to fix the same issue. Because the names were different we ended up merging both of them when we should have realized they were attempting to do the same thing and only merged one.

This PR attempts to do that.

abadger:
@ryanwalls @shin- @larsks You all were involved in the original PRs adding HostConfig.  Any of you want to make sure that I'm not doing something deeply wrong here?  There's a PR waiting on this one so I'd like to merge it quickly but we can always apply a bugfix if you notice something is wrong here.

ryanwalls:
LGTM, but I don't have any tests that I can run.  @smashwilson had some tests.  Maybe he can run them?

smashwilson:
Sadly the box that I have my tests configured on hasn't had a NIC for about a month, then I'm on travel next week. (One of the reasons I've been quiet in these issues recently.)

My tests are [in this repository](https://github.com/smashwilson/docker-module-test) if someone else has a chance to run them before I do, although it may take some tweaking (I think I still have hardcoded paths and such.)

abadger:
Okay.  I'll merge, see what I can test myself, and rely on people reporting breakage for finding any cornercases.  Thanks guys!

waisbrot:
Before this merge, I was able to do `log_driver: syslog`. After pulling the new 'devel', I now get 

```
"msg": "Docker API Error: Cannot start container 4d33889daab615e89a4aaf74b66449563055fbe4e1d65229e6ae97f8036d0f38: Failed to initialize logging driver: Unix syslog delivery error"
```

The -vvv flag doesn't give me any additional information. How could I help debug?

abadger:
@waisbrot, open a new ticket.  (If you are on irc, ping me abadger1999 when you do) Tell me what versions of the docker server and docker-py python library you have installed in the ticket.

Issue #2013, opened 2015-09-04
unexpected state in yum module (1.9.3)


steynovich:
Updating YUM packages is broken in Ansible 1.9.3.

env:
Ubuntu 15.04 vivid
Ansible version: 1.9.3-1ppa~vivid (deb-src http://ppa.launchpad.net/ansible/ansible/ubuntu)

Snippet of my playbook:

``` yaml
- name: YUM upgrade
  yum: name=* state=latest update_cache=yes
```

Unexpected result:

```
TASK: [centos | YUM upgrade] **************************************************
failed: [centos_host] => {"changed": false, "errors": "unexpected state", "failed": true, "results": ""}
msg: we should never get here unless this all failed
```

Debugging showed that the value state is 'latest', which is an expected state.

Expected result (was working in 1.9.2):

```
TASK: [centos | YUM upgrade] **************************************************
ok: [centos_host]
```

gunzy83:
Confirming the above, updating all packages using the yum module is broken in 1.9.3.

My environment:
Amazon Linux AMI 2015.03.1 x86_64
Ansible version 1.9.3 installed via pip

Broken task:

```
- name: Ensure all base packages are updated
  yum:
    name: "*"
    state: "latest"
```

Result:

```
TASK: [base | Ensure all base packages are updated] ***********************
failed: [127.0.0.1] => {"changed": false, "errors": "unexpected state", "failed": true, "results": ""}
msg: we should never get here unless this all failed
```

Downgrading to 1.9.2 fixes the issue.

abadger:
I just pushed https://github.com/ansible/ansible-modules-core/commit/411adff00687842dd1c379fd78b6b2562c6f97e4 which I think will fix this.  Testing now.  If you'd like to test as well, that would be appreciated.

steynovich:
Your commit 411adff seems to fix the issue for me.

```
TASK: [centos | YUM upgrade] **************************************************
ok: [centos_host]
```

Thanks for your swift reply.

abadger:
Thanks for the quick test.  Works here too.  We're working out when to make a 1.9.4 release with this fix in it now.

abadger:
Okay, we've tagged 1.9.4-rc1.  If no more major bugs are found a 1.9.4 final should be out soon with this fix in it.

josebv:
This only affects when yum update is used inside roles, not at the top level yml

lero:
[...]

PR #2015, opened 2015-09-04
Update yum.py

steynovich:
Bug fix for issue 2013 https://github.com/ansible/ansible-modules-core/issues/2013


abadger:
Thanks... looking now to see if this is better done by cherry-picking a commit from devel instead of applying a wholy new patch.

abadger:
I like your approach -- it's what we're doing in the devel branch.  However, after looking at the stable-1.9 code again, I decided to go with this: https://github.com/ansible/ansible-modules-core/commit/411adff00687842dd1c379fd78b6b2562c6f97e4  It doesn't feel as good to me from a "clean code" standpoint but it fits in with the existing code better.  So if someone besides me has to look at what the stable-1.9 code is doing, they won't have to figure out why install() and remove() never seem to return a result even though we expect it in ensure() or why latest() would return a result half the time and module.exit_json() the other times.

Thanks for working up the fix, though.  It let me very quickly see what the root cause of the error was!

Issue #1937, opened 2015-08-19
Linode create server: Change underscore to dash in server name prefix


arturmartins:
Bumped into this because I was trying to setup the hostname with the same name created on linode, which resulted on the following error:

`msg: Command failed rc=1, out=, err=hostname: the specified hostname is invalid`

Investigating the problem, it lead into here: http://ubuntuforums.org/showthread.php?t=1640679

> Avoid using the underscore (_) character in machine names. Internet standards dictate that domain names conform to the host name requirements described in **Internet Official Protocol Standards RFC 952 and RFC 1123**. Domain names must contain only letters (upper or lower case) and digits. Domain names can also contain dash characters ( - ) as long as the dashes are not on the ends of the name. Underscore characters ( _ ) are not supported in the host name

Related code: https://github.com/ansible/ansible-modules-core/blob/devel/cloud/linode/linode.py#L259

ansibot:
@zbal ping, this issue is waiting for your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@zbal, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@zbal, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@zbal, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
This repository has been locked. All new issues and pull requests should be filed in https://github.com/ansible/ansible

Please read through the <a href="http://docs.ansible.com/ansible/dev_guide/repomerge.html#move-issues-and-prs-to-new-repo">repomerge</a> page in the dev guide. The guide contains links to tools which automatically move your issue or pull request to the ansible/ansible repo.

ansibot:
This issue was migrated to https://github.com/ansible/ansible/issues/30059

PR #2020, opened 2015-09-04
Adding in Linode private IP support, display group support, missing indempotence fixed

jaddison:
... and optional prepending of "&lt;linode_id&gt;-".

These issues have been brought up before in #7223, #6025, #5991, #5992, #7086. Issues and pull requests have been closed due to moving repositories - it seems like quite a bit has been lost in this move...

### DIFF
--- a/cloud/linode/linode.py
+++ b/cloud/linode/linode.py
@@ -39,7 +39,7 @@ options:
     type: string
   prefix_name:
     description:
-     - Bool, prefix the instance name with 'LinodeID_'
+     - Bool, prefix the instance name with the Linode's ID - ie. '<linode_id>-'
     default: "yes"
     choices: [ "yes", "no" ]
   display_group:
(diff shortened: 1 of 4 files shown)


abadger:
@zbal is this something you have time to review?

arturmartins:
This is great!

I like the LinodeID prefix idea, but I bumped into a problem to set the server's /etc/hostname with the exactly same linode name (#1937).

According to **RFC 952** and **RFC 1123**: _Domain names must contain **only letters** (upper or lower case) and **digits**._ 

Therefore I recommend to change `LinodeID_` to `LinodeID-`.

arturmartins:
BTW, the idempotence is only fixed to the scenario that no prefix is used or the complete linode name is given.

jaddison:
Thanks for the comments. The prefix isn't new, just the option to disable
it is. I can change the prefix to use a dash. It shouldn't cause any issues
with existing users, seeing how the indempotence want there before anyhow.

Regarding indempotence itself, I actually noticed that last night. A fix is
forthcoming, along with Linode NodeBalancer support via three new
modules.... In a day or two.

jaddison:
With that last commit, I've address @arturmartins comments and also added in Linode NodeBalancer  (their load balancing 'appliance' functionality) support with 3 modules based largely on @DistilledLtd's efforts.

buunguyen:
:+1: I look forward to see the private IP and display group support gets merged, need them badly.

@jaddison maybe it's a good idea to reissue the Linode NodeBalancer from another branch to a separate PR. That probably makes the job of the reviewers easier and ensures that the private IP fix can get merged even if the balancer needs further revision.

tarqd:
Display group support is essential for getting the most out of this and the dynamic Linode inventory (which can group by display group). Really needs to be merged

gregdek:
So this is essentially three new modules for linode. Is there any reason these shouldn't be in extras rather than core? cc: @ansible/core

jaddison:
The main reason to put them in core was that the original Linode module is there, and these new modules rely on (and this PR changes) it.

Edit: If there was a desire to move all Linode modules to extras, then that would be ok too, but I think it makes sense to group them together.

gregdek:
Thanks @jaddison for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

gregdek:
Sorry for missing this one for so long @jaddison. This one has been sitting for a while because it's a weird corner case that our bots mishandled -- a combination of old modules and new modules in core.

I'm going to close this PR. My recommendation at this point is to split it into multiple PRs:
- A PR in this repository to address the issue in cloud/linode/linode.py; 
- Separate PRs in the Extras repo, one for each of the new modules proposed (since we now strongly discourage new modules in Core).

Thanks, and sorry again for the painful delay.

PR #2021, opened 2015-09-04
ec2: allow ENIs to be specified at instance launch

bretmartin:
This adds a `network_interfaces` parameter to the `ec2` module to allow existing Elastic Network Interfaces (ENIs) to be attached to an EC2 instance at launch time.

ENIs are an important mechanism for high availability solutions and for things like MAC-address-based software licensing. Without this new parameter, it is not possible to use Ansible to manage instances where the primary network interface needs to be an existing ENI (because after launch, the primary network interface can't be detached).

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -246,7 +246,6 @@ options:
       - A list of existing network interfaces to attach to the instance at launch. When specifying existing network interfaces, none of the assign_public_ip, private_ip, vpc_subnet_id, group, or group_id parameters may be used. (Those parameters are for creating a new network interface at launch.)
     required: false
     default: null
-    aliases: []
 
 author:
     - "Tim Gerla (@tgerla)"
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -852,15 +852,6 @@ def create_instances(module, ec2, vpc, override_count=None):
     if group_id and group_name:
         module.fail_json(msg = str("Use only one type of parameter (group_name) or (group_id)"))
 
-    if (network_interfaces and
-            (assign_public_ip or private_ip or vpc_subnet_id
-                or group_name or group_id)):
-        module.fail_json(
-            msg=str("network_interfaces must not be set when specifying " +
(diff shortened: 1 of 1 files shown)


bretmartin:
@evanccnyc @rickmendes Hoping you may be able to comment on this since you've contributed/commented on ENI-related functionality in the past

evanccnyc:
Looks good to me but havent had time to test.

cloud/amazon/ec2.py
```diff
@@ -240,6 +240,13 @@
     required: false
     default: null
     aliases: []
+  network_interfaces:
+    version_added: "2.0"
+    description:
+      - A list of existing network interfaces to attach to the instance at launch. When specifying existing network interfaces, none of the assign_public_ip, private_ip, vpc_subnet_id, group, or group_id parameters may be used. (Those parameters are for creating a new network interface at launch.)
+    required: false
+    default: null
+    aliases: []
```
ghost: Aliases don't need to be specified when there are none. You don't have to continue the bad habit of the original code.

ghost:
I think you should add an example showing the use of an ENI. Other than that it looks good. I also have not had time to test.

bretmartin:
Good point, will add an example.

bretmartin:
So arguments like `group` will let you specify a string if you're only specifying one group.

Should I make this work similarly and add an alias `network_interface`?

bretmartin:
I just noticed there is some general-purpose code in `main()` for mutually-exclusive arguments. Do we think it's OK that I'm handling these in `create_instances()` separately, so the specific error message can be provided?

ghost:
@bretmartin 

network_interface: I like the idea of it working similarly to group. Keeps it consistent with the existing module.

mutually- exclusive: I tend to lean towards using mutually exclusive because it is cleaner. You should end up with a context based error message any way. It may not be the exact language you want but it will illustrate the conflict.

ghost:
@bretmartin Thanks for making the changes. When I am back home from the weekend I will test this out and let you know how it goes.

ghost:
@bretmartin @gregdek @bcoca 

Tested this and it works as expected. +1

PR #2027, opened 2015-09-08
New version of ec2_elb module supporting multi instances

vengaar:
- Change way to wait. Instead of looping on each ELB to register or deregister it, we try to register/deregister them all at once and then wait for all to be in the wanted state => Optimise time to wait
- Increase waiting time between 2 retries with minimal healthcheck interval (5s), to avoid reaching "Exceeded rate limit" message in EC2 API. Could be optimized by taking the healthcheck interval value of the LB.
- Add new parameter "instances" aliasing "instance" for compatibility with previous version.
- Add new returned values and add documentation on it
- Behaviour for _enable_availailability_zone method in multi instances mode has not been tested

### DIFF
--- a/cloud/amazon/ec2_elb.py
+++ b/cloud/amazon/ec2_elb.py
@@ -19,22 +19,22 @@ DOCUMENTATION = """
 module: ec2_elb
 short_description: De-registers or registers instances from EC2 ELBs
 description:
-  - This module de-registers or registers an AWS EC2 instance from the ELBs
+  - This module de-registers or registers an AWS EC2 instance(s) from the ELBs
     that it belongs to.
-  - Returns fact "ec2_elbs" which is a list of elbs attached to the instance
+  - Returns fact "ec2_elbs" which is a list of elbs attached to the instance(s)
     if state=absent is passed as an argument.
   - Will be marked changed when called only if there are ELBs found to operate on.
 version_added: "1.2"
-author: "John Jarvis (@jarv)"
+author: John Jarvis
 options:
   state:
     description:
-      - register or deregister the instance
+      - register or deregister the instance(s)
     required: true
     choices: ['present', 'absent']
-  instance_id:
+  instance_ids:
     description:
-      - EC2 Instance ID
+      - list of EC2 Instance ID (alias of instance_id)
     required: true
   ec2_elbs:
     description:
@@ -48,14 +48,14 @@ options:
     aliases: ['aws_region', 'ec2_region']
   enable_availability_zone:
     description:
-      - Whether to enable the availability zone of the instance on the target ELB if the availability zone has not already
+      - Whether to enable the availability zone of the instance(s) on the target ELB if the availability zone has not already
         been enabled. If set to no, the task will fail if the availability zone is not enabled on the ELB.
     required: false
     default: yes
     choices: [ "yes", "no" ]
   wait:
     description:
-      - Wait for instance registration or deregistration to complete successfully before returning.  
+      - Wait for instance(s) registration or deregistration to complete successfully before returning.  
     required: false
     default: yes
     choices: [ "yes", "no" ] 
@@ -69,7 +69,7 @@ options:
     version_added: "1.5"
   wait_timeout:
     description:
-      - Number of seconds to wait for an instance to change state. If 0 then this module may return an error if a transient error occurs. If non-zero then any transient errors are ignored until the timeout is reached. Ignored when wait=no.
+      - Number of seconds to wait for an instance(s) to change state. If 0 then this module may return an error if a transient error occurs. If non-zero then any transient errors are ignored until the timeout is reached. Ignored when wait=no.
     required: false
     default: 0
     version_added: "1.6"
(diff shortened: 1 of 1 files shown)


bcoca:
cc @jarv @defionscode @bennojoy

cloud/amazon/ec2_elb.py
```diff
@@ -19,22 +19,22 @@
...
-  - This module de-registers or registers an AWS EC2 instance from the ELBs
+  - This module de-registers or registers an AWS EC2 instance(s) from the ELBs
     that it belongs to.
-  - Returns fact "ec2_elbs" which is a list of elbs attached to the instance
+  - Returns fact "ec2_elbs" which is a list of elbs attached to the instance(s)
     if state=absent is passed as an argument.
   - Will be marked changed when called only if there are ELBs found to operate on.
 version_added: "1.2"
-author: "John Jarvis (@jarv)"
+author: John Jarvis
```
bcoca: do not remove the github handle

cloud/amazon/ec2_elb.py
```diff
@@ -19,22 +19,22 @@
...
 options:
   state:
     description:
-      - register or deregister the instance
+      - register or deregister the instance(s)
     required: true
     choices: ['present', 'absent']
-  instance_id:
+  instance_ids:
```
bcoca, line 35: this is a backwards incompatible change, use aliases to keep the old instance_id usable and not break existing playbooks

I see that you did add the alias below, this should be documented here.

cloud/amazon/ec2_elb.py
```diff
@@ -98,69 +98,87 @@
...
     import boto
     import boto.ec2
-    import boto.ec2.autoscale
     import boto.ec2.elb
     from boto.regioninfo import RegionInfo
-    HAS_BOTO = True
 except ImportError:
-    HAS_BOTO = False
-
+    print "failed=True msg='boto required for this module'"
```
bcoca: don't print or exit from a module, ALWAYS use fail_json as this module was doing before your changes

cloud/amazon/ec2_elb.py
```diff
@@ -98,69 +98,87 @@
...
+    sys.exit(1)
 
 class ElbManager:
     """Handles EC2 instance ELB registration and de-registration"""
 
-    def __init__(self, module, instance_id=None, ec2_elbs=None,
+    def __init__(self, module, instance_ids=[], ec2_elbs=None,
```
bcoca: it is not a good idea to initialize list in the method definition, keep it None and check for it inside the method

cloud/amazon/ec2_elb.py
```diff
@@ -176,64 +194,53 @@ def _enable_availailability_zone(self, lb):
...
         lb: load balancer"""
-        instance = self._get_instance()
-        if instance.placement in lb.availability_zones:
-            return False
-
-        lb.enable_zones(zones=instance.placement)
+        instances = self._get_instance()
+        for instance in instances:
+            if instance.placement in lb.availability_zones:
+                return False
```
bcoca: does it still make sense to return here now that it is a loop?

cloud/amazon/ec2_elb.py
```diff
@@ -176,64 +194,53 @@ def _enable_availailability_zone(self, lb):
...
+                    lb.name
+                )
+                self.module.fail_json(
+                    msg=msg,
+                    wait_attempt=self.wait_attempt
+                )
+            
+            self.wait_attempt[lb.name] = self.wait_attempt[lb.name] + 1
+            # min healthcheck interval
+            time.sleep(5)
```
bcoca: this should not be hardcoded, we might want to make it a parameter that defaults to 5

cloud/amazon/ec2_elb.py
vengaar, line 35: Done, instance_id added as alias of instance_ids in the documentation.
I have also done the other changes requested, I hope it is ok for you.
Don't hesitate to do other feedbacks, as it is the first time I do a pull request, I am not very confortable with the process

juliedavila:
`Behaviour for _enable_availailability_zone method in multi instances mode has not been tested`, why has it not been tested?

broferek:
Our current Amazon infrastructure relies only on 1 availibility zone which is already associated to our ELBs.
So we always run our playbooks with the "enable_availability_zone" set to No by default.

We have modified the method "_enable_availailability_zone" to loop on the list of instances. So it should do what it is supposed to.
If someone has instances in several availibilty zones, it would be good to run a test.

gregdek:
Thanks @vengaar for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

robynbergeron:
@vengaar A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

gregdek:
@vengaar Another friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open. If we don't hear from you within another 14 days, we will close this pull request.

gregdek:
Closed due to inactivity.

PR #2028, opened 2015-09-08
Added option to set multiple ACLs for S3 objects

rosmo:
It also verifies the selected permissions against Boto's canned ACL strings list. Combined with a patch to Boto, this allows you to set the "aws-read-exec" permission to a bucket and its files.

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -95,7 +95,7 @@ options:
     default: null
   permission:
     description:
-      - This option let's the user set the canned permissions on the object/bucket that are created. The permissions that can be set are 'private', 'public-read', 'public-read-write', 'authenticated-read'.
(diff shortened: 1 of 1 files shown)


bcoca:
cc @lwade @ralph-tice

cloud/amazon/s3.py
```diff
@@ -409,6 +412,10 @@ def main():
     s3_url = module.params.get('s3_url')
     src = module.params.get('src')
 
+    for acl in module.params.get('permission').split(','):
```
bcoca: make permission a list and you won't have do do this, also you don't check for , first so this can cause an exception.

rosmo: You mean make permission accept a list as well and convert string to list? I was thinking backwards compatibility.

bcoca: if you define it as a list in the spec, that is done for you

rosmo: Indeed it was, I've amended the commit.

lwade:
Looks good, thank you for this!

PR #2034, opened 2015-09-09
Fix indentation in os_server documentation

pwnall:
The indentation bug has the unfortunate effect of making it seem like `root_volume` deletes an instance's volume.


pwnall:
@bcoca Thank you for the quick turnaround! :+1:

PR #2062, opened 2015-09-14
Update iam_policy.py doc and state default parameter

wimnat:
Fixed doc for policy_name - it is a required field
Removed empty aliases
Added default for state
Removed unnecessary aws_secret parameters and boto requirement as aws_doc fragment is included

@defionscode

### DIFF
--- a/cloud/amazon/iam_policy.py
+++ b/cloud/amazon/iam_policy.py
@@ -27,54 +27,34 @@ options:
     required: true
     default: null
     choices: [ "user", "group", "role"]
-    aliases: []
   iam_name:
     description:
       - Name of IAM resource you wish to target for policy actions. In other words, the user name, group name or role name.
     required: true
-    aliases: []
   policy_name:
     description:
       - The name label for the policy to create or remove.
-    required: false
-    aliases: []
+    required: true
   policy_document:
     description:
       - The path to the properly json formatted policy file (mutually exclusive with C(policy_json))
     required: false
-    aliases: []
   policy_json:
     description:
       - A properly json formatted policy as string (mutually exclusive with C(policy_document), see https://github.com/ansible/ansible/issues/7005#issuecomment-42894813 on how to use it properly)
     required: false
-    aliases: []
   state:
     description:
       - Whether to create or delete the IAM policy.
     required: true
     default: null
     choices: [ "present", "absent"]
-    aliases: []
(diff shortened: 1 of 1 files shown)


evanccnyc:
Why is state not required?

cloud/amazon/iam_policy.py
```diff
@@ -27,54 +27,34 @@
...
     description:
       - A properly json formatted policy as string (mutually exclusive with C(policy_document), see https://github.com/ansible/ansible/issues/7005#issuecomment-42894813 on how to use it properly)
     required: false
-    aliases: []
   state:
     description:
       - Whether to create or delete the IAM policy.
-    required: true
-    default: null
+    required: false
```
jimi-c: I don't believe this is correct, as we have in the past stated that explicit is better than implicit when it comes to state parameters. I'm ok with correcting the above doc mistake, but I don't believe we should make this state implicit.

juliedavila:
+1, thanks!

@evanccnyc you cannot have a state=required and a default at the same time. @wimnat made a proper correction in that for most if not all other CRUD type modules the default state is always 'present' .

cloud/amazon/iam_policy.py
juliedavila: @jimi-c, re: my last comment: it goes against the UX present in probably most modules. Though if we're pushing for a new standard, then I agree.

wimnat: As @defionscode says, i've updated to match most, if not all, other aws modules.  I agree that including state in a playbook is recommended practice but a departure from setting the default of state now goes against most other committed code.

juliedavila: @wimnat, 

@jimi-c and I discussed this offline. Let's go ahead and remove the default and re-add required: true. The long term plan is to over time deprecate all `implicitly present` logic across all modules. Given that, it makes sense to stop the practice.

wimnat:
Reverted the change to state.  Should probably update developer guidelines regarding requiring state from here on in.

juliedavila:
+1 Thanks.

gregdek:
Thanks @wimnat for the PR, and thanks @defionscode for the review. Marking for inclusion.

gregdek:
(cc @ansible/core team)

jimi-c:
Merged, thanks!

PR #2064, opened 2015-09-14
Remove unnecessary json.loads

wimnat:


juliedavila:
+1, simple enough patch

PR #2069, opened 2015-09-14
Add doc for gce's return object

phanect:
Maybe we need "Return Values" sections on doc, but for now I added this in "Notes" section


phanect:
The error on CI seems derived from current `devel` head [813053c](https://travis-ci.org/ansible/ansible-modules-core/builds/79919189).
I will rebase after the issue is fixed.

phanect:
Error seems to be fixed and I rebased my fork.

bcoca:
don't document this in the notes, use the RETURN_DOCS facility (check copy, stat or any of the cloud stack modules for examples)

phanect:
@bcoca Oh, already there is parameter to state return values.

I rewrote partially. Above commit is in progress. I will rebase after I finished it.

phanect:
@bcoca Sorry for late reply.
I found I may add wrong info because of lacking understanding for gce module.
Let me close this PR.

PR #2070, opened 2015-09-14
We are copying from S3,  the bucket is the source, not the target.

fvant:
We are copying from S3,  the bucket is the source, not the target.


gregdek:
@lwade @ralph-tice Please review this PR to make sure it adheres to the following guidelines:

gregdek:
Actually, this is a trivial fix and I will +1 it myself. @fvant thanks for the PR.

PR #2076, opened 2015-09-14
Adding list state to os_network module

dagnello:
There can be instances during an Ansible play where the list of networks
currently avaialble from OpenStack is required.  This update provides
network list functionalty to the current os_network module.


bcoca:
this makes more sense as part of a _facts module

dagnello:
was thinking about that as well, got feedback to have it as part of existing modules.  We have os_server_facts which sets facts associated with one server, along these lines we can have a os_servers_facts, os_networks_facts, etc.. that will set facts for all respective resources.  I am ok with either approach.

Shrews:
I'm going to say go with whatever @bcoca suggests here as he is definitely the domain expert and actually knows what he is talking about.  :)

bcoca:
@dagnello wan't to then replace this PR? I think a os_networks_facts might be a good new module

@Shrews I just play a developer on TV

dagnello:
@bcoca will be pushing up a revised patch shortly.  thanks

PR #2075, opened 2015-09-14
fixing termination policies for ec2_asg

evanccnyc:
This was missing which would cause an error if you specified the termination policies.

PR #2080, opened 2015-09-15
Fix crash in ec2_asg

evax:


evax:
This is a dupe of #2075, closing this

PR #2082, opened 2015-09-15
Update apache2_module.py

n0trax:
Fix item change on a2enmod/a2dismod multiple line output.

In some cases a2enmod/a2dismod response message has multiple lines. E.g:

"Your MPM seems to be threaded. Selecting cgid instead of cgi.
Module cgid already enabled"

Because of that an item like
`- apache2_module name=cgi state=absent`
is always changed.


gregdek:
@berendt Please review this PR to make sure it adheres to the following guidelines:

robinro:
Multiline output is already covered by `re.S`. Your example would need some `.*` inbetween as it is looking for `cgi already enabled` but gets `cgid already enabled`

It would be probably better to use `a2enmod -l` to check if something changed instead of parsting the output. This would also render my fix, #2091, unnecessary.

n0trax:
Of course, you're right. The problem in my example is `cgid` and not the multiple lines.
Unfortunately there is no `a2enmod -l` on debian based systems, afaik.

PR #2086, opened 2015-09-16
Remove 'str' type so that json is properly quoted

wimnat:


wimnat:
@defionscode

juliedavila:
What was this causing? Duplicate quotes?

wimnat:
It would replace double quotes in my json file with single quotes, thus breaking the json

juliedavila:
+1

Ah, weird it didn't come up earlier. Thanks

gregdek:
Thanks @wimnat for the PR and thanks @defionscode for the review. Marking for inclusion.

cc: @ansible/core team.

PR #2089, opened 2015-09-16
Cast the Docker API version to a float where it is used in comparison tests.

homme:
By default the Docker API version is a string which resulted in bad logic in two
cases where this was compared to a float.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1019,7 +1019,7 @@ class DockerManager(object):
         current = self.get_inspect_containers(running)
 
         #Get API version
-        api_version = self.client.version()['ApiVersion']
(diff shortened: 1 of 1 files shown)


gregdek:

smashwilson:
I think #2239 has a better solution to this by using an explicit version comparison method from the Docker library, so I just gave that one the go-ahead. I just gave that one the go-ahead, so hopefully it'll be fixed soon.

Thanks for taking a stab at it! Using floats for the comparison is pretty clever.

gregdek:
Closing as dupe of #2239. Thanks @homme for the effort in putting together this PR.

homme:
@gregdek No problem - just as long as it's fixed one way or another!

PR #2090, opened 2015-09-16
docs - ec2_tag.py

halberom:
order params, add missing tag param. update examples

The current examples are confusing as they apply tags via `args:` and everything else via normal key=value.  Also the ec2 instance example isn't very useful because the ec2 module has tagging built-in.  Using it to tag volumes (which currently don't!) is imo a better example.


evanccnyc:
This is great I was meaning to write something like this since the ec2_tags docs arent very helpful. @jimi-c

jimi-c:
Merged, thanks!

PR #2099, opened 2015-09-17
update quantum modules for non-admin users

bregman-arie:
Not every tenant allowed to use tenant-list
which may result in long wait time and an
failure at the end.

This change allows regular user to run the
module without waiting for timeout because
of using tenant-list.

It also support insecure connections in case
resolving SSL certificate not working.


robynbergeron:
Hi, @bregman-arie -- thanks for the PR.

robynbergeron:
Greetings, lovely openstack smart folks, @emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz -- 

Thanks!

omgjlk:
This adds new options to these modules, but does not update the documentation for them. Is this the intention, to add a hidden option?

Also this PR is against `stable-1.9`, and the change isn't on master yet.

Otherwise, this looks okay.

gregdek:
I think our practice here is to put new PRs into devel, and then the Ansible team makes the decision to backport as needed. So seems like, at the very least, we should be seeing this PR against devel first -- since it would still apply there, right? Pinging @ansible/core for confirmation.

bregman-arie:
Hmm so what I need to do? =)

Shrews:
@gregdek It's worth pointing out that all of these modules are deprecated in 2.0. Not sure what your policy is on that.

bregman-arie:
FYI I'm aware those deprecated in Ansible 2.0
But atm we are using ansible 1.9.x

gregdek:
Thanks @bregman-arie. All backport requests must be reviewed by the core team, and this can take time. We appreciate your patience.

PR #2104, opened 2015-09-17
Fixing ec2_asg termination_policy

evanccnyc:
If this isnt set, it wont launch the instances because it needs to default to "Default" despite what boto docs say.


jimi-c:
Merged, thanks!

PR #2105, opened 2015-09-18
Adding os_networks_facts module

dagnello:
There can be instances during an Ansible play where the list of networks
currently available from OpenStack is required.  This update provides
network list functionality as a new os_networks_facts module.

### DIFF
--- a/cloud/openstack/os_networks_facts.py
+++ b/cloud/openstack/os_networks_facts.py
@@ -1,0 +1,139 @@
+#!/usr/bin/python
+
+# Copyright (c) 2015 Hewlett-Packard Development Company, L.P.
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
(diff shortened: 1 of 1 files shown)


bcoca:
cc @emonty @Shrews @j2sol

cloud/openstack/os_networks_facts.py
```diff
@@ -0,0 +1,72 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this software.  If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    import shade
+    from shade import meta
```
Shrews: meta is unused. Remove this import.

Shrews:
@dagnello Looks pretty good, except for the unnecessary import. Also, let's document the return (see os_keypair for an example).

Shrews:
@bcoca Is the RETURN documentation intended for facts modules? If not, then @dagnello can ignore my comment.

omgjlk:
I would think documenting what a fact module gives access to would be even more important than other modules.

omgjlk:
This module potentially return significantly more networks if the user is an admin. The command line tool has an argument to control whether all networks, or just networks associated with a user's project get returned. It'd be good to expose that here as well.

bcoca:
return docs are intended for ALL modules, we do understand that mapping every possible return is not always possible so you should find a nice middle ground to let users know what to expect but not need to detail every single possible outcome.

dagnello:
@bcoca @emonty @Shrews @j2sol ready

Shrews:
Seems ok to me.

:+1:

Shrews:
:shipit:

omgjlk:
shipit

gregdek:
Thanks @dagnello for the PR, and thanks @shrews and @j2sol for the reviews.

PR #2106, opened 2015-09-18
Adding os_port Module to openstack ansible modules

dagnello:


bcoca:
cc @emonty @Shrews @j2sol

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,302 @@
...
+     required: false
+     default: None
+   mac_address:
+     description:
+        - MAC address of this port.
+     required: false
+     default: None
+   security_groups:
+     description:
+        - Security group(s) associated with the port.
```
omgjlk: describe how to supply multiple (comma separated or however)

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,302 @@
...
+     required: false
+     default: None
+   no_security_groups:
+     description:
+        - Associate no security groups with this port.
```
omgjlk: `Do not associate a security group with this port`

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,302 @@
...
+
+def _compose_port_args(module):
+    port_kwargs = {}
+    optional_parameters = ['name',
+                           'fixed_ips',
+                           'device_id',
+                           'admin_state_up',
+                           'mac_address',
+                           'extra_dhcp_opt']
+    #print json.dumps({"module.params: ": module.params})
```
omgjlk: Might want to get rid of this code.

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,302 @@
...
+        if module.params[optional_param] is not None:
+            port_kwargs[optional_param] = module.params[optional_param]
+
+    if module.params['no_security_groups']:
+        port_kwargs['security_groups'] = []
+    elif module.params['security_groups']:
+        port_kwargs['security_groups'] = module.params[
+            'security_groups'].split(',')
+
+    #print json.dumps({"portkwargs: ": port_kwargs})
```
omgjlk: More code to remove.

omgjlk:
Some arguments are hashes, some are comma separated lists. These should be explained in the docs better.

The RETURN data should be documented as well.

dagnello:
@bcoca @emonty @Shrews @j2sol ready

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,316 @@
...
+                cloud.delete_port(port['id'])
+                changed = True
+            module.exit_json(changed=changed)
+
+    except shade.OpenStackCloudException as e:
+        module.fail_json(msg=e.message)
+
+# this is magic, see lib/ansible/module_common.py
+from ansible.module_utils.basic import *
+from ansible.module_utils.openstack import *
```
Shrews, line 393: Need to add:

<pre>
   if __name__ == '__main__':
</pre>

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,316 @@
+#!/usr/bin/python
+
+# Copyright (c) 2015 Hewlett-Packard Development Company, L.P.
```
Shrews: I don't think you are part of HP any more, so just remove this line.

Shrews:
RETURN documentation still needs to be added.

cloud/openstack/os_port.py
dagnello: I am part of HP.

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,357 @@
...
+        changed = False
+        if state == 'present':
+            if not port:
+                network = module.params['network']
+                if not network:
+                    module.fail_json(
+                        msg="Parameter 'network' is required in Port Create"
+                    )
+                port_kwargs = _compose_port_args(module)
+                network_id = cloud.get_network(network)['id']
```
Shrews: get_network() can return None if no matching network is found. You probably should call module.fail_json() in that case.

dagnello: ok

dagnello:
@bcoca @emonty @Shrews @j2sol all comments have been addressed.

omgjlk:
This looks fine to me, but I'd like somebody else to verify it as well.

dagnello:
@gregdek another patch ready

Shrews:
I want to manually test this one before approving. Will do so soon.

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,368 @@
...
+     description:
+        - Security group(s) ID(s) associated with the port (comma separated for
+          multiple security groups - no spaces between comma(s)).
+     required: false
+     default: None
+   no_security_groups:
+     description:
+        - Do not associate a security group with this port.
+     required: false
+     default: None
```
Shrews: You need to explicity define no_security_groups as a bool type. Since you are allowing it to be a string, any string value will be interpreted as true with the logic you have in this module.

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,368 @@
...
+   mac_address:
+     description:
+        - MAC address of this port.
+     required: false
+     default: None
+   security_groups:
+     description:
+        - Security group(s) ID(s) associated with the port (comma separated for
+          multiple security groups - no spaces between comma(s)).
```
Shrews: Why a comma separated list and not a standard YAML list? I'd actually prefer the YAML list, but if you need comma separated values, then support both. PR #2078 has such an example.

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,368 @@
...
+
+def main():
+    argument_spec = openstack_full_argument_spec(
+        network=dict(required=False),
+        name=dict(required=False),
+        fixed_ips=dict(default=None),
+        admin_state_up=dict(default=None),
+        mac_address=dict(default=None),
+        security_groups=dict(default=None),
+        no_security_groups=dict(default=None),
```
Shrews: See comment above. This should be a bool type.

Shrews:
I've tested basic functionality. I think it needs a few changes.

cloud/openstack/os_port.py
dagnello: I can support both.  For comparison the os_server accepts security_groups in the same manner as this.

cloud/openstack/os_port.py
dagnello: done

cloud/openstack/os_port.py
dagnello, line 393: done

dagnello:
@Shrews comments addressed.

cloud/openstack/os_port.py
Shrews: None is not a good default here. If I'm a user reading this doc, which value will I get if I don't supply one?

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,382 @@
...
+
+    if module.params['no_security_groups']:
+        port_kwargs['security_groups'] = []
+    elif module.params['security_groups']:
+        if type(module.params['security_groups']) == str:
+            port_kwargs['security_groups'] = module.params[
+                'security_groups'].split(',')
+        else:
+            port_kwargs['security_groups'] = module.params['security_groups']
+
```
caphrim007, line 293: Would it be possible to further interpret this list of security groups to get their ID from the cloud.get_security_group() method so that we could provide the list of security groups by name as well as id?

Something akin to this, or however you would want to implement it.

port_kwargs['security_groups'] = map(lambda v: cloud.get_security_group(v)['id'], security_groups)

caphrim007:
I can verify that the following example works with this module

```
- name: Assign client port
  os_port:
      name: "client-port"
      network: "client-network"
      fixed_ips:
          - ip_address: 10.2.2.2
```

Additionally specifying security groups by ID works and using the no_security_groups correctly removes the security groups.

cloud/openstack/os_port.py
dagnello, line 293: ok, this looks simple and useful to add.  Will look into this shortly.

Shrews, line 293: You _could_ add this, but it might slow the module quite a bit. In order to support passing name or ID to get_security_group(), a list of _all_ security groups is returned and then searched for a match on name or ID. This would be done for each security group passed in to this module. This isn't very efficient, and something we hope to change in later versions of shade.

I would suggest leaving it as is, and maybe we could add support for sec group names later.

caphrim007, line 293: ahh. I looked at the OpenStack API notes and noticed it's done this way in OpenStack, so shade is just mirroring that behavior. I was just considering it because one can specify security groups by name in the os_server and os_security_group_rule modules but not the os_port module. So it was purely a feature parity request.

Shrews, line 293: Yeah. I'm not opposed to @dagnello adding this, if he wants to go ahead and do that. Just wanted to put that warning out there.

caphrim007, line 293: Understood. I was not aware of the shade behavior, so thanks for the heads up. If @dagnello can come up with a tenable solution, then by all means. Otherwise I'm not opposed to the current behavior if we can address it in the future with enhancements to shade.

Thanks @Shrews!

dagnello, line 293: @Shrews @caphrim007 already added it, missed this discussion :)  See what you think, I can revert back if needed.

cloud/openstack/os_port.py
dagnello: looking into this more, we don't want supply a default value here.  no_security_group=False is already assumed by neutron when creating a new port when not indicated otherwise.  Basically this configuration flag/value is not included at all in the request, hence 'None'.

dagnello:
@caphrim007 @Shrews patch updated

cloud/openstack/os_port.py
```diff
@@ -0,0 +1,391 @@
...
+        if module.params[key] is not None and cmp(module.params[key],
+                                                  port[key]) != 0:
+            return True
+    for key in compare_comma_separated_list:
+        if type(module.params[key]) == str:
+                module.params[key] = module.params[key].split(',')
+        if key == 'security_groups':
+            # Retrieve security group id's since security group names are also
+            # supported as valid input.
+            module.params[key] = map(lambda v: cloud.get_security_group(v)['id'], module.params[key])
```
Shrews: Almost there!  :)

get_security_group() will return None if a match is not found. So if a user mistypes a secgroup name, they'll get a weird TypeError exception:

<pre>
TypeError: 'NoneType' object has no attribute '__getitem__'
</pre>


 So, it would probably be best to look for a None return and fail_json() with the bad group.

Also, since we are having to translate group names to IDs twice, maybe we should just do this step up front and pass the UUIDs down to _needs_update() and _compose_port_args()?

dagnello: agreed, will convert up-front.

dagnello:
@caphrim007 @Shrews patch updated

Shrews:
This looks good to me now! Thanks for your contribution.

Shrews:
shipit

dagnello:
thanks!

caphrim007:
tested on my end too. lgtm. thanks!

dagnello:
cc @robynbergeron

dagnello:
cc @bcoca this is ready

PR #2123, opened 2015-09-21
standardize tags param

halberom:
Most of the new ec2 modules use `tags:` as the tags parameter name.  However several of the core modules use `snapshot_tags`, `instance_tags`, `ami_tags` etc.

Updating them so they all accept `tags:`, and use aliases for backward compatibility.

Tested all 3 with before and after tag names.  See [here](https://gist.github.com/halberom/9a440ae21e88ff18fd19) for testing output

I've not updated the ec2_vpc module, because the new ones (ec2_vpc_net etc) replace it, so I'm assuming it's deprecated although I can do that too if desired.


halberom:
I could probably have just added 'tags' as the alias, instead of the find/replace I did, but again, all the other modules use `tags` in the code too.

gregdek:
Since this affects multiple AWS modules, pinging all the AWS folks here for review:

@jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode @tastychutney @bpennypacker @loia

cloud/amazon/ec2.py
```diff
@@ -229,17 +229,17 @@
...
     aliases: []
-  count_tag:
+  count_tags:
     version_added: "1.5"
     description:
       - Used with 'exact_count' to determine how many nodes based on a specific tag criteria should be running.  This can be expressed in multiple ways and is shown in the EXAMPLES section.  For instance, one can request 25 servers that are tagged with "class=webserver".
     required: false
     default: null
-    aliases: []
+    aliases: ['count_tags']
```
manicminer: Should the alias here perhaps be `count_tag` (i.e. singular) like the original arg?

needs_revision ?

jarv:
shipit

halberom:
good catch, thanks Tom.

manicminer:
No problem! This is a sensible change. shipit

gregdek:
Thanks for the quick review gents, and thanks to @halberom for the PR.

juliedavila:
+1

bcoca:
version_added needs to be updated for all. I would suggest instead to make 'tags' as an alias and make a comment 'tags alias added in 2.0'

juliedavila:
@halberom friendly ping

halberom:
Busy with other stuff atm, will redo at a later date when have time.

Issue #2040, opened 2015-09-09
EBS volume_type ignored


jfindley:
## Issue type

Bug report
## Summary

When creating an ec2 volume, volume_type is ignored.

The config:

``` yaml
volumes:
  device_name: /dev/sdb
        volume_size: 100
        volume_type: gp2
```

creates a volume using standard disks, not gp2.

The config:

``` yaml
volumes:
  device_name: /dev/sdb
        volume_size: 100
        volume_type: io1
        iops: 150
```

produces an error message: "Instance creation failed => InvalidParameterCombination: The parameter iops is not supported for standard volumes."
## Ansible version:

```
ansible --version
ansible 2.0.0 (detached HEAD 43cc27084d) last updated 2015/08/13 15:18:18 (GMT +100)
  lib/ansible/modules/core: (devel 1c7e8ef1a0) last updated 2015/09/09 17:21:32 (GMT +100)
  lib/ansible/modules/extras: (detached HEAD ff2386faf4) last updated 2015/07/21 18:18:28 (GMT +100)
  v1/ansible/modules/core: (detached HEAD f8d8af17cd) last updated 2015/07/21 18:18:36 (GMT +100)
  v1/ansible/modules/extras: (detached HEAD 495ad450e5) last updated 2015/07/21 18:18:40 (GMT +100)
  config file = /etc/ansible/ansible.cfg
  configured module search path = None
```
## Environment/config:

OSX source install.  Default config.

evanccnyc:
This appears to be because you are using volume_type instead of device_type under the volumes settings. Why volume_type is allowed there I dont know.

evanccnyc:
If anyone wanted to we could add the following lines around 729:

```
    if 'volume_type' in volume:
        volume_type = volume.get('volume_type')
    elif 'device_type' in volume:
        volume_type = volume.get('device_type')
    return BlockDeviceType(snapshot_id=volume.get('snapshot'),
                           ephemeral_name=volume.get('ephemeral'),
                           size=volume.get('volume_size'),
                           volume_type=volume_type,
                           delete_on_termination=volume.get('delete_on_termination', False),
                           iops=volume.get('iops'),
                           encrypted=volume.get('encrypted', None))
```

Not sure its worth it though

evanccnyc:
The bigger problem is that ec2_vol uses 'volume_type' where as ec2 and ec2_lc uses 'device_type'

willthames:
[...]

PR #2124, opened 2015-09-21
ec2 - allow use of volume_type in volumes dict

halberom:
For some reason, `device_type` was used to specify the volume_type.  This differs with both the ec2_vol module, and also the BlockDeviceType parameter name.   This PR adds handling for `volume_type` while keeping `device_type` for backward compatibility.

Test output is [here](https://gist.github.com/halberom/523f788633ff74e97b20)

I didn't note an alias in the param description, because it's a sub key, and I wasn't sure how to specify it.

I also have no idea about deprecation warnings, how to add, or when to - as it might be nice to deprecate the device_type.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -216,7 +216,7 @@ options:
   volumes:
     version_added: "1.5"
     description:
-      - "a list of volume dicts, each containing device name and optionally ephemeral id or snapshot id. Size and type (and number of iops for io device type) must be specified for a new volume or a root volume, and may be passed for a snapshot volume. For any volume, a volume size less than 1 will be interpreted as a request not to create the volume. Encrypt the volume by passing 'encrypted: true' in the volume dict."
+      - a list of hash/dictionaries of volumes to add to the new instance; '[{"key":"value", "key":"value"}]'; keys allowed are - device_name (str; required), delete_on_termination (bool; False), device_type (deprecated), ephemeral (str), encrypted (bool; False), snapshot (str), volume_type (str), iops (int) - device_type is deprecated use volume_type, iops must be set when volume_type='io1', ephemeral and snapshot are mutually exclusive.
     required: false
     default: null
     aliases: []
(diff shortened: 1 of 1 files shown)


willthames:
This addresses #2040, which does suggest a couple of methods of 'deprecating' `device_type` - unfortunately there's no official mechanism for deprecating subkeys afaik. 

This looks good to me (and apologies for calling it `device_type` in the first place)

halberom:
Thanks, and no worries, I'm sure it made perfect sense at the time!

evanccnyc:
+1 on this

abadger:
Looks good code wise  Probably should mention the deprecation in the documentation and in the error message that's returned if both volume_type and device_type are specified.

gregdek:
Going to consider @willthames comment to be the needed +1 since git blame confirms he wrote the code in the first place. :)

Special thanks to @halberom for including test output.

@abadger should we block on the docs change, or mark for inclusion now?

gregdek:
@abadger if it's ready to go, put "shipit" in the text. If not, put "needs_revision" in the text. Thx!

halberom:
@abadger ok, rather than go back and forward lots, can you confirm whether this seems reasonable
- a list of hash/dictionaries of volumes to add to the new instance; "[{key:value, key:value}]"; keys allowed are - device_name (str; required), delete_on_termination (bool; False), device_type (deprecated), ephemeral (str), encrypted (bool; False), snapshot (str), volume_type (str), iops (int) - device_type is deprecated use volume_type, iops must be set when volume_type='io1', ephemeral and snapshot are mutually exclusive.

with the example being

```
    volumes:
      - device_name: /dev/sdb
        snapshot: snap-abcdef12
        volume_type: io1                  # device_type is deprecated
        iops: 1000
        volume_size: 100
        delete_on_termination: true
        encrypted: no
```

I'm thinking a rewrite of the description is in order as the previous one doesn't list allowed keys, value types or defaults.

(I think the reason it was device_type originally might be due to boto's returning root_device_type for the instance call).

abadger:
@halberom  The rewritten description looks good.

No need to add the device_type comment to the example

I'd also change line: 718 like this:

``` python
       module.fail_json(msg = 'device_type is a deprecated name for volume_type. Do not use both device_type and volume_type.')
```

abadger:
Thanks for doing this @halberom Thanks for reviewing @willthames and @evanccnyc !

# Merged

Hi!

issue is resolved.

Thank you!

PR #2126, opened 2015-09-22
Adding support for access logs in ELB

nfelsen:
screen_shot_2015-09-21_at_6_40_10_pm

I added support to turn on the access log options as shown in the screenshot above. 
The configuration is more or less identical how you add a healcheck for example:

```
# Configure a health check and the access logs
- local_action:
    module: ec2_elb_lb
    name: "test-please-delete"
    state: present
    zones:
      - us-east-1d
    listeners:
      - protocol: http
        load_balancer_port: 80
        instance_port: 80
    health_check:
        ping_protocol: http # options are http, https, ssl, tcp
        ping_port: 80
        ping_path: "/index.html" # not required for tcp or ssl
        response_timeout: 5 # seconds
        interval: 30 # seconds
        unhealthy_threshold: 2
        healthy_threshold: 10
    access_logs:
        interval: 5 # minutes
        s3_location: "my-bucket"
        s3_prefix: "logs"
```


evanccnyc:
This would be nice to add but I havent had a chance to test it.

gregdek:
@jsdalton @rickmendes Please review this PR to make sure it adheres to the following guidelines:

jsdalton:
I have not tested this code, but it looks mostly good. I think the implementation of `_set_access_logs()` could use some improvement. It's mostly minor stuff, but I'll try to call it out in line comments.

cloud/amazon/ec2_elb_lb.py
```diff
@@ -182,6 +187,10 @@
         interval: 30 # seconds
         unhealthy_threshold: 2
         healthy_threshold: 10
+    access_logs:
+        interval: 5 # minutes
```
jsdalton: Since we set a default explicitly below, it might make sense to call that out in the comment e.g.

```
interval: 5 # minutes (defaults to 60)
```

cloud/amazon/ec2_elb_lb.py
```diff
@@ -182,6 +187,10 @@
...
+        s3_location: "my-bucket"
```
jsdalton: Call out that this is required of access_logs is set

```
s3_location: "my-bucket" # This value is required if access_logs is set
```

cloud/amazon/ec2_elb_lb.py
```diff
@@ -698,6 +710,28 @@ def _set_cross_az_load_balancing(self):
         self.elb_conn.modify_lb_attribute(self.name, 'CrossZoneLoadBalancing',
                                           attributes.cross_zone_load_balancing.enabled)
 
+    def _set_access_log(self):
+        attributes = self.elb.get_attributes()
+        if self.access_logs is not None:
```
jsdalton: How about just

``` python
if self.access_logs:
    ...
```

cloud/amazon/ec2_elb_lb.py
```diff
@@ -698,6 +710,28 @@ def _set_cross_az_load_balancing(self):
...
+            access_log = boto.ec2.elb.attributes.AccessLogAttribute()
+            access_log.enabled = True
+            access_log.s3_bucket_prefix = ''
+            access_log.emit_interval = 60
```
jsdalton: Does this value need to be set explicitly here? I see from the AWS docs that the default is already 60.

cloud/amazon/ec2_elb_lb.py
```diff
@@ -698,6 +710,28 @@ def _set_cross_az_load_balancing(self):
...
+            if 's3_location' not in self.access_logs:
+              self.module.fail_json(msg='s3_location information required')
+            access_log.s3_bucket_name = self.access_logs['s3_location']
+            if 's3_prefix' in self.access_logs:
+              access_log.s3_bucket_prefix = self.access_logs['s3_prefix']
+            if 'interval' in self.access_logs:
+              access_log.emit_interval = self.access_logs['interval']
+            if str(access_log) != str(attributes.access_log):
```
jsdalton: Coercing these into a string and doing a comparison is almost certainly not what we want to do here. Is there a reason we can't just do `access_log == attributes.access_log'?

If those don't compare for some reason there is probably a better type we can coerce these into for comparison's sake, e.g. a dict.

Alternatively you could keep a `changed` boolean initialized to `False` and set it to `True` any of the values are set.

cloud/amazon/ec2_elb_lb.py
nfelsen: Yeah, they don't set default values but complain when something required isn't set:

```
% python
Python 2.7.10 (default, Jul 14 2015, 19:46:27)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import boto
>>> import boto.ec2.elb
>>> import boto.ec2.elb.attributes
>>> from boto.regioninfo import RegionInfo
>>> from boto.ec2.elb.attributes import AccessLogAttribute
>>> access_log = AccessLogAttribute()
>>> access_log
AccessLog(None, None, None, None)
>>> access_log.emit_interval
>>>
```

cloud/amazon/ec2_elb_lb.py
nfelsen: Thanks for looking at my code...
There is no good way to compare those objects unfortunately:

```
>>> dir(AccessLogAttribute())
['__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__module__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'emit_interval', 'enabled', 'endElement', 's3_bucket_name', 's3_bucket_prefix', 'startElement']
>>> a = AccessLogAttribute()
>>> b = AccessLogAttribute()
>>> a
AccessLog(None, None, None, None)
>>> b
AccessLog(None, None, None, None)
>>> a == b
False
```

Following your advices, I changed the code to use `__dict()__` instead... let me know.

nfelsen:
Thanks for reviewing my code. I changed and rebase the code following your advices. Let me know...

cloud/amazon/ec2_elb_lb.py
```diff
@@ -698,6 +710,28 @@ def _set_cross_az_load_balancing(self):
...
+            access_log.s3_bucket_prefix = ''
+            access_log.emit_interval = 60
+            if 's3_location' not in self.access_logs:
+              self.module.fail_json(msg='s3_location information required')
+            access_log.s3_bucket_name = self.access_logs['s3_location']
+            if 's3_prefix' in self.access_logs:
+              access_log.s3_bucket_prefix = self.access_logs['s3_prefix']
+            if 'interval' in self.access_logs:
+              access_log.emit_interval = self.access_logs['interval']
+            if access_log.__dict__ != attributes.access_log.__dict__:
```
jsdalton: This works better although I still think it's not ideal. If one or the other object accumulates extra data attributes stored in `__dict__` we might end up in a situation where the comparison always returns false and the module always thinks there has been a change.

Consider that in the code above we are explicitly walking over each attribute that's meaningfully shared across both object. It might make sense to just check if one of those has changed while we're doing that check.

Checkout  `__set_health_check` for a way you could approach it.

nfelsen:
Hey, @jsdalton let me know if it's more to your liking 
Thanks,
Nat

jsdalton:
@nfelsen :+1: looks good to me.

nfelsen:
Awesome thanks again for the code review. What's the next step to get the "shipit" ?

ghost:
Looks good to me. shipit

gregdek:
Thanks @jsdalton and @rickmendes for the reviews, and thanks @nfelsen for the PR. Marking for potential inclusion.

nfelsen:
rebased

ghost:
At the risk of being redundtant: shipit

PR #2127, opened 2015-09-22
ec2_group.py: added ICMP rule example

selivan:
ec2_group.py: added ICMP rule example to documentation.


evanccnyc:
+1 on this. Took me forever to hunt down how to do this.

abadger:
Thanks for the doc update!

Issue #1277, opened 2015-05-07
Add support for docker labels


bobrik:
I can come up with a PR, but I already have 2 from March 22 that were not merged yet.

diefans:
+1

ryanwalls:
+1

MrMMorris:
@bobrik I can't find any PR's from you on this. If you have a patch I would LOVE this functionality as I don't name my containers and need to differentiate with Prometheus. 

Let me know if there is a PR I can throw a fat +1 at :yum:

MrMMorris:
@smashwilson this would also be nice if you have time :smile:

bobrik:
@MrMMorris I have tho PRs hanging from Mar 22, can't see any reason to add more. Anyway, check out [collectd-docker](https://github.com/bobrik/collectd-docker) it might be sufficient for you :)

0x736861:
+1

lbradeen:
+1

Cactusbone:
:+1:

MrMMorris:
has there been any progress on this?

leonty:
Someone please review.

leonty:
Created a PR for v2 devel.

solher:
+1

paudam:
+3

PR #2133, opened 2015-09-22
Support 'labels' parameter for docker.

leonty:
This PR adds support of the 'label' parameter to the 'docker run' command as it stated in the docker documentation:

```
-l, --label=[]                Set metadata on the container (e.g., --label=com.example.key=value)
```

The new parameter is called 'labels' and has the 'dict' type. Example:

```
    - name: Running a container from new build
      docker: 
        name: "{{ container_name }}"
        image: "{{ container_name }}:{{ build_name }}"
        hostname: "{{ container_name }}.{{ dev_domain }}"
        labels:
          com.mycorp.type: dev
          com.mycorp.name: MyProject
        state: running
```


j1r1k:
This fixes #1277. Thanks

gregdek:

AndreiPashkin:
The PR doesn't introduce matching logic for labels. Without it, if you change labels in the playbook, the modules will start a new container, instead changing existing one.

leonty:
Isn't this code does so? Line 1015:
-  if actual_labels != expected_labels:
-                self.reload_reasons.append('labels {0} => {1}'.format(actual_labels, expected_labels))
-                differing.append(container)
-                continue

I did everything similarily to 'env' parameter processing.

AndreiPashkin:
@leonty 
You talking about `get_differing_containers()`, which is used for `state=reloaded` case and I'm talking about `get_deployed_containers()`, which is responsible for finding running containers that are matches with parameters passed to Ansible task. However, maybe you right, and labels should not participate in containers matching.

leonty:
Hi @AndrewPashkin, 
Below is a real deploy script from my work environment that uses the labels functionality from this PR. Isn't the case you describe matches the step of finding and removing (the very first 'docker' task) the container? While the running container already has labels, the remove task doesn't and still it finds the container correctly.

```
---
- hosts: "{{ deploy_host }}"
  vars:
    release_name: "{{ lookup('env', 'bamboo_deploy_environment') }}"
    build_name: "{{ lookup('env', 'bamboo_buildNumber') }}"
    container_name: myproject-{{ release_name }}
    build_dir: /opt/docker_build/{{ container_name }}
    work_dir: "{{ lookup('env', 'bamboo_build_working_directory') }}"

  tasks:
    - file: path={{ build_dir }}/ state=absent
    - copy: src=docker/ dest={{ build_dir }}
    - file: path={{ build_dir }}/apps state=directory
    - copy: src={{ item }} dest={{ build_dir }}/apps
      with_fileglob: [ "{{ work_dir }}/*.tar" ]
    - docker_image: name={{ container_name }}
        tag={{ build_name }}
        path={{ build_dir }}
        state=build
    - docker: name={{ container_name }}
        image={{ container_name }}
        state=absent
    - docker: 
        name: "{{ container_name }}"
        image: "{{ container_name }}:{{ build_name }}"
        hostname: "{{ container_name }}.{{ dev_domain }}"
        volumes_from: myproject-images
        labels:
          my.type: dev
          my.name: MyProject
        state: running
    - docker_facts: name={{ container_name }}
    - file: path={{ share_folder }}/{{ container_name }} state=absent
    - file: path={{ share_folder }}/{{ container_name }} state=directory
    - file: path={{ share_folder }}/{{ container_name }}/logs
        src={{docker_containers[container_name].docker_volumes['/opt/myproject/logs']}}
        state=link
```

AndreiPashkin:
@leonty I didn't got the point, why do you need to delete the container?

My point was, that some parameters to `docker` module are participate in container matching, and some are not. If you look at the documentation for each parameter, you will see, that it mentions whether or not a parameter used for matching. So if you have "matching" parameter and `state` equal to `started`, the `docker` module will first check if a container with a setting, that corresponds to the parameter, set to equal value is running, and start a new container only if there is no such container.

leonty:
@AndrewPashkin So you think labels should be used for matching or not? I've implemented this for my case that doesn't demand such a behavior, I always refer the name...

ps: I delete the container because the 'reloaded' state didn't work as expected (no matter of labels) and I didn't dig it out. Probably because the new image has a new tag.

AndreiPashkin:
@leonty I don't know, actually, I'm in doubt.

leonty:
@AndrewPashkin so let it evolve according to the users' needs :) As the labels for the 'run' command is a quite fresh functionality I believe most existing devops use cases rely on the 'name' matching (as mine).

cloud/docker/docker.py
```diff
@@ -252,6 +252,12 @@
         docker-py >= 0.5.0.
     default: false
     version_added: "1.9"
+  labels:
+    description:
+      - Set container labels. Requires docker >= 1.6 and docker-py >= 1.2.0.
+    requered: false
```
smashwilson: Spelling: "required" :nail_care:

smashwilson:
I'm also inclined to favor not using labels for container matching, at least initially. I think there's value in using as few attributes as possible to keep matching predictable. Also, different hosting environments can use labels for different purposes (affinity and so forth), so it would feel weird to me to interpret them as a flag for container identity.

I could pretty easily be persuaded otherwise, though. I don't actually use them myself so I'd gladly listen to anyone who is using them in anger.

@leonty: Ping me once you've had a chance to touch up that spelling fix and I'll give the go ahead.

smashwilson:
:+1:

shipit

gregdek:
Thanks @leonty for the PR, and thanks @smashwilson for the review. Marking for inclusion.

bcoca:
duplicate of #2159, but closing this one as the PR is not against devel, please check that PR in case its missing something this implementation has.

y8:
Hum, in the [docs](http://docs.ansible.com/ansible/docker_module.html) it's marked like this `labels` are added to the `1.9.4`, but now it's obvious that it's not even merged in the 1.9 branch.

Should I open an issue or I'm missing something? :)

leonty:
@y8 Before the 1.9.4 was released I created 2 PRs for 1.9 and 2.0 branches. In the docs I wrote the closest release I expected to merge this functionality.

y8:
So your PR will be avilable in next 1.9.x release in the 2.0? I'm new to the ansible branch layout, and can't find any info.

Anyway, if it's not in the 1.9.x, it's better to change the docs to expected version, which is seems like 2.0.

By the way, thanks for PR!

leonty:
@y8 this one is closed and not merged due to "closing this one as the PR is not against devel," bcoca's decision, see comments above. Another PR for 2.0 is already merged.

PR #2135, opened 2015-09-23
Revert "win_get_url re-download file only if modified"

jugalporwal:
Faced the following issue while using win_get_url for getting a remote build 
fatal: [remote_winodws_ip]: FAILED! => {"changed": false, "failed": true, "msg": "Error downloading http://<jenkins_server_ip>/jenkins/view/Trunk/job/router/lastSuccessfulBuild/artifact/router/conf/router-service-context.xml to D:\router\router-service-context.xml Exception calling \"DownloadFile\" with \"2\" argument(s): \"An exception occurred during a WebClient request.\""}

### DIFF
--- a/windows/win_get_url.ps1
+++ b/windows/win_get_url.ps1
@@ -1,7 +1,7 @@
 #!powershell
 # This file is part of Ansible.
 #
-# (c)) 2015, Paul Durivage <paul.durivage@rackspace.com>, Tal Auslander <tal@cloudshare.com>
+# Copyright 2014, Paul Durivage <paul.durivage@rackspace.com>
 #
 # Ansible is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


gregdek:
@jugalporwal Thanks for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

cchurch:
According to your issue description, you're trying to download `http:///jenkins/view/Trunk/job/router/lastSuccessfulBuild/artifact/router/conf/router-service-context.xml`, with 3 slashes after http.  **Since it isn't a correctly-formatted URL, I wouldn't expect it to work.**

When I run against current devel using a similar URL, I also get an error with the following similar message:

> Error downloading http:///docs.ansible.com to C:\Users\Administrator\docs_index.html Exception calling \"DownloadFile\" with \"2\" argument(s): \"URI formats are not supported.\"

Even with your PR applied, I get the following message:

> Error downloading http:///docs.ansible.com to C:\test.html

jhawkesworth:
Wasn't aware of this when I raised https://github.com/ansible/ansible-modules-core/issues/2146 which covers the same issue but describes a fix, rather than reverting force=no logic.

PR #2142, opened 2015-09-24
Fix hashbang for openstack novaclient modules

pawelkowalak:
Hardcoded `/usr/bin/python` hashbang makes it unsuable in virtualenv.
It results with error:
`failed=True msg='novaclient is required for this module'` if ansible
is not installed globally.

### DIFF
--- a/cloud/openstack/_glance_image.py
+++ b/cloud/openstack/_glance_image.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 # (c) 2013, Benno Joy <benno@ansible.com>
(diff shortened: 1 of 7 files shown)


sivel:
# Possible Misunderstanding

Hi!

We believe the ticket you have filed is being somewhat misunderstood, as one thing works a little differently than stated.

Using a hardcoded value of `/usr/bin/python` is expected here.

Please see following FAQ entry for how to deal with the python binary living in another location: http://docs.ansible.com/ansible/faq.html#how-do-i-handle-python-pathing-not-having-a-python-2-x-in-usr-bin-python-on-a-remote-machine

In the future, this might be a topic more well suited for the user list, which you can also post here if you'd like some more help with the above.
- https://groups.google.com/forum/#!forum/ansible-project

pawelkowalak:
Ah, thanks for clarification. I guess I was asking wrong questions to google as I couldn't find the answer :)

PR #2149, opened 2015-09-24
Allow must_exist to function with systemd

retr0h:
Currently, the `must_exit` option will fail when a systemd unit does not
exist on the system.  Extended the functionality provided in #751 to
support systemd.

```
msg: systemd could not find the requested service "'openstack-nova-scheduler'":
```

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -520,13 +520,15 @@ class LinuxService(Service):
             return False
 
     def get_systemd_status_dict(self):
-
         # Check status first as show will not fail if service does not exist
         (rc, out, err) = self.execute_command("%s show '%s'" % (self.enable_cmd, self.__systemd_unit,))
         if rc != 0:
             self.module.fail_json(msg='failure %d running systemctl show for %r: %s' % (rc, self.__systemd_unit, err))
         elif 'LoadState=not-found' in out:
-            self.module.fail_json(msg='systemd could not find the requested service "%r": %s' % (self.__systemd_unit, err))
+            if self.module.params['must_exist']:
+                self.module.fail_json(msg='systemd could not find the requested service "%r": %s' % (self.__systemd_unit, err))
+            else:
+                self.module.exit_json(changed=False, exists=False)
 
         key = None
         value_buffer = []


system/service.py
```diff
@@ -520,13 +520,15 @@ def sysv_is_enabled(name):
...
-            self.module.fail_json(msg='failure %d running systemctl show for %r: %s' % (rc, self.__systemd_unit, err))
-        elif 'LoadState=not-found' in out:
-            self.module.fail_json(msg='systemd could not find the requested service "%r": %s' % (self.__systemd_unit, err))
+        if self.module.params['must_exist']:
+            if rc != 0:
+                self.module.fail_json(msg='failure %d running systemctl show for %r: %s' % (rc, self.__systemd_unit, err))
+            elif 'LoadState=not-found' in out:
+                self.module.fail_json(msg='systemd could not find the requested service "%r": %s' % (self.__systemd_unit, err))
+        else:
+            self.module.exit_json(changed=False, exists=False)
```
omgjlk: I'm assuming from previous code that a not-found case would still exit 0, so I think in both both `must_exist` or not we should catch the `rc != 0` case. Whether or not a not-found scenario leads to a failure should be the only thing inside of the `if self.module.params['must_exist']:` if statement.

As it stands in this change, a run with `must_exist` set to False could simply exit_json even on a non-0 return code.

retr0h: Not found doesn't exit 0.  Systemctl returns a non 0 exit code when the unit file does not exist.

```
[root@aio-01-rhel-7 ~]# systemctl restart foo
Failed to issue method call: Unit foo.service failed to load: No such file or directory.
[root@aio-01-rhel-7 ~]# echo $?
6
```

Since it is not 0 we get a failure, and bombs out the module.  I have throughly tested this patch in my environment, and works exactly as expected now.

omgjlk: Huh, I guess I didn't follow the logic.

So if `must_exist` is set, and the exit is non-zero, you get the failure from line 527. Since failure to find the service always results in non-zero, how would line 529 ever be reached?

Lets walk through when `must_exist` is set to false, it looks like the code immediately jumps to line 531, regardless of the exit code. How would this actually work if `must_exist` is set to false, but the service actually does exist?

omgjlk:
This works. `systemctl show foo` does actually result in a `rc` of 0 when a service doesn't exist. Now the logic flows correctly and when in that case checks the `must_exist` argument.

I'd say just squash the two commits into one and this is good to merge!

gregdek:
@retr0h thanks for the PR. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

cc: @ansible/core team.

retr0h:
Closing in favor of #2724, since #751 was removed off master, thus making this invalid.

PR #2160, opened 2015-09-25
Fix to handle user directory within parameter 'template'.

j-carl:


bcoca:
actually, if we just change it to a 'path' in the spec this happens automatically.

gregdek:
I presume you're actively recommending this change @bcoca.

@j-carl Please make the suggested revisions, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

j-carl:
I made the change recommend by @bcoca and it's "ready_for_review".

PR #2163, opened 2015-09-26
misc: remove superfluous trailing semicolon

pheanex:


gregdek:
cc: @ansible/core -- looks like a trivial fix

bcoca:
IMO it is not worth changing the attribution in the line for dealing with an optional character.

pheanex:
@bcoca sry you are probably right, this was part of a cleanup task in the complete ansible probject.
=> didn't notice that is was just this one line in core :-)

PR #2168, opened 2015-09-28
Updated 'stat' module examples to use 'failed_when' syntax

jeffwidman:
Just made the examples a little cleaner using the newer syntax


bcoca:
not that these examples were made for displaying error conditions, the debug is just to show the info, the fail allows for custom failure message. I would add a example with failed when but not remove the existing ones.

gregdek:
@jeffwidman Please make the suggested revisions, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

gregdek:
@jeffwidman A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

jeffwidman:
Thanks for the reminder. 

After looking at this a little more, I'm not too excited about keeping the existing error messages as I think the `failed_when` is what most people should use if they're writing a `stat` task. 

The custom error messages are a little more advanced and don't need to be documented here as they're a separate `fail` task and anyone who wants them can just search for 'how to produce custom error message ansible'. 

Not a big deal, just my personal opinion. Go ahead and close this PR if you prefer to keep existing examples.

bcoca:
they are just very simple examples on how to use the stat module with other modules,

PR #2173, opened 2015-09-29
Powershell throwing 'The variable cannot be retrieved because it has not been set' error

mcalhoun:
When a variable is previously unitialized in the script, powershell is throwing an error message: The variable '$foo' cannot be retrieved because it has not been set.

This was happening for the following in windows/setup.ps1:

$winrm_https_listener_path
$https_listener
$winrm_cert_thumbprint
$uppercase_cert_thumbprint


cchurch:
Merged, thanks!

PR #2174, opened 2015-09-29
Fix a argument mismatch in elasticache

voidfiles:
I think in this commit 720aeffca2bd2ae1eca158abc2d1463a8597afb6

There was bug introduced where the ElastiCacheManager init method has
a number of positional arguments like so.

``` py
    def __init__(self, module, name, engine, cache_engine_version, node_type,
                 num_nodes, cache_port, parameter_group, cache_subnet_group,
                 cache_security_groups, security_group_ids, zone, wait,
                 hard_modify, region, **aws_connect_kwargs):
```

But then later in the code the positional arguments are passed in
like this.

``` py
    elasticache_manager = ElastiCacheManager(module, name, engine,
                                             cache_engine_version, node_type,
                                             num_nodes, cache_port,
                                             cache_subnet_group,
                                             cache_security_groups,
                                             security_group_ids, parameter_group, zone, wait,
                                             hard_modify, region, **aws_connect_kwargs)
```

If you count, you can see that cache_subnet_group, is being passed in
where the manager expects to see parameter_group.

### DIFF
--- a/cloud/amazon/elasticache.py
+++ b/cloud/amazon/elasticache.py
@@ -445,6 +445,7 @@ class ElastiCacheManager(object):
 
     def _refresh_data(self, cache_cluster_data=None):
         """Refresh data about this cache cluster"""
+
         if cache_cluster_data is None:
             try:
                 response = self.conn.describe_cache_clusters(cache_cluster_id=self.name,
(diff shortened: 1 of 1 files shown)


gregdek:
@voidfiles thanks for the PR.

@jsdalton Please review this PR to make sure it adheres to the following guidelines:

jsdalton:
:+1:  "shipit"

voidfiles:
@jsdalton thanks.

PR #2196, opened 2015-10-03
Support cloning VMs into a specific VM folder

fennm:
The pysphere VIVirtualMachine.clone() method supports specifying a VM
folder to place the VM in after the clone has completed.  This exposes
that functionality to playbooks.

Also documents that creating VMs could always place VMs in a specific
folder.

### DIFF
--- a/cloud/vmware/vsphere_guest.py
+++ b/cloud/vmware/vsphere_guest.py
@@ -170,6 +170,7 @@ EXAMPLES = '''
       vcpu.hotadd: yes
       mem.hotadd:  yes
       notes: This is a test VM
+      folder: MyFolder
     vm_disk:
       disk1:
         size_gb: 10
@@ -241,6 +242,8 @@ EXAMPLES = '''
     template_src: centosTemplate
     cluster: MainCluster
     resource_pool: "/Resources"
+    vm_extra_config:
+      folder: MyFolder
 
 # Task to gather facts from a vSphere cluster only if the system is a VMWare guest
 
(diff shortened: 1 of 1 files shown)


cloud/vmware/vsphere_guest.py
```diff
@@ -689,6 +692,10 @@ def deploy_template(vsphere_client, guest, resource_pool, template_src, esxi, mo
                 cloneArgs["linked"] = True
                 cloneArgs["snapshot"] = snapshot_to_clone
 
+            if vm_extra_config.get("folder") is not None:
```
marvinpinto, line 695: I _think_ there may be an edge case here where `vm_extra_config.get("folder")` returns an empty string (for example, if the user supplied an empty string). Will this feature degrade gracefully in this scenario?

If not, another option might be `if not vm_extra_config.get("folder"):` which would evaluate as a truthy/falsey, vs specifically testing for `None`.

fennm, line 695: @marvinpinto Good catch!  I tested and an empty string results in the same behavior as None, which is that the VM is cloned into the same folder as the template.  So I would say that yes the feature does degrade gracefully.

marvinpinto, line 695: Cool :+1:

gregdek:

gregdek:
@chrrrles @rhoop This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

chrrrles:
The code looks good so I am merging for more testing and community feedback.

fennm:
Awesome, thank you!

PR #2198, opened 2015-10-04
make a text link into a actual hyperlink

hutchic:
the text link doesn't fit on some screen resolutions. Making it into sphinx hyperlink will solve that


marvinpinto:
Makes sense to me :+1:

Issue #1419, opened 2015-05-25
Yum can't downgrade packages


g--:
As of 1.9.1, the yum module can't downgrade packages.

I saw a 77e66cb01e9121e5fcbdc08e87d248c21bdad497 attempting to fix this, but it had to be reverted in d77f1976a6cbf30cd2284817e92a838f7f7e47ef .

(I don't know the story behind the revert, and I couldn't find an open bug or public discussion.. so I'm opening an issue!)

solomongifford:
We are very interested in downgrading too - or to be more specific and indempotent, when a specific version of a package is given in the ansible task and the yum module is greater than that, the ansible module should downgrade to that version.

claudiomll:
I'm also interested in this feature, downgrades are key to rollbacks. :+1: 

Update: after updating ansible to version 1.9.2 , and be sure the package I wanted to downgrade doesn't have any conflict (like i686 packages installed on x86_64 arch) I was able to perform the downgrade.

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

pporada-gl:
This would help my org out as well.

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

g--:
Dear ansibot: it appears this has been resolved based on the commits.  I haven't tested it personally though.

g--:
judging from https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py it looks like this bug/feature omission is still unresolved.

It looks like @willusher is playing with a fix on a fork.  But I'm not sure why the original feature was reverted in the first place.. so.. :shrug:

ansibot:
@ansible, ping. This issue is still waiting on your response.
[...]

PR #2199, opened 2015-10-04
Allow RPMs to be downgraded within Ansible

marvinpinto:
## Overview

Ansible is not currently able to install an older version of an RPM (if this older version was previously installed on the system) because of the way yum handles downgrading RPMs.

In order to downgrade to an older version of an RPM, one needs to use the `yum downgrade` plugin, since the intuitive `yum install <older version>` does not work in a CentOS world.

As of now, yum downgrading _will not_ work on lists of packages. This is a necessary workaround since building the logic for this will take a larger refactoring effort.
## References

This work builds on the work done in previous PRs:
- https://github.com/ansible/ansible/pull/6033
- https://github.com/ansible/ansible/pull/6333
## Testing

I decided to use a CentOS 6 Docker image to do my testing since that seemed like the easiest thing to build + have someone else verify. All the files are available here https://gist.github.com/marvinpinto/326f3f3750fac1121cf4

Running that sample playbook should demonstrate the RPM upgrading and downgrading process (a full sample output is available there)

Looks something like:

```
localhost                  : ok=44   changed=21   unreachable=0    failed=0
```
## Feedback

I would love to hear your feedback on how we could improve this module!

### DIFF
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -544,7 +544,7 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
                     f.close()
                     pkg = package
                 except Exception, e:
-                    shutil.rmtree(tempdir)
+                    delete_temporary_directory(tempdir)
                     module.fail_json(msg="Failure downloading %s, %s" % (spec, e))
 
         #groups :(
@@ -611,11 +611,7 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
 
         if module.check_mode:
             # Remove rpms downloaded for EL5 via url
-            try:
-                shutil.rmtree(tempdir)
-            except Exception, e:
-                module.fail_json(msg="Failure deleting temp directory %s, %s" % (tempdir, e))
-
+            delete_temporary_directory(tempdir)
             module.exit_json(changed=True, results=res['results'], changes=dict(installed=pkgs))
 
         changed = True
@@ -650,10 +646,7 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
         res['changed'] = changed
 
     # Remove rpms downloaded for EL5 via url
-    try:
-        shutil.rmtree(tempdir)
-    except Exception, e:
-        module.fail_json(msg="Failure deleting temp directory %s, %s" % (tempdir, e))
+    delete_temporary_directory(tempdir)
 
     return res
 
@@ -915,6 +908,12 @@ def ensure(module, state, pkgs, conf_file, enablerepo, disablerepo,
                 " failed", changed=False, results='', errors='unepected state')
 
     return res
+
+def delete_temporary_directory(tempdir):
+    try:
+        shutil.rmtree(tempdir)
+    except Exception, e:
+        module.fail_json(msg="Failure deleting temp directory %s, %s" % (tempdir, e))
 
 def main():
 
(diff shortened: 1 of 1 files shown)


des4maisons:
Seems reasonable :paw_prints: :dog:

I'm not sure what https://github.com/ansible/ansible/pull/6033#issuecomment-38474780 is referring to, but your test cases look legitimate and seem to cover the specific case mentioned in the comment.

hutchic:
looks good to me :+1:

packaging/os/yum.py
```diff
@@ -544,7 +546,7 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
                     f.close()
                     pkg = package
                 except Exception, e:
-                    shutil.rmtree(tempdir)
+                    delete_temporary_directory(tempdir)
```
ddub, line 549: This is a change of behaviour in the case that the download failed AND the directory couldn't be removed the download error would be "hidden" behind the directory removal error. I'm 20% thinking this is ok, 20% thinking this is wrong but 60% thinking that it's such a small probability that it actually happens it's not been worth writing this out.

packaging/os/yum.py
```diff
@@ -606,54 +628,56 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
...
-                module.fail_json(msg="Failure deleting temp directory %s, %s" % (tempdir, e))
+    if not pkgs:
+        delete_temporary_directory(tempdir)
+        return res
 
-            module.exit_json(changed=True, results=res['results'], changes=dict(installed=pkgs))
+    # yum downgrading will not work on lists of packages. This is a necessary
+    # workaround for now since building the logic for this will take a larger
+    # refactoring effort.
+    if downgrade and len(pkgs) == 1:
```
ddub, line 638: What's the anticipated behaviour in this circumstance? Does the yum command fail and pass that back to the module or does it silently continue with no change recorded?

marvinpinto, line 638: My thinking was that depending on the output status of this yum command, it would either fail or pass the ansible play.

https://github.com/marvinpinto/ansible-modules-core/blob/allow-rpm-downgrading/packaging/os/yum.py#L650-L652

This logic is identical to what it did before.

Does this make sense?

(Thank you for reviewing this PR @ddub!)

packaging/os/yum.py
marvinpinto, line 549: Yeah, I was thinking the same :smile: 

Anyone else think this should be called out explicitly? I am open to discussion.

packaging/os/yum.py
ddub, line 638: Yup, makes sense to me!

marvinpinto:
Any chance I could get some ansible :eyes: on this PR?

cc @gregdek @bcoca

g--:
this would close #1419

robynbergeron:
Thanks @marvinpinto for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

marvinpinto:
Thanks for letting me know @robynbergeron! I will work on rebasing this and I'll update this PR at some point during the next few weeks or so.

marvinpinto:
I apologize folks, I'm afraid I don't have the time/resources to see this through.

In testing, there were a slew of edge cases that `splitFilename` function and we decided this wasn't a route worth pursing for FreshBooks.

If anyone else would love to build off this work, I will leave the branch around so have it.

Issue #2197, opened 2015-10-03
ec2 module: launch group for spot requests


kai11:
I needed spot launch group support for my cluster, so I updated ec2 module.

kai11:
Pull request https://github.com/ansible/ansible-modules-core/pull/2211

bcoca:
closing this in favor of the PR

PR #2211, opened 2015-10-06
Added launch group support for ec2 module

kai11:
https://github.com/ansible/ansible-modules-core/issues/2197


kai11:
To test this update, add "spot_launch_group: group_name" to any spot request call.
EC2 console will show group in spot request properties.

amenonsen:
Just adding a note to say that this patch worked well for us.

Issue #868, opened 2015-02-27
AttributeError: 'NoneType' object has no attribute 'describe_alarms' when creating new ec2_metric_alarm


lostmimic:
##### ISSUE TYPE

Bug Report
##### COMPONENT NAME

ec2_metric_alarm module
##### ANSIBLE VERSION

ansible (1.8.4)
##### SUMMARY

Versions run on:
ansible (1.8.4)
boto (2.36.0)
botocore (0.93.0)

Playbook run:
    - name: Create a Cloudwatch metric alarm for up scaling and associate it with a Scaling Policy
      ec2_metric_alarm:
        name: Product 15.1.1.0.0 opsdev upscale
        description: Triggered when the CPU of a node is more than 50% for 5 minutes
        namespace: "AWS/EC2"
        metric: CPUUtilization
        comparison: ">"
        threshold: 50.0
        unit: Percent
        period: 100
        evaluation_periods: 3
        statistic: Average
        dimensions: {'AutoScalingGroupName':'Product opsdev'}
        state: present
        alarm_actions: "{{ sp_up_result.arn }}"

The sp_up_result variable is an ec2_scaling_policy task register as demonstrated in this link (http://stackoverflow.com/questions/24686407/unable-to-retrieve-aws-scaling-policy-information-from-ec2-scaling-policy-module).

Error is the following:
failed: [localhost] => {"failed": true, "parsed": false}
Traceback (most recent call last):
  File "/home/lostmimic/.ansible/tmp/ansible-tmp-1425074403.05-269505839436583/ec2_metric_alarm", line 2069, in <module>
    main()
  File "/home/lostmimic/.ansible/tmp/ansible-tmp-1425074403.05-269505839436583/ec2_metric_alarm", line 2065, in main
    create_metric_alarm(connection, module)
  File "/home/lostmimic/.ansible/tmp/ansible-tmp-1425074403.05-269505839436583/ec2_metric_alarm", line 1934, in create_metric_alarm
    alarms = connection.describe_alarms(alarm_names=[name])
AttributeError: 'NoneType' object has no attribute 'describe_alarms'

lostmimic:
Figured out the error.

The documentation http://docs.ansible.com/ec2_metric_alarm_module.html does not list region as a required parameter but the call will fail without it. (this same thing was in ec2_scaling_policy)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[...]

PR #2217, opened 2015-10-07
Add region to doc and handle missing region parameter

wimnat:


wimnat:
Resolves https://github.com/ansible/ansible-modules-core/issues/868

bcoca:
we should not be adding this to each module that needs this, it should be part of module doc fragments (guessing we need an ec2one vs aws one)

cloud/amazon/ec2_metric_alarm.py
```diff
@@ -260,7 +265,6 @@ def main():
             insufficient_data_actions=dict(type='list'),
             ok_actions=dict(type='list'),
             state=dict(default='present', choices=['present', 'absent']),
-            region=dict(aliases=['aws_region', 'ec2_region']),
```
bcoca, line 262: yes, this is redundant as we import it in ec2_argument_spec

bcoca:
fixed via c4f64d822ce5ef9300c06433ac85a0bbfee0a75c

bcoca:
actually only fixed docs, you can update to fix the code side

wimnat:
All modules in one PR?

bcoca:
not even a PR, direct commit

wimnat:
I don't have access to direct commit.  I'll update this PR to include the fixes code side for appropriate modules.

wimnat:
Also, i don't think the iam modules need region

wimnat:
Done for extras - https://github.com/ansible/ansible-modules-extras/pull/1069

wimnat:
@bcoca discarded commit for the rest of the modules.

PR #2223, opened 2015-10-07
make chdir a path so it resolves shell aliases

bcoca:
also removed this_dir logic as it is not needed, chdir is None by default and run_command can handle that.


abadger:
+1

PR #2230, opened 2015-10-08
Documentation fix for ec2_asg

adisharoon:


cloud/amazon/ec2_asg.py
```diff
@@ -112,8 +112,8 @@
     version_added: "2.0"
   wait_timeout:
     description:
-      - how long before wait instances to become viable when replaced.  Used in concjunction with instance_ids option.
-    default: 300
+      - The number of seconds to wait for instances to become healthy after they are replaced.  Used in conjunction with the replace_instances option.
+    default: 300 seconds
```
axilleas: This should be just the number, no? The way you have it is like saying

```
wait_timeout: 300 seconds
```

is a valid option.

adisharoon: @axilleas I was just copying the style that appears in the rest of the documentation for this module, e.g. for 'default_cooldown' and 'health_check_period'

axilleas: Ok, I see. @bcoca any comment on this? I would expect the default value to be the value I eventually enter when running the module. Maybe the word `seconds` should be removed alltogether since you already state that this is measured in seconds in the description.

bcoca: agreed, remove 'seconds' from default: entry, it already states the unit in the description

gregdek:
Thanks @adisharoon-socialware for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

PR #2240, opened 2015-10-10
Add min_count and max_count options to ec2 module

wrouesnel:
These two options fulfill the same purpose as the exact_count option, but allow
for loosely specifying the number of instances or bounding it.

min_count ensures at least a certain number of instances are running, max_count
ensures no more then a certain number are running.

min_count and max_count can be used together, but are mutually exclusive to
exact_count (which is functionally identical to setting both values to the same
number).

These options make performing non Amazon autoscaling much easier, and
in particular allow idempotent deploy and bootstrapping ("ensure at least 1
production machine exists" can now be written into a play.)


robynbergeron:
Hi @wrouesnel -- thanks for submitting this PR.

@tgerla @lwade -- Please review this PR to make sure it adheres to the following guidelines:

Also pinging the AWS crowd en masse for eyeballs -- @jsmartin @scicoin-project @tombamford @garethr @lorin @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @carsongee @defionscode @tastychutney @bpennypacker @loia

cloud/amazon/ec2.py
```diff
@@ -233,6 +233,20 @@
     required: false
     default: null
     aliases: []
+  min_count:
+    version_added: "2.0"
+    description:
+      - An integer value which indicates the minimum number of instances that match the 'count_tag' parameter should be running. Instances are created if less then this number are running.
+    required: false
+    default: null
+    aliases: []
```
juliedavila: Plz remove the `[]` aliases

juliedavila:
This looks good, but this begs the question, why would you want to use this instead of an auto scaling group which carried the first class citizen support from AWS?

wrouesnel:
I created this to deal with scaling up a database cluster which shouldn't really be scaled down. It was easier to setup the playbook so it would bootstrap the cluster if not there (for dev purposes - i.e. min 1) but then if I've added more nodes I want it to also run and provision those nodes separately.

It hugely simplifies giving instructions to someone else on how to use it since you can just say "run this" rather then "run this file once, then you need to run this if you want to shut it down etc. etc."

wrouesnel:
Also re: the `[]` aliases - that seems to be the pattern in most of the rest of the help txt in the file?

wimnat:
As per the module checklist, http://docs.ansible.com/ansible/developing_modules.html, empty aliases are not required

wrouesnel:
Empty aliases removed and rebased to latest devel.

gregdek:
Thanks @wrouesnel for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

cloud/amazon/ec2.py
```diff
@@ -1359,6 +1384,8 @@ def main():
             termination_protection = dict(type='bool', default=False),
             state = dict(default='present', choices=['present', 'absent', 'running', 'stopped']),
             exact_count = dict(type='int', default=None),
+            min_count = dict(type='int', default=None),
```
juliedavila, line 1387: the default of `min_count` should be 0 here as that is what the code sets

cloud/amazon/ec2.py
```diff
@@ -233,6 +233,18 @@
     required: false
     default: null
     aliases: []
+  min_count:
+    version_added: "2.0"
+    description:
+      - An integer value which indicates the minimum number of instances that match the 'count_tag' parameter should be running. Instances are created if less then this number are running.
+    required: false
+    default: null
```
juliedavila: default should be 0

juliedavila:
@wrouesnel just update the default value for `min_count` and update `version_added` to 2.1 and we should be good to go

wrouesnel:
Rebased with documentation fixes.

gregdek:
Moving to shipit per @defionscode feedback above. @wrouesnel thanks for the PR!

gregdek:
Thanks @wrouesnel for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2243, opened 2015-10-11
Fix v2 breakage: Add missing variable initializations

amenonsen:
Without this, «ec2: state=stopped instance_ids=…» would fail with a
traceback like this:

```
if inst.get_attribute('sourceDestCheck')['sourceDestCheck'] != source_dest_check:
```

NameError: global name 'source_dest_check' is not defined


jimi-c:
Merged, thanks!

PR #2244, opened 2015-10-11
allow for sshpass support for `synchronize`

ritzk:
Needs - https://github.com/ansible/ansible/pull/12704 .Also has more detailed info
- add sshpass support to sync module

PR #2247, opened 2015-10-12
add sshpass support for `synchronize`: v2

elvismacak:
like https://github.com/ansible/ansible-modules-core/pull/1755, it's a safety improvement


robynbergeron:
Hi @elvis-macak -- thanks for submitting this PR!

@tima -- Please review this PR to make sure it adheres to the following guidelines:

robynbergeron:
Also: Not sure what is up with this one and #2244 -- trying to sort it out by reading #1755 but not getting very far just yet. If @elvis-macak @ritzk @tima and others can just make sure to indicate if anything needs closing in favor of other PRs, that would be great! :) Thanks, everyone.

files/synchronize.py
```diff
@@ -345,15 +352,22 @@ def main():
...
+        ssh_opts = '%s -o Port=%d' % (ssh_opts, dest_port)
+
+    if private_key:
+        cmd += " --rsh 'ssh %s %s'" % (private_key, ssh_opts)
+    elif ssh_passwd:
+        import os
+        sshpass_pipe = os.pipe()
+        os.write(sshpass_pipe[1], "{0}\n".format(ssh_passwd))
+        map(os.close, sshpass_pipe)
+        cmd += " --rsh 'sshpass -d %d ssh -l ass %s'" % (sshpass_pipe[0], ssh_opts)
```
ritzk, line 370: - Does this work reliably ?  run_command ( b/lib/ansible/module_utils/basic.py +1591) creates a pipe and writes to it, and this (stdin fd ) number can be pretty random.
  
  ```
      st_in = subprocess.PIPE
  ```
- Is there a reason to set username to ass ?

files/synchronize.py
```diff
@@ -261,6 +266,7 @@ def main():
             partial = dict(default='no', type='bool'),
             verify_host = dict(default='no', type='bool'),
             mode = dict(default='push', choices=['push', 'pull']),
+            ssh_passwd = dict(default=None),
```
ritzk, line 269: Update this to include no_log option, and you might want to standardize to login_password var name.
ansible uses login_password heuristics filter to ensure that it is not logged.

Will allow this to be used with https://github.com/ansible/ansible/pull/12704 .

files/synchronize.py
```diff
@@ -164,6 +164,11 @@
     default: no
     required: false
     version_added: "2.0"
+  ssh_passwd:
+    description:
+      - Tells rsync to use sshpass with password
+    required: false
+    version_added: "2.0"
```
sivel, line 171: I still believe this should not be a module argument, at least not one that accepts an actual password.  Since we know the password already if we are performing all other actions via password, we should be able to inject this via the `synchronize` action plugin in ansible itself.

ritzk:
@sivel - This is done by https://github.com/ansible/ansible/pull/11657

## 

And I use this (login_password) in https://github.com/ansible/ansible-modules-core/pull/2244 , with https://github.com/ansible/ansible/pull/12704 ( for inject)

files/synchronize.py
elvismacak, line 370: <pre>
elvis@dev-elvis:~$ sshpass -h
Usage: sshpass [-f|-d|-p|-e] [-hV] command parameters
   -f filename   Take password to use from file
   -d number     Use number as file descriptor for getting password
   -p password   Provide password as argument (security unwise)
   -e            Password is passed as env-var "SSHPASS"
   With no parameters - password will be taken from stdin

   -h            Show help (this screen)
   -V            Print version information
At most one of -f, -d, -p or -e should be used
</pre>


though it's random fd, but sshpass can read the password from the pipe.

files/synchronize.py
elvismacak, line 171: @sivel   as @ritzk said, it was injected by https://github.com/ansible/ansible/pull/11657. 
but not sure  module argument can be used like this, if not, then my PR should be closed, it gets some bad styles.

elvismacak:
@ritzk , there is a potential harm to use env in `subprocess.Popen`
if you don't pass env, the new process can inherit all the env from it's parent process,  but when you pass  a env, it'll override it's parent, and use the new `env` variable you passed in. they are exclusive.

more or less read the ssh_pass from env seems easier than a pipe

ritzk:
@elvis-macak  Thanks for catching this. Now fixed in https://github.com/ansible/ansible/pull/12704/files .

files/synchronize.py
ritzk, line 370: This does not work on my system

```
hosts: 192.168.56.102
staging_server: 192.168.56.101
- synchronize: src=/etc
               dest=/tmp
               login_password=true
  delegate_to: "{{ staging_server }}"
```

From ps

```
$ ps aux|grep sshpass
   sshpass -d 4 ssh -S none -o StrictHostKeyChecking=no -l root 192.168.56.102 rsync --server -logDtprze.iLs --log-format=%i --delay-updates . /tmp
$  file /proc/pid_of_sshpass/fd
0: broken symbolic link to `socket:[13047]'
1: broken symbolic link to `socket:[13050]'
2: broken symbolic link to `pipe:[13045]'
3: symbolic link to `/dev/ptmx'
4: symbolic link to `/dev/tty'
5: symbolic link to `/dev/pts/1'
```

## 

The most reliable way to fix this would be to use env variables, or to fix run_command function to support passing fd for input.

```
$ vim   lib/ansible/module_utils/basic.py
1700         if data:
1701             st_in = subprocess.PIPE
```

This creates pipe for input to write data. This is usually <0>, but this can't be guaranteed. 

Alternatively, you could pass this fd no sshpass to read from.

```
$ vi lib/ansible/modules/core/files/synchronize.py
369         cmd += " --rsh 'sshpass -d 0 ssh %s'" %  ssh_opts)
...
394     if login_password:
395         (rc, out, err) = module.run_command(cmd, data = login_password)
396     else:
```

bcoca:
closing both these tickets as they are flawed implementations, opening a feature request and a discussion on how to do this securely and in a way that functions for all

PR #2249, opened 2015-10-12
find.py: add the ability to search file name based on regex

ritzk:
pushed https://github.com/ansible/ansible-modules-core/pull/1590 to a separate branch


files/find.py
```diff
@@ -153,12 +155,19 @@
...
 
 def pfilter(f, patterns=None):
-    '''filter using glob patterns'''
     if patterns is None:
         return True
-    for p in patterns:
-        if fnmatch.fnmatch(f, p):
-             return True
+        
+    if type(patterns) == type([]):
```
bcoca, line 161: patterns are 'list' in the spec, no need to check here again as we always get a list

files/find.py
```diff
@@ -276,6 +285,12 @@ def main():
         else:
             module.fail_json(size=params['size'], msg="failed to process size")
 
+    patterns = params['patterns']
+    if patterns and patterns[0].startswith("~"):
+        patterns = re.compile(patterns[0][1:])
```
bcoca, line 290: im not sure if I'm reading this correctly, if the first pattern starts with ~ all patterns are considered regexes?

abadger:
Probably better to have a separate parameter for regex and fnmatch patterns.  Could implement them as default=None.  That way if we have None for that type of pattern, we skip filtering by that test altogether.

ritzk:
@bcoca 
There is no way to safely differentiate between glob and regex  pattern, as mentioned earlier.

  pattern -> .*.txt
  glob -> find hidden txt file
  regex -> find files with txt extension .

as per your suggestion of overloading patterns op using "~" , I have implemented this code. Personally, I would opt for pattern (for regex), and patterns for glob.

## 

patterns would be set to regex type if the prefix is ~
otherwise, it would be a list.

bcoca:
ok, since it seems easier for me to implement what i meant than communicate it effectively, i opened #2262

ritzk:
fair enough. closing this for #2262

PR #2253, opened 2015-10-12
Allow nics to be a string in os_server

### LINKED ISSUE #2231: nics parameter in os_servers module should not be fixed to 'list' type
Reference:  https://github.com/ansible/ansible-modules-core/blob/devel/cloud/openstack/os_server.py#L438

nics parameter to this module can be accepted as either a list:
nics:
         - net-id: 34605f38-e52a-25d2-b6ec-754a13ffb723
         - net-name: another_network
or a string:
nics: "net-id=4cb08b20-62fe-11e5-9d70-feff819cdc9f,net-id=542f0430-62fe-11e5-9d70-feff819cdc9f..."

By defining nics as a list type, this breaks the string parsing for nics.  The parameter is then received in this format:
nics: ["net-id=4cb08b20-62fe-11e5-9d70-feff819cdc9f", "net-id=542f0430-62fe-11e5-9d70-feff819cdc9f...", ...]

Since this module supports both list and string type for this parameter, the 'list' type should be omitted.

omgjlk:
The module docs and code handle comma separated strings as well as
actual lists. Making the argument a list type turns the string into a
single entry list which is undesireable.

Resolves #2231

### DIFF
--- a/cloud/openstack/os_server.py
+++ b/cloud/openstack/os_server.py
@@ -461,7 +461,7 @@ def main():
         flavor_include                  = dict(default=None),
         key_name                        = dict(default=None),
         security_groups                 = dict(default=['default'], type='list'),
-        nics                            = dict(default=[], type='list'),
+        nics                            = dict(default=None),
         meta                            = dict(default=None),
         userdata                        = dict(default=None),
         config_drive                    = dict(default=False, type='bool'),


bcoca:
I'm not sure I follow, the list type is so you are guaranteed you always get a list and don' t have to worry with detection and/or splitting

we can actually remove this code:
https://github.com/j2sol/ansible-modules-core/blob/nics-as-string/cloud/openstack/os_server.py#L278

omgjlk:
Guess I need to do some more testing on this issue. 
- jlk

larsks:
I just ran into this problem.  When called as an ad-hoc command, like this:

```
ansible all -i localhost, -c local -m os_server -a \
  'name=cirros
  image=cirros-0.3.4-x64 flavor=m1.tiny
  nics=net-id=70014eda-8767-4c1d-9d16-37252deea978'
```

The `_network_args` functions up getting `module.params['nics']` set
to:

```
['nics=net-id=70014eda-8767-4c1d-9d16-37252deea978']
```

That is, it's a _list_ containing a _string_.  The way the code is
currently written...

```
if type(nics) == str :
    for kv_str in nics.split(","):
        nic = {}
        k, v = kv_str.split("=")
        nic[k] = v
        args.append(nic)
else:
    for net in module.params['nics']:
        if net.get('net-id'):
            args.append(net)
```

...if the `nics` param is not a string then it assumed
to be a list of dictionaries, which leads to this failure:

```
AttributeError: 'str' object has no attribute 'get'
```

emonty:
shipit

gregdek:
Well, shoot. Between now and @emonty saying "shipit", this got out of sync and needs to be rebased. Sorry @j2sol -- if you can rebase, we'll get it back into shipit asap. Comment again with "shipit" after rebase and we'll take care of it.

omgjlk:
shipit

gregdek:
Thanks @j2sol, sorry for the delay. Hope you're enjoying Japan. :)

Marked for inclusion.

larsks:
This is now a no-op, right? I think the code submitted in #2275 addressed this as well.

bcoca:
would this not counter act the fix in #2275 at this point?

omgjlk:
Looks like you're right. 2275 keeps the type as a list but allows for a single item list that may be comma separated.

Issue #2257, opened 2015-10-13
Docker 1.8.3 is broken under certain conditions


ghost:
The API now returns different json data depending on the conditions of the variable. Specifically if the variables are empty it will not return them.

This change happened between 1.8.2 and 1.8.3. In a commit[1] now ExposedPorts will not return if it is None (or empty). This effectively kills the module since that is hardcoded. This is the only variable that is affected by this patch.

[1] https://github.com/docker/docker/commit/9098628b2901ae8585ba4c66ee6e14759d2119da

Glueon:
I confirm this bug. Here is a traceback:

```
Traceback (most recent call last):  
  File "/root/.ansible/tmp/ansible-tmp-1444786751.04-168544313354701/docker", line 3185, in <module>  
    main()  
  File "/root/.ansible/tmp/ansible-tmp-1444786751.04-168544313354701/docker", line 1542, in main  
    reloaded(manager, containers, count, name)  
  File "/root/.ansible/tmp/ansible-tmp-1444786751.04-168544313354701/docker", line 1419, in reloaded  
    for container in manager.get_differing_containers():
  File "/root/.ansible/tmp/ansible-tmp-1444786751.04-168544313354701/docker", line 933, in get_differing_containers
    expected_exposed_ports = set((image['ContainerConfig']['ExposedPorts'] or {}).keys())
KeyError: 'ExposedPorts'
```

robynbergeron:
Hi @SamYaple -- thanks for filing this bug report (and the accompanying PR in #2258 :D). Also, thanks to @glueon for confirming! 

@cove @softzilla @joshuaconner @smashwilson -- thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module.

Fei-Guang:
ansible 1.9.4  still has no this patch?  

i still got this error

Fei-Guang:
i've installed ansigle on ubuntu using 

$sudo pip install -U ansible
$ ansible --version
ansible 1.9.4
  configured module search path = None

Fei-Guang:
have to install ansible from source

ghost:
@Fei-Guang This was backported but there has been no new 1.9.x tag cut. I have been told there will not be either. This fix will only be available in ansible 2.0

bcoca:
there will be a 1.9.5 release, if the patch made it into the stable-1.9 tree it will be in that release

ghost:
[...]

PR #2258, opened 2015-10-13
Don't assume ExposedPorts exists (bug #2257)

ghost:
A recent change [1] in docker between v1.8.2 and v1.8.3 changed what
is returned in the json when inspecting an image. Five variables which
could have been expected before will now be omited when empty. Only
one of those variables is being addressed in the docker, ExposedPorts.

Unfortunately there was also no API version change on this so this
can't be easily corrected with pinning the API to the older version.

This does a get() which will return None if the variable is not in the
dict formed from the json that was returned. Everything else works the
same way.

[1] https://github.com/docker/docker/commit/9098628b2901ae8585ba4c66ee6e14759d2119da

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1062,7 +1062,7 @@ class DockerManager(object):
                     continue
 
             # EXPOSED PORTS
-            expected_exposed_ports = set((image['ContainerConfig']['ExposedPorts'] or {}).keys())
+            expected_exposed_ports = set((image['ContainerConfig'].get('ExposedPorts') or {}).keys())
             for p in (self.exposed_ports or []):
                 expected_exposed_ports.add("/".join(p))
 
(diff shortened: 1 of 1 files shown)


brk3:
+1

nhlfr:
+1

robynbergeron:
Greetings, @SamYaple :)  Thanks for filing this PR, and for filing the initial bug report.  #2257 

@cove @smashwilson @softzilla @joshuaconner -- Please review this PR to make sure it adheres to the following guidelines:

robynbergeron:
Also -- @brk3 @nihilifer @magnars @Glueon -- thanks for confirming that this fix is helping you out, much appreciated!

mandre:
I was just hit by this bug, I'm going to stick with docker 1.8.2 for now.

Like @SamYaple said, it's a safe change that ensures consistent behavior between different versions of docker.

PR #2262, opened 2015-10-13
added regex support to find

bcoca:
Also added 'singular' aliasess to patterns and paths

### DIFF
--- a/files/find.py
+++ b/files/find.py
@@ -25,8 +25,6 @@ import stat
 import fnmatch
 import time
 import re
-import shutil
-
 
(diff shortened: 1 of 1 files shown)


abadger:
-1 I really think we should have separate parameters for regexes and shell globs.  Combining them into a single parameter is a bad standard to set.

abadger:
Here's my alternative implementation: https://github.com/ansible/ansible-modules-core/pull/2263

robynbergeron:
[module: files/find.py]

abadger:
+1  Looks good to me!

PR #2263, opened 2015-10-13
Alternative to #2262 that uses two parameters

abadger:


bcoca:
this includes hidden files by default, also it 'ANDS' the patterns when they are supposed to be 'ORd'

bcoca:
i don't think we need to diff options in this case, both are patterns and both should work transparently, you in the other PR we are just flagging some patterns as regexes to use a diff matching than globbing. I think it makes for a better user interface.

abadger:
No, they're filters.  So it's supposed to be AND'd.

EDIT: I'm dumb nad doing too many things at once.  Yes, they should be OR'd.

abadger:
Taking the time to examine this -- hidden files are okay as they're taken care of by the hidden parameter, not by the glob pattern.

abadger:
- Corrcted example documentation
- Corrected the regex filter

Had a conversation with @bcoca about two parameters v2 one and AND vs OR.  Came to the conclusion that (1) AND and OR are linked to whether it's two parameters or one.  We agree that if it's two parameters than it is an AND; if it is one parameter then it's an OR.

We do not agree about whether there should be two parameters or one.  I feel that one parameter is bad UI because you can't tell with certainty whether a pattern is a regex or a glob.  Using a magic marker character means that you have to switch to a regex if you try to start your shell-style pattern with that character.

bcoca contends that having separate parameters for the patterns is bad UI as they're both patterns and to him it then becomes confusing that the patterns are AND'd rather than OR'd.

Latest code in this PR marks the two parameters as mutually exclusive as the only thing I can think of as a compromise between the two.  If you want regexs, use regexs and if you want shell style use shell style.  Don't mix the two.

ritzk:
imho, 
- use pattern for regex pattern
- use patterns (list) for glob

regex supports multiple argument using pipe(|)

abadger:
@ritzk that would be backwards incompatible but find was added for 2.0 so I suppose it would be possible.  I find that overloading of the parameter to be bad UI too, though.  In ansible, if a list value is given a single string in the playbook, that string is made into a one element list so people are not used to specifying a list explicitly being different than specifying a string.  (note: this was also a problem with your other PR.  The argspec specified it was a list so it was automatically made a list.  So your conditional never found a string there.)

Also, one of the nice things about regex as a list is that you don't have to use ( | ) to specify multiple patterns... nesting of various patterns gets very complex very quickly.

robynbergeron:
@abadger -- thanks for the PR :D (/me appreciates the transparency and all that! :D)

@ritzk and @abadger -- 

Just trying to determine where this is at so @gregdek and I can apply labels and nag or hug as needed. :)

I think maybe @abadger is waiting on feedback from @ritzk -- so just going to keep this in community_review for now.

robynbergeron:
Also I am guessing that this and #2262 are ... in competition? Either way: Merge one or the other but not both, yes? @bcoca @abadger (not that either of you would do both, lol)

robynbergeron:
[module: files/find.py]

bcoca:
so I updated #2262 to take both regexes and shell globs transparently, what do you guys think?

cc @abadger @ritzk

abadger:
The current implementation of #2262 looks good to me so I'm closing this one.

Issue #1901, opened 2015-08-11
error when using profile under route53


glnds:
**Issue Type:** Bug Report

**Ansible Version:**  1.9.2

**boto version:** 2.38.0

**Ansible Configuration:** nothing special

**Environment:** OX 10.10.4

**Summary:**

When specifying **profile** and **region** the playbook throws an error. If using AWS environment variables to connect to AWS the playbook runs fine.

**Steps To Reproduce:**

playbook:

```
# vim:ft=ansible:

---

- name: Set DNS Record for New Instance
    route53:
      profile: "{{ profile }}"
      region: "{{ region }}"
      command: create
      zone: example.be
      record: ansible.example.be
      type: A
      ttl: 7200
      value: "{{ eip.public_ip }}"
```

**Expected Results:**

```
TASK: [Set DNS Record for New Instance] ***************************************

changed: [localhost]

PLAY RECAP ********************************************************************
```

**Actual Results**:

```
TASK: [Set DNS Record for New Instance] ***************************************
failed: [localhost] => {"failed": true, "parsed": false}
Traceback (most recent call last):
  File "/Users/glnd/.ansible/tmp/ansible-tmp-1439288241.09-151779888771650/route53", line 2147, in <module>
    main()
  File "/Users/glnd/.ansible/tmp/ansible-tmp-1439288241.09-151779888771650/route53", line 227, in main
    ec2_url, aws_access_key, aws_secret_key, region = get_ec2_creds(module)
  File "/Users/glnd/.ansible/tmp/ansible-tmp-1439288241.09-151779888771650/route53", line 2100, in get_ec2_creds
    region, ec2_url, boto_params = get_aws_connection_info(module)
  File "/Users/glnd/.ansible/tmp/ansible-tmp-1439288241.09-151779888771650/route53", line 2087, in get_aws_connection_info
    if not boto_supports_profile_name():
  File "/Users/glnd/.ansible/tmp/ansible-tmp-1439288241.09-151779888771650/route53", line 2018, in boto_supports_profile_name
    return hasattr(boto.ec2.EC2Connection, 'profile_name')
AttributeError: 'module' object has no attribute 'ec2'


FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
```

Seraf:
Hello, same problem here.

Subscribing

glnds:
Same problem with the iam_policy module. It seems to be a more general problem with the ec2 modules.
if i use:

```
[...]

PR #2272, opened 2015-10-14
Fix route53 issue when boto profile used.

jonhadfield:
This is a fix for issue #1901 which occurs when a profile is specified.
The route53 module imports the ec2 utils:
`from ansible.module_utils.ec2 import *`
Which has a method that checks for boto support calling:
`return hasattr(boto.ec2.EC2Connection, 'profile_name')`
Although boto IS imported, it's the boto.ec2 module that contains the EC2Connection class, so it will fail with:
`AttributeError: 'module' object has no attribute 'ec2'`
The PR adds the necessary import to the route53.py module to be consistent with others.


robynbergeron:
Hi, @jonhadfield -- thanks for this PR. 

@bpennypacker -- Please review this PR to make sure it adheres to the following guidelines:

Note that a few folks have reported seeing this issue in #1901, so it would be great to get fixed up. Thanks! :D

willthames:
I've suggested an alternative fix on #2405. Such a fix would address this for all modules, not just piecemeal. Props to @twaggs for spotting the link.

gregdek:
So @willthames is it your opinion that we should close this PR in favor of #2405?

willthames:
@gregdek, that is indeed my opinion

mscherer:
remove needinfo, since the info was provided by willthames

gregdek:
Closing in favor of #2405.

PR #2275, opened 2015-10-14
fix handling of nics argument

larsks:
The existing code was receiving a list of strings and erroneously
assuming it was being given a list of dictionaries, leading it to fail
with:

```
AttributeError: 'str' object has no attribute 'get'
```

This commit corrects the list handling code to check the type of each
item and handle it appropriately.   Also, based on bcoca's comment
in #2253, thie code removes the special case for a string-only argument.

### DIFF
--- a/cloud/openstack/os_server.py
+++ b/cloud/openstack/os_server.py
@@ -84,8 +84,8 @@ options:
         - A list of networks to which the instance's interface should
           be attached. Networks may be referenced by net-id/net-name/port-id
           or port-name.
-        - 'Also this accepts a string containing a list of net-id/port-id.
-          Eg: nics: "net-id=uuid-1,net-id=uuid-2"'
+        - 'Also this accepts a string containing a list of (net/port)-(id/name)
+          Eg: nics: "net-id=uuid-1,port-name=myport"'
      required: false
      default: None
    auto_ip:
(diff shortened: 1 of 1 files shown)


larsks:
Also, how do we feel about testing?  I created https://github.com/larsks/ansible-modules-core/blob/bug/2253-tests/tests/unit/cloud/openstack/test_os_server.py in order to validate the changes I was making to this code, but I since there weren't any other tests here I wasn't sure whether or not I should include it.

emonty:
@larsks I love the idea of getting some actual unit tests in the ansible tree. FWIW - we've been talking recently about getting a comprehensive set of fakes in the shade library itself similar to what you have here - but also containing real dicts that we've fetched from the various divergent clouds. SO - I'd say let's land your tests - and then once we have fakes in shade, we can write more like it.

emonty:
shipit

larsks:
@emonty Sounds good.  In order to keep things simple, I'll wait for this PR to land first, and then I'll submit the tests as a new PR.

larsks:
@emonty Actually, I went ahead and submitted #2286, which adds the tests (which will fail until this PR is merged).

robynbergeron:
Hi guys! :)

@larsks -- thanks for the PR. And the others. And glad to see you are settling in nicely in core and extras :)

I started reading this and was thinking.... wow, I should ping @emonty so you guys can chat. And then I kept reading and HOORAY you guys are already doing All The Things. :)

Marking for inclusion as this has already been reviewed by @emonty.

robynbergeron:
@ansible/core team -- this is ready for shipping. Note that inclusion of this will permit #2286 to pass :)

Thanks! :)

PR #2303, opened 2015-10-19
MariaDB instances can be created using RDS module

pnuz3n:
Amazon released new fully managed database engine, MariaDB. Ansible`s RDS module could not be used to create this database type because "MariaDB" was not recognised as a valid value.

See http://www.businesswire.com/news/home/20151007006313/en#.VhV3uUa7k0W


pnuz3n:
This PR should be deleted. The author email is wrong in the commit.

Issue #2176, opened 2015-09-30
Support SELinux volume modes.


jfindley:
##### ISSUE TYPE

Feature Idea
##### COMPONENT NAME

docker module
##### ANSIBLE VERSION

N/A
##### SUMMARY

Since docker 1.7, two additional volume modes are supported on top of 'ro' and 'rw': 'z' and 'Z'.  These set the SELinux mode of the volume to be accessible from every container or just this container, respectively.

See http://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/ for more info on this.

The docker-py does seem to support this option - setting mode: Z works fine in my testing, but ansible is coded to only accept 'ro' or 'rw'.

robynbergeron:
Hi @jfindley -- thanks for filing this feature request. 

@cove @softzilla @smashwilson @joshuaconner -- pinging you to let you know about this feature request for the docker module.

robynbergeron:
Also -- pinging my lovely pals from the land of Atomic,  @jasonbrooks @dustymabe @nzwulfin -- in case any of you are interested in maybe making this happen or know someone who might be :)

robynbergeron:
[module: cloud/docker/docker.py]

dustymabe:
@robynbergeron Thanks for thinking of me. I don't have time right now to look at this but I'm interested in having this work in the future so I may be back soon!

robynbergeron:
Hi @jfindley -- note that #2312 looks like it may be a fix for this -- it would be great to get your feedback there. 

@dustymabe -- thanks for the comment, just figured I'd give y'all a heads-up! :)

alvaroaleman:
I think this can be closed, as a patch to add this feature is merged into devel already.

ansibot:
@cove, @softzilla, @zfil, @ThomasSteinbach, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@cove, @softzilla, @zfil, @ThomasSteinbach, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@jfindley We appreciate you taking the time to file this issue, but the module you've opened the issue about has been deprecated.

[...]

PR #2312, opened 2015-10-19
Add support for :z flag on Docker volumes

alvaroaleman:
This pr adds support for the `:z` flag on Docker volumes.

This is used to automatically set the secontext on selinux enabled systems. More details on this can be found [here](http://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/).

~~The link above also states there is a `:Z` mode available. This is not the case on my up-to-date CentOS7 host, which is why I didn't implement that feature.~~
`:Z` does work fine on CentOS 7


robynbergeron:
Hi @alvaroaleman -- thanks for this PR. (And hey: it looks like it will fix an outstanding issue, #2176, which is awesome!)

@cove @softzilla @smashwilson @joshuaconner -- Please review this PR to make sure it adheres to the following guidelines:

robynbergeron:
@alvaroaleman -- might be worth looking at #2176 just for info -- it looks like @jfindley was possibly able to use the :Z option, so maybe it's something you can collectively wrap your minds around... :)

Thanks! :D

jfindley:
You might need to use something a bit more modern to get :Z -- latest fedora version works fine for example.

alvaroaleman:
Okay, added the `:Z` flag which does indeed work on my Fedora. I'm currently testing what happens if you use those flags when they are not supported.

alvaroaleman:
Well, apparently I did something wrong when trying this out on my CentOS 7 box, because the upercase `:Z` does indeed work there. So after all this means:
- CentOS 7: both `:z` and `:Z` work
- Fedora 23: both `:z` and `:Z` work
- Ubuntu Trusty: Both params work, but don't have any effect

But: If the folder does not exist on the Docker host, you get a `404` exception. This is different from `ro` or `rw` which both cause the folder to be created in that case.

This is however an issue in Docker itself, because a `docker run -it -v /nonexistant:/tmp:Z debian:jessie bash` causes a `Error response from daemon: no such file or directory`.
For this, I'll create an issue at the Docker project.

Can you test if my implementation solves your issue, @jfindley ?

cloud/docker/docker.py
```diff
@@ -624,14 +624,14 @@ def __init__(self, module):
...
                     if len(parts) == 3:
-                        if parts[2] not in ['ro', 'rw']:
-                            self.module.fail_json(msg='bind mode needs to either be "ro" or "rw"')
+                        if parts[2] not in ['ro', 'rw', 'z', 'Z']:
+                            self.module.fail_json(msg='bind mode needs to be "ro", "rw", "z" or "Z"')
                         else:
-                            ro = parts[2] == 'ro'
-                    self.binds[parts[0]] = {'bind': parts[1], 'ro': ro }
+                            mode = parts[2]
+                    self.binds[parts[0]] = {'bind': parts[1], 'mode': mode }
```
smashwilson, line 634: I suspect this will break the module for older Docker versions.

Do you know when `ro` was changed to `mode` in the Docker API? I'm not finding anything in the release notes. We might be able to gate this on the API version, which would be awkward but at least keep things working.

lerenn:
Thanks @alvaroaleman, for proposing this feature ! (I'm looking forward issue #2176 resolution).

cloud/docker/docker.py
alvaroaleman, line 634: What do you mean? Before my patch, `ro` was passed as a boolean between the part in line 613-636 and the part in line 1196-1205, but then passed as `String mode` to Docker on line 1205

cloud/docker/docker.py
```diff
@@ -1193,23 +1193,11 @@ def get_differing_containers(self):
...
-                    expected_binds.add("{0}:{1}:{2}".format(host_path, container_path, mode))
+                    expected_binds.add("{0}:{1}:{2}".format(host_path, config['bind'], config['mode']))
 
             actual_binds = set()
             for bind in (container['HostConfig']['Binds'] or []):
-                if len(bind.split(':')) == 2:
-                    actual_binds.add(bind + ":rw")
-                else:
-                    actual_binds.add(bind)
+                actual_binds.add(bind + ":" + bind.split(':')[1])
```
alvaroaleman: As a colleague pointed out, this is plain broken. I'll revert it.

alvaroaleman:
If someone got some spare time, I'd be really glad if this could get reviewed @cove @joshuaconner @softzilla @smashwilson @ThomasSteinbach

lerenn:
Don't want to bother you, but if this could get reviewed. Thank you :)
@cove @joshuaconner @softzilla @smashwilson @ThomasSteinbach

alvaroaleman:
At the first glance #2455 looks like the better implementation for me, since it includes all Z modes, whereas mine "only" includes the two documented ones.

alvaroaleman:
If I had the feeling anyone who can merge stuff/give a "lgfm" would care id go ahead and actually test @solidm implemenation. But the (no) reactions so far make me not want to invest any more time into this.

bcoca:
@alvaroaleman so I'm going to close this for now unless that implementation happens to not be as good, we can always reopen this one

PR #2326, opened 2015-10-21
Mark this as a string so it is rendered in the docs

jibsheet:
When this was treated as a boolean, sphinx was leaving the Default
column on http://docs.ansible.com/ansible/ec2_module.html blank,
implying it would use AWS's default.  In reality, it passes False, which
overrides the defaults at AWS (it's possible to boot an instance which
AWS claims will always have EBS optimization without it because of this
silently passed False).

Tested by running make webdocs in a local checkout and looking at ec2_module.html
Note that other docs use default: true which don't have this problem.  I have not looked into the doc building process in depth.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -225,7 +225,7 @@ options:
     description:
       - whether instance is using optimized EBS volumes, see U(http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html)
     required: false
-    default: false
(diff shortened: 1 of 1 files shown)


gregdek:
@jsmartin @scicoin-project @tombamford @garethr @jarv @jsdalton @silviud @adq @zbal @zeekin @willthames @lwade @carsongee @defionscode @tastychutney @bpennypacker @loia

juliedavila:
shipit

Issue #2260, opened 2015-10-13
Stat module does not return sha256


ryansydnor:

robynbergeron:
[module: files/stat.py]

robynbergeron:
Hi @SL4YT1M3 -- Thanks for submitting this feature request. It might be helpful to elaborate more on why you think this would be useful (even if it seems obvious... at least to be clear :D)

@bpennypacker - Pinging you to bring this feature request to your attention. Please note that you are under no particular obligation to add this feature.

We encourage all Ansible community members to submit pull requests to satisfy feature requests. Consult the module guidelines (http://docs.ansible.com/ansible/developing_modules.html) for more information.

ryansydnor:
To sum it up, a sha256 is more trustworthy than a sha1. (source: https://sites.google.com/site/itstheshappening/). 

What I want is to use the stat module to do a comparison of a file's actual sha256 vs. an expected sha256. 

The get_url module is already capable of this behavior, so the actual implementation of this is minimal (I've already completed a little spike and should have a PR opened shortly). Given this fact, I'm considering implementing similar behavior on the s3 module as well (which is where my above file originated).

ryansydnor:
PR: https://github.com/ansible/ansible-modules-core/pull/2327

ryansydnor:
PR merged. Closing issue.

PR #2327, opened 2015-10-21
Add capability for stat module to use more hash algorithms

ryansydnor:
Specifically, the stat module now has a checksum_algorithm parameter.
This lets the module utilize one of the hash algorithms available on the host
to return the checksum of the file.

This change is backwards compatible. The checksum_algorithm defaults to
sha1 and still returns its result to the stat.checksum property.

I contemplated using the "get_checksum" parameter to specify the algorithm used, but opted to be explicit rather than overload an existing interface.

### DIFF
--- a/files/stat.py
+++ b/files/stat.py
@@ -42,11 +42,18 @@ options:
     aliases: []
   get_checksum:
     description:
-      - Whether to return a checksum of the file (currently sha1)
+      - Whether to return a checksum of the file (default sha1)
     required: false
     default: yes
     aliases: []
     version_added: "1.8"
+  checksum_algorithm:
+    description:
+      - Algorithm to determine checksum of file. Will throw an error if the host is unable to use specified algorithm.
+    required: false
+    choices: [ 'sha1', 'sha224', 'sha256', 'sha384', 'sha512' ]
+    default: sha1
+    version_added: "2.0"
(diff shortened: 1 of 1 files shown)


ryansydnor:
related issue: https://github.com/ansible/ansible-modules-core/issues/2260

gregdek:
@bpennypacker Please review this PR to make sure it adheres to the following guidelines:

tima:
I'm not @bpennypacker, but for what it's worth I think this is a useful function worth inclusion. 

I spoke to @ryansydnor about this offline. I'm wondering if `checksum_algorithm` should be folded in to `get_checksum` where the latter takes non-boolean values like "sha512" etc while a "yes" (true) value == "sha1" and "no" still means don't return a checksum. In some ways I see that streamlining the use of the module and in others I see it complicating it's code. 

Food for thought on this PR. Thanks for submitting.

bpennypacker:
Sorry, been very busy of late.  This looks good to me.  shipit.

tima:
@bpennypacker: Any thoughts on folding the `checksum_algorithm` in? @bcoca? @abadger?

bcoca:
i like as is, making a boolean into a enum/boolean is error prone and turns ugly fast

tima:
fair enough @bcoca.

PR #2332, opened 2015-10-22
Set default value of `purge` to use `yes`

wpg4665:
Force purging of configuration files if the module state is set to I(absent) by default to allow for default purging (a desired outcome in my opinion) from the generalized package.py module.


bcoca:
This is not the expected default, it will 'surprise' people that were using the module previously and new users as it is also not the default behavior of apt.

PR #2341, opened 2015-10-22
simpler way to check if systemd is the init system

reactormonk:
According to
http://www.freedesktop.org/software/systemd/man/sd_booted.html
check if the directory /run/systemd/system/ exists.

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -403,22 +403,7 @@ class LinuxService(Service):
                 self.svc_initscript = initscript
 
         def check_systemd():
-            # verify systemd is installed (by finding systemctl)
-            if not location.get('systemctl', False):
-                return False
-
-            # Check if init is the systemd command, using comm as cmdline could be symlink
-            try:
-                f = open('/proc/1/comm', 'r')
-            except IOError, err:
-                # If comm doesn't exist, old kernel, no systemd
-                return False
-
-            for line in f:
-                if 'systemd' in line:
-                    return True
-
-            return False
+            return os.path.exists("/run/systemd/system/")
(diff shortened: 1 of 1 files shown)


gregdek:
@reactormonk Thanks for your pull request. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

bcoca:
is this applicable to older versions? most of our issues stem from how systemd has changed over time and made it hard to recognize, will it also work inside containers?

reactormonk:
Yes it will work inside containers. The older version checks for cgroups, but has a false positive in case of containers. https://github.com/systemd/systemd/commit/66e4118

Explicit creation of `/run/systemd/system` in https://github.com/systemd/systemd/commit/3a90ae0 - the directory will also be implicity created after startup.

Earlier name: `/dev/.run/systemd/system`. https://github.com/systemd/systemd/commit/2b583ce6576d4a074ce6f1570b3e60b65c64ae7d

eliasp:
SaltStack uses the same approach:

``` Python
            # This check does the same as sd_booted() from libsystemd-daemon:
            # http://www.freedesktop.org/software/systemd/man/sd_booted.html
            if os.stat('/run/systemd/system'):
                context['systemd.sd_booted'] = True
```

reactormonk:
Exposing `sd_booted` as a fact would be nice, but I don't know how to do that.

bcoca:
I"m going to merge this but I'm going to restore the other checks in case this one fails

PR #2343, opened 2015-10-23
allow empty description attribute for os_security_group

larsks:
The `os_security_group` module would fail if there was no `description:`
attribute:

```
localhost | FAILED! => {
    "changed": false,
    "failed": true,
"msg": "Error creating security group larstest: Invalid input for
description. Reason: 'None' is not a valid string."
}
```

This commit makes the default description `''` rather than `None`.


gregdek:

Shrews:
shipit

PR #2344, opened 2015-10-23
allow os_server to accept a list of security groups

larsks:
This commit allows the `security_groups` parameter of the `os_server`
module to be either a YAML list or a common-delimited string (much like
the `nics` attribute).  E.g., this:

```
- os_nova_server:
    [...]
    security_groups:
      - default
      - webserver
```

Or this:

```
- os_nova_server:
    [...]
    security_groups: default,webserver
```

### DIFF
--- a/cloud/openstack/os_server.py
+++ b/cloud/openstack/os_server.py
@@ -76,7 +76,8 @@ options:
      default: None
    security_groups:
      description:
-        - The name of the security group to which the instance should be added
+        - Names of the security groups to which the instance should be
+          added.  This may be a YAML list or a common separated string.
      required: false
      default: None
    nics:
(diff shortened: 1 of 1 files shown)


cloud/openstack/os_server.py
```diff
@@ -360,13 +361,18 @@ def _create_server(module, cloud):
             metas[k] = v
         module.params['meta'] = metas
 
+    if type(module.params['security_groups']) is str:
+        module.params['security_groups'] = module.params['security_groups'].split(',')
```
sivel: Instead of this,  it would be better to set `type='list'` in `openstack_full_argument_spec`.  `type='list'` automatically handles this functionality.

sivel: Also, just for future reference, it is better to use `isinstance` instead of running `type` on a variable and comparing it to a type.

larsks: I'm just trying to stick with the existing code style in this module.  I agree totally, but I'd like to make compartmentalized changes.

gregdek:

Shrews:
shipit

gregdek:
Thanks @larsks for the PR, and thanks @Shrews for the timely review. Marking for potential inclusion.

PR #2349, opened 2015-10-23
make os_router return a top level 'id' key

larsks:
make os_router return a top-level 'id' key, much like other
os_\* resources.


Shrews:
shipit

PR #2350, opened 2015-10-23
rearranged systemd check, removed redundant systemctl check

bcoca:
also fixed unused cmd and state var assignements

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -395,7 +395,7 @@ class LinuxService(Service):
         location = dict()
 
         for binary in binaries:
-            location[binary] = self.module.get_bin_path(binary)
+            location[binary] = self.module.get_bin_path(binary, opt_dirs=paths)
 
         for initdir in initpaths:
             initscript = "%s/%s" % (initdir,self.name)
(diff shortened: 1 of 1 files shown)


bcoca:
cc @reactormonk @eliasp

system/service.py
```diff
@@ -403,10 +403,31 @@ def get_service_tools(self):
...
 
         def check_systemd():
-            return os.path.exists("/run/systemd/system/") or os.path.exists("/dev/.run/systemd/") or os.path.exists("/dev/.systemd/")
+
+            # tools must be installed
+            if location.get('systemctl',False):
+
+                # this should show if systemd is the boot init system
+                # these mirror systemd's own sd_boot test http://www.freedesktop.org/software/systemd/man/sd_booted.html
+                for canary in ["/run/systemd/system/", "/dev/.run/systemd/", "/dev/.systemd/"]:
```
reactormonk, line 412: I think the absolute correct way would be to check for `systemd/system`, because systemd might be running as user, so `systemd/user` could exist, but `systemd/system` doesn't.

bcoca, line 412: not understanding the comment, are you saying we should add a check for systemd/user?

reactormonk, line 412: No, checking for `systemd/user` doesn't tell us anything about `systemd` running as init service. I suggest always checking for `systemd/system`.

Issue #1855, opened 2015-08-03
iam_policy doesn't update policies


garthk:
- Run a playbook using `iam_policy`
- Observe in EC2 IAM console the group has the policy with the correct document
- Edit the policy document a) in EC2 IAM console, OR b) in the local file
- Re-run the playbook
- Observe Ansible reports `ok`, not `changed`
- Observe either a) the local file is not re-applied, OR b) the group policy is not updated

This is with `iam.py` and `iam_policy.py` from around 1268a83eb7 dropped into `library/cloud/amazon` and run under Ansible 1.9.2.

```
- name: "IAM groups"
  iam:
    state: present
    iam_type: group
    name: "{{ item }}"
  with_items:
  - "{{ worker_name }}"
  - "{{ worker_control_name }}"

- name: "IAM policies"
  iam_policy:
    state: present
    iam_type: group
    iam_name: "{{ item.iam_name }}"
    policy_name: "{{ item.policy_name }}"
    policy_document: "{{ item.policy_document }}"
  with_items:
  -
    iam_name: "{{ worker_name }}"
    policy_name: "{{ worker_name }}"
    policy_document: "{{ worker_policy_file }}"
  -
    iam_name: "{{ worker_control_name }}"
    policy_name: "{{ worker_control_name }}"
    policy_document: "{{ worker_control_policy_file }}"
```

garthk:
Perhaps relevant, this item from the return structure:

```
The policy document you specified already exists under the name X.
```

… where X is the name of the policy file.

mseiwald:
Any news on this? It makes the iam_policy module pretty much unusable for us..

halberom:
Looks like there's a PR pending review - https://github.com/ansible/ansible-modules-core/pull/2352.

ryanwalls:
+1.  Experiencing the same problem.

ryanwalls:
Looks like the PR has been merged, so I think this can be closed.

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[...]

PR #2352, opened 2015-10-25
Feature/iam policy present state includes policy changes

bermudezjd:
This PR adds a minor change to the iam_policy module that allows existing policies to be updated without having to disable the `skip_duplicates` flag.  I introduced the idea on the devel boards here - https://groups.google.com/forum/#!topic/ansible-devel/sumBECXO4hY.  Please let me know whether this can remain part of the `present` state or whether an `updated` state is needed.


gregdek:
@defionscode Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@defionscode This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
@defionscode still waiting on your review.  Please comment with text 'shipit' or 'needs_revision' as appropriate. If we don't hear from you within 14 days, we will start to look for additional maintainers for this module.

gregdek:
Thanks @bermudezjd for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

bermudezjd:
ok, rebased and resolved merge conflicts so ready_for_review again... looks like folks are looking for this via #1855 so hoping that @defionscode will look at this soon.

robynbergeron:
Thanks @bermudezjd. @defionscode please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
Thanks @bermudezjd. @defionscode please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

bermudezjd:
@robynbergeron just checking in on this since it's been well over 2 weeks since you had reached out to @defionscode to review this.  is there someone else who can take a look?

waky79:
First I would like to say that great that someone has already fixing this. My only concern with the whole implementation is that what is the use of the do not skip duplicates flag. Now the new implementation will work better in case the policy changes. Still the new implementation will not update the policy name in case it has changed.

I am OK with the change but this will cause cases where you update the policy name but it stays as is, and if you allow duplicate, there will be two policies with the same name. The first one, in case you make ansible delete the policy, it does not find the one with name you have currently in your configuration. For the latter it leaves the older policy behind.

It would be, in my opinion, good to understand why the policy name is even part of the configuration as the implementation does not bind it to the policy content.

Still I would like to say that I am waiting this fix to be merged as it fixes even bigger issue with policy updating.

Jmainguy:
I am no familiar with the iam module at all. That being said, this change looks pretty minor. 6 total non comment lines being added. Slight refactor to make it a little easier to read. Looks good to me from my limited understanding. Only nitpitck comment would be to squash commits to make history easier to read when this does get merged. Nice work.

abadger:
@defionscode @Jmainguy has reviewed the code for this and thinks it looks sane.  Are you able to review whether the feature itself makes sense?  With that out of the way I think we could merge this. https://groups.google.com/forum/#!topic/ansible-devel/sumBECXO4hY

juliedavila:
This is fine with me, and as far as @waky79 s concern, I just verified, and it is not possible to have two IAM policies with the same name. Worst case, you'll have two policies with different names but with the same rules, which I think is an 'acceptable' scenario.

abadger:
Thanks @defionscode @Jmainguy Merged to devel!

waky79:
@defionscode yes true, it fails with same name. And like said I was fine with the fix. Just liked to know/debate what would be the desired optimal behavior, looks like I expressed myself badly.

Issue #2246, opened 2015-10-12
Documentation bug: mysql_user password parameter required


hut8:
While adding a user to mysql, [the documentation](http://docs.ansible.com/ansible/mysql_user_module.html) says that the `password` parameter is optional. Excluding it, I get:

```
TASK: [mysql_user name=your_mom priv=*.*:ALL state=present] **********************
failed: [1.3.3.7] => {"failed": true}
msg: password parameter required when adding a user
```

I should also point out that adding the parameter:

```
password=
```

works fine

robynbergeron:
Hi @hut8 -- Thanks for submitting this bug report. You might want to consider giving us the following additional information to help the module maintainer identify the issue: https://github.com/ansible/ansible/blob/devel/ISSUE_TEMPLATE.md

@marktheunissen -- thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module.

robynbergeron:
[module: database/mysql/mysql_user.py]

marktheunissen:
Hi, I don't maintain this module. The copyright holder and maintainer are not necessarily the same person.

famousgarkin:
+1, didn't work for me even with just `password:`, had to set an empty string explicitly:

```
- mysql_user:
    ...
    password: ''
```

Ansible 1.9.4

Jmainguy:
@famousgarkin @hut8 I think #2367 should solve this issue, what do yall think?

hut8:
Good enough!

bcoca:
closing as per above

PR #2367, opened 2015-10-27
Update doc to reflect password is required if adding a new user

Jmainguy:


Jmainguy:
To fix https://github.com/ansible/ansible-modules-core/issues/2246

PR #2379, opened 2015-10-28
os_server_facts returns facts about multiple servers

larsks:
have `os_server_facts` call `list_servers` rather than `get_server`, and
treat the `server` parameter as a wildcard pattern.  This permits one to
get facts on a single server:

```
- os_server:
    server: webserver1
```

On mutiple servers:

```
- os_server:
    server: webserver*
```

Or on all servers:

```
- os_server:
```

Introduces a `detailed` parameter to request additional server details
at the cost of additional API calls.

### DIFF
--- a/cloud/openstack/os_server_facts.py
+++ b/cloud/openstack/os_server_facts.py
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this software.  If not, see <http://www.gnu.org/licenses/>.
 
+import fnmatch
+
 try:
     import shade
     from shade import meta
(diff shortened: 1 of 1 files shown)


larsks:
@emonty, if this looks like a good idea, I'd like to extend a similar model to other openstack objects (images, routers, networks, etc.), and ultimately extend the filtering to support more than name-based queries.  Let me know what you think.

gregdek:

omgjlk:
Doing it this way can introduce significant delays in operation. Getting a full list can take a while on cloud with lots of instances. Your code makes you pay that penalty even if you're attempting to get information about a single server. I believe we should retain the `server:` parameter as a single entity and use the `get_server` API call, while introducing a new argument in order to pass in a glob.

Alternatively, the compute API capability of filtering by regex could be piped through http://developer.openstack.org/api-ref-compute-v2.1.html#listServers which may prevent paying the penalty of a whole list grab to filter by the glob.

larsks:
@j2sol  The shade module is already performing internally a similar operation in the `get_server` method (getting a list of servers, and then filtering by name).  This code will be exactly as performant (or not) as the current code, while permitting a wider range of uses.

Taking advantage of server-side filtering would indeed be better, but I think this module is not the right place for that code.

cloud/openstack/os_server_facts.py
```diff
@@ -64,10 +77,16 @@ def main():
 
     try:
         cloud = shade.openstack_cloud(**module.params)
-        server = cloud.get_server(module.params['server'])
-        hostvars = dict(openstack=meta.get_hostvars_from_server(
-            cloud, server))
```
omgjlk, line 69: Is this part lost, `get_hostvars_from_server` ? Is that automatically grabbed by `list_servers` ?

omgjlk:
You're right. My mistake. I have a code question, and then the general unease that this changes the return value a bit (from a list of vars to a list of servers [plus vars?]). Making that change should have some extra information provided, even though we are going to 2.0.

cloud/openstack/os_server_facts.py
larsks, line 69: It is called automatically by `list_servers` when `detailed=True`.  So you can get faster but less detail with `detailed=False`, or slower but more detail with `detailed=True`.

omgjlk, line 69: So this is also changing the default behavior of the module, which is to provide all the details in the facts. The default now would be to just provide a list. Is there a strong reason to change the default?

larsks, line 69: I guess it is changing the default, but this module has not yet seen a release, so now seems the right time to be doing this.  My reasoning is that it is better not to impose the load ot the additional API calls without explicitly requesitng that behavior, but if folks feel strongly we can default detailed to True.

omgjlk, line 69: You make good points. I don't feel strongly.

omgjlk:
shipit

PR #2380, opened 2015-10-28
Adding Facter Support to Setup.ps1

blakfeld:
The Linux "setup" module also queries facter (as well as a few other services). I added a bit at the end to add facter facts.

It checks to see if it can find the facter executable in the $PATH, and if so, runs `facter -j`, parses the json, and spits them out in the "facter_fact_name" format.


gregdek:
@blakfeld thanks for this PR. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

nitzmahone:
LGTM- thanks for the update!

PR #2382, opened 2015-10-28
Module now returns a list existing keys that exist on the target but …

Vitiate:
Module now returns a list existing keys that exist on the target but were not specified. These are the keys that would be removed in enforce mode.

Returning this list allows us to identify keys that perhaps have been missed or keys that are used by services that we are not aware of.


gregdek:
Thanks @Vitiate. This module is currently owned by the Core team, so it can take a while for patches to be reviewed. Thanks for your patience here.

PR #2384, opened 2015-10-28
PAM config module

juliedavila:
This a module for managing PAM configs.


system/pam.py
```diff
@@ -0,0 +1,551 @@
...
+            if set( re.split(r'[\s\t\n]*(?![^\[\]]*\])', config_lines[ref_location + 1]) ) == set(entry_args):
+                changed = False
+            else:
+                config_lines.insert(ref_location + 1, pam_entry + '\n')
+                changed = True
+        except IndexError:
+            config_lines.append(pam_entry + '\n')
+            changed = True
+
+    with open(config_path, 'w') as conf:
```
bcoca: won't work with 2.4

gregdek:
Why does this belong in core and not extras?

juliedavila:
@gregdek my thought process has to do with the declared intent to make extras a separately installed 'thing' in the future. Assuming that is still the case, it would be prudent to have this in core, imo, as a big user base of this will be public sector, especially those that leverage the Lockdown project. I can foresee occasional public sector types being limited to only core modules.

juliedavila:
If that's a problem, of course I can resubmit to extras

juliedavila:
Found a couple of bugs, will update when ready

juliedavila:
Updated

juliedavila:
will resubmit later

Issue #2390, opened 2015-10-30
need for --disableplugin=rhnplugin option for the YUM module (patch provided)


r1k0:
**Issue Type:**

feature request

**Ansible Version:**

1.9.4

**Ansible Configuration:**

``` [defaults]
roles_path            = roles:roles/datacenter:roles/shared:roles/customers/viapass:roles/customers/transwide:roles/customers/bmwallet:roles/customers/digipolis:roles/customers/mobilefor
inventory             = hosts
remote_tmp            = $HOME/.ansible/tmp
pattern               = *
remote_port           = 22
forks                 = 100
poll_interval         = 15
remote_user           = root
sudo_user             = root
transport             = smart
module_lang           = C
gathering             = smart
host_key_checking     = False
sudo_exe              = sudo
timeout               = 10
log_path              = /var/log/ansible.log
ansible_managed       = Ansible managed: {file} modified on %Y-%m-%d %H:%M:%S by {uid} on {host}
action_plugins        = /usr/share/ansible_plugins/action_plugins
callback_plugins      = callback_plugins/
connection_plugins    = /usr/share/ansible_plugins/connection_plugins
lookup_plugins        = /usr/share/ansible_plugins/lookup_plugins
vars_plugins          = /usr/share/ansible_plugins/vars_plugins
filter_plugins        = /usr/share/ansible_plugins/filter_plugins
bin_ansible_callbacks = True # enables callbacks for ansible executable (not just ansible-playbook)
fact_caching          = memory
display_skipped_hosts = False

[ssh_connection] 
# INFO: This option is there for cases where connections start being dropped with
#       'SSH Error: mux_client_hello_exchange: write packet: Broken pipe'
#       Using this option will ensure connections won't be closed but
#       the downside is that ansible-playbook slows down to almost 10 fold...
#ssh_args              = -o ServerAliveInterval=60
```

**Environment:**

remote hosts RHEL6/7, local client gentoo

**Summary:**

Need for --disableplugin=rhnplugin option for the YUM module.

**Steps To Reproduce:**

Using this:

```
 61 - name: rhn_register | install rhn-org-trusted-ssl-cert
 62   yum: name=http://rhns.mgnt.belbone.be/pub/rhn-org-trusted-ssl-cert-1.0-1.noarch.rpm state=present
```

**Expected Results:**

Ability to disable a yum plugin while installing an RPM from an http URL.

**Actual Results:**

Gives:

```
[...]

PR #2391, opened 2015-10-30
support --disableplugin and --enableplugin YUM options

r1k0:
more details in https://github.com/ansible/ansible-modules-core/issues/2390

Let me know if it's "merge-able". If not how can I further help?
Thanks


r1k0:
I'm now wondering if added parameters to install() shouldn't be passed as well to latest().
Let me know

abadger:
yeas, we'd need it both for install and latest -- possibly to other subcommands as well.  I'm not sure if yum will shortciruit the repos if you do an erase or if it has to load that code no matter what, for instance.

Do these also need to be used with repoquery?  Can these be used to replace the repoquery code (which seems to exist because of exactly this error...)  What versions of RHEL are you testing this on (Need to make sure this works on rhel5 and above).

Instead of comma separated strings, use type=list in the argument spec.

abadger:
And yes, this seems like a useful feature, so once we get questions answered and code tweaked we should be able to merge this.

I'm abadger1999 on irc.freenode.net in case this drops off my radar and you need to remind me to look at a new version.

gregdek:
@r1k0 It looks as though @abadger has recommended some revisions. 

Please make the suggested revisions, and when you're done, please add a comment with the text "ready_for_review" and we will continue with the review process.  Thanks!

gregdek:
@r1k0 A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

PR #2399, opened 2015-11-02
Change behavior of apt.py around installing recommended packages (#1916) with requested fixes

abadger:
Here's @hlieberman's changes to apt.py with the changes requested by @bcoca.


abadger:
bcoca reviewed and approved.

PR #2401, opened 2015-11-03
total_seconds not present on timedelta on python2.6

larsla:
wait_for with timeout doesn't work on python 2.6 with ansible 2.0 since timedelta doesn't have total_seconds.
this patch will make it compatible with python 2.6.


gregdek:
Thanks @larsla.

@jhoekx @jarv @AnderEnder Please review this PR to make sure it adheres to the following guidelines:

PR #2405, opened 2015-11-04
Add missing import on cloudformation for boto profile check

twaggs:
This is the same issue that affects route53: #1901 #2272

"The  module imports the ec2 utils:
from ansible.module_utils.ec2 import *
Which has a method that checks for boto support calling:
return hasattr(boto.ec2.EC2Connection, 'profile_name')"

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -105,9 +105,9 @@ EXAMPLES = '''
 tasks:
 - name: launch ansible cloudformation example
   cloudformation:
-    stack_name: "ansible-cloudformation" 
+    stack_name: "ansible-cloudformation"
     state: "present"
-    region: "us-east-1" 
+    region: "us-east-1"
     disable_rollback: true
     template: "files/cloudformation-example.json"
     template_parameters:
@@ -124,6 +124,7 @@ import time
 
 try:
     import boto
+    import boto.ec2
     import boto.cloudformation.connection
     HAS_BOTO = True
 except ImportError:


willthames:
I think it'd be better if `module_utils/ec2.py` was rewritten to avoid the need for the `boto.ec2` import.
I was trying to write elegant checking with the `hasattr` but the `LooseVersion` check used for `validate_certs` would work just as well and avoid the unnecessary import.

bcoca:
since boto.ec2 is not used here, does it make more sense to be included in the shared code?

willthames:
@bcoca see my comment above. I think we should just get rid of the reliance on boto.ec2 in module_utils/ec2.py (um, and maybe rename ec2.py to aws.py but that's probably a larger battle)

bcoca:
we can rename and keep a symlink for those that have custom modules still using it and give a deprecation notice

gregdek:
Assigning to core team since it's proposing a backport.

gregdek:
Thanks @twaggs. All backport requests must be reviewed by the core team, and this can take time. We appreciate your patience.

willthames:
@gregdek I think ansible/ansible#15059 should remove the need for this. Should also be backported, of course

ryansb:
This is fixes with the boto3 rewrite of the `cloudformation` module.

gregdek:
Thanks @twaggs for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

PR #2406, opened 2015-11-04
Changed ec2 connection method

wimnat:
Pull request to resolve issue https://github.com/ansible/ansible/issues/12959

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -1353,9 +1353,12 @@ def main():
     if not HAS_BOTO:
         module.fail_json(msg='boto required for this module')
 
-    ec2 = ec2_connect(module)
-
     region, ec2_url, aws_connect_kwargs = get_aws_connection_info(module)
+
+    try:
+        ec2 = connect_to_aws(boto.ec2, region, **aws_connect_params)
+    except (boto.exception.NoAuthHandlerFound, StandardError), e:
+        module.fail_json(msg=str(e))
 
     if region:
         try:


willthames:
I'm not 100% sure why this change works, as I thought `ec2_connect` was just an special case of `connect_to_aws('boto.ec2',...)` (see https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/ec2.py#L202-L213)

However, if `ec2_connect` is this broken, should we look at removing it completely?

wimnat:
I do feel the option of connect_to_aws and ec2_connect is unnecessary.  I would prefer to just use the connect_to_aws method all the time in all modules.

cloud/amazon/ec2.py
```diff
@@ -1353,10 +1353,13 @@ def main():
     if not HAS_BOTO:
         module.fail_json(msg='boto required for this module')
 
-    ec2 = ec2_connect(module)
-
     region, ec2_url, aws_connect_kwargs = get_aws_connection_info(module)
 
+    try:
+        ec2 = connect_to_aws(boto.ec2, region, **aws_connect_params)
```
willthames, line 1359: Note that the previous line returns `aws_connect_kwargs`

You're effectively passing it a dict from somewhere else

willthames:
I think we should fix whatever is going wrong in `ec2_connect`. `ec2_connect` eliminates a lot of boilerplate (it handles the `get_aws_connection_info`, it does exception handling, it even handles the `ec2_url` which most alternatives do not). 

I suspect better security_token handling in `get_aws_connection_info` might get closer to a genuine fix.

wimnat:
Hmmm I'll close this for now but it warrants more research.  You're right, it shouldn't fix it but i've seen a similar approach fix other connection problems too.

willthames:
I just wonder if throwing all the `aws_connect_params` away makes it work (as that's really the only difference - you're passing a completely different dict, although one that must exist, and therefore be populated with reasonable values)

PR #2410, opened 2015-11-04
Added support for insecure parameter

stewrutledge:
Adds a boolean insecure for ignoring SSL certificate validation errors when connecting to vcenter.

### DIFF
--- a/cloud/amazon/iam_policy.py
+++ b/cloud/amazon/iam_policy.py
@@ -64,9 +64,9 @@ extends_documentation_fragment:
 '''
 
 EXAMPLES = '''
-# Create and policy with the name of 'Admin' to the group 'administrators'
+# Create a policy with the name of 'Admin' to the group 'administrators'
 tasks:
-- name: Create two new IAM users with API keys
+- name: Assign a policy called Admin to the administrators group
   iam_policy:
     iam_type: group
     iam_name: administrators
@@ -87,7 +87,7 @@ task:
      - Luigi
   register: new_groups
 
-- name:
+- name: Apply READ-ONLY policy to new groups that have been recently created
   iam_policy:
     iam_type: group
     iam_name: "{{ item.created_group.group_name }}"
--- a/cloud/amazon/rds.py
+++ b/cloud/amazon/rds.py
@@ -829,13 +829,17 @@ def promote_db_instance(module, conn):
     instance_name = module.params.get('instance_name')
     
     result = conn.get_db_instance(instance_name)
+    if not result:
+        module.fail_json(msg="DB Instance %s does not exist" % instance_name)
+
     if result.get_data().get('replication_source'):
-        changed = False
-    else:
         try:
             result = conn.promote_read_replica(instance_name, **params)
+            changed = True
         except RDSException, e:
             module.fail_json(msg=e.message)
+    else:
+        changed = False
 
     if module.params.get('wait'):
         resource = await_resource(conn, result, 'available', module)
--- a/cloud/digital_ocean/digital_ocean.py
+++ b/cloud/digital_ocean/digital_ocean.py
@@ -37,6 +37,7 @@ options:
   api_token:
     description:
      - DigitalOcean api token.
+    version_added: "1.9.5"
   id:
     description:
      - Numeric, the droplet id you want to operate on.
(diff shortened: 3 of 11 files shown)


gregdek:

stewrutledge:
shipit

gregdek:
Sorry @stewrutledge -- only the maintainer of a module can give the "shipit" command.

rhoop:
@gregdek @stewrutledge Not sure I feel comfortable allowing an insecure connection to VCenter. Can you give an example of why this would be needed?

stewrutledge:
Sorry about the shipit, misunderstanding.

stewrutledge:
@rhoop maybe insecure is the wrong way to phrase it, it should be "verify_ssl."

The connection is still encrypted, it just suppresses warnings for an unverified ssl connection.

With recent Python versions (I think 2.7.9 and up) the default behavior is to throw an exception when connecting to unverified ssl, and this behavior effects pyvsphere.

stewrutledge:
Looks like it's 2.7.10:

http://legacy.python.org/dev/peps/pep-0476/

https://lwn.net/Articles/611243/

There are surely many cases of users running self signed vsphere instances out there.

rhoop:
@stewrutledge I am away from the office to test this with a self signed cert. Can you 1) Post the exception and 2) Try the original module after running `pip install requests[security]`.

I can get to it tomorrow otherwise

stewrutledge:
I won't be able to test it tonight, but I can give it a try tomorrow morning at the office.

The exception is something like:

```
ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)
```

I can post the full exception tomorrow.

stewrutledge:
Yeah, that is exactly the exception:

``` python
raise Exception(sslerr)\nException: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)
```

I'm working on getting requests[security] installed however I got some compilation issues and didn't have time to work with that.

A thought hit me though: if this exception is no longer thrown wouldn't that mean it silently creates an insecure connection (or silently ignored self signed certificates)? If that is so, I think it may be better to explicitly allow self signed certs rather than removing the exception all together.

luto:
@stewrutledge when getting the `_create_unverified_context` factory, you should also catch any `AttributeError`s, because that attribute doesn't exist in older python versions. Currently you're relying on the assumption that a pre-2.7.9 python will never raise a `CERTIFICATE_VERIFY_FAILED` exception.

stewrutledge:
Ah I can see how that would be a problem!

What do you think the behavior should be then? Except on attribute error when attempting to set the unverified context and raise a more specific exception like "Cannot create insecure connection"?

I would even prefer to see a scenario where a self signed cert can be overridden with insecure, but a public/private mismatch isn't (as that indicates something more nefarious/an actual misconfiguration on the serverside). What other conditions would generate that kind of exception?

The CERTIFICATE_VERIFY_FAILED exception won't be handled unless the user wants to create an insecure connection, so I am assuming that a message relating specifically to that failure would be best?

bcoca:
To keep consistency across modules, you should rename this option to 'validate_certs' and default to 'yes' (which reverses the current logic of insecure).

stewrutledge:
Looks like I screwed up a rebase here. I've implemented the validate_cert logic anyway.

I assume I'll need to redo this with a clean version of the repo?

bcoca:
yep, this is not mergable as is

tic1release:
Has a decision been made to merge this branch?    My company uses self-signed certs and this would be helpful

gregdek:
Thanks @stewrutledge. @jhawkesworth please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

jhawkesworth:
@gregdek my git-fu is weak looks like the change to win_copy has already been merged and certainly isn't relevant to the actual described change.

stewrutledge:
@jhawkesworth

So what do I need to do here?

I am assuming a new pull request is the only way to get rid of all the commits I've accidentally included?

Beyond that are there any thoughts on handling Python <=2.8?

nitzmahone:
@stewrutledge - I'm good to merge this, but yeah, we need to clean up the commit stuff before it's mergeable. Easiest thing to do would be to do a clean checkout of ansible-modules-core/devel, then create and apply a single patch with your final changes, commit, then force-push to your branch. That will auto-update the PR with the new single patch.

Alternatively, I can create and merge the patch for you, but your name won't be on it anymore in the resultant code... (or you could create a new PR from a clean branch and close this one- CC @nitzmahone if you do that and I'll merge it ASAP before it drifts again)

stewrutledge:
@nitzmahone It's alright, I can take care of it. I'll get a new PR thrown together during the weekend.

Thanks for the help!

gregdek:
Thanks @stewrutledge. @jhawkesworth please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

nitzmahone:
@stewrutledge Sounds good- I'll close this PR, just CC me on your new one and I'll merge it. Thanks!

Issue #2389, opened 2015-10-29
temporary AWS Access Keys results in auth failures


preflightsiren:
#### Issue Type:

 Bug Report
#### Component Name:

ec2 module
#### Ansible Version:

ansible 1.9.4
#### Ansible Configuration:

none 
#### Environment:

Mac OSX 10.11 / Not applicable
#### Summary:

Using temporary AWS Access Keys results in auth failures
#### Steps to reproduce:

generate temporary access keys (eg. via STS or SAML provider)
attempt to create ec2 resource
#### Expected results:

ec2 resource is created
#### Actual results:

AWS AuthFailure exception

I originally lodged this against ansible/ansible as https://github.com/ansible/ansible/issues/12959 but I think maybe this should be resolved in the core modules code. 

There's a comment in the code that suggests, that perhaps the modules just need to be modified to use  `connect_to_aws()` 

```
def get_ec2_creds(module):
    ''' for compatibility mode with old modules that don't/can't yet
        use ec2_connect method '''
```

```
failed: [localhost] => {"failed": true, "parsed": false}
Traceback (most recent call last):
  File "/Users/secole/.ansible/tmp/ansible-tmp-1446095011.34-234704611276689/ec2", line 3070, in <module>
    main()
  File "/Users/secole/.ansible/tmp/ansible-tmp-1446095011.34-234704611276689/ec2", line 1249, in main
    (instance_dict_array, new_instance_ids, changed) = create_instances(module, ec2, vpc)
  File "/Users/secole/.ansible/tmp/ansible-tmp-1446095011.34-234704611276689/ec2", line 792, in create_instances
    vpc_id = vpc.get_all_subnets(subnet_ids=[vpc_subnet_id])[0].vpc_id
  File "/Library/Python/2.7/site-packages/boto/vpc/__init__.py", line 1153, in get_all_subnets
    return self.get_list('DescribeSubnets', params, [('item', Subnet)])
  File "/Library/Python/2.7/site-packages/boto/connection.py", line 1186, in get_list
    raise self.ResponseError(response.status, response.reason, body)
boto.exception.EC2ResponseError: EC2ResponseError: 401 Unauthorized
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>AuthFailure</Code><Message>AWS was not able to validate the provided access credentials</Message></Error></Errors><RequestID>bcce0f14-b8d4-46e0-a582-17993365b18b</RequestID></Response>
[...]

PR #2421, opened 2015-11-06
Use connect_to_aws where possible

willthames:
`connect_to_aws` fixes a bug with security tokens in AWS.
Modules should use that rather than calling
`boto.x.connect_to_region`

Addresses #2389 

Needs testing by people with security tokens, but I'm confident that `connect_to_aws` is widely tested across the board.

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -305,10 +305,7 @@ def main():
     stack_outputs = {}
 
     try:
-        cfn = boto.cloudformation.connect_to_region(
-                  region,
-                  **aws_connect_kwargs
-              )
+        cfn = connect_to_aws(boto.cloudformation, region, **aws_connect_kwargs)
     except boto.exception.NoAuthHandlerFound, e:
         module.fail_json(msg=str(e))
     update = False
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -1361,7 +1361,7 @@ def main():
 
     if region:
         try:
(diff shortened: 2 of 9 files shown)


gregdek:
Thanks @willthames. @lwade please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
Also adding a backport request, per #2557. Which means this belong in core review instead. cc: @ansible/core

abadger:
Code looks good to me but I'm not a user of the amazon modules so would like to have someone confirm the changes work.  Willing to merge with that confirmation.

backporting I'm not sure of... @willthames wrt 1.9.x do you consider this a bugfix or just a missing feature?  If we backport are there other modules that need this change applied in the 1.9.x tree or are these the only modules that needed to be changed to use connect_to_aws there as well?

abadger:
Merged to stable-1.9 for 1.9.5..  If someone would like to test the 1.9.5 release candidates (or current checkouts of stable-1.9) to make sure this works, that would be greatly appreciated.

romanrev:
@abadger just installed the Ansible 1.9.5 using PIP and got the same error as before when using temporary credentials with cloudformation module:

```
msg: The security token included in the request is invalid.

```

1.9.6 did not improve things either. Is there any chance for this to be finally released in 1.9.x?

romanrev:
I went ahead and created the [Ansible 1.9.6 STS assume role backport patch](https://gist.github.com/romanrev/8bacbad568ced64754de7f9ca8f6df16) and it seems to be working.

PR #2425, opened 2015-11-06
Update ec2_elb_lb.py

aioue:
add connection draining default


gregdek:
Thanks @aioue! Trivial fix, marking for potential inclusion.

bcoca:
actually the default is null or None, not 'no'

abadger:
Thanks for fixing that!

PR #2427, opened 2015-11-06
Allow htpasswd to update/not update passwords

martingalloar:
Add option to choose if password for the user should be updated when setting state 'present' and the user was already in the file. This allows using the module for ensuring that certain users are there but not change their passwords every time.


gregdek:
Thanks @martingalloar for your pull request. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

martingalloar:
Sure, no problem! In the meantime, I'm using it as a custom module for a role.

PR #2428, opened 2015-11-06
fix regression introduced by f38186ce8b49ea98e29241712da45917a3154e73…

abadger:
…, and propose a fix for docker facts

Cherry-pick docker facts fix from devel


abadger:
Took a look through history and f38186c hasn't yet been pushed to stable-1.9.  So either this is not needed on 1.9 or we have to pull more commits as well.  Going to close this for now as 2.0 is getting close.  If someone actually has a problem on 1.9 that would be fixed by this and other commits being cherrypicked we can look into backporting then.

PR #2433, opened 2015-11-08
Add openstack project module

agireud:
Added os_project module to support openstack projects.
Originally submitted to ansible-modules-extras but I was told it should be submitted here instead.


bcoca:
you were told wrong, it should go to extras, that is the rule for new modules. We made a short exception with some openstack modules that were replacing the old neutro/quantum/etc ones already in core, but this one does not fall into that category.

PR #2436, opened 2015-11-09
Use add_ips_to_server API

emonty:
This module is still using an old pre-release API and needs to update
to use the current API.

Co-Authored-By: Marton Kiss marton.kiss@gmail.com


Shrews:
shipit

PR #2438, opened 2015-11-10
unarchive: read absolute tempdir path for downloads (fixes #2250)

### LINKED ISSUE #2250: unarchive can't download archives when pipelining is enabled
##### Issue Type:

Bug Report
##### Ansible Version:

```
$ ansible --version
2.0.0
```

Latest git as of a few minutes ago, meaning ansible-modules-core 92d2501.
##### Ansible Configuration:

```
[defaults]
inventory = inventory

[ssh_connection]
pipelining = True
```
##### Environment:

Running on Ubuntu 15.04, against Debian Jessie.
##### Summary:

The unarchive module fails to unpack tar files retrieved from http when pipelining is enabled. It works fine when pipelining is disabled.
##### Steps To Reproduce:

```

---
- name: Test
  hosts: whatever
  remote_user: admin
  tasks:
    - unarchive:
        src: https://github.com/ansible/ansible/archive/v1.9.4-1.tar.gz
        dest: /tmp
        copy: no
```
##### Expected Results:

A new release of ansible in /tmp
##### Actual Results:

The unpack step fails with `{"cmd": "/bin/tar -xzf \"v1.9.4-1.tar.gz\"", "err": "tar (child): v1.9.4-1.tar.gz: Cannot open: No such file or directory\ntar (child): Error is not recoverable: exiting now\n/bin/tar: Child returned status 2\n/bin/tar: Error is not recoverable: exiting now\n", "out": "", "rc": 2}`

I did some digging, and on [https://github.com/ansible/ansible-modules-core/blob/92d2501/files/unarchive.py#L281](line 281 in unarchive.py), `__file__` is `<stdin>`, so `os.path.dirname(__file__)` returns `''`, which means the file is uploaded to `/home/admin` in my reproduction case, while `src` is set to the relative path `v1.9.4-1.tar.gz`, and `TgzArchive`'s `unarchive` specifies the `dest` directory as working dir and thus can't find the file.

If I disable pipelining, everything works fine, however.

mmatuska:
The tempdir variable for unarchive downloads needs to be determined absolutely.
Fixes bug #2250

### DIFF
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -278,7 +278,7 @@ def main():
             module.fail_json(msg="Source '%s' failed to transfer" % src)
         # If copy=false, and src= contains ://, try and download the file to a temp directory.
         elif '://' in src:
-            tempdir = os.path.dirname(__file__)
+            tempdir = os.path.dirname(os.path.abspath(__file__))
             package = os.path.join(tempdir, str(src.rsplit('/', 1)[1]))
             try:
                 rsp, info = fetch_url(module, src)


bcoca:
I cannot reproduce the error on current devel, so I cannot verify that this fix is needed, do you have another example aside from the one in #2250 that reproduces this?

mmatuska:
On my setup I am getting this issue using latest 2.0.0 beta with ssh pipelining enabled, login with the root user and use a "dest" folder that is not users home (dest="/tmp").

bcoca:
the only differences with my test is that i was using the latest devel and logging in as a normal user, let me try with root

bcoca:
so it still works for me as root, can you test against current devel? it might have been fixed already.

mmatuska:
No, it doesn't work with latest devel. Target system is RHEL 7.1 with Python 2.7.5.

The real problem here is the usage of __file__. If you are pipelining your "__file__" is "<stdin>" which is actually the wrong way of dealing with this problem here. I would prefer os.getcwd()

Since Python 3.4, __file__ always returns the absolute path. Are you testing with Python 3.4+?
Reference: https://docs.python.org/3.4/whatsnew/3.4.html#other-language-changes

To stay compatible with older Python versions, there are 2 possible solutions, you can choose whatever you like:
1.) tempdir = os.path.dirname(os.path.abspath(__file__))
2.) tempdir = os.getcwd()

Here is an interesting discussion regarding this problem:
http://stackoverflow.com/questions/7116889/python-file-attribute-absolute-or-relative

But now as I consider the whole problem, shouldn't we use a real tempdir, anyway?
tempdir = tempfile.gettempdir()

bcoca:
considering that ansible does NOT work with python 3 yet, no, I'm not testing with python 3.4, using Python 2.7.10

mmatuska:
Okay, didn't know that. But back to my question, why are you using __file__ when the scripts are pipelined via stdin? And is it standard to put temporary files to the home directory of the remote user?

bcoca:
well, I imagine the `__file__` was used before we added pipelining into the mix, not sure how that works.

as for the temp files, yes, by default they go into a subdir of the user's home, but that is configurable and can change with privilege escalation in the mix. But pipelining avoids all that.

I did testing against centos7.1 target, I still cannot reproduce the issue.

Jmainguy:
@bcoca @mmatuska I am able to reproduce on devel 2.1

Set pipelining true, workstation == Centos 7, Target == Debian 8

Changing to 

```
tempdir = os.path.dirname(os.path.abspath(__file__))
```

Fixed the issue.

shipit

Jmainguy:
Actually, while that does fix the error, it introduces a new one. It changes the tempdir to the $HOME directory, not the temp directory, so that the file never gets deleted. Looking at a way to make it work correctly with pipelining.

Jmainguy:
Looks like with pipelining there is no longer a temp dir, if this is true, this problem will also exist in the yum module. And will exist in two open PR's affecting uncompress and apt module. 

@bcoca @abadger Is there no longer a temp dir in pipelining, and if I am wrong, how do I access the temp dir in pipelining?

bcoca:
pipelning avoids writing to disk, so a tmpdir is not needed, right now thinking we would want to create one anyways (so modules can use) or should we add a default tmpdir function in modules that will use it if it exists or create a new one if not (pipelining=true)

hryamzik:
Folks, please see some more details in #3228. I'm still not sure why does unarchive download files directly to the home directory but this code prevents it from failing.

I've done mostly the same fix in #3229. By the way, I don't see any real difference between `abspath` and `realpath`, both should work here. =)

@bcoca can we just merge this and fire a separate ticket for tmp directory issues? Failing module is a real pain.

> pipelning avoids writing to disk, so a tmpdir is not needed

Well, after running two failing unarchive tasks I get two new empty directories in `.ansible/tmp`. And they never get cleaned.

hryamzik:
@Jmainguy 

> Actually, while that does fix the error, it introduces a new one. It changes the tempdir to the $HOME directory, not the temp directory, so that the file never gets deleted. Looking at a way to make it work correctly with pipelining.

That's not true now. If you don't make any changes you'll still find downloaded file in your home directory after task fail.

UPD: here's some results of template + pipelining:

```
level=error msg="Error reading file \"/etc/prometheus/file_sd_configs/.ansible_tmpENSLenfile_sd_config.yml\": open /etc/prometheus/file_sd_configs/.ansible_tmpENSLenfile_sd_config.yml: permission denied" source="file.go:196"
```

gregdek:
Thanks @mmatuska. To the current maintainers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

Jmainguy:
@bcoca so, we should create a new tmpdir function which yum,apt, unarchive, and others can utilize instead of assuming the dir the file is called from is tmp right? and then delete that tmp dir at the end of the play?

gregdek:
Thanks @mmatuska for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

Jmainguy:
@gregdek because #3307 got merged, its quite possible this PR is now redundant.

gregdek:
Closing as dup of #3307. @mmatuska if you feel that #3307 does not handle this situation, please feel free to open a new PR. Thanks for your effort!

PR #2451, opened 2015-11-11
Update facts when hostname is changed

jordidg:
When successfully changing the hostname, return an updated ansible_hostname fact.
ansible_hostname should contain the unqualified hostname, not the full one

### DIFF
--- a/system/hostname.py
+++ b/system/hostname.py
@@ -42,6 +42,7 @@ EXAMPLES = '''
 - hostname: name=web01
 '''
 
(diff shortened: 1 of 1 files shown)


bcoca:
shouldn't ansible_node and ansible_fqdn also be updated?

jordidg:
I think you mean ansible_nodename?
They are both updated, I just forgot to mention it in the description.
ansible_domain is updated as well.

PR #2453, opened 2015-11-12
Add optional 'spot_launch_group' parameter to ec2 module

amenonsen:
This allows instances to be placed into a specific launch group:

```
«Specify a launch group in your Spot instance request to tell Amazon
EC2 to launch a set of Spot instances only if it can launch them
all. In addition, if the Spot service must terminate one of the
instances in a launch group (for example, if the Spot price rises
above your bid price), it must terminate them all.»
```

Patch by Pavan Deolasee (@pavanvd).

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -247,6 +247,13 @@ options:
     required: false
     default: null
     aliases: ['network_interface']
+  spot_launch_group:
+    version_added: "2.0"
+    description:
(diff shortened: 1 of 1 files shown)


willthames:
looks good to me. :+1:

amenonsen:
It seems I misunderstood the situation. The patch isn't actually by @pavanvd, it's actually the same thing as the PR #2211 from @kai11. I'm very sorry for the confusion. I'll close this PR now.

PR #2454, opened 2015-11-12
Small doc change in EXAMPLES for lineinfile.py

parberge:
I.e remove one case of '='. ( I said it was small, didn't I?)


parberge:
Wow, just realized it was intended.. nevermind.

Issue #2449, opened 2015-11-11
Git.py is unable to fetch submodule updates due to locales


glenux:
##### Issue Type:
- Bug Report
##### Ansible Version:

```
$ ansible --version
ansible 1.9.4
  configured module search path = None
```
##### Ansible Configuration:

Default debian configuration for ansible (nothing changed).

Here is the diff with the current upstream version : 

```
$ diff -Nur examples/ansible.cfg /etc/ansible/ansible.cfg
--- examples/ansible.cfg    2015-11-11 11:23:15.701009848 +0100
+++ /etc/ansible/ansible.cfg        2015-09-04 01:03:08.000000000 +0200
@@ -11,18 +11,18 @@

 # some basic default values...

-#inventory      = /etc/ansible/hosts
+inventory      = /etc/ansible/hosts
 #library        = /usr/share/my_modules/
-#remote_tmp     = $HOME/.ansible/tmp
-#pattern        = *
-#forks          = 5
-#poll_interval  = 15
-#sudo_user      = root
+remote_tmp     = $HOME/.ansible/tmp
+pattern        = *
+forks          = 5
+poll_interval  = 15
+sudo_user      = root
 #ask_sudo_pass = True
 #ask_pass      = True
-#transport      = smart
+transport      = smart
 #remote_port    = 22
-#module_lang    = C
+module_lang    = C

 # plays will gather facts by default, which contain information about
 # the remote system.
@@ -30,7 +30,7 @@
 # smart - gather by default, but don't regather if already gathered
 # implicit - gather by default, turn off with gather_facts: False
 # explicit - do not gather by default, must say gather_facts: True
-#gathering = implicit
+gathering = implicit

 # additional paths to search for roles in, colon separated
 #roles_path    = /etc/ansible/roles
@@ -38,20 +38,14 @@
 # uncomment this to disable SSH key host checking
 #host_key_checking = False

-# change the default callback
-#stdout_callback = skippy
-# enable additional callbacks
-#callback_whitelist = timer, mail
-
 # change this for alternative sudo implementations
-#sudo_exe = sudo
+sudo_exe = sudo

-# What flags to pass to sudo
-# WARNING: leaving out the defaults might create unexpected behaviours
-#sudo_flags = -H -S -n
+# what flags to pass to sudo
+#sudo_flags = -H

 # SSH timeout
-#timeout = 10
+timeout = 10

 # default user to use for playbooks if user is not specified
[...]

PR #2462, opened 2015-11-12
Reset locales before calling any (git) command.

glenux:
Closes bug #2449 .

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -707,6 +707,11 @@ def main():
         set_git_ssh(ssh_wrapper, key_file, ssh_opts)
         module.add_cleanup_file(path=ssh_wrapper)
 
+    # make sure no locales are used (that could break output matching)
+    os.environ["LANG"] = "C"
+    os.environ["LC_ALL"] = "C"
+    os.environ["LC_MESSAGES"] = "C"
+
     # add the git repo's hostkey 
     if module.params['ssh_opts'] is not None:
         if not "-o StrictHostKeyChecking=no" in module.params['ssh_opts']:


bcoca:
this should already be happening when executing the module.

glenux:
As you can see using the [proof-of-concept](https://github.com/glenux/bugreport-ansible-git-locales) playbook, locales are not OK when calling git.

I added two more utils to the POC repository :
- a Vagrantfile + build script that replicates my environment (Debian + french locales)
- a wrapper for git that dumps environment variables into /var/log/git.log

If you run the vagrant box (as described in the README.md) you get the following result :
- ansible submodule update fails (with latest ansible-module-core version)
- /var/log/git.log shows french settings when git is called (in  LANG / LC_MESSAGES / etc.)

glenux:
I also noticed that other modules ([subversion.py](https://github.com/ansible/ansible-modules-core/blob/devel/source_control/subversion.py#L219), sh.py, etc.) are using a similiar environment reset for locales.

I think it would be better to do a such environment reset before dispatching commands to modules. That would prevent similar issues on other modules that parse system commands output.

bcoca:
That is what i mentioned above, we already do this when calling any/all modules.

gregdek:
So if we already do this, what's the issue that @glenux is seeing? Perhaps something to be discussed in #2449 instead of here?

@bcoca should we close this one?

abadger:
@glenux going through backlog of tickets and found this one.  I remembered the same thing as @bcoca but on reviewing the code we realized that ansible has evolved over time and the current code only overrides the controller's locale settings if those locale settings are not found on the client.

A few weeks ago I even added some code to basic.py to help with this something like the following change to the yum module is probably what we want in 2.0  for git, apt, yum, and anything else that screen scrapes : e2a7ba35dbc2b9af5ad1858a64a7bb460e109a0b

I'll work on a fix that utilizes that.

abadger:
Okay, the alternate fix is now applied in devel and stable-2.0.

PR #2463, opened 2015-11-13
mark failed async_status as finished

nitzmahone:
Running async_status in an "until: result.finished" loop will mask a module failure (eg, traceback) with a
template failure, because the fail dict doesn't include "finished" (eg, you'll see "ERROR! The conditional check 'bogus_out.finished' failed. The error was: ERROR! error while evaluating conditional: bogus_out.finished ({% if bogus_out.finished %} True {% else %} False {% endif %}") instead of the traceback from the module. Because the failure dict still includes "failed: true",
this change has no effect on stoppage/failure reporting, it just prevents the common usage pattern from masking the underlying error message.

### DIFF
--- a/utilities/logic/async_status.py
+++ b/utilities/logic/async_status.py
@@ -66,7 +66,7 @@ def main():
     log_path = os.path.join(logdir, jid)
 
     if not os.path.exists(log_path):
-        module.fail_json(msg="could not find job", ansible_job_id=jid)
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @nitzmahone for your pull request. This module is critical to Ansible's functionality, so it's owned by the Core team, and it can take a while for patches to be reviewed. Thanks for your patience here.

nitzmahone:
also cherry-picked to stable-2.0 (target 2.0.1.0)

PR #2467, opened 2015-11-13
It may be string with int comparison, if ansible user specifies ident…

cinek810:
…ifier as int

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -389,7 +389,7 @@ def main():
         decoded_name = rset.name.replace(r'\052', '*')
         decoded_name = decoded_name.replace(r'\100', '@')
 
-        if rset.type == type_in and decoded_name.lower() == record_in.lower() and rset.identifier == identifier_in:
+        if rset.type == type_in and decoded_name.lower() == record_in.lower() and rset.identifier == str(identifier_in):
             found_record = True
             record['zone'] = zone_in
             record['type'] = rset.type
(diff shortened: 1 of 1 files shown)


bcoca:
cc @bpennypacker for review

bpennypacker:
Looks good to me.  shipit!

Etherdaemon:
I think this commit has broken route53 delete's for me:

Changing back to this:

```
    if rset.type == type_in and decoded_name.lower() == record_in.lower() and rset.identifier == identifier_in:
```

From this:

```
if rset.type == type_in and decoded_name.lower() == record_in.lower() and rset.identifier == str(identifier_in):
```

And my delete's work again. With this commit, my records alway come back as not found.

Example of the command that was working but doesn't with this change is:

```
route53:
    command: delete
    zone: vc.test.solutions.
    record: spectre-ha.vc.test.solutions.
    value: 10.2.2.2
    type: A
    ttl: '60'
```

This would be because identifier is only required when you want to use weighted/latency based resources. If identifier is not supplied this the value is null, and rset.identifier is also null. But setting str(identifier) actually changes that statement to null == "None" which is false.

Changing that line to the following fixes it for me - so maybe an option?

```
    if rset.type == type_in and decoded_name.lower() == record_in.lower() and (rset.identifier == identifier_in or rset.identifier == str(identifier_in)):
```

Or:

```
 if rset.type == type_in and decoded_name.lower() == record_in.lower() and str(rset.identifier) ==  str(identifier_in):
```

cinek810:
Can you please check #2477 it's my commit solving this issue without regression, but it's not yet merged. Your solution in #2501 is generally OK, however what if someone will use "None" as an identifier? In such cache non existing identifier will match this string. 

You also removed the `rset.name=dedoded_name` line, which solves different issue with separate commit b6223ca.

Etherdaemon:
@cinek810 no worries - I'll cancel my merge - tested yours and it works for my stuff so :+1:

PR #2477, opened 2015-11-16
Change always convert identifier_in to string

cinek810:
Because it breaks behaviour for all tasks without identifier_in. 
In this cases obviously `None !=str(None)`, but we want to match already existing DNS records with unset identifier with requests without identifier.


cinek810:
a kind of community review done by @Etherdaemon, maybe the commit should be merged faster to prevent problems like in #2501

gregdek:
@cinek810 Thanks for submitting this pull request. Unfortunately, it appears that the underlying code has changed since your submission, and this PR now needs to be rebased. 

cinek810:
rebased in #2581

PR #2486, opened 2015-11-18
Fix copy with the remote_src option

mcls:
It was broken in 6e37f1dcef0 when the `remote_src` was added. 
`os.basedir` doesn't seem to exist and `shutil.copy2` expects a path instead of a tuple.

Not sure if this is related to a specific python version? I'm using Python 2.7.10
I'm not a python developer and I'm new to ansible, but this is what I did to get it to work for me.

Otherwise I got this error:

```
AttributeError: 'module' object has no attribute 'basedir'
```


bcoca:
applied to devel and stable-2.0

PR #2492, opened 2015-11-18
Fix os_user_group module

Shrews:
This module had a couple of errors in it. Also added check mode
support.


Shrews:
ping @emonty @j2sol

emonty:
shipit

emonty:
and thanks!

PR #2501, opened 2015-11-20
Revert #2467 fix ident

Etherdaemon:
By updating to make it compare 2 strings so that null values match


Etherdaemon:
Not required - #2477 will rectify the issue

PR #2505, opened 2015-11-20
Fix #822 - Correct failure to overwrite existing records

### LINKED ISSUE #822: route53 module: overwrite set to yes but still fails
I have the following in my playbook:

``` yaml
- name: Set Up Route53 IPv4
  route53: >
      aws_access_key=""
      aws_secret_key=""
      command=create
      zone=domain.com
      record=site.domain.com
      type=A
      ttl=7200
      value={{ ansible_default_ipv4.address }}
      overwrite=yes
```

Yet it still fails if the record already exists:

```
Tried to create resource record set [name='site.domain.com.', type='A'] but it already exists
```

Please let me know if that's not how it was intended to use, otherwise it seems like a bug?
Thanks!

donckers:
The boto.route53.record.Record class defaults the identifier attribute to None.  So, by setting the default of this value in the Ansible module and then ensuring that we are comparing the proper types in the conditional this should be fixed.

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -295,7 +295,7 @@ def main():
             overwrite            = dict(required=False, type='bool'),
             retry_interval       = dict(required=False, default=500),
             private_zone         = dict(required=False, type='bool', default=False),
-            identifier           = dict(required=False),
+            identifier           = dict(required=False, default=None),
             weight               = dict(required=False, type='int'),
             region               = dict(required=False),
             health_check         = dict(required=False),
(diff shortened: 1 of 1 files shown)


erulabs:
:+1: This fixed the issue for me!

erulabs:
This fixes the issue for me! :+1:  Thanks!

cloud/amazon/route53.py
```diff
@@ -295,7 +295,7 @@ def main():
             overwrite            = dict(required=False, type='bool'),
             retry_interval       = dict(required=False, default=500),
             private_zone         = dict(required=False, type='bool', default=False),
-            identifier           = dict(required=False),
+            identifier           = dict(required=False, default=None),
```
bcoca, line 298: this should already be the default

PR #2512, opened 2015-11-23
Fail ec2 module when group names cannot resolve to IDs

ghost:
This is a bugfix pull request.

Ansible version:

```
ansible 2.0.0 (devel 08c9e70307) last updated 2015/10/21 22:56:33 (GMT -400)
```

Summary:
The `ec2` module allows for a list of named security groups via the `group` parameter.  The module resolves the corresponding IDs from those names, but will silently succeed when any of the named security groups do not exist.  This is an unexpected result and can lead to problems in subsequent plays (and could also be a security problem as the instance will start with the `default` security group, which may leave the instance exposed).

Expected results:
fails with a message

Actual results:
succeeds without all the specified security groups

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -883,6 +883,9 @@ def create_instances(module, ec2, vpc, override_count=None):
                 grp_details = ec2.get_all_security_groups()
             if isinstance(group_name, basestring):
                 group_name = [group_name]
+            unmatched = list(set(group_name) - set([str(grp.name) for grp in grp_details]))
+            if len(unmatched) > 0:
+                module.fail_json(msg="the following group names are not valid: %s" % ','.join(unmatched))
             group_id = [ str(grp.id) for grp in grp_details if str(grp.name) in group_name ]
         # Now we try to lookup the group id testing if group exists.
         elif group_id:


willthames:
I think this could be fixed to give the user more information.

```
# Use dictionary generator for python 2.6 compatibility
group_map = dict((grp.name, grp.id) for grp in grp_details)
group_id = []
for group_name in group_names:
     if group_name not in group_map:
        module.fail_json('%s is not a valid security group name' % group_name)
     else:
         group_id.append(group_map['group_name'])
```

ghost:
good idea, @willthames.  The error message now reports all unmatched security group names.

cloud/amazon/ec2.py
```diff
@@ -883,6 +883,10 @@ def create_instances(module, ec2, vpc, override_count=None):
                 grp_details = ec2.get_all_security_groups()
             if isinstance(group_name, basestring):
                 group_name = [group_name]
+            unmatched = list(set(group_name) - set([str(grp.name) for grp in grp_details]))
+            if len(unmatched) > 0:
+                module.fail_json(
```
willthames: wouldn't bother with the line break here - longer but readable lines are preferred over strict PEP-8 line length compliance.

willthames:
Nice and simple, I like it.

willthames:
:+1:

ghost:
Got rid of the line break.  Thanks for reviewing.

abadger:
Looks good.  I would simplify this a little, though:

``` python
unmatched = set(group_name).difference(str(grp.name) for grp in grp_details)
```

len() and join() should both operate on the set fine.  This is a simple enough change that I'll apply it after merging.

PR #2516, opened 2015-11-23
Missing documentation fix

Petrox:
Extending the documentation to state, that this module does not handle externals well.

### DIFF
--- a/source_control/subversion.py
+++ b/source_control/subversion.py
@@ -28,6 +28,7 @@ version_added: "0.7"
 author: "Dane Summers (@dsummersl) <njharman@gmail.com>"
 notes:
    - Requires I(svn) to be installed on the client.
+   - This module does not handle externals
 requirements: []
 options:
   repo:


gregdek:
@dsummersl Please review this PR to make sure it adheres to the following guidelines:

(I know this is a pretty simple PR, but would be nice to get validation that this new assertion in the docs is correct and useful.)

dsummersl:
The change in this PR looks fine WRT the module checklist.

I did notice one additional change that should probably happen though:
- `For password / secret arguments no_log=True should be set`: This isn't set for the `password` documentation section.
- `Requirements should be documented, using the requirements=[] field`: The `svn` requirement is documented under `notes`, but it looks like that should migrate to `requirements`.
- `Does module use check_mode? Could it be modified to use it? Document it`: It does, its not documented.

Open a separate issues for these?

gregdek:
Missed this. @dsummersl I would say open a separate issue for the others; we'll just put this one into shipit per your advice. Thanks @Petrox for the PR.

dsummersl:
will do.

Petrox:
@gregdek Although correct docs are important, but I believe implementing the handling of externals would be a better solution, as the current subversion module is just half-working. And the funnyest bug existing is that ansible hits me with warnings for using shell: svn :) :beetle: 

Maybe in a later release module svn gets a little treatment. :+1:

PR #2521, opened 2015-11-23
docker: Fix state=reloaded to detect default LoggingDriver for docker

### LINKED ISSUE #2479: docker module reloads unnecessary if log_driver for daemon is not default
Running devel, this of code is the exact problem:

``` python
if self.ensure_capability('log_driver', False):
                expected_log_driver = self.module.params.get('log_driver') or 'json-file'
                actual_log_driver = container['HostConfig']['LogConfig']['Type']
                if actual_log_driver != expected_log_driver:
                    self.reload_reasons.append('log_driver ({0} => {1})'.format(actual_log_driver, expected_log_driver))
                    differing.append(container)
                    continue
```

When running with `state=reloaded`, if your docker-daemon has a default log and you have not specified the log_driver then a comparison is always made to `'json-file'`. So in my case where I have journald as the default on all my hosts, the driver will always restart my containers - even though it doesn't actually pass `'json-file'` to the containers on the restart.

If expected log_driver is `self.module.params.get('log_driver')` is None, it should be assumed that no changes are actually specified - or (but I do not yet know how to do this) - the daemon default driver should be used instead.

Related to #1251

EDIT: Looks like the driver should just add a call to docker_py's info() module to check the default settings.

wrouesnel:
Previously the logging module hard coded the default logging driver. This means if the docker daemon is started with a different logging driver, the ansible module would continually restart it when run.

This fix adds a call to docker.Client.info(), which is inspected if a logging driver is not supplied in the playbook, and the container only restarted if the logging driver applied differs from the configured default.

In usage, this has solved issues with using alternative logging drivers.

Closes #2479.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -1017,6 +1017,7 @@ class DockerManager(object):
 
         running = self.get_running_containers()
         current = self.get_inspect_containers(running)
+        defaults = self.client.info()
 
         #Get API version
         api_version = self.client.version()['ApiVersion']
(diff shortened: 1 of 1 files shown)


gregdek:

smashwilson:
:+1: Yep, looks good to me! :shipit:

gregdek:
Thanks again to @wrouesnel for this PR, and thanks @cove @joshuaconner @softzilla @smashwilson @zfil @ThomasSteinbach for reviewing. Marking for inclusion.

PR #2530, opened 2015-11-23
Update documentation of the 'pkg' and 'state' parameters in yum.

gerph:
The yum module allows the 'name' parameter to be given as 'pkg', in
a similar way to some of the other package managers. This change
documents this alias.

The module's 'state' parameter has two other aliases, in line with
the 'apt' action; the 'state' parameter can take 'installed' as an
alias for 'present', and 'removed' as an alias for 'absent'. These
aliases are documented.


abadger:
Thanks for the updated documentation!

PR #2532, opened 2015-11-24
initial add of nxos_config module

privateip:
This commit adds initial support for network/nxos/nxos_config.  This module
is used to work with Cisco NXOS configurations both online and offline.


gregdek:
Thanks for submitting this new module to Ansible Extras! This module is now in community review, a process that is open to all Ansible users. In order for this module to be approved, it must gain the following votes, which can be contributed by anyone other than the author of the new module:

“works_for_me”: If you have tested the module thoroughly, including testing of all of the module’s options, and if the module works for you, please add “works_for_me” in the comments.

“passes_guidelines”: If you have gone through the module guidelines and the module meets all of the requirements, please add “passes_guidelines” in the comments. Guidelines are available here: http://docs.ansible.com/developing_modules.html#module-checklist

“needs_revision”: If the module fails to work for you, or if it doesn’t meet guidelines, please add “needs_revision” in the comments with details about what needs to be fixed.

When a module has both “works_for_me” and “passes_guidelines” tags, we will promote the module for inclusion in Ansible Extras.  At this point, you will be expected to maintain the module by fixing bugs and evaluating pull requests in a timely manner. 

More information about the review process can be seen here: 
https://github.com/ansible/ansible-modules-extras/blob/devel/REVIEWERS.md

Thanks again for submitting your Ansible module!

Issue #2537, opened 2015-11-24
CRITICAL BUG REPORT: using the user module on Mac OS X will kill your groups


domibarton:
This bug report is just for tracking purposes.
Please have a look at the [origin bug report](https://github.com/ansible/ansible/issues/13266) in the ansible repo.

robynbergeron:
[module: system/user.py]

robynbergeron:
@domibarton --Thanks for submitting this bug report, and for the accompanying PR :)

@sfromm --  thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module. Please note there is an associated PR to help fix things up :)

nitzmahone:
merged to devel as https://github.com/ansible/ansible-modules-core/commit/6128845b696b41d90862f6255b9a0e08557101c3 (I added a case in the next commit to handle the additional groups as well). Thanks!

domibarton:
@robynbergeron: You're welcome. 

Thanks for your work guys!

bcoca:
backported to stable-1.9 via 611084c92f70944346ebd3f2154884148fdddec0

## 

Brian Coca

PR #2538, opened 2015-11-24
prevents groups from being wiped out on user update on OS X

domibarton:
This merge request will fix the issue #2537, where the `user` module will reset the users primary group on Mac OS X when no `group` parameter is defined.

Please also have a look at the [origin issue](https://github.com/ansible/ansible/issues/13266).


bcoca:
in current devel this seems to be fixed by : 

```

    def _make_group_numerical(self):
        '''Convert SELF.GROUP to is stringed numerical value suitable for dscl.'''
        if self.group is None:
            self.group = 'nogroup'

```

which seems better as the function is called from both create and modify user

bcoca:
hmm, nvmd, that is also in 1.9, the diff just pushed it down

gregdek:
@sfromm Please review this PR to make sure it adheres to the following guidelines:

gregdek:
@sfromm This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

tima:
@gregdek I think (someone correct me if I'm wrong) this was resolved by the commit @nitzmahone made to close out #2537. I'm not sure why this PR didn't get marked merged here. @nitzmahone's commit looks identical to the diff of what's here.

nitzmahone:
Yeah, the way I merged it didn't close the PR, but yes, it's fixed and merged to both devel and 2.0-stable.

PR #2541, opened 2015-11-24
Update ec2.py remove state tag

joshuaeke:
'exact_count' and 'state' are mutually exclusive options they should not be in the following examples:
- # Enforce that 5 running instances named "database" with a "dbtype" of "postgres" example and 
- # Enforce that 5 instances with a tag "foo" are running

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -488,7 +488,6 @@ EXAMPLES = '''
 #
 
 - ec2:
-    state: running
     key_name: mykey
     instance_type: c1.medium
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @joshuaeke. @tgerla @lwade please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

joshuaeke:
shipit

joshuaeke:
shipit

PR #2542, opened 2015-11-24
Fix rds "promote" command never promoting

maxrothman:
Previously, the `promote` command in the `rds` module would always return OK and never actually promote an instance. This was because `promote_db_instance()` had its conditions backwards: if the instance had the `replication_source` attribute indicating that it **was** a replica, it would set `changed = False` and do nothing. If the instance **wasn't** a replica, it would attempt to run `boto.rds.promote_read_replica()`, which would always fail.


willthames:
I'm not surprised I got this wrong - `replication_source` is very misleading (surely a replica is a replication destination).

The change and explanation look good to me, :+1:

PR #2544, opened 2015-11-24
Make sure all calls to ec2.get_image are in a try/catch

jjshoe:
and simplify this rats nest of code at the same time


jjshoe:
Have some touch ups to do.

PR #2549, opened 2015-11-25
Handles BotoServerError 404 error in iam_cert

### LINKED ISSUE #2547: iam_cert module occasionally returns BotoServerError 404 NotFound
As part of our automation testing environment we create and destroy IAM server certs in rapid succession.  Occasionally we encounter a race condition where a server certificate is present on AWS at the time that IAMConnection.get_all_server_certs() is executed.  But the certificate is destroyed on AWS just before IAMConnection.get_server_certificate(certificate-name-that-no-longer-exists) is executed.  This results in the following error:

```
boto.exception.BotoServerError: BotoServerError: 404 Not Found
    Sender
    NoSuchEntity
    The Server Certificate with name certificate-that-no-longer-exists cannot be found.
```

sclausson:
Fixes #2547


gregdek:
Thanks @sclausson. @defionscode please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/amazon/iam_cert.py
```diff
@@ -273,10 +273,15 @@ def main():
...
-                  certificate_body \
-                  for thing in orig_certs]
+    orig_bodies = []
+    for thing in orig_certs:
+        try:
+            orig_bodies.append(iam.get_server_certificate(thing).\
+                               get_server_certificate_result.\
+                               certificate_body)
+        except boto.exception.BotoServerError:
+            continue
```
bcoca: this ignores all errors, not a good way to 'handle' this

a more sensible one would be to use module.fail_json to send a reasonable error to the user or at least capture it and send it as part of warnings

sclausson:
@bcoca I have modified the patch to include better error handling based on your previous comment.  Please verify if this is what you had in mind.

cloud/amazon/iam_cert.py
```diff
@@ -273,10 +273,18 @@ def main():
...
-                  for thing in orig_certs]
+    orig_bodies = []
+    for thing in orig_certs:
+        try:
+            orig_bodies.append(iam.get_server_certificate(thing).\
+                               get_server_certificate_result.\
+                               certificate_body)
+        except boto.exception.BotoServerError, err:
+            if err.status == 404 and err.code == 'NoSuchEntity':
+                module.exit_json(msg="Warning: %s" % err.message)
```
bcoca: this will exit the module, when i mentioned 'warning's i meant it as part of a normal return if you did not want to exit at this point. Also it is it's own key in 2.0 `exit_json(warnings=warnings, msg=msg, ...)`

sclausson:
@bcoca ok, hopefully this latest change is what you had in mind.  If not, can you please point me to another module as an example?  thanks.

gregdek:
Thanks @sclausson for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

defionscode:
shipit

joelthompson:
FYI @defionscode @sclausson -- this has a bug in it; I've submitted #3259 to fix this bug.

juliedavila:
@sclausson can you close this? @joelthompson made a good catch on the race condition and has incorporated your commits into his mods so you still get credit.

PR #2550, opened 2015-11-25
Make digital_ocean_domain use API v2

lippylee:
digital_ocean_domain currently uses DigitalOcean's API v1 which is now no longer in use. Update it to use API v2.

### DIFF
--- a/cloud/digital_ocean/digital_ocean_domain.py
+++ b/cloud/digital_ocean/digital_ocean_domain.py
@@ -29,12 +29,9 @@ options:
      - Indicate desired state of the target.
     default: present
     choices: ['present', 'absent']
-  client_id:
-     description:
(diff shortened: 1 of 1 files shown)


abadger:
- Code looks good -- I don't use digital ocean so I can't check that the API calls are correct.  Due to DO disabling the v1 API altogether I'm willing to push this into the tree for the 2.0 release and if there's bugs we can correct them post-release.
- We need to make a note of the changes in the documentation.  You can copy the format of what we said for the main digital_ocean module: https://github.com/ansible/ansible-modules-core/commit/9c30ef8926b869e6f5b16c1a484dfbcb1b1ccb3a

bcoca:
making doc changes ourselfs

PR #2556, opened 2015-11-26
Update in Amazon IAM Policy Documentation

jaydrogers:
There were typos in the documentation that made the examples seem misleading of what was being demonstrated. This update fixes that.

### DIFF
--- a/cloud/amazon/iam_policy.py
+++ b/cloud/amazon/iam_policy.py
@@ -64,9 +64,9 @@ extends_documentation_fragment:
 '''
 
 EXAMPLES = '''
-# Create and policy with the name of 'Admin' to the group 'administrators'
+# Create a policy with the name of 'Admin' to the group 'administrators'
 tasks:
-- name: Create two new IAM users with API keys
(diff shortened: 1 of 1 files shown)


cloud/amazon/iam_policy.py
```diff
@@ -64,9 +64,9 @@
 '''
 
 EXAMPLES = '''
-# Create and policy with the name of 'Admin' to the group 'administrators'
+# Create a policy with the name of 'Admin' to the group 'administrators'
```
willthames, line 67: I think your description in the name of the task is better, even though this is an improvement over the previous version

cloud/amazon/iam_policy.py
```diff
@@ -64,9 +64,9 @@
...
 tasks:
```
willthames, line 68: Let's take the opportunity to delete the unnecessary `tasks:` and `task:` lines

willthames:
Oops, I hadn't noticed this was merged before my comments!

Issue #2330, opened 2015-10-22
win_msi troubles with Exit-Json and logcontent


davidobrien1985:
Hi,

anybody come across the issue that win_msi just fails because of Exit-Json not being able to properly return the log content?

Got an MSI that created a 9kb log file, it takes Exit-Json about 5mins (on a decent VM) to convert that to Json and then fails with this message:

``` powershell
failed: [sv0000101np.domain.local] => {"failed": true, "parsed": false}
#< CLIXML
<Objs Version="1.1.0.1" xmlns="http://schemas.microsoft.com/powershell/2004/04"><S S="Error">ConvertTo-Json : Cannot convert value to type System.String._x000D__x000A_</S><S S="Error">At C:\Users\svc_go\AppData\Local\Temp\ansible-tmp-1445485201.78-60445853908721\_x000D__x000A_</S><S S="Error">win_msi.ps1:87 char:17_x000D__x000A_</S><S S="Error">+     echo $obj | ConvertTo-Json -Depth 99_x000D__x000A_</S><S S="Error">+                 ~~~~~~~~~~~~~~~~~~~~~~~~_x000D__x000A_</S><S S="Error">    + CategoryInfo          : InvalidArgument: (:) [ConvertTo-Json], PSInvalid _x000D__x000A_</S><S S="Error">   CastException_x000D__x000A_</S><S S="Error">    + FullyQualifiedErrorId : InvalidCastFromAnyTypeToString,Microsoft.PowerSh _x000D__x000A_</S><S S="Error">   ell.Commands.ConvertToJsonCommand_x000D__x000A_</S><S S="Error"> _x000D__x000A_</S></Objs>
```

Doesn't matter how large the log file is, in my tests it always fails.

It works fine with a straight string.

Seen that before?

cchurch:
@davidobrien1985 Do you have a publicly available (i.e. one I can use to test) MSI that causes this problem?

davidobrien1985:
Unfortunately, no. Also, I can only reproduce it in the customer's environment.
They are on WS2k8R2 with PowerShell version 4.

For now I am overwriting the module and not returning any log.

In my test environment, with the same MSI, the $result.log is null by the way, which is very odd.

h0nIg:
reproduced with Pscx-3.2.0.msi , see https://pscx.codeplex.com/releases

h0nIg:
[...]

PR #2563, opened 2015-11-26
fix problem with file content and high cpu/out of memory errors

h0nIg:
fix problem with file content and high cpu/out of memory errors. i took a look and the reason is some weird stuff with ConvertTo-Json and the depth of 99. the user can still fetch the file and examine it. fixxes https://github.com/ansible/ansible-modules-core/issues/2330

### DIFF
--- a/windows/win_msi.ps1
+++ b/windows/win_msi.ps1
@@ -47,7 +47,7 @@ Else
 
 Set-Attr $result "changed" $true;
 
-$logcontents = Get-Content $logfile;
+$logcontents = Get-Content $logfile | Out-String;
 Remove-Item $logfile;
 
 Set-Attr $result "log" $logcontents;


jhawkesworth:
I have not tried the following but I wonder if doing 

$logcontents = Get-Content $logfile | Select-Object -Property \* ;

Might reduce the memory and cpu usage 

I got the idea from this page about ConvertTo-Json : 

https://technet.microsoft.com/en-us/library/hh849922.aspx

h0nIg:
not working. also tried with the following, but failed:

```
Set-Attr $result "log" $logcontents[0]
```

the funny part is: if i'm doing the following, its fast:

```
Set-Attr $result "log" $logcontents[0].Clone()
```

but the sad part is, this is slow too:

```
Set-Attr $result "log" $logcontents.Clone()
```

gregdek:
Thanks @h0nIg. @sivel please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

jhawkesworth:
@h0nIg this was still bothering me.

I found a post suggesting you use export-clixml to examine what objects and methods are contained in a structure, and then I found an example here https://technet.microsoft.com/en-us/library/hh849952.aspx so it looks like something like the following would reduce the memory usage as $logcontents is just an array of strings, rather than an array of object references.

$logcontents = Get-Content $logfile | Out-String -stream

Worth a try?

h0nIg:
omg... thats the solution. corrected commit, but this time the output will not be an array but an concatenated string of the stdout. the reason for changing is, that the user may want to .find('error string') == -1 on the full output instead of walking through the array (which you have to join and find() again). I think this change is ok, since ansible major version upgrade can break older versions

h0nIg:
since @sivel is not responding, can somebody else review the change?

ping: @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

jhawkesworth:
This change looks fine to me and if I were PR owner I would approve it.

Jon

jhawkesworth:
@nitzmahone can you take a look at this at some point?

mojo4life:
works_for_me

koloros:
i would appreciate the inclusion.

works_for_me

ping: @trondhindenes @petemounce @elventear @smadam813 @jhawkesworth @angstwad @sivel @chrishoffman @cchurch

nitzmahone:
This is a good stopgap fix until we get around to re-doing the JSON serialization to stop using Convert-ToJson. Merging to devel and backporting to 2.0.x.

jhawkesworth:
Many thanks

PR #2564, opened 2015-11-26
vsphere_guest: support putting a guest into a nested folder

luto:
## current situation

Assume the following folder structure:
- test
  - developer1
    - productC
- prod
  - customerA
  - customerB

Currently the `vm_extra_config.folder` parameter does the following:
1. vsphere_guest constructs a flat list like this: `test, developer1, productC, prod, customerA, customerB`.
2. the last entry matching `vm_extra_config.folder` in this list is used as folder

It is not possible to give a nested path like `prod/customerA`. Instead you'd give `customerA` and hope, that this name is never duplicated.
## this PR

This PR allows to give a full path to the folder, like `prod/customerB`. This is done by reconstructing the whole folder-tree in the background, then finding the proper folder, according to the path and using that.

Please consider that this is a breaking change, since just supplying `customerB` won't yield the same result as before, but an error. Supplying a top-level folder like `prod` will still work like before. Since I wasn't able to find anything in the [contributor guidelines](http://docs.ansible.com/ansible/community.html) regarding breaking changes, I guess we have the following ways to go about this:
- just merge it and let it break
- fallback to the old behavior, when the folder could not be found, then error
- add a 2nd parameter, `folder_path`, which uses the new behavior. The current parameter `folder` would behave the same, as to does now.

Just tell me which you'd prefer and I'll update this PR accordingly.

Please also note that the current behavior doesn't seem to be documented anywhere. Once you decided how we will go about this, I'll add the documentation.

### DIFF
--- a/cloud/vmware/vsphere_guest.py
+++ b/cloud/vmware/vsphere_guest.py
@@ -175,6 +175,10 @@ EXAMPLES = '''
         size_gb: 10
         type: thin
         datastore: storage001
+        # VMs can be put into folders. The value given here is either the full path
+        # to the folder (e.g. production/customerA/lamp) or just the last component
+        # of the path (e.g. lamp):
+        folder: production/customerA/lamp
     vm_nic:
       nic1:
         type: vmxnet3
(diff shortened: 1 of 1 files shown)


chrrrles:
Thanks for the pull request.   I would prefer your second option, of falling back to the old folder logic in the case that a full path search fails.

luto:
I changed the PR. So it does now first try to match the full path, then it falls back to the old behavior, of just matching the last part. I've also added a comment in the documentation.

Like before, this does not report in any way, when there are two possible folders, e.g. `customerA/lamp` and `customerB/lamp` for `lamp`. But I guess this is a different story, requiring a different PR.

Let me know if there is anything else you'd like to see changed about this.

gregdek:
Thanks @luto. @rhoop please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
@chrrrles @rhoop This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

rhoop:
@luto I agree with @chrrrles on using the second option.LGTM

rhoop:
shipit

gregdek:
Thanks again to @luto for this PR, and thanks @chrrrles @rhoop for reviewing. Marking for inclusion.

luto:
Thank you @gregdek, @chrrrles and @rhoop!

PR #2565, opened 2015-11-26
Removed unneccessary permission greed

suonto:
Commit message:
"fixed quantum_ modules to work with minimum access rights if greater access rights are not needed"

The quantum modules should not require tenant listing permissions when they do not need to list tenants. This upgrade removes the need for that when target tenant name is not explicitly provided. Instead, tenant id that is already acquired by keystone during authentication is used.

### DIFF
--- a/cloud/openstack/_quantum_network.py
+++ b/cloud/openstack/_quantum_network.py
@@ -164,17 +164,16 @@ def _get_neutron_client(module, kwargs):
 def _set_tenant_id(module):
     global _os_tenant_id
     if not module.params['tenant_name']:
-        tenant_name = module.params['login_tenant_name']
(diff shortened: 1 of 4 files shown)


bcoca:
cc @emonty @Shrews @j2sol, any sense to update the deprecated modules?

this does seem like a nice feature for those still using the older APIs

emonty:
Seems reasonable to me - doesn't change the interface, so feels like a fine bugfix. @Shrews @j2sol  ?

Shrews:
I don't think anyone wants to accept responsibility for maintaining the deprecated modules (at least I know I don't). I won't test changes to them, nor will I try to fix them since their replacements should be the focus of our development. Given those conditions, if others want this change to go in, I won't block that.

PR #2569, opened 2015-11-26
Update documentation for 'file' module to include 'diff_peek'.

gerph:
diff_peek and validate are options that are for internal use - they shouldn't be used by end users.
So we document them in the code as such.

### DIFF
--- a/files/file.py
+++ b/files/file.py
@@ -87,6 +87,12 @@ options:
       - 'force the creation of the symlinks in two cases: the source file does
         not exist (but will appear later); the destination exists and is a file (so, we need to unlink the
         "path" file and create symlink to the "src" file in place of it).'
+  diff_peek:
+    required: false
+    description:
+      - "Only check whether the file looks like binary. Returns with the parameter
+         'appears_binary' set to True or False depending on the initial content of the
+         file. This option is enabled when the option is set (to any value)."
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


bcoca:
diff_peek is not there for users, but for internal checks in the action plugins. 

The validate option is also there for compatibility with template and copy modules, as they use the file module behind the scenes.

gerph:
Ah, right; that wasn't clear.
I can think of 3 ways to capture that information...
1. Add some comments near the AnsibleModule() creation to indicate their internal use.
2. Add a new property to the dictionary for the options in the form 'internal="For use by template and copy which delegate their work to this module"', which wouldn't be used, but could identify the options in a way that /could/ be used in the future.
3. Include in the documentation section for those options 'Internal Use Only'.

(there is also "option 4: do nothing" but I believe that if there are reasons for something that aren't obvious, you should explain yourself)

What do you think? I personally like option 2, as that's the most programatic way of explaining things, and doesn't interfere with anything else.

bcoca:
currently i'm leaning for 1, but 2 might make a lot of sense when we move to using module specs to 'autodocument' more.

gerph:
I'll go with 1, and let other people argue on how it might be more programatically used.

gerph:
How's that look?

bcoca:
lgtm

PR #2570, opened 2015-11-27
Update documentation of 'assemble' module for 'remote_src' and 'ignore_hidden'.

gerph:
The assemble module documents the default value for 'remote_src' as 'True'.
Actually, it's implemented as 'False', so we correct this.

Ansible documentation commonly uses 'yes' and 'no' as the boolean choices,
so update the choices, description and default value to make this option
more consistent.

The 'ignore_hidden' option is also a boolean, but doesn't document its
choices, so we add them here.

### DIFF
--- a/files/assemble.py
+++ b/files/assemble.py
@@ -65,11 +65,11 @@ options:
     default: null
   remote_src:
     description:
-      - If False, it will search for src at originating/master machine, if True it will
-        go to the remote/target machine for the src. Default is True.
-    choices: [ "True", "False" ]
+      - If C(no), it will search for src at originating/master machine, if C(yes) it will
+        go to the remote/target machine for the src. Default is C(yes).
+    choices: [ "yes", "no" ]
     required: false
-    default: "True"
+    default: "no"
     version_added: "1.4"
   regexp:
     description:
@@ -82,6 +82,7 @@ options:
   ignore_hidden:
     description:
       - A boolean that controls if files that start with a '.' will be included or not.
+    choices: [ "yes", "no" ]
     required: false
     default: false
     version_added: "2.0"


gregdek:
Thanks @gerph. @sfromm please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

files/assemble.py
```diff
@@ -65,11 +65,11 @@
...
     description:
-      - If False, it will search for src at originating/master machine, if True it will
-        go to the remote/target machine for the src. Default is True.
-    choices: [ "True", "False" ]
+      - If C(no), it will search for src at originating/master machine, if C(yes) it will
+        go to the remote/target machine for the src. Default is C(yes).
+    choices: [ "yes", "no" ]
     required: false
-    default: "True"
+    default: "no"
```
bcoca, line 72: this is incorrect, the default is true or yes

gerph, line 72: You are mistaken; the default is 'no'.

The implementation (lines 161 onward) states:

```
     module = AnsibleModule(
         # not checking because of daisy chain to file module
         argument_spec = dict(
             src = dict(required=True),
             delimiter = dict(required=False),
             dest = dict(required=True),
             backup=dict(default=False, type='bool'),
             remote_src=dict(default=False, type='bool'),
             regexp = dict(required=False),
             ignore_hidden = dict(default=False, type='bool'),
             validate = dict(required=False, type='str'),
         ),
         add_file_common_args=True
     )
```

The default value for remote_src is 'no'. This change corrects the documentation to match the implementation.

bcoca:
It is still wrong, consider that `remote_src` is not used at all in the module (not sure why its in the spec, I need to look into that). It is used by the action plugin (of same name) which then either assembles locally and then uses the copy module or uses the assemble module otherwise.

gregdek:
@gerph A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes.

robynbergeron:
@gerph Another friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open. If we don't hear from you within another 14 days, we will close this pull request.

gregdek:
Closing due to contributor inaction.

PR #2577, opened 2015-11-28
Pass volume args to OpenStack Server creation

emonty:
There are several parameters that OpenStack and shade support but that we're not passing from ansible. That's sad for people.


gregdek:
Thanks @emonty. Since you are the owner of this module, we are marking this PR for immediate inclusion.

bcoca:
can you flatten the commits?

emonty:
Flattened/rebased

PR #2586, opened 2015-11-30
initial add of the ios_facts module

privateip:
This commit is the initial add of the ios_facts module.  This module has a dependency on ansible/ansible PR 13357 and PR 13358 being merged


gregdek:

ogenstad:
I've done some testing and I'm seeing some issues with the do_vlans, do_routing, do_interfaces functions. Basically you can't switch them off. Though I'm not quite sure the problem comes from this module.

Looking at this function module.params['include_interfaces'] always evaluates to True.
def do_interfaces(module):
    if module.params['include_interfaces']:
        return ('interfaces', 'show interfaces')

I think it's some issue with the boolean lookup. Below is the output from a -vvv

{"module_args": {"include_interfaces": "false", "include_routing": "no", "include_vlans": "false"

I believe that the param gets set to "false" or "no" instead of False in Python and anything but False or None evaluates to True.

Let me know if you need any more information about this.

ogenstad:
Another issue with this module is that I tested it against a Catalyst 2960 where the command to show vlans is "show vlan" instead of "show vlans". If I just remove the "s" it works.

However I think it would be better to have a module which can return structured data.

PR #2587, opened 2015-11-30
initial add of ios_command module 

privateip:
This commit is the initial add of the ios_command module.  This module has a dependency on ansible/ansible PR 13357 and PR 13358 being merged


gregdek:

PR #2588, opened 2015-11-30
initial add of eos_command module

privateip:
This is the initial add of a new module for sending commands to eos devices
over eapi.  This module requires the eapi shared module in order to properly
run.


gregdek:

PR #2589, opened 2015-11-30
initial add of eos_facts module

privateip:
This commit push the initial add of a module for collecting facts from
eos devices over eapi.   This module requires the eapi shared module be
available in order to properly execute.


gregdek:

PR #2590, opened 2015-11-30
Fix ini_file module, avoid bracklets in key break idempotence

### LINKED ISSUE #2475: ini_file module: bracklets in key break idempotence
Issue Type: Bug Report
Ansible Version: ansible-playbook 2.0.0 (devel 5d645f3a8b) last updated 2015/11/13 13:27:13 (GMT +200)
Environment: Ubuntu 15.04
Summary: When using ini_file for modifying /etc/php5/fpm/pool.d/www.conf it will not recognize strings  which include "[" or "]" in key and append it again to the file. Using Ansible 1.9.4 this works just as expected.

Example: You want to set

```
php_admin_value[sendmail_path]: "/usr/sbin/sendmail -t -i -f an@onymo.us"
```

using this hash:

``` yaml
php_admin_value[sendmail_path]:
    value: '"/usr/sbin/sendmail -t -i -f an@onymo.us"'
```

On first run Ansible will append this string to the file just as expected. On second run it will append it again, even though it's already present. When removing the bracklets it will only be inserted once.

Removing the bracklets from value will work as expected (not changed one second run)

noseka1:
fixes #2475

### DIFF
--- a/files/ini_file.py
+++ b/files/ini_file.py
@@ -100,6 +100,22 @@ import sys
 import os
 
 # ==============================================================
+# match_opt
+
+def match_opt(option, line):
+  option = re.escape(option)
+  return re.match('%s *=' % option, line) \
+    or re.match('# *%s *=' % option, line) \
+    or re.match('; *%s *=' % option, line)
+
+# ==============================================================
+# match_active_opt
+
+def match_active_opt(option, line):
+  option = re.escape(option)
+  return re.match('%s *=' % option, line)
+
+# ==============================================================
 # do_ini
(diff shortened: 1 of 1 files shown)


robynbergeron:
Hi @noseka1 -- thanks for submitting this PR. Since you're the owner, marking it for inclusion --- cc @ansible/core team :)

bcoca:
@noseka1 can you flatten the commits?

noseka1:
@bcoca Ok, flattened as requested.

PR #2592, opened 2015-11-30
Fix ec2.py bug when it does not respect network_id

akazakov:
If instance already exists, but in a differnet VPC, we should
not count towards exact_count
## Example

Imagine the followin situation

VPC_A:
  Subnet_A:
    Instance1:
      Tags:
        Name: my_instance

VPC_B:
  Subnet_B: {}

Now if we call ec2 module:

ec2:
  image_id: ...
  vpc_subnet_id: Subnet_B
  exact_count: 1
    count_tag:
      Name: "my_instance"

Without the fix, this will result in no-op, as if instance was present in
Subnet_B

Even weirder behaviour would be observed if you call it with exact_count

> 1, because now you'd end up with some copies in the right subnet, and
> 1 in the wrong (Subnet_A)

With the fix, it will create necessary count in Subnet_B and ignore
Subnet_A

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -553,10 +553,10 @@ except ImportError:
     HAS_BOTO = False
 
 
-def find_running_instances_by_count_tag(module, ec2, count_tag, zone=None):
+def find_running_instances_by_count_tag(module, ec2, vpc, count_tag, zone=None):
 
     # get reservations for instances that match tag(s) and are running
-    reservations = get_reservations(module, ec2, tags=count_tag, state="running", zone=zone)
+    reservations = get_reservations(module, ec2, vpc, tags=count_tag, state="running", zone=zone)
 
     instances = []
     for res in reservations:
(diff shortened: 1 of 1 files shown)


cserafin1124:
This code is useful and seems good. Please merge this!

akazakov:
@bcoca Is there anything I can do to help this merged?

copumpkin:
:shipit:

gregdek:
Thanks @akazakov. @tgerla @lwade please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

akazakov:
@tgerla @lwade  @gregdek ping?

akazakov:
Can someone please look at this???

jsmartin:
I'm not a maintainer (and do not aspire to be), but this makes logical sense to me.  +1

akazakov:
@gregdek Can you please merge this? It has 3 votes from other people to merge this.

ryansydnor:
Agree with @jsmartin here. :shipit: !

gregdek:
Moving this module to @ansible/core review. @akazakov sorry for the delay.

akazakov:
ping?

PR #2593, opened 2015-11-30
initial add of eos_config module

privateip:
This commit adds a new module for working with Arista EOS configurations. The
eos_config module requires module_utils/eapi.py shared module for proper
operation


gregdek:

Issue #2377, opened 2015-10-28
unarchive: confusing error message when src points to directory instead of archive file


rodush:
**Ansible Installation**

> ansible 1.9.2
>    configured module search path = None`

**OS**

> MacOS X El Capitan 10.11.1

When I made a mistake in my task - 

``` yml
- name: Unpack archive
  unarchive: src={{ java_download_dir }}/ dest={{java_download_dir}} copy=no
```

where `java_download_dir = /tmp`
I got this message:
**Current Result**

> msg: Failed to find handler to unarchive. Make sure the required command to extract the file is installed.

which is basically confusing.
**Expected Results**
 It is better to say that the `src` is not an archive file, or something like that.

robynbergeron:
[module: files/unarchive.py]

robynbergeron:
Hi @rodush -- thanks for submitting this bug or possibly documentation bug report :)

@pileofrogs -- thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module. Marking as a regular bug report for now.

pileofrogs:
Yeah, that's a pretty lame error message.  I'll go fix that.

pileofrogs:
Just made PR #2594 in response to this bug

Jmainguy:
@rodush does #2594 fix this for you? Looks like it got merged.

ansibot:
@dagwieers, @pileofrogs, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

dagwieers:
Fix has been merged.
Do we need to keep this open in order to get a response from the original reporter @rodush ?

resolved_by_pr #2594

PR #2594, opened 2015-11-30
improved error message when no handler found

pileofrogs:
In response to issue #2377


gregdek:
Thanks @pileofrogs. Since you are the owner of this module, we are marking this PR for inclusion.

PR #2596, opened 2015-12-01
initial add of the ios_config module

privateip:
This commit adds a new module for working with cisco IOS device configurations.
It requires both module_utils/ssh.py and module_utils/ios.py.  This module
has an external dependency on paramiko (inhereted from module_utils/ssh.py)


gregdek:

gregdek:

PR #2520, opened 2015-11-23
Add stop_timeout option to docker-module.

wrouesnel:
stop_timeout allows specifying a timeout for commands which can stop or restart docker containers (the time before SIGKILL is sent to the container).

The default behavior is the docker-py default (currently 10s).

I've tested and deployed a local version of this module in production, and it fills a big gap in ansible functionality with docker when working with containers with longer shutdown times.

PR #2606, opened 2015-12-02
State == "stopped" timeout override support

daviddyball:
This is a simple change to make use of `docker.client.Client.stop()`'s `timeout` option that allows you to override the amount of time to wait when setting a containers state to `stopped`.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -297,6 +297,12 @@ options:
         retries.
     default: 0
     version_added: "1.9"
+  stop_timeout:
+    description:
+      - Maximum number of seconds to wait for a container top stop when state ==
+        "stopped"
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @daviddyball. @cove @joshuaconner @softzilla @smashwilson please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

smashwilson:
Already inbound as #2520. Comparing the two, I think I prefer that one because it uses the timeout for `state=restarted` and `state=reloaded` as well.

daviddyball:
@smashwilson  NP. Feel free to close this one.

PR #2655, opened 2015-12-09
Set the argument type for ec2_vol's encrypted parameter

michel-slm:
If this is not set, Ansible parses the parameter as a string.
This is fine if the parameter is not provided by the caller, but
if it is set to False or True explicitly, ec2_vol receives this as
the string 'False' or the string 'True', both of which are truthy.

Thus, without this fix, setting the parameter results in encryption
always enabled.

[Review on Reviewable](https://reviewable.io/reviews/ansible/ansible-modules-core/2655)


willthames:
Yep, makes sense, :+1:

gregdek:
Thanks @willthames for the review and thanks @michel-slm for the PR. Marking for inclusion.

PR #2657, opened 2015-12-09
Add assume_role_policy option to iam role creation

rcarton:


gregdek:
Thanks @rcarton. @defionscode @seiffert please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
@defionscode @seiffert This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

fiunchinho:
:+1: I really need this!

gregdek:
@defionscode @seiffert still waiting on your review.  Please comment with text 'shipit' or 'needs_revision' as appropriate. If we don't hear from you within 14 days, we will start to look for additional maintainers for this module.

captainkerk:
Awesome! I was just needing this ability earlier today. Ended up just editing the dang thing through the aws console (not good!)

I like this, however, I was thinking of a different way to pass in the assume_role_policy_document, using just straight YAML. Here's how i thought the task definition would look:

```
- name: test
  iam:
    iam_type: role
    name: test5
    state: present
    assume_role_policy:
      Version: "2012-10-17"
      Statement:
        - Sid: test1
          Effect: Allow
          Principal: {'AWS': 'arn:aws:iam::xyz:root'}
          Action: 'sts:AssumeRole'
```

I got this working on my fork. It might not conform to the ansible norm (as i recall args are generally of the key=value type) but I believe this has advantages compared to both types proposed here:
1. from external file: separating the document and the task makes sense when its a large document, like a cloudformation template. Since these policies are generally small, placing it inline is beneficial in terms of maintainability.
2. json string: while they are small, i think the character count of these things could make for some long lines? 

Thoughts?

gregdek:
@defionscode ping?

robynbergeron:
@defionscode @seiffert This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
@defionscode @seiffert still waiting on your review.  Please comment with text 'shipit' or 'needs_revision' as appropriate. If we don't hear from you within 14 days, we will start to look for additional maintainers for this module.

gregdek:
Thanks @rcarton for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

rcarton:
I have kind of given on the PR tbh, we use our own set of modules internally now since it takes so long to get things merged/fixed. I think I made new revisions for this in our local code, but it has diverged from the original ansible core module. Maybe you guys should adopt a different policy when it comes to importing modules? More like actual libraries than a core set of libraries.

defionscode:
@rcarton Apologies it takes a while, it's something that will be improved relatively soon.

As far as this PR, similar to what i said in #2864 would you be willing to add the logic/code to support YAML defined policies? 

I definitely want to get this merged in, it'll just come down to who is more available/willing to do the above either you or @captainkerk

defionscode:
Ok so update on this. I'd like for this to be merged in BEFORE #2864 at which point @captainkerk has stated his willingness to add YAML policy support after this gets merged.

cc/ @bcoca

juliedavila:
Let's go ahead and close this in favor of  #2864

juliedavila:
Closing in favor of #2864

PR #2659, opened 2015-12-09
Remove a unneeded use of use_unsafe_shell

mscherer:
code we have to toroughly inspect for a security audit.


hloeffler:
:+1:

PR #2661, opened 2015-12-09
Split a shell snippet in 2 to avoid using use_unsafe_shell=True

mscherer:
code we have to toroughly inspect for a security audit.

In this case, the '&&' can be replaced by doing 2 calls to run_command.

### DIFF
--- a/packaging/os/apt_repository.py
+++ b/packaging/os/apt_repository.py
@@ -106,7 +106,10 @@ def install_python_apt(module):
     if not module.check_mode:
         apt_get_path = module.get_bin_path('apt-get')
         if apt_get_path:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @mscherer. @sashka please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
@sashka This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

sashka:
@robynbergeron: code looks OK, but it needs testing.
If someone tested and it worked, ship it then.  Sorry, I have almost no free time to test it on my own.

gregdek:
I'm going to assume that @mscherer tested this, and it looks to me like it's important enough to warrant a shipit, so that's what I'm doing.

PR #2668, opened 2015-12-10
added a reference to the template module for clarity

fields:
The current documentation for the copy module does not include any reference to the template module, and it should.


gregdek:
Thanks @fields for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2676, opened 2015-12-12
make the system module call daemon-reload on systemctl control

oaken-source:
I would like to propose a change in the way the service module works on systemd based machines, like arch linux.

With systemd, when you change a service file on disk, systemctl will not pick up the new file, until you restart the machine or call `systemctl daemon-reload` explicitly. This can become a problem when you change e a service file significantly through ansible and then `service restart` the service, because the changes are not picked up and it could either fail, or worse, succeed with a warning the ansible user never sees and do something entirely unexpected.

IMO, ansible should hide this complexity and call `systemctl daemon-reload` implicitly on control tasks. See the (very crude) attached patch.

this issue has been discussed in #191 and seems to have gained some support.


gregdek:
Thanks @oaken-source for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

djmattyg007:
What does this do in check mode? Is there a way to easily check to see whether or not the cached unit data for a systemd unit (service or otherwise) is out of date and needs to be reloaded?

oaken-source:
it has not been incorporated yet, but `systemctl status foo.service > /dev/null` prints a warning like 

```
Warning: foo.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

which could serve as an indicator in check mode.

xenithorb:
Is this going to gain any traction? I too am wondering why the service module doesn't already deal with this. If the point of the service module is to abstract dealing with services between various types of systems, then this is an essential part of dealing with systemd systems. Otherwise we have to add additional checks.

bcoca:
There is a systemd module that allows you to control this now, the service module should not do this automatically as it can be problematic on existing systemd systems to reload too often, we are leaving it as an explicit option instead of making it implicit. 

So I'm going to close this PR in favor of using the systemd module when this is needed.

PR #2679, opened 2015-12-13
Added state=check option

tembaby:
Adding support to check the existence of a line in a file without changing the file by adding or deleting the matched line. Reasons behind this change:
- It's very logical to have this functionality in the "lineinfile" module
- Similar functionality is found in the "find" module, but it's not available ansible prior to version 2.0, version 1.X is still widely used, so a lot of people will benefit from this functionality in lineinfile.


bcoca:
Even if this were added to lineinfile, it would not be available until Ansible 2.1 and file already has this in 2.0.

PR #2681, opened 2015-12-13
Pass network argument to add_ips_to_server

timrchavez:
The `network` argument needs to be passed through add_ips_to_server so that the default value can be properly overridden from ansible.

### DIFF
--- a/cloud/openstack/os_floating_ip.py
+++ b/cloud/openstack/os_floating_ip.py
@@ -155,8 +155,9 @@ def main():
 
         if state == 'present':
             server = cloud.add_ips_to_server(
-                server=server, ips=floating_ip_address, reuse=reuse,
-                fixed_address=fixed_address, wait=wait, timeout=timeout)
+                server=server, ips=floating_ip_address, ip_pool=network,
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @timrchavez. @emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

omgjlk:
shipit

gregdek:
Thanks again to @timrchavez for this PR, and thanks @emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz for reviewing. Marking for inclusion.

bcoca:
is this a new feature? does it fix a bug? does the user need to be notified of the change in behavior?

timrchavez:
@bcoca I'd say it fixes a bug.  Shade allows you to override the default network and `os_floating_ip` allows you to specify the the network you wish to override the default with, but it never gets passed into shade, and so you'll always get the default.

PR #2683, opened 2015-12-14
Do not escape pipe in the command, fix #2597

### LINKED ISSUE #2597: Impossible to use pipe in validate option of copy module
##### Issue Type:

Bug Report
##### Ansible Version:

1.9.2 although this is true for older versions.
##### Ansible Configuration:

N/A
##### Environment:

N/A
##### Summary:

When using a pipe (`|`) in a `validate` command of the `copy` module, the command run because the pipe gets escaped.

This is a problem when the command being given does not discriminate on the return code but its output differs whether the check passes or fails. An example of this is `openssl dhparam -check`, which outputs `DH parameters appear to be ok` when passing and `p value is not prime` and/or `the g value is not a generator` when failing.

I haven't checked on other modules having a `validate` option but my guess is that the issue applies there too.
##### Steps To Reproduce:

Assuming `validate_command` always return `0` and outputs `passing`/`failing`:

``` yaml
- copy:
  src: src_path
  dest: dest_path
  validate: validate_command %s | grep -q 'passing'
```
##### Expected Results:

The file being copied if `validate_command` outputs `passing`, stoping the playbook execution otherwise.
##### Actual Results:

``` sh
failed: [default] => {"checksum": "66b...039", "cmd": "validate_command /home/user/.ansible/tmp/ansible-tmp-144...013/source '|' grep -q 'passing'", "failed": true, "rc": 2}
```

The pipe gets escaped, preventing the validation command to pass at all.

mscherer:
However, I just want to open the discussion on the fix (see last comment on the bug for why), so do not merge unless confident this doesn't add security issue.

### DIFF
--- a/files/copy.py
+++ b/files/copy.py
@@ -306,7 +306,7 @@ def main():
                     module.set_mode_if_different(src, mode, False)
                 if "%s" not in validate:
                     module.fail_json(msg="validate must contain %%s: %s" % (validate))
-                (rc,out,err) = module.run_command(validate % src)
+                (rc,out,err) = module.run_command(validate % src, use_unsafe_shell=True)
                 if rc != 0:
                     module.fail_json(msg="failed to validate: rc:%s error:%s" % (rc,err))
             if remote_src:


gregdek:

abadger:
@mscherer Thanks for taking a look at this but we've decided that we didn't want to introduce unsafe_shell here.  I wrote up a reply in the bug ticket.  The security issues you raised there and the difficulty of writing good shell scripts made us decide that we really wanted validate to be used for simple commands.  I added some examples of workarounds to the ticket so that the original reporter could see how to workaround this limitation.

PR #2692, opened 2015-12-15
Adding gathering environment variables to gather_facts on Windows

hantonov:
This functionality is missing from setup.ps1 and exists in gather_facts on Linux. 
Will get all defined env variables for the user at the time this module is running.

### DIFF
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -81,6 +81,16 @@ $ips = @()
 Foreach ($ip in $netcfg.IPAddress) { If ($ip) { $ips += $ip } }
 Set-Attr $result.ansible_facts "ansible_ip_addresses" $ips
 
+$env_vars = New-Object psobject
+foreach ($item in Get-ChildItem Env:)
+{
+    $name = $item | select -ExpandProperty Name
+    # Powershell ConvertTo-Json fails if string ends with \
+    $value = ($item | select -ExpandProperty Value).TrimEnd("\")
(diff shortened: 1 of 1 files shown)


windows/setup.ps1
```diff
@@ -81,6 +81,16 @@ $ips = @()
 Foreach ($ip in $netcfg.IPAddress) { If ($ip) { $ips += $ip } }
 Set-Attr $result.ansible_facts "ansible_ip_addresses" $ips
 
+$env = New-Object psobject
```
ryansydnor: I wouldn't use $env as a variable name. It could interfere with an attempt to retrieve a particular environment variable further down in the file.

The following will now not function as expected.

``` powershell
$env:MY_ENV_VAR
```

ryansydnor:
looks good to me! shipit

copumpkin:
:shipit:

gregdek:
Thanks @hantonov for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

gregdek:
(But we do have multiple community folks +1 here, so may be worth a quicker look than usual; cc: @ansible/core here.)

trondhindenes:
Nice addition! Looks good here, shipit.

jhawkesworth:
Yes, just tested this change locally against 2.0.0. rc4, so please shipit

PR #2693, opened 2015-12-15
Revert "Fixes #4448 reloaded services should be started if not running"

### LINKED ISSUE #350: service: state=reloaded starts service if not running
This differs from what `service mydaemon reload` or `/etc/init.d/mydaemon reload` do in most cases (because these usually try to send SIGHUP to a running process and do nothing if they can't find one).

For example when using Nginx, `service nginx reload` will not do anything if Nginx isn't already running. Ansible, instead of replicating this behaviour, [starts the service](https://github.com/ansible/ansible-modules-core/blob/devel/system/service.py#L295) when asked only to reload and it isn't found running.

pabelanger:
This breaks how service mydaemon or /etc/init.d/mydaemon reload
actually work.

Closes #350

This reverts commit f27e2fd285c6014666249b6a71480c2b7af2b465.

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -280,7 +280,7 @@ class Service(object):
         if self.state and self.running is None:
             self.module.fail_json(msg="failed determining service state, possible typo of service name?")
         # Find out if state has changed
-        if not self.running and self.state in ["started", "running", "reloaded"]:
+        if not self.running and self.state in ["started", "running"]:
             self.svc_change = True
         elif self.running and self.state in ["stopped","reloaded"]:
             self.svc_change = True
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @pabelanger for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

omame:
Is there any news about this pull request?
It's an easy fix for a very bad behaviour but it's taking one month for each line changed.

JoePricey:
@omame Comments in the pull request suggest it will probably be left as it is: https://github.com/ansible/ansible-modules-core/issues/350#issuecomment-215507274

Issue #2708, opened 2015-12-18
unarchive permission issues with become and http src


gregsidelinger:
##### ISSUE TYPE

Bug Report
##### COMPONENT NAME

unarchive module
##### ANSIBLE VERSION

N/A
##### SUMMARY

The unarchive module does not appear to honor become permissions when downloading a remote file.  It looks like the module tries to download the file to ansible tmp directory as the become_user however that user does not have permissions to write there since they are owned by the user ansible is running as. It only appears to work when become_user is root.

Here is an example from my playbook that fails.

```
- name: Extract Zip
  unarchive: src=http://someserver.com/files/zipfile.zip dest=/opt/exploded copy=no
  become: yes
  become_user: user2
```

Jmainguy:
using latest 2.1 devel, I am unable to reproduce this. Can you reproduce this on latest devel?

gregsidelinger:
Yes the bug still exists in the 2.1 devel branch.

Here is the error message I get.

```
TASK [Extract Zip] *************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Failure downloading https://github.com/ansible/ansible-modules-core/archive/devel.zip, [Errno 13] Permission denied: '/tmp/ansible-tmp-1451356456.51-72820795886431/devel.zip'"}
```

Here is my test playbook which I ran on a pretty vanilla CentOS 7.1 system.  I created two local users, user1 and user2.  Granted user1 full sudo access in /etc/sudoers and setup a ssh key.

The playbook works fine if I unarchive the files as root because root is able to write to the ansible tmp directory unlike user2. 

```
---
- name: Reproduce bug 2708
  hosts: localhost
  remote_user: user1

  tasks:
    - name: Create /opt/exploded
      file: path=/opt/exploded owner=user2 state=directory
      become: yes
      become_user: root

    - name: Extract Zip
      unarchive: src=https://github.com/ansible/ansible-modules-core/archive/devel.zip dest=/opt/exploded copy=no
      become: yes
      become_user: user2
```

I did submit a patch last week to get around the issue of the module using the ansible tmp dir to download the file too.

abadger:
Does this still happen with the recently released 2.1?  I changed the way permissions on the temporary directory are assigned so I'm not certain this is still a problem (I tested locally by becoming nobody but I see that it was sometimes tricky to reproduce so I want to make sure I'm not missing something.)

gregsidelinger:
[...]

PR #2712, opened 2015-12-20
Fixed an issue with unarchive using the ansible temporty directory

gregsidelinger:
Here is my simple fix for issue #2708 which I reported a few days ago.  The unarchive module now creates it's own temp directory when needing to download the file instead of using the directory the module is running out of.  The cleanup of the temporary dir most likely should be handled in a cleaner way as I'm sure there are a few cases where it could be left around.


robynbergeron:
Thanks @gregsidelinger. @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

pileofrogs:
shipit

Looks good.  Thanks!

gregdek:
Thanks again to @gregsidelinger for this PR, and thanks @pileofrogs for reviewing. Marking for inclusion.

bcoca:
this changes the current behavior w/o any of it being apparent to the user, in any case this might need a more generic fix if become + temp files are in play.

robynbergeron:
@gregsidelinger Looks like @bcoca thinks this needs_revision -- marking as such. 

@pileofrogs -- just pinging as an FYI to you on this, in case you have any thoughts.

robynbergeron:
@gregsidelinger -- when this is ready for review, please let us know using the ready_for_review comment, and should probably ping @bcoca as well so he knows to take a look as he had the revision request.

gregsidelinger:
@bcoca can you give some details on your issues with the patch.  I actually though it was the cleanest way to get around the issue as it did not require a bunch of extra logic.

However I was thinking of adding a new option to the module that would let you specify the tempory location to download the file to.  That way the current behavior could be maintained and users could specify the location if needed.   

Here is the description of the option as I'm thinking of it.

```
  tmp_path:
    description:
      - If copy=no and src contains :// the location the file will be downloaded to before being unarchived. The file will be removed after it's unarchived.
      - tmp_path supports two special values which will pick location for you. 'ansible_tmp' will use the ansible temp directory while 'os_tmp' will use the standard OS tmp location as returned by a call to mkdtemp.
    required: false
    default: "ansible_tmp"
    version_added: "2.0"
```

gregdek:
Thanks @gregsidelinger for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregdek:
Thanks @gregsidelinger. @nitzmahone please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
OK, so something very bad has happened with this PR, as it has 156 file changes and 250+ commits in it -- and since it seemed like @bcoca was giving it a -1 anyway, I'm going to close this and recommend you create a new PR.

PR #2713, opened 2015-12-20
Fixed issue 1113 to enable updating ec2 subnet tags

autotune:
##### Issue Type:
- Bugfix Pull Request
##### Ansible Version:

```
[autotune@ansibledev amazon]$ ansible --version
ansible 2.1.0 (devel 15135f3c16) last updated 2015/12/16 16:15:58 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 0e043f8c58) last updated 2015/11/26 16:32:37 (GMT     -700)
  lib/ansible/modules/extras: (detached HEAD e46e2e1d6f) last updated 2015/11/26 16:32:42 (GMT -700)
  config file =
```
##### Ansible Configuration:

Nothing special besides using localhost or a local IP. Full example config with test account on AWS for testing the module is located under the section "Steps to Reproduce."
##### Environment:

```
[autotune@ansibledev 1113]$ cat /etc/redhat-release
CentOS Linux release 7.1.1503 (Core)
```
##### Summary:

This fix is in regards to bug 1113 in which existing subnets are not updated when new tags are added or modified for said subnets. This works for both the Name format:

```
resource_tags:
    Name: 'Fixed Awesomeness'
```

and the custom resource_tags format:

```
resource_tags: {'We good?':'Oh yeah'}
```
## awskeys.yml encrypted with ansible-vault

```

---
  - hosts: 10.0.2.15
    tasks:
      - include_vars: awskeys.yml
      - name: Modify ec2 subnet tags
        ec2_vpc:
          state: present
          cidr_block: 192.168.0.0/16
          resource_tags:
              Name: 'Alpha Testing2'
          internet_gateway: True
          region: us-east-1
          subnets:
            - cidr: 192.168.3.0/24
              az: us-east-1a
              resource_tags: {'we good?':'oh yeah'} 
...
```
##### Expected Results:

PLAY RECAP will show changed=1 and AWS account will show new or modified tags under subnet 192.168.3.0/24 in az us-east-1a if they did not exist with those values before. 
##### Actual Results:

PLAY RECAP will show changed=1 and AWS account will show new or modified tags under subnet 192.168.3.0/24 in az us-east-1a if they did not exist with those values before.

### DIFF
--- a/cloud/amazon/elasticache.py
+++ b/cloud/amazon/elasticache.py
@@ -50,7 +50,7 @@ options:
     default: cache.m1.small
   num_nodes:
     description:
-      - The initial number of cache nodes that the cache cluster will have
+      - The initial number of cache nodes that the cache cluster will have. Required when state=present.
     required: false
   cache_port:
     description:


robynbergeron:
Thanks @autotune. @carsongee please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/amazon/ec2_vpc.py
```diff
@@ -406,9 +406,21 @@ def create_vpc(module, vpc_conn):
...
                 if subnet['cidr'] == csn.cidr_block:
                     add_subnet = False
+                subnet_tags_current = (set(new_subnet_tags.items()).issubset(set(csn.tags.items())))
+                if subnet_tags_current != True:
+                    try:
+                        changed = True
+                        create_subnet_tag = vpc_conn.create_tags(csn.id, new_subnet_tags)
+
+                    except EC2ResponseError, e:
+                        module.fail_json(msg='Unable to create resource tag {0}, error: {1}'.format(new_subnet_tags, e))
```
carsongee: This doesn't need to change, but this line is a bit long, maybe:

``` python
module.fail_json(
    msg='Unable to create resource tag {0}, error: {1}'.format(new_subnet_tags, e)
)
```

carsongee:
Thanks for finding this and looking to fix it. I'd like to see this handle deleting tags as well instead of just adding missing ones only. So doing a set intersection of tags for each subnet and if there are diffs, set the tags in AWS equal to the ones defined in the play would be nice.

autotune:
Sure thing! I'll aim to add that feature under a "state=absent" condition so tags aren't unexpectedly deleted, similar to vpc_id.

carsongee:
The module deletes all subnets completely with `state=absent` so tags won't matter with that action as all of them will be removed.  This should be to support changing subnet resource tags as a modification, i.e. I have a subnet that currently has `foo: bar` and `a: b`, I decide I don't need `a: b`, but I want to add `c: b` so I both remove a tag and add one from the subnet with the action.  That isn't a absent state, but just a small property change to an existing resource

autotune:
@carsongee this should be more along the lines of what you're looking for. Now requires subnet resource tags to be specified as a resource tag, and if they aren't then those tags are deleted. New tags are added if they do not currently exist and have been specified as a resource tag.

autotune:
Hey folks. See new branch for this issue at https://github.com/ansible/ansible-modules-core/issues/2750.

PR #2715, opened 2015-12-21
Use JSON credentials for GCE modules

viglesiasce:
This PR enables the use of JSON credentials for the GCE modules. It is part of the fix for https://github.com/ansible/ansible/issues/13515 and is paired with https://github.com/ansible/ansible/pull/13623. 

cc @erjohnso

### DIFF
--- a/cloud/google/gce.py
+++ b/cloud/google/gce.py
@@ -63,9 +63,15 @@ options:
   pem_file:
     version_added: "1.5.1"
     description:
-      - path to the pem file associated with the service account email
+      - (deprecated) path to the pem file associated with the service account email
     required: false
     default: null
+  credentials_file:
+    version_added: 2.1.0
+    description:
+      - path to the JSON file associated with the service account email
+    default: null
+    required: false
   project_id:
     version_added: "1.5.1"
     description:
@@ -129,7 +135,7 @@ options:
 
 requirements:
     - "python >= 2.6"
-    - "apache-libcloud >= 0.13.3"
+    - "apache-libcloud >= 0.17.0"
 notes:
   - Either I(name) or I(instance_names) is required.
 author: "Eric Johnson (@erjohnso) <erjohnso@google.com>"
@@ -162,13 +168,13 @@ EXAMPLES = '''
     image: debian-6
     zone: us-central1-a
     service_account_email: unique-email@developer.gserviceaccount.com
-    pem_file: /path/to/pem_file
+    credentials_file: /path/to/json_file
     project_id: project-id
   tasks:
     - name: Launch instances
       local_action: gce instance_names={{names}} machine_type={{machine_type}}
                     image={{image}} zone={{zone}} service_account_email={{ service_account_email }}
-                    pem_file={{ pem_file }} project_id={{ project_id }}
+                    credentials_file={{ credentials_file }} project_id={{ project_id }}
       register: gce
     - name: Wait for SSH to come up
       local_action: wait_for host={{item.public_ip}} port=22 delay=10
@@ -445,6 +451,7 @@ def main():
             service_account_email = dict(),
             service_account_permissions = dict(type='list'),
             pem_file = dict(),
+            credentials_file = dict(),
             project_id = dict(),
             ip_forward = dict(type='bool', default=False),
             external_ip = dict(choices=['ephemeral', 'none'],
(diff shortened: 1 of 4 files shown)


cloud/google/gce.py
```diff
@@ -71,9 +71,15 @@
   pem_file:
     version_added: "1.5.1"
     description:
-      - path to the pem file associated with the service account email
+      - (deprecated) path to the pem file associated with the service account email
```
bcoca: when adding a deprecation message you should note what takes its place

cloud/google/gce.py
```diff
@@ -141,7 +147,7 @@
 
 requirements:
     - "python >= 2.6"
-    - "apache-libcloud >= 0.13.3"
+    - "apache-libcloud >= 0.17.0"
```
bcoca: this is not a change we do lightly, we normally prefer to have version detection code in the modules that will act according to which version is available. Otherwise this forces existing installations to upgrade libcloud when updating Ansible and that is not something we want to do.

cloud/google/gce.py
```diff
@@ -174,14 +180,15 @@
     image: debian-6
     zone: us-central1-a
     service_account_email: unique-email@developer.gserviceaccount.com
-    pem_file: /path/to/pem_file
+    credentials_file: /path/to/json_file
```
bcoca, line 187: instead of just changing it, have old and new examlpe explaining the deprecation and the library version difference

bcoca:
cc @erjohnso

bcoca:
@viglesiasce, in general we don't like breaking changes, may i suggest?:
- make credentials_file the new option and keep pem_file as an alias
- detect the library version and give an error when attempting to use credential file with old version
- warn user when using pem file, this seems to already be the case as mentioned in ansible/ansible#13515 

Having it deprecated does not mean we remove it right away, if possible we would prefer to support those that cannot change/update in a fluid manner as well as those that can be in the cutting edge.

viglesiasce:
Thanks for the feedback @bcoca!! 
- Yep the PR to ansible core https://github.com/ansible/ansible/pull/13623 does the aliasing [here](https://github.com/ansible/ansible/pull/13623/files#diff-b6bc17b51c779b2ac4eefcfc5504a83bR53) so there shouldnt be any backwards incompatible change. Is there a better/more official way to do it? I wasn't sure the "right way" to deprecate something in the Ansible docs so thanks for the pointer. Is there an example of deprecated config params? Any formatting conventions?
- Agree on the lib version check and giving an error when JSON won't work, I'll update accordingly

Once again, thanks for your help!

cloud/google/gce.py
viglesiasce: Does this actually change the Ansible installation process? I was under the impression that this was just guidance since the GCE guide had [users installing libcloud](http://docs.ansible.com/ansible/guide_gce.html#introduction) outside of the Ansible process.

bcoca:
an example of the deprecation can be `sha256sum` in the `get_url` module, though in this case it might make sense to just make it an alias, either way works for me.

cloud/google/gce.py
bcoca: no, not 'yet'. but this documents the code so if users see this they'll feel they HAVE to upgrade
`- "apache-libcloud >= 0.13.3, >= 0.17.0 if using JSON credentials"`
might be a better change

erjohnso:
Outside of @bcoca's feedback, this lgtm. Thanks @viglesiasce!

bcoca:
another thought i had, does it make sense to 'centralize' this code in module_utils? it seems to be repeated across all the modules.

viglesiasce:
@bcoca agree on the centralization but I'd prefer to do it in a separate PR after this one makes it in. 

One question I had with aliases, is it mutually exclusive with deprecation? Namely do I leave in the docs with the deprecation warnings in AND add pem_file as an alias to credentials_file? Or do I do the deprecation stuff first then on a next pass, when removing the docs, add pem_file as an alias to credentials_file?

viglesiasce:
I believe everything has been addressed. Aliasing is out for now as I assume that it should be done when pem_file is fully removed.

erjohnso:
I think this is good to go.

shipit

robynbergeron:
Thanks @viglesiasce. @erjohnso please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

robynbergeron:
Thanks again to @viglesiasce for this PR, and thanks @erjohnso for reviewing. Marking for inclusion.

viglesiasce:
Thanks folks!

PR #2719, opened 2015-12-22
command: Add warnings for dnf and zypper

alxgu:


robynbergeron:
Thanks @alxgu for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2720, opened 2015-12-22
Fix #2629, replace choices=BOOLEANS by type='bool'

### LINKED ISSUE #2629: Replace outdated idiom choices=BOOLEANS
Argument specs used to use `choices=BOOLEANS` to specify that an argument should be True or False.  We've since enhanced the argument_spec code so that `type='bool'` does the job.  We need to go through the modules and change this over at some point.

Find modules needing updating this way:

```
grep BOOLEAN modules/core -r
```

Change lines within the module like this:

```
- is_ready = dict(choices=BOOLEANS, default=False),
+ is_ready = dict(type='bool', default=False),
```

mscherer:


robynbergeron:
Thanks @mscherer. @emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

emonty:
Awesome. Could you also update _choose_if_password_only and delete _is_value_true?

mscherer:
Indeed, didn't see it, I pushed a new commit.

PR #2722, opened 2015-12-22
Fix 2630

mscherer:


robynbergeron:

abadger:
Thanks @mscherer!

PR #2723, opened 2015-12-22
Add more precise documentation on the requirements parameter

### LINKED ISSUE #427: pip module "requirements" parameter documentation is incomplete.
For the "requirements" parameter, the comment "The path to a pip requirements file" is incomplete. I am left with the following questions (I am a very new Ansible user):
- Is this a local or remote path?
- If local, is there a way to refer to the path relatively? There doesn't appear to be, and if there is, it is not documented.
- If the path is local and must be absolute, that should be clearly stated instead of being inferred by the example (which uses an absolute path, making the role unmoveable which seems broken).

mscherer:
Fix #427

### DIFF
--- a/packaging/language/pip.py
+++ b/packaging/language/pip.py
@@ -44,7 +44,8 @@ options:
     default: null
   requirements:
     description:
-      - The path to a pip requirements file
+      - The path to a pip requirements file, which should be local to the remote system. 
+        File can be specified as a relative path if using the chdir option.  
     required: false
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @mscherer. @mattupstate please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

PR #2724, opened 2015-12-22
Service must exist

retr0h:
This brings a couple pull request to stable/1.9 which were originally accepted onto master, but later reverted, likely b/c ansible 2.0 will handle this use case.  However, those running 1.9 could use this functionality, and I don't expect people will jump to 2.0 immediately, especially those with custom vars or callback plugins.

https://github.com/ansible/ansible-modules-core/pull/751
https://github.com/ansible/ansible-modules-core/pull/2149

/cc @j2sol

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -508,9 +508,16 @@ class LinuxService(Service):
         return False
 
     def get_systemd_status_dict(self):
-        (rc, out, err) = self.execute_command("%s show %s" % (self.enable_cmd, self.__systemd_unit,))
+        # Check status first as show will not fail if service does not exist
+        (rc, out, err) = self.execute_command("%s show '%s'" % (self.enable_cmd, self.__systemd_unit,))
         if rc != 0:
             self.module.fail_json(msg='failure %d running systemctl show for %r: %s' % (rc, self.__systemd_unit, err))
+        elif 'LoadState=not-found' in out:
+            if self.module.params['must_exist']:
+                self.module.fail_json(msg='systemd could not find the requested service "%r": %s' % (self.__systemd_unit, err))
+            else:
+                self.module.exit_json(changed=False, exists=False)
+
         key = None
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @retr0h. All backport requests must be reviewed by the core team, and this can take time. We appreciate your patience.

retr0h:
@robynbergeron Any news?  ~8 months seems excessive.

retr0h:
@robynbergeron Any news?

resmo:
I would just expect this to be solved like:

``` yaml
- service: name={{ item }} state=restarted
  register: result
  with_items:
  - bla
  - foo
  failed_when: "result|failed and not 'Could not find the requested service' in result.msg"
```

results in 

```
TASK [service] *****************************************************************
ok: [localhost] => (item=bla) => {"failed": false, "failed_when_result": false, "item": "bla", "msg": "Could not find the requested service \"'bla'\": "}
ok: [localhost] => (item=foo) => {"failed": false, "failed_when_result": false, "item": "foo", "msg": "Could not find the requested service \"'foo'\": "}

localhost                  : ok=4    changed=0    unreachable=0    failed=0 
```

omgjlk:
That type of solution does not scale across all the different init providers and versions of them. Each will have a different output string to deal with.

resmo:
Another approach would be using the listen feature of ansible 2.2 but this would need some playbook code change I assume.

retr0h:
Going to abandon this patch.  Everyone should at least be on a 2.1.4 or 2.2.1 after [CVE-2016-9587](https://www.computest.nl/advisories/CT-2017-0109_Ansible.txt).

Since [1.9 appears vulnerable](https://bugs.gentoo.org/show_bug.cgi?id=605342#c4), a fix hasn't been issued, and I no longer am using 1.9 will close.

PR #2726, opened 2015-12-23
Update default value  documentation to use None instead of null

barnabycourt:
Updating from null to None per guidelines at http://docs.ansible.com/ansible/developing_modules.html


bcoca:
the None there is an example, null is fine and 'non python' people will understand it better.

barnabycourt:
@bcoca It might be good to update the doc's then. I figured there had been a change to the guidelines and the python type is what should be in the docs. I feel bad for the guy I asked to change his PR to use None instead of null

PR #2729, opened 2015-12-23
removed unneeded json.dumps when using policy_json

timnee:
policy_json is serializing a string (already in json format) to a json string.


timnee:
the problem was something else, closing

PR #2730, opened 2015-12-24
be smarter when dealing with policy_json input

### LINKED ISSUE #2009: RFE: iam_policy support for jinja templates for policy_document parameter
##### Issue Type:
- Feature Idea
##### Summary:

A good use-case for iam_policy module is to create roles with commonly used policies.  If the policy_document parameter supported templates, this could be much more useful.

For example, a playbook:

``` yaml
vars:
  aws_resource: arn::s3::somebucket/*
tasks:
  - iam_policy: policy_document=somepolicy.json.j2 iam_type=group state=present name=somepolicy
```

With a template somepolicy.json.j2:

``` jinja
{
    ...
    'policy': {
       resource: "{{ aws_resouce }}"
    }
}
```

The resource could be swapped out by variable and reused for many playbooks.

bcoca:
Now module will assume that if the argument is a string it is already formated as json
and will only try to convert non strings into json string.
Also removed unused 'msg' var declarations and the ifs that set them
fixes #2009

### DIFF
--- a/cloud/amazon/iam_policy.py
+++ b/cloud/amazon/iam_policy.py
@@ -146,9 +146,7 @@ def user_action(module, iam, name, policy_name, skip, pdoc, state):
       if urllib.unquote(iam.get_user_policy(name, pol).
                         get_user_policy_result.policy_document) == pdoc:
         policy_match = True
-        if policy_match:
-          msg=("The policy document you specified already exists "
-               "under the name %s." % pol)
+
     if state == 'present' and skip:
       if policy_name not in current_policies and not policy_match:
         changed = True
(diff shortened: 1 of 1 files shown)


bcoca:
@timnee see if this works for all cases

robynbergeron:
Thanks @bcoca. @defionscode please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

ryanwalls:
This change causes an error when using a policy json.  

```
  File "/root/.ansible/tmp/ansible-tmp-1457652747.27-251976645440330/iam_policy", line 2528, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1457652747.27-251976645440330/iam_policy", line 306, in main
    if not isinstance(pdoc, basestring):
UnboundLocalError: local variable 'pdoc' referenced before assignment

fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_name": "iam_policy"}, "parsed": false}

```

PR #2733, opened 2015-12-26
Github uses HTTPS (not HTTP) protocol

Labutin:


bcoca:
http also works, so `HTTP(S)` seems like the most complete entry here

PR #2737, opened 2015-12-28
use custom ismount function in mount module

### LINKED ISSUE #2186: Mount fails if ansible user has no read permission inside mount
##### Summary:

A bug in python os.path makes the mount module fail when
- state=mounted
- filesystem: nfs (probably most others as well)
- the mount is already present
- the ansible user has no read permission inside the mount

The reason for this is that ansible uses os.path.ismount(mountpath) which fails if pointed at a mount inside which the user has no read permission, see http://bugs.python.org/issue2466.
This is quite common if you use nfs with root_squash or kerberos.

The python-bug is quite old and unfixed, so maybe it's worthwhile to have a workaround in ansible. The function in question is 10 lines of python.
##### Steps To Reproduce:

See Summary
##### Expected Results:

Should return 'ok'
##### Actual Results:

returns failed

robinro:
- based on cpython os.path.ismount
- includes patch from http://bugs.python.org/issue2466
- fixes #2186
- when the upstream bug is fixed this should be removed/rewritten


bcoca:
code breaks 2.4 compatibility

bcoca:
once upstream fixes it, we should add a version check, but still keep this code when older versions of python are involved

robinro:
I never thought of 2.4, just tested 2.7 and 3.5. How long will you keep 2.4 around? I couldn't even find it packaged anymore for my current system.

Nevertheless, the fix saves 2 more lines of code and should actually also move upstream.

bcoca:
as long as people have it in their production systems (has been waaay too long already)

robynbergeron:

jimi-c:
Hi @robinro, as this code is copied from upstream Python, we will need to preserve the Python license. I would prefer for this code to live in a new file in `module_utils/` (ie. `module_utils/ismount.py`) which preserves the license there, and then this module would add `from ansible.module_utils.ismount import *`.

robinro:
@jimi-c thanks for the feedback
I implemented your suggestion and reran my testcase. To works this now need this pull request and the one at https://github.com/ansible/ansible/pull/14374

robinro:
btw: the function "copied" is less than 20 lines of actual python. Half of them I touch in the patch submitted to cpython and the rest could probably be rewritten so if including the license is somehow blocking, there are other ways to get this fixed. Including the license is obviously the better approach.

abadger:
At the moment, the import has to look like this:

```
from ansible.module_utils.ismount import *
```

These imports are really preprocessor-like statements where ansible substitutes in a file from module_utils into that spot in the module file before shipping the file across the wire to the remote machine.  The replacement only happens for wildcard imports from a file in the module_utils directory.

(I have plans to add a new method that uses real python imports for 2.1 but that doesn't exist yet.)

robinro:
the import statement is fixed now as described by @abadger

gregdek:
Thanks @robinro for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

robinro:
@gregdek I rebased and squashed the commits. The conflicts were only due to whitespace/linebreak cleanups. I reduced those.

robinro:
Please let me know if there is anything else to do to help this getting merged.
The helper function https://github.com/ansible/ansible/pull/14374 made it into 2.0.1, but the actual fix (this PR) didn't.

gregdek:

gregdek:
ping @ansible/core team to see if this can be moved along

abadger:
Discussed at today's Ansible Core Meeting.  Merged to devel and we'll cherry-pick to stable-2.0 once 2.0.2 is out the door.  Sorry for the delay in looking this over and getting it in and thanks for submitting it!

robinro:
@abadger 2.0.2 is out of the door, so this is your requested reminder to cherry-pick this to stable-2.0 :bell:

PR #2738, opened 2015-12-28
GCE: Use the Python installation from the env

viglesiasce:
I am using virtualenvs to develop and the GCE modules are not picking up libcloud from the virtualenv. This fix uses the virtualenv's Python for the modules to use the proper paths and such for their dependencies.

### DIFF
--- a/cloud/google/gc_storage.py
+++ b/cloud/google/gc_storage.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 # This file is part of Ansible
 #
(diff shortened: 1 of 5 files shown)


bcoca:
use ansible_python_interpreter, this breaks module usage in other environments.

viglesiasce:
If anybody ends up here, the context for ansible_python_interpreter is here:
https://github.com/ansible/ansible/issues/6345

I used the following inventory file since only the localhost path needs to change:

```
127.0.0.1 ansible_python_interpreter=/path/to/my/python
```

PR #2748, opened 2015-12-31
Neutron should manage router ports and not addresses

yfried:
Fails to remove gateway or detected existing one when not
admin.
Otherwise fails with overlapping private ips

Instead of searching gw ports (visible to admin only) search router
gw info.
Also - make network_name redundant for "absent" mode.

### DIFF
--- a/cloud/openstack/quantum_floating_ip.py
+++ b/cloud/openstack/quantum_floating_ip.py
@@ -171,9 +171,9 @@ def _get_port_info(neutron, module, instance_id, internal_network_name=None):
         return None, None
     return fixed_ip_address, port_id
 
-def _get_floating_ip(module, neutron, fixed_ip_address):
+def _get_floating_ip(module, neutron, port_id):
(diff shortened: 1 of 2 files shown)


robynbergeron:
Hi @yfried-redhat -- 

the quantum_floating_ip.py module is deprecated -- you might try os_floating_ip.py, which i believe is the replacement. 

Pinging @emonty @Shrews @j2sol for confirmation :)

gregdek:
So it doesn't look like it's yet officially deprecated, because it doesn't have an underscore in front of it. Putting into core review since it's also a backport request.

gregdek:
Thanks @yfried-redhat. All backport requests must be reviewed by the core team, and this can take time. We appreciate your patience.

emonty:
Hey @gregdek ... Quantum_floating_IP is/should be deprecated as of 2.0. It shows as _ prefixed in the repo for me?

PR #2749, opened 2015-12-31
check either ec2_group source or destination (egress) parameter is specified

jonhadfield:
**Issue Type:** Bug report
**Summary:** 
There is no check for a rule source (ingress) or destination (egress).
If a new group is to be created with a rule not specifying cidr_ip, group_id or group_name, it will create a group without any rules.
If creating a group with the same name of an existing with a rule, it will delete the existing rule.
Probably an unlikely issue to come across but it frustrated me for a time when specifying a number of rules (one with a missing source) and succeeding with changed=1.

**Steps to reproduce:**

```
- name: create a test security group
  ec2_group:
    name: "my_test_group"
    description: "my test description"
    region: "eu-west-1"
    rules:
       - proto: tcp
         from_port: 443
         to_port: 443
```

**Expected result:**

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Specify either cidr_ip, group_id or group_name"}
```

**Actual result:**

```
TASK [create a test security group] ********************************************
changed: [localhost]

```

Again, if a group of the same name exists, the result will be as above but with the rule removed.

**Change:**
The PR adds another check to the validate_rule method to ensure cidr_ip, group_id or group_name are specified.


robynbergeron:
Thanks @jonhadfield for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2765, opened 2016-01-06
apt: export env vars before run dpkg

no2a:
Without this change, some trouble may occur when "deb" parameter
is used as env vars controlling dpkg are not set. For example,
installing a package that requires user input will never end since
DEBIAN_FRONTEND=noninteractive is not set.

So export env vars in APT_ENV_VARS before run dpkg, like in cases
using apt-get/aptitude.


bcoca:
tempted to just do that in 'main'

PR #2768, opened 2016-01-06
Allow cron_file to be an absolute path

quinot:
Support specifying an absolute path (typically /etc/crontab) rather than
a path relative to /etc/cron.d, to allow modifying the main system crontab.
Particularly useful for target systems that have /etc/crontab but no
/etc/cron.d.

### DIFF
--- a/system/cron.py
+++ b/system/cron.py
@@ -66,7 +66,9 @@ options:
     choices: [ "present", "absent" ]
   cron_file:
     description:
-      - If specified, uses this file in cron.d instead of an individual user's crontab.
+      - If specified, uses this file instead of an individual user's crontab.
+        If this is a relative path, it is interpreted with respect to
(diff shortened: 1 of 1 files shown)


tyleha:
Woah, we did basically the same thing at the same time! See PR #2770. 

I like this method better...let's do it this way.

+1

tyleha:
Tested locally, works for me.

gregdek:
Thanks @quinot. @dsummersl @EvanK please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

EvanK:
shipit

robynbergeron:
Thanks again to @quinot for this PR, and thanks @dsummersl @EvanK for reviewing. Marking for inclusion.

Issue #2763, opened 2016-01-05
cron module: add jobs into /etc/crontab, not /etc/cron.d/...


tyleha:
**Issue Type:**

Feature Pull Request

I'll be submitting a PR for this shortly, be on the lookout.

**Ansible Version:**

```
ansible 2.1.0 (devel 095790ac5c) last updated 2016/01/05 15:15:08 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 33014c6db1) last updated 2016/01/05 15:43:25 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 82a4cf84be) last updated 2016/01/05 15:43:33 (GMT -700)
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```

**Environment:**
Ubuntu 14.04.3

**Summary:**

Reopening an incorrectly closed issue from https://github.com/ansible/ansible/issues/4881.

Unfortunately commit [4384114ed6fc901e1503310f23aac4e5589bf283](https://github.com/ansible/ansible/commit/4384114ed6fc901e1503310f23aac4e5589bf283) only allows you to specify a relative path to a cron file within `/etc/cron.d`. This did not address the initial concern in [#4481](https://github.com/ansible/ansible/issues/4881) which looked to add cron entries to `/etc/crontab`. As mentioned in that issue, not all distros of cron support the `cron.d/` folder like BusyBox which runs on Raspberry Pi and Synology hardware.

**Expected Results:**

When passing the `cron_file` field my crontab file at `/etc/crontab`, I expect crontab to be updated.

**Actual Results:**

IOError: [Errno 2] No such file or directory: '/etc/cron.d//etc/crontab'\r\n"

Running a `cron` command without specifying `cron_file` gives

```
[...]

PR #2770, opened 2016-01-06
Add a flag to register cron jobs at /etc/crontab. For #2763.

tyleha:
See #2763 for more info.

Not all distros of cron support the cron.d/ folder. BusyBox, which runs on Raspberry Pi and Synology hardware, supports only the system crontab file at `/etc/crontab`. This PR provides support for such an OS.

Also fixes a bug with the `state` field which was argparsed as
a boolean but is a choice between present and absent. This means
that using cron_file to add a line to a user's crontab has
been broken. Honestly, that makes me question whether anyone has even been using the `cron_file` option, and if it should simply be made to be an absolute path such that you could specify `/etc/crontab` or `/etc/cron.d/myfile` through the same variable.

### DIFF
--- a/system/cron.py
+++ b/system/cron.py
@@ -70,6 +70,11 @@ options:
         To use the C(cron_file) parameter you must specify the C(user) as well.
     required: false
     default: null
+  cron_file:
+    description:
+      - If 'yes', use the system crontab file at /etc/crontab instead of an individual's crontab.
+    required: false
+    default: no
   backup:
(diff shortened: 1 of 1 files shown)


tyleha:
Actually, see #2768 for a more general and possibly superior solution.

gregdek:
Closing in favor of #2768 per your suggestion. Thanks @tylerhartley.

PR #2773, opened 2016-01-06
initial add of new module ops_config

privateip:
This commit adds a new module ops_config to network/openswitch.  The
ops_config module will push a new configuration to an OpenSwitch instance
using the OpenSwitch declarative configuration library


gregdek:
Putting in core review.

PR #2783, opened 2016-01-08
win_feature: source option

trondhindenes:
This non-required option allows win_feature to specify the install source when adding features. Usually used with "Net-Framework-Core" since that package isn't included with Windows.

`ansible -m "win_feature" -a "name=NET-Framework-Core source=C:/Temp/iso/sources/sxs" windows -vv`

### DIFF
--- a/windows/win_feature.ps1
+++ b/windows/win_feature.ps1
@@ -28,6 +28,7 @@ $result = New-Object PSObject -Property @{
 }
 
 $name = Get-Attr $params "name" -failifempty $true
+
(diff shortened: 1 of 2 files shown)


jhawkesworth:
I have tested this successfully today.
Ansible controller: Centos 6.5 running 2.0.0 rc4
Target window hosts: Windows Server 2012 R2

gregdek:
Thanks @trondhindenes. Since you are a maintainer of this module, we are marking this PR for inclusion.

PR #2489, opened 2015-11-18
Add evaluate taget health for alias based records

jimbydamonk:

PR #2790, opened 2016-01-11
Add wait option to route53 module

jcassee:
The new `wait` option (disabled by default) makes the module wait for the changes to replicate to all Amazon DNS servers. This is useful if you want to make sure DNS queries work in subsequent tasks.

@bpennypacker As the original author, could you take a look at the proposed changes?

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -137,6 +137,12 @@ options:
     required: false
     default: null
     version_added: "2.0"
+  wait:
+    description:
+      - Wait until the changes have been replicated to all Amazon Route 53 DNS servers.
+    required: false
+    default: no
+    version_added: "2.0"
 author: 
   - "Bruce Pennypacker (@bpennypacker)"
   - "Mike Buzzetti <mike.buzzetti@gmail.com>"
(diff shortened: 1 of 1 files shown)


bpennypacker:
This looks good to me, however since the wait option turns the task into a blocking one, it would probably be a good idea to also add a wait_timeout option that returns a timeout error in the event something unexpected ever occurs.  Other modules like rds, ec2, etc. have a wait_timeout option to prevent Ansible hanging on those tasks if something unexpected happens, so replicating that behavior here probably makes sense.

jcassee:
Good point. I will add a `wait_timeout` option.

jcassee:
@bpennypacker I added the `wait_timeout` option. Are you comfortable with the exception I added to be able to fail the module?

robynbergeron:
Thanks @jcassee. @bpennypacker please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

bpennypacker:
Thanks, @jcassee, this looks great! 

shipit

jcassee:
I rebased the PR after #2489 got merged. It may be wise to review the changes again, as there were some conflicts.

ready_for_review

bpennypacker:
The rebase looks good.  Thanks.  shipit.

jcassee:
Thanks for picking this up so quickly @bpennypacker and @bcoca!

PR #2796, opened 2016-01-12
guard main-execution

jvantuyl:
The Developing Modules Guide mentions that the main function should be guarded
(as is the common idiom for Python scripts).  Most of the modules were not
written that way.  This patch updates most of them to comply with this policy.

See here: http://docs.ansible.com/ansible/developing_modules.html#common-module-boilerplate

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -403,4 +403,5 @@ def main():
 from ansible.module_utils.basic import *
 from ansible.module_utils.ec2 import *
 
-main()
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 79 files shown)


bcoca:
This is not really needed, it was added to the policy for the new modules as we intend to change the module construction to allow for actual importing, currently the `from ansible.module_utils.basic import *` is not an actual python import but a 'copy and replace' marker. If we get to that point we will look at this again.

kstrauser:
I'm a little confused. This makes those modules more Pythonic without changing their functionality in any way. Even if the benefit isn't big for everyone today, isn't it worth the zero price?

jvantuyl:
Or perhaps should we update the policy so nobody else wastes their time?  I mean, I'm a robot; but I'd hate to put off other people if they were generally trying to help with some "low-hanging fruit."

PR #2806, opened 2016-01-13
Unarchrefactor

pileofrogs:
This is the module side of a refactor to allow untar and unzip (and any future unarchivers) to have their own modules, thus allowing their own arguments. There is a corresponding PR for ansible.  Don't accept one without the other!

This PR only refactors and does not add new arguments.

After doing it, I realized a lot of ways to make it better. I added a _copy method to the ActionBase class, thinking this would be good because other future modules could use it. However, it might be better to make an unarchive subclass of ActionBase and stick stuff in their. Also, I wound up duplicating a fair amount of code between the unarchive module and the untar and unzip modules. I would like to separate out some of the logic. Possibly into ansible.module_utils somewhere... sigh...

So, basically, if you aren't picky, take this PR. It should allow folks who want to submit PRs for new tar/zip features to be able to do so. If you're even remotely picky, fire it back with comments and I'll unscrew whatever parts you don't like.

There is a nearly identical sounding PR for ansible. One doesn't work without the other.


gregdek:
Thanks @pileofrogs for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

pileofrogs:
Hi all.  Anyone out there have anything to say at all about these PRs?  There are other PRs for unarchive waiting and I can't really get to them until I know if this PR is going in the right direction or not.

Jmainguy:
Even this functionality is not new, just being split out. I think the version_added should reflect 2.1 as people trying to call these modules on older versions will get an error.

gregdek:
Thanks @pileofrogs for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

pileofrogs:
Cripes... is a rebase supposed to look like that?

Jmainguy:
=) nope

Jmainguy:
@pileofrogs Try (backup first just in case before doing this)

```
git reset c5fc39b15eb3d425fe5ca0b90071902c6cc9d701
git add .
git commit -m 'yarg'
git rebase -i HEAD~4
```

Leave splitting unarchive into three parts as pick, change the other 3 to fixup

pileofrogs:
I already totally screwed up my fork.  Am going to make new fork and manually merge in stuff for this PR .

PR #2808, opened 2016-01-14
S3 Module: Adding S3 website hosting

noqcks:
[Static website hosting](http://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html) on s3 is something that i've wanted out of an ansible s3 module for a while now.

Here im adding two parameters which are related - `website` and `website_index`. Let me know what you think of the module names. I can change if needed.

:beers:  :beers:  @lwade :beers: :beers:

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -144,9 +144,15 @@ options:
     version_added: "1.3"
   website:
     description:
-      - Enable bucket as static website hosting.
+      - Enable static website hosting for the S3 bucket.
     required: false
     default: false
+  website_index:
+    description:
+      - The document to be returned for each directory of the static website. If someone requests c(samplebucket/images) and the website index is index.html, the object returned will be c(samplebucket/images/index.html).
+      - Used when c(website=true).
+    required: false
+    default: index.html
 
 requirements: [ "boto" ]
 author:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @noqcks. @lwade please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

wimnat:
This PR has been open for over 6 months. If you can review maybe it'll get merged https://github.com/ansible/ansible-modules-extras/pull/662. I prefer a separate module as I think it's better to leave the current s3 module to handle keys only

noqcks:
@wimnat the current s3 module also creates and delete buckets, which is when you can specify static website hosting. I don't feel like an extra module is needed :8ball: 

@lwade any input?

lwade:
Looks like a simple change, would be nice to have this feature. I'd say shipit. Thanks @noqcks

wimnat:
@noqcks create bucket functionality is also in the module s3_bucket along with stuff not available in the s3 module such as versioning and policy.  That's the reason s3_website was created - to try and stop the s3 module becoming a complex behemoth like the old ec2_vpc module.

I would appreciate it if you could test the s3_website module anyway and give it the thumbs up even if you do decide to proceed with this PR.

gregdek:
Thanks @noqcks for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2814, opened 2016-01-14
Add purge option to os_floating_ip module

Shrews:
Add the ability to completely delete a floating IP from the pool
when disassociating it from a server. When state is absent and
purge is true, the IP will be completely deleted. The default
keeps the current behavior, which is to only disassociate the IP
from the server.

### DIFF
--- a/cloud/openstack/os_floating_ip.py
+++ b/cloud/openstack/os_floating_ip.py
@@ -43,13 +43,13 @@ options:
      required: false
    floating_ip_address:
      description:
-        - A floating IP address to attach or to detach. Required only if state
-          is absent. When state is present can be used to specify a IP address
+        - A floating IP address to attach or to detach. Required only if I(state)
+          is absent. When I(state) is present can be used to specify a IP address
           to attach.
      required: false
    reuse:
      description:
-        - When state is present, and floating_ip_address is not present,
+        - When I(state) is present, and I(floating_ip_address) is not present,
           this parameter can be used to specify whether we should try to reuse
           a floating IP address already allocated to the project.
      required: false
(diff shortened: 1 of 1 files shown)


yfried:
I think default should be the same as any other ansible module, where absent means - not found

cloud/openstack/os_floating_ip.py
```diff
@@ -76,6 +76,12 @@
      choices: [present, absent]
      required: false
      default: present
+   purge:
+     description:
+        - When I(state) is absent, indicates whether or not to delete the floating
+          IP completely, or only detach it from the server. Default is to detach only.
+     required: false
+     default: false
```
yfried-redhat: suggest `true`, which is more intuitive

Shrews, line 84: Completely disagree here. We cannot change the default behaviour as others may already be depending on it.

cloud/openstack/os_floating_ip.py
```diff
@@ -173,6 +181,9 @@ def main():
                 server_id=server['id'], floating_ip_id=f_ip['id'])
```
yfried, line 185: L79:

``` python
if not fip and purge:
    module.exit_json(changed=False)

```

gregdek:
Thanks @Shrews. @emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

emonty:
shipit

yfried:
:+1: 
LTGM

rcarrillocruz:
shipit

omgjlk:
shipit

gregdek:
Thanks again to @Shrews for this PR, and thanks @emonty @shrews @dguerri @juliakreger @j2sol @rcarrillocruz for reviewing. Marking for inclusion.

PR #2815, opened 2016-01-14
Fix issue with mysql_user missing argument

### LINKED ISSUE #2804: mysql_user is broken in stable-2.0
It appears that commit e8777119b239cda2c6dccab41de0d7e62a0e7d68 introduced code, such as an additional parameter for user_exists that breaks the module.  In addition to issues with user_exist, the module no longer appears to authenticate as expected to do so previously.

TypeError: user_exists() takes exactly 4 arguments (3 given)

If I attempt to fix the just that error by reverting the lines changed in that method and execute a playbook that uses the module that worked prior to the change, I get:

> fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_args": {"append_privs": false, "check_implicit_admin": false, "config_file": "~/.my.cnf", "encrypted": false, "host": "localhost", "login_host": "localhost", "login_password": "", "login_port": 3306, "login_unix_socket": null, "login_user": "root", "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "priv": "database.*:ALL,GRANT", "ssl_ca": null, "ssl_cert": null, "ssl_key": null, "state": "present", "update_password": "always", "user": "mydatabaseuser"}, "module_name": "mysql_user"}, "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: (1044, \"Access denied for user 'root'@'localhost' to database 'mysql'\")"}

When executing the following step:

```
- name: "MySQL - Creating user"
  mysql_user:
    user: "mydatabaseuser"
    password: "{{ db_password }}"
    priv: "mydatabase.*:ALL,GRANT"
    state: present
    login_user: "{{ mysql_username | default(None) }}"
    login_password: "{{ mysql_password | default(None) }}"
```

miguel250:
Closes #2804


miguel250:
This was fixed already.

juliakreger:
In devel yes, not in stable-2.0. :(

Issue #2314, opened 2015-10-20
cloudformation state: absent requires template or template_url


adenot:
##### Summary:

Using the latest (9eb0c178ec2f4e94ae29329fa9694a381157b413) cloudformation module, below **does not** works:

``` yaml
    - name: Removing RDS instances
      cloudformation2:
        stack_name: "{{ db_stack_name }}"
        state: "absent"
        region: "{{ aws_region }}"
```

But passing a template or template_url, even if dummy, works:

``` yaml
    - name: Removing RDS instances
      cloudformation2:
        stack_name: "{{ db_stack_name }}"
        state: "absent"
        region: "{{ aws_region }}"
        template_url: "it's a bug"
```
##### Expected Results:

According to docs, template/template_url should only be required when state: present.
##### Actual Results:

```
TASK: [Removing RDS instances] ************************************************
failed: [localhost] => {"failed": true}
msg: Either template or template_url expected

cloudformation2 [localhost] =>
{
  "msg": "Either template or template_url expected",
  "failed": true
}
```

robynbergeron:
[module: cloud/amazon/cloudformation.py]

robynbergeron:
Greetings, @adenot -- thanks for this bug report. 

@jsmartin -- thanks in advance for taking a look at this bug report and for your ongoing work in maintaining this module.

ansibot:
@jsmartin, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@jsmartin, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@jsmartin, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@tedder, @ryansb, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

tedder:
resolved_by_pr #5319.

ansibot:
@tedder, @ryansb, ping. This issue is still waiting on your response.
[...]

PR #2826, opened 2016-01-16
The parameters 'template' and 'template_url' are incorrectly required in all cases in the cloudformation module

### LINKED ISSUE #2575: Cloudformation module: requiring template during stack deletion
ansible 1.9.4, OS X 10.11

According to the documentation 'template' argument for cloudformation module required only when state=present. But execution failed when I try to delete stack without specifying template:

```
 ansible -i hosts localhost -m cloudformation -a 'stack_name=my state=absent'
 127.0.0.1 | FAILED >> {
     "failed": true, 
     "msg": "missing required arguments: template"
 }
```

shawnsilva:
The parameters 'template' and 'template_url' are incorrectly required in all cases in the cloudformation module

If the state is 'absent' they should not be required. A few lines below the correct check is already in place.

This commit fixes #2575, #2314

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -258,9 +258,6 @@ def main():
     if not HAS_BOTO:
         module.fail_json(msg='boto required for this module')
 
-    if module.params['template'] is None and module.params['template_url'] is None:
-        module.fail_json(msg='Either template or template_url expected')
-
     state = module.params['state']
     stack_name = module.params['stack_name']
 


shawnsilva:
I don't think a specific check for state == 'absent' is required in the case of 'template' or 'template_url'.
If the state is 'absent', nothing would be done if neither 'template' or 'template_url' are set as they aren't required to delete a stack. 

The lines that I removed were extra, and failing the module before getting to the real check that was already in place a few lines below: https://github.com/shawnsilva/ansible-modules-core/blob/cloudformation-req-params/cloud/amazon/cloudformation.py#L264

gregdek:
Thanks @shawnsilva. @jsmartin please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

jsmartin:
I concur with @shawnsilva .  These lines appear to be redundant.   shipit.

gregdek:
Thanks again to @shawnsilva for this PR, and thanks @jsmartin for reviewing. Marking for inclusion.

abadger:
Thanks.  merged to devel and stable-2.0.

PR #2829, opened 2016-01-17
Add wait support to os_ironic_node

juliakreger:
Add support for the os_ironic_node module to be capable of waiting
until the bare metal node activation or deactivation has completed.


gregdek:
Thanks @juliakreger. Since you are a maintainer of this module, we are marking this PR for inclusion.

PR #2845, opened 2016-01-19
Execution fails when manage.py is not executable

danpilch:
there has been a regression with this module that makes old playbooks incompatible as it now forces manage.py to be executable. This should not be the default behavior.


gregdek:
Thanks @danpilch. All backport requests must be reviewed by the core team, and this can take time. We appreciate your patience.

danpilch:
any news on this?

PR #2857, opened 2016-01-22
initial add of new nxos_config module

privateip:
This adds a new module for managing configuraiton files for Cisco NXOS
devices.  It provides configuration file management including templating
and backing up the current configuration.

This PR has a dependency on ansible/ansible PR # 14012

### DIFF
--- a/network/nxos/nxos_config.py
+++ b/network/nxos/nxos_config.py
@@ -1,0 +1,274 @@
+#!/usr/bin/python
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+DOCUMENTATION = """
+---
+module: nxos_config
+version_added: "2.1"
+author: "Peter sprygada (@privateip)"
+short_description: Manage Cisco NXOS configuration sections
+description:
+  - Cisco NXOS configurations use a simple block indent file sytanx
+    for segementing configuration into sections.  This module provides
+    an implementation for working with NXOS configuration sections in
+    a deterministic way.  This module works with either CLI or NXAPI
+    transports.
+extends_documentation_fragment: nxos
+options:
+  lines:
+    description:
+      - The ordered set of commands that should be configured in the
+        section.  The commands must be the exact same commands as found
+        in the device running-config.  Be sure to note the configuration
+        command syntanx as some commands are automatically modified by the
+        device config parser.
+    required: true
+  parents:
+    description:
+      - The ordered set of parents that uniquely identify the section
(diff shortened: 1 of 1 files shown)


gregdek:

network/nxos/nxos_config.py
```diff
@@ -0,0 +1,268 @@
...
+short_description: Manage Cisco NXOS configuration sections
+description:
+  - Cisco NXOS configurations use a simple block indent file sytanx
+    for segementing configuration into sections.  This module provides
+    an implementation for working with NXOS configuration sections in
+    a deterministic way.  This module works with either CLI or NXAPI
+    transports.
+extends_documentation_fragment: nxos
+options:
+  commands:
```
bcoca: 'rules' might be a better name as commands seems to indicate that they are 'always run' and not checked against existing configuration to determine if a change is needed.

bcoca: ramblings: another thought would be to make it 'rule' so order can be ignored and single rules be passed to the module (but this can become inefficient ... squashing would help but then we are back to checking order).

privateip: Rules mean something else in the network space (specifically most network savvy individuals equate rules with ACL rules).  The entries are indeed commands so I would assert the argument is properly named.

WRT to a singular rule, context and order are indeed important depending on the configuration block that is being implemented so creating a scalar argument wont work

bcoca: config then? we can keep commands as an alias, but i would like to move away from having the implication even if most network folks understand the difference. 

the single item probably merits it's own module, but would require more complexity about positioning (iptables module does some of this already). Again this is just a random thought, not a requirement.

network/nxos/nxos_config.py
```diff
@@ -0,0 +1,268 @@
...
+- nxos_config:
+    commands:
+      - 10 permit ip 1.1.1.1/32 any log
+      - 20 permit ip 2.2.2.2/32 any log
+      - 30 permit ip 3.3.3.3/32 any log
+      - 40 permit ip 4.4.4.4/32 any log
+    parents: ['ip access-list test']
+    before: ['no ip access-list test']
+    replace: block
+
```
bcoca, line 135: i would add a couple of useful examples here using lookups and vars, most network admins will not be learning ansible through other modules and might need hints that these things are possible.

```
- nxos_config:
    commands: "{{lookup('file', 'datcenter1.txt'}}"
    parents: ['ip access-list test']
    before: ['no ip access-list test']
    replace: block
```

network/nxos/nxos_config.py
```diff
@@ -0,0 +1,268 @@
...
+"""
+
+RETURN = """
+
+commands:
```
bcoca: I would return this using the new 'diff' facility modules have, also i would include 'removed' commands when using the 'block' manner of updating.

network/nxos/nxos_config.py
privateip: there is already an argument named config that serves a different purpose.  I'm not following the concern here with using commands as an argument name.  The reality is the values inserted from commands truly are commands that are updating the configuration in real time.

bcoca: In this case they are closer to 'settings' though in the 'particular domain' of these devices those terms can be interchangeable. It is not an issue about this module in particular but general UX across the modules, where command/commands really signifies an execution that cannot be controlled for change or reporting on it.

privateip: Thx, that helps my feeble brain to understand  :)    As a compromise, how about "lines" and commands to be an alias?  agree?

bcoca: ok, i suck at naming, just want to avoid 'commands' as the main one

PR #2871, opened 2016-01-25
Fix glance_client module

yfried:
Wrong variable name.


yfried:
How do I backport this to 1.X?

bcoca:
well, the patch itself won't work for 1.x as the code is pretty different, an alternate patch against the stable-1.9 branch will be considered

PR #2872, opened 2016-01-25
initial add of eos_command module

privateip:
This adds a new module eos_command to network/eos.  The eos_command module
is used for sending arbitrary commands to Arista EOS devices.  It includes
arguments that allow the module to wait for specific values before the
module returns control to the playbook or fails


gregdek:

PR #2887, opened 2016-01-26
Add support for additional parameter 'pkg_opts' in package action.

indrajitr:
Documentation for ansible/ansible#14129.

NB: the parameter `state` isn't actually used in any way in `package.py`. And in any case, all the underlying package modules have a default value for `state` (`present`). Does it require to be explicitly documented and marked 'required'?

### DIFF
--- a/packaging/os/package.py
+++ b/packaging/os/package.py
@@ -40,6 +40,11 @@ options:
     description:
       - Whether to install (C(present), C(latest)), or remove (C(absent)) a package.
     required: true
+  pkg_args:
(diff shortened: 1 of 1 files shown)


bcoca:
Hi!

Unfortunately, we're not sure if we want this feature in the program, and I don't want this to seem confrontational.  Our reasons for this are:

  The package module is meant to be generic, adding package manager specific opts is not something we want. If you want to use package manager specific options, use the module for that package manager.

PR #2897, opened 2016-01-27
initial add of new module nxos_command

privateip:
This adds a new module nxos_command that can be used to send arbitrary
commands to NXOS devices.  The module includes an argument that allows
the responses to be evaluated and causes the module not to return
control to the playbook until a set of conditions has been met.


gregdek:

PR #2902, opened 2016-01-27
Add support for encrypted EBS volumes in launch configuration

steynovich:
Add support for encrypted EBS volumes in launch configuration. This uses the same logic as encrypted volumes in the ec2 module.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -755,7 +755,7 @@ def create_block_device(module, ec2, volume):
             if int(volume['iops']) > MAX_IOPS_TO_SIZE_RATIO * size:
                 module.fail_json(msg = 'IOPS must be at most %d times greater than size' % MAX_IOPS_TO_SIZE_RATIO)
         if 'encrypted' in volume:
-            module.fail_json(msg = 'You can not set encyrption when creating a volume from a snapshot')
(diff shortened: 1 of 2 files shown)


gregdek:
Thanks @steynovich for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

fuhrysteve:
If you want an EC2 instance running on an encrypted volume, you can to start from an encrypted AMI.

Here's a poor man's guide:
1. decide what AMI you want encrypted. Example: `ami-93e1cf84` (happens to be an Ubuntu 16.04 image)
2. In the AWS control panel, go to EC2 => AMI => Public Images => search for `ami-93e1cf84` => Right click => Copy AMI => enable encryption => click "Copy AMI"
3. Wait 20mins or so for it to copy & encrypt your new AMI image

Now you can use this new private encrypted AMI instead of whatever AMI you were using before, and it will magically be using an encrypted volume. You don't need this pull request to spin up an encrypted EC2 instance, if that's all you're trying to do.

steynovich:
The purpose of this PR is to add support for encryption of additional EBS volumes (e.g. /dev/xvdf). Those blank volumes are created from scratch, not from snapshot or AMI.

fuhrysteve:
Ah I see - that makes sense.

PR #2919, opened 2016-01-29
Nxos nxapi module

chouseknecht:
New module for managing NXAPI access on an NXOS device. If enabling NXAPI, configure http state and port, https state and port, and sandbox state.

### DIFF
--- a/network/nxos/nxos_nxapi.py
+++ b/network/nxos/nxos_nxapi.py
@@ -1,0 +1,258 @@
+#!/usr/bin/python
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+DOCUMENTATION = """
+---
+module: nxos_nxapi 
+version_added: "2.1"
+author: "Chris Houseknecht (@chouseknecht)"
+short_description: Manage NXAPI configuration on an NXOS device.
+description:
+  - Use to enable or disable NXAPI access, set the port and state
+    of http and https servers, and enable or disable the sandbox.
+  - When enabling NXAPI access the default is to enable HTTP on port
+    80, enable HTTPS on port 443, and enable the web based UI sandbox.
+    Use the options below to override the default configuration.
+extends_documentation_fragment: nxos
+options:
+    state:
+        description:
+            - Set to started or stopped. A state of started will
+              enable NXAPI access, and a state of stopped will
+              disable or shutdown all NXAPI access.
+        choices:
+            - started
+            - stopped
+        requred: false
+        default: started
+    http_port:
+        description:
+            - Port on which the HTTP server will listen.
+        required: false
+        default: 80
+    https_port:
+        description:
+            - Port on which the HTTPS server will listen.
+        required: false
+        default: 443
+    http:
+        description:
+            - Enable/disable HTTP server.
+        required: false
+        default: true
+        aliases:
+            - enable_http
+    https:
+        description:
+            - Enable/disable HTTPS server.
+        required: false
+        default: true
+        aliases:
+            - enable_https
+    sandbox:
+        description:
+            - Enable/disable NXAPI web based UI for entering commands.
+        required: false
+        default: true
+        aliases:
+            - enable_sandbox
+"""
+
+EXAMPLES = """
+    - name: Enable NXAPI access with default configuration
+      nxos_nxapi:
+          provider: {{ provider }}
+
+    - name: Enable NXAPI with no HTTP, HTTPS at port 9443 and sandbox disabled
+      nxos_nxapi:
+          enable_http: false
+          https_port: 9443
+          enable_sandbox: no
+          provider: {{ provider }}
+
+    - name: shutdown NXAPI access
+      nxos_nxapi:
+          state: stopped
+          provider: {{ provider }}
+"""
+
+RETURN = """
+changed:
+    description:
+        - Indicates if commands were sent to the device.
+    returned: always
+    type: boolean
+    sample: false
+
+commands:
+    description:
+        - Set of commands to be executed on remote device. If run in check mode,
+          commands will not be executed.
+    returned: always
+    type: list
+    sample: [
+        'nxapi feature',
+        'nxapi http port 8080'
+    ]
+
+_config:
+    description:
+       - Configuration found on the device prior ro any commands being executed.
+    returned: always
+    type: object
+    sample: {...}
+"""
+
+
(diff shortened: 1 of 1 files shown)


network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,237 @@
...
+    Use the options below to override the default configuration.
+extends_documentation_fragment: eos
+options:
+    state:
+        description:
+            - Set to running or stopped. A state of running will
+              enable NXAPI access, and a state of stopped will
+              disable or shutdown all NXAPI access.
+        requred: false
+        default: running
```
bcoca: to be consistent with other modules states shoudl be started/stopped

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,237 @@
...
+    enable_https:
+        description:
+            - Enable HTTPS server access.
+        required: false
+        default: true
+    enable_sandbox:
+        description:
+            - Enable NXAPI web based UI for entering commands
+        required: false
+        default: true
```
bcoca, line 72: a better pattern might be `http: enabled/disabled`

gregdek:
Thanks @chouseknecht for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,251 @@
...
+        elif port_config != port:
+            # update port
+            commands.append('nxapi {0} port {1}'.format(protocol, port))
+            changed = True
+    return commands, changed
+
+
+def execute_commands(module, commands):
+    if not module.params.get('check_mode'):
+        module.execute(commands)
```
privateip: use module.configure instead of module.execute and remove config t from the commands list

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,251 @@
...
+version_added: "2.1"
+author: "Chris Houseknecht (@chouseknecht)"
+short_description: Manage NXAPI configuration on an NXOS device.
+description:
+  - Use to enable or disable NXAPI access, set the port and state
+    of http and https servers, and enable or disable the sandbox.
+  - When enabling NXAPI access the default is to enable HTTP on port
+    80, enable HTTPS on port 443, and enable the web based UI sandbox.
+    Use the options below to override the default configuration.
+extends_documentation_fragment: eos
```
privateip: fix doc fragment

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,251 @@
...
+options:
+    state:
```
privateip, line 34: add choices to doc string

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,251 @@
...
+    result = config_server(module)
+
+    return module.exit_json(**result)
+
+
+from ansible.module_utils.basic import *
+from ansible.module_utils.urls import *
+from ansible.module_utils.shell import *
+from ansible.module_utils.netcfg import *
+from ansible.module_utils.eos import *
```
privateip: fix import statement

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,251 @@
...
+    module = get_module(argument_spec=argument_spec,
+                        supports_check_mode=True)
+
+    result = config_server(module)
+
+    return module.exit_json(**result)
+
+
+from ansible.module_utils.basic import *
+from ansible.module_utils.urls import *
```
privateip: not needed

privateip:
please squash commits to a single add message since this an initial add of a new module and remove **init** from the commit

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,251 @@
...
+from ansible.module_utils.shell import *
+from ansible.module_utils.netcfg import *
```
privateip: not needed

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,250 @@
...
+        description:
+            - Enable/disable HTTP server.
+        required: false
+        default: true
+    https:
+        description:
+            - Enable/disable HTTPS server.
+        required: false
+        default: true
+    sandbox:
```
privateip, line 68: add aliases

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,250 @@
...
+        description:
+            - Port on which the HTTPS server will listen.
+        required: false
+        default: 443
+    http:
+        description:
+            - Enable/disable HTTP server.
+        required: false
+        default: true
+    https:
```
privateip, line 61: add aliases

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,250 @@
...
+        description:
+            - Port on which the HTTP server will listen.
+        required: false
+        default: 80
+    https_port:
+        description:
+            - Port on which the HTTPS server will listen.
+        required: false
+        default: 443
+    http:
```
privateip, line 54: add aliases

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,250 @@
...
+    sample: false
+
+commands:
+    description:
+        - Set of commands to be executed on remote device. If run in check mode,
+          commands will not be executed.
+    returned: always
+    type: list
+    sample: [
+        'config terminal',
```
privateip: remove config terminal

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,250 @@
...
+    argument_spec = dict(
+        state=dict(default='started', choices=['started','stopped']),
+        http_port=dict(default=80, type='int'),
+        https_port=dict(default=443, type='int'),
+        http=dict(aliases=['enable_http'], default=True, type='bool'),
+        https=dict(aliases=['enable_https'], default=True, type='bool'),
+        sandbox=dict(aliases=['enable_sandbox'], default=True, type='bool')
+    )
+
+    module = get_module(argument_spec=argument_spec,
```
privateip, line 245: check and fail on transport=nxapi

network/nxos/nxos_nxapi.py
chouseknecht, line 54: Added

network/nxos/nxos_nxapi.py
```diff
@@ -0,0 +1,258 @@
...
+        state=dict(default='started', choices=['started','stopped']),
+        http_port=dict(default=80, type='int'),
+        https_port=dict(default=443, type='int'),
+        http=dict(aliases=['enable_http'], default=True, type='bool'),
+        https=dict(aliases=['enable_https'], default=True, type='bool'),
+        sandbox=dict(aliases=['enable_sandbox'], default=True, type='bool'),
+
+        # Only allow configuration of NXAPI using cli transpsort
+        transport=dict(required=True, choices=['cli'])
+    )
```
chouseknecht, line 243: Overriding transport param raises the exception before nxos.py gets evaluated. If nxos.py is evaluated by the interpreter, then we need urls.py from module_utils included so that fetch_url is available. And even with that there is still some other exception getting raised that I couldn't track down. But it's an exception that makes no sense to the end user, so we would have to include something else or fix whatever it is. Might be including net_cfg.py.

Anyway, this throws an exception letting user know that the only transport choice is 'cli', which seems reasonable enough. Added 'badtransport.py' test to test-nxos/test_nxos_nxapi to accompany this.

PR #2926, opened 2016-01-31
Split shell command to avoid use_unsafe_shell.

mattclay:
This mirrors a nearly identical change made to apt_repository.py in 
commit https://github.com/mscherer/ansible-modules-core/commit/8c5e8f042524ab464f6e46e435f2932abba1ad5c (PR https://github.com/ansible/ansible-modules-core/pull/2661).

Also removes the use of apt-get --force-yes as it can be dangerous
and should not be necessary (apt_repository.py does not use it).

Repeating the explanation from the apt_respository change below:

code we have to thoroughly inspect for a security audit.

In this case, the '&&' can be replaced by doing 2 calls to run_command.

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -567,7 +567,8 @@ def main():
 
     if not HAS_PYTHON_APT:
         try:
-            module.run_command('apt-get update && apt-get install python-apt -y -q --force-yes', use_unsafe_shell=True, check_rc=True)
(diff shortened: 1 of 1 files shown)


mattclay:
Tested on Ubuntu 14.04 LTS.

gregdek:
Thanks @mattclay for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2155, opened 2015-09-25
WIP: add 'env_file' support. Refs #792.

vpetersson:
Suggested/untested env-file support for public feedback.

Issue #792, opened 2015-02-14
add support for env-file to the docker module


kampka:
Docker run supports the use of env-files for passing environment variables to containers.

```
  --env-file=[]              Read in a line delimited file of environment variables
```

It would be convenient if the docker module would support this as well.

mahnunchik:
+1

Any news?

MrMMorris:
+100 on this. It's a lot easier to use conditionals in a template than it is in a task if I want to omit some env vars

vpetersson:
+1

vpetersson:
@mpdehaan I'm inclined to write a fix for this, however before I do spend the time on that, can you tell me if you will accept this or if this is something that is already in the pipeline?

bcoca:
@vpetersson I cannot find a PR in queue for this functionality

vpetersson:
Ok, I'll try to carve out some time this weekend to write a patch then.

vpetersson:
I took a closer look at this, and it appears as this is blocked by the lack of env-file support in docker-py.

I've gone ahead and opened an [issue](https://github.com/docker/docker-py/issues/565) over at the docker-py repo.

dexterbt1:
+1 to this

liquidvapour:
+1 We need this too. Now that it is in docker-py is that all that was needed to start work on getting it into ansible?

@vpetersson are you still up for doing this?

It would really clean up our build pipeline ;)

Thanks

vpetersson:
@liquidvapour The code has been accepted and merged into docker-py and will be in the next version (I presume). I'm not really sure how closely Ansible tracks these dependencies, but we'd at the very least have to wait until the next build of docker-py is out.

liquidvapour:
@vpetersson ok. Well I guess it is just a waiting game then ;) seems like you are all over it though. 

Looking forward to it.

vpetersson:
Added an experimental version in PR #2155. Not sure if this is the right approach, but would love to get some feedback on it.

denschu:
Any news here? Would love to see that feature in Ansible :+1:

gesellix:
@denschu I'm going to check the PR #2155 tonight /cc @vpetersson

gesellix:
@denschu see https://github.com/ansible/ansible-modules-core/pull/2155#issuecomment-177606142 for the current progress.

denschu:
[...]

PR #2929, opened 2016-02-01
add `env_file` support

gesellix:
This adds a new configuration parameter `env_file`. When `env_file` and `env` properties are present, they will be merged, while `env` parameters override the same keys of `env_file`. This is identical to the behaviour of the official Docker client.

_note_: credits go to @vpetersson, see #2155 and #792. This PR obsoletes #2155.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -223,6 +223,15 @@ options:
     description:
       - Pass a dict of environment variables to the container.
     default: null
+  env_file:
+    version_added: "2.1"
+    description:
+      - Pass in a path to a file with environment variable (FOO=BAR).
+        If a key value is present in both explicitly presented (i.e. as 'env')
+        and in the environment file, the explicit value will override.
+        Requires docker-py >= 1.4.0.
+    default: null
+    required: false
   dns:
     description:
       - List of custom DNS servers for the container.
@@ -661,7 +670,8 @@ class DockerManager(object):
             'labels': ((1, 2, 0), '1.18'),
             'stop_timeout': ((0, 5, 0), '1.0'),
             # Clientside only
-            'insecure_registry': ((0, 5, 0), '0.0')
+            'insecure_registry': ((0, 5, 0), '0.0'),
+            'env_file': ((1, 4, 0), '0.0')
             }
 
     def __init__(self, module):
@@ -712,7 +722,9 @@ class DockerManager(object):
         if self.module.params.get('links'):
             self.links = self.get_links(self.module.params.get('links'))
 
-        self.env = self.module.params.get('env', None)
+        env = self.module.params.get('env', None)
+        env_file = self.module.params.get('env_file', None)
+        self.environment = self.get_environment(env, env_file)
 
         # Connect to the docker server using any configured host and TLS settings.
 
@@ -837,6 +849,25 @@ class DockerManager(object):
                     self._cap_ver_req[capability][1],
                     '.'.join(map(str, self.docker_py_versioninfo)),
                     api_version))
+
+    def get_environment(self, env, env_file):
+        """
+        If environment files are combined with explicit environment variables, the explicit environment variables will override the key from the env file.
+        """
+        final_env = {}
+
+        if env_file:
+            self.ensure_capability('env_file')
+            parsed_env_file = docker.utils.parse_env_file(env_file)
+
+            for name, value in parsed_env_file.iteritems():
+                final_env[name] = str(value)
+
+        if env:
+            for name, value in env.iteritems():
+                final_env[name] = str(value)
+
+        return final_env
 
     def get_links(self, links):
         """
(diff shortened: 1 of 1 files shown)


cloud/docker/docker.py
```diff
@@ -713,6 +721,8 @@ def __init__(self, module):
             self.links = self.get_links(self.module.params.get('links'))
 
         self.env = self.module.params.get('env', None)
+        self.env_file = self.module.params.get('env_file', None)
```
gesellix: We might discuss about removing the `env` and `env_file` properties from `self`. As far as I can see they won't be needed after `self.environment` has been created as a merged result of both params. WDYT?

vpetersson: That's a good point. More DRY.

gesellix: I (force) pushed another commit which keeps the variables local to keep the code readable, but doesn't expose them in the module.

gregdek:
Thanks @gesellix. @cove @joshuaconner @softzilla @smashwilson @zfil @ThomasSteinbach please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/docker/docker.py
```diff
@@ -838,6 +848,24 @@ def ensure_capability(self, capability, fail=True):
...
                     api_version))
 
+    def get_environment(self, env, env_file):
+        """
+        If environment files are combined with explicit environment variables, the explicit environment variables will override the KW from the env file.
+        """
+        final_env = {}
+
+        if env_file:
+            parsed_env_file = docker.utils.parse_env_file(env_file)
```
smashwilson, line 861: This'll fail with a nasty exception if your docker-py is too old. The way to fail more gracefully (with a more user-friendly error message) would be to do an explicit docker-py version check; see [`_cap_ver_req`](https://github.com/gesellix/ansible-modules-core/blob/add_env_file_support_v2/cloud/docker/docker.py#L655-L673) and [`ensure_capability()`](https://github.com/gesellix/ansible-modules-core/blob/add_env_file_support_v2/cloud/docker/docker.py#L822-L849).

cloud/docker/docker.py
```diff
@@ -223,6 +223,14 @@
...
       - Pass a dict of environment variables to the container.
     default: null
+  env_file:
+    description:
+      - Pass in a path to a file with environment variable (FOO=BAR).
+        If a key value is present in both explicitly presented (i.e. as 'env')
+        and in the environment file, the explicit value will override.
+        Requires docker-py >= 1.4.0.
+    default: null
+    required: false
```
smashwilson, line 234: :point_up: While we're at it, this doc section should have a `version_added:` key in it.

smashwilson:
Nicely done. Just a pair of minor issues and we'll be good to go.

needs_revision

cloud/docker/docker.py
gesellix, line 234: `version_added` would be the Ansible version? So, a good value is `2.1`?

smashwilson, line 234: Yup!

gesellix:
@smashwilson thx for the review!
The fixups have been added in a squashed commit.

smashwilson:
:+1: :100: :sparkles: Perfect, thanks for the fixes!

shipit

gesellix:
:dancer: :smile:

denschu:
Awesome!

gesellix:
@smashwilson @gregdek I don't have to do anything here, right? If you need a rebase, just give me a ping. Thanks!

the-frey:
What's the status of this PR currently?

gesellix:
would this one have a chance to be included in Ansible 2.1? According to https://github.com/ansible/ansible-modules-core/pull/2929#issuecomment-180371895 it was shippable?
ping @bcoca

gregdek:
Hmmm... prbot is missing this case for some reason.

Marking for inclusion. Thanks @smashwilson for the review and thanks @gesellix for the PR.

gesellix:
:+1: :bow:

ballPointPenguin:
It is not clear how to use this.
For example, if I have a task at roles/dataserver/tasks/postgres.yml
that uses docker: env_file: myvars.env,
and the myvars.env file is in roles/dataserver/files/myvars.env
it is not being loaded.

I'm afraid the description:
`Pass in a path to a file with environment variable (FOO=BAR).`
has left me unsure of how to use this. Could somebody provide an example?

chouseknecht:
@ballPointPenguin in the case of the role example, the path would be specified as "{{ role_path }}/files/myvars.env". Outside of a role, use _playbook_dir_.

faustoFF:
Is it posible to specify multiple env files to use? env_file param is not a "list".

vpetersson:
@fabfuel It must be a single file in the current implementation.

PR #2938, opened 2016-02-01
added ability to control sleep between attempts

bcoca:
the default was 1s but it makes sense to make this configurable


gregdek:
Thanks @bcoca for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2941, opened 2016-02-02
initial add of new module eos_template

privateip:
The eos_template module works by allowing configurations to be pushed
to Arista EOS devices that can be templated by the Ansible Jinja2
template engine


gregdek:
cc: @ansible/core team for new core modules.

PR #2954, opened 2016-02-04
Make modify_user_usermod honor check mode on AIX.

### LINKED ISSUE #2953: User module fails or makes changes in check mode on AIX
##### Issue Type:

Bug Report
##### Ansible Version:

```
ansible 2.1.0 (detached HEAD 37eac4592e) last updated 2016/02/02 06:56:57 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 93d02189f6) last updated 2016/02/02 06:03:39 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD fff5ae6994) last updated 2016/02/02 06:03:40 (GMT +000)
  config file =
```
##### Ansible Configuration:

N/A
##### Environment:

AIX
##### Summary:

Reviewing the code for AIX in the user module I discovered bugs in check mode handling:
1. If changes to the user are needed then the task fails in check mode.
2. If only a password change is needed then the password is changed in check mode.

I don't have an AIX system to test on, so I tested the code by making the following changes:
1. Changed the platform in the AIX subclass from 'AIX' to 'Linux'.
2. Changed execute_command to log instead of execute.

This allowed me to verify the incorrect behavior on Ubuntu 14.04 LTS instead.
##### Steps To Reproduce:

Change UID in check mode:

```
ansible localhost -m user -a 'name=root uid=1000' --check
```

Change password in check mode:

```
ansible localhost -m user -a 'name=root update_password=always password=test' --check
```
##### Expected Results:

Change UID in check mode:

```
localhost | SUCCESS => {
    "append": false,
    "changed": true,
    "comment": "root",
    "group": 0,
    "home": "/root",
    "move_home": false,
    "name": "root",
    "shell": "/bin/bash",
    "state": "present",
    "uid": 0
}
```

Change password in check mode:

```
localhost | SUCCESS => {
    "append": false,
    "changed": true,
    "comment": "root",
    "group": 0,
    "home": "/root",
    "move_home": false,
    "name": "root",
    "password": "NOT_LOGGING_PASSWORD",
    "shell": "/bin/bash",
    "state": "present",
    "uid": 0
}
```
##### Actual Results:

Change UID in check mode:

```
localhost | FAILED! => {
    "changed": false,
    "failed": true,
    "msg": "",
    "name": "root",
    "rc": true
}
```

This is due to the module using `return (True, '', '')` instead of `return (0, '', '')` in check mode.

Change password in check mode:

The module executes:

```
/usr/sbin/chpasswd -e -c
```

This is due to the module not testing for check mode when setting the password.

mattclay:
This should resolve #2953. However, I'm only able to test this by running the AIX subclass on Linux. If someone could test this on AIX that would be good.

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -1869,13 +1869,15 @@ class AIX(User):
         if len(cmd) == 1:
             (rc, out, err) = (None, '', '')
         elif self.module.check_mode:
-            return (True, '', '')
+            return (0, '', '')
         else:
             cmd.append(self.name)
             (rc, out, err) = self.execute_command(cmd)
 
         # set password with chpasswd
         if self.update_password == 'always' and self.password is not None and info[1] != self.password:
+            if self.module.check_mode:
+                return (0, '', '')
             cmd = []
             cmd.append(self.module.get_bin_path('chpasswd', True))
             cmd.append('-e')


gregdek:
Thanks @mattclay for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #2959, opened 2016-02-04
Double imports, removed one

dagwieers:
time and signal were imported twice.

### DIFF
--- a/utilities/helper/fireball.py
+++ b/utilities/helper/fireball.py
@@ -74,8 +74,6 @@ import time
 import base64
 import syslog
 import signal
-import time
-import signal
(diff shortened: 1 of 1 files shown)


abadger:
Thanks @dagwieers. Turns out that we removed the fireball action plugin in 2.0 but forgot to remove the helper module.  I've now replaced the helper module with some documentation saying that fireball is gone so we don't need to make fixes to the helper module anymore.

dagwieers:
Ha, I didn't know fireball was removed ;-)
Not that this was very important.
I just noticed one such case by accident, and then found these too.

PR #2968, opened 2016-02-05
add retry with exponential backoff on route53

ashwanthkumar:
Similar to #569 but for route53.

### DIFF
--- a/cloud/amazon/route53.py
+++ b/cloud/amazon/route53.py
@@ -335,6 +335,21 @@ def commit(changes, retry_interval, wait, wait_timeout):
         raise TimeoutError()
     return result
 
+# Shamelessly copied over from https://git.io/vgmDG
+IGNORE_CODE = 'Throttling'
+MAX_RETRIES=5
+def invoke_with_throttling_retries(function_ref, *argv):
+    retries=0
+    while True:
+        try:
+            retval=function_ref(*argv)
+            return retval
+        except boto.exception.BotoServerError, e:
+            if e.code != IGNORE_CODE or retries==MAX_RETRIES:
+                raise e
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @ashwanthkumar. @bpennypacker please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

bpennypacker:
Looks good.  shipit.

Thanks, @ashwanthkumar!

ashwanthkumar:
@bpennypacker Thanks! One small question - Do you think, would it be useful if we can extract this retry method separately and use it across all AWS modules? Since they've throttling implemented for all the services.

bcoca:
that is something i was trying out here https://github.com/bcoca/ansible/tree/retry_rate (also rate limiting and result 'paging')

bpennypacker:
I like the retry idea.  I do think it would make for a good method for other modules if it can be abstracted properly. It might need a bit more robust exception handling so other exceptions can be handled - I'd have to think about that a bit.

ashwanthkumar:
@gregdek @bpennypacker Would we be merging this?

gregdek:
Thanks again to @ashwanthkumar for this PR, and thanks @bpennypacker for reviewing. Marking for inclusion.

joelthompson:
@bcoca -- just wondering about your previous comment. I just submitted #3379 for much the same thing, though in a different way (using a decorator), and was also thinking that we should do something more standard across AWS modules. It looks like you deleted the branch you referenced, so I'm not sure what you were experimenting with, but what about potentially putting something like this in ansible.module_utils.ec2?

PR #2974, opened 2016-02-05
adding autoremove to apt

chrrrles:


basictheprogram:
I see autoclean is an alias for autoremove. They are very different options for apt-get.

```
       autoclean
           Like clean, autoclean clears out the local repository of retrieved
           package files. The difference is that it only removes package files
           that can no longer be downloaded, and are largely useless. This allows
           a cache to be maintained over a long period without it growing out of
           control. The configuration option APT::Clean-Installed will prevent
           installed packages from being erased if it is set to off.
```

```
       autoremove
           autoremove is used to remove packages that were automatically installed
           to satisfy dependencies for other packages and are now no longer
           needed.
```

rotanid:
i agree with @basictheprogram - autoclean really shouldn't be an alias but a separate command! this is totally misleading. i created a bugreport: https://github.com/ansible/ansible/issues/24718

Issue #2465, opened 2015-11-13
Need to support ulimit on run command


kenwdelong:
##### ISSUE TYPE

Feature Idea
##### COMPONENT NAME

docker module
##### ANSIBLE VERSION

N/A
##### SUMMARY

Docker run command has an option --ulimit.  This does not appear to be settable from Ansible's docker module.  I would like to see this added.

jctanner:
resolved_by_pr https://github.com/ansible/ansible-modules-core/pull/2992

PR #2992, opened 2016-02-09
Adds ulimit to the docker module

dusdanig:
The ulimit will be specified as a list and separated by colons. The
hard limit is optional, in which case it is equal to the soft limit.

The ulimits are compared to the ulimits of the container and added
or adjusted accordingly on by a reload.

### DIFF
--- a/cloud/docker/docker.py
+++ b/cloud/docker/docker.py
@@ -370,6 +370,14 @@ options:
     required: false
     default: 0
     version_added: "2.1"
+  ulimits:
+    description:
+      - ulimits, list ulimits with name, soft and optionally
+        hard limit separated by colons. e.g. nofile:1024:2048
+        Requires docker-py >= 1.2.0 and docker >= 1.6.0
+    required: false
+    default: null
+    version_added: "2.1"
 
 author:
     - "Cove Schneider (@cove)"
@@ -378,6 +386,7 @@ author:
     - "Ash Wilson (@smashwilson)"
     - "Thomas Steinbach (@ThomasSteinbach)"
     - "Philippe Jandot (@zfil)"
+    - "Daan Oosterveld (@dusdanig)
 requirements:
     - "python >= 2.6"
     - "docker-py >= 0.3.0"
@@ -660,6 +669,7 @@ class DockerManager(object):
             'read_only': ((1, 0, 0), '1.17'),
             'labels': ((1, 2, 0), '1.18'),
             'stop_timeout': ((0, 5, 0), '1.0'),
+            'ulimits': ((1, 2, 0), '1.18'),
             # Clientside only
             'insecure_registry': ((0, 5, 0), '0.0')
             }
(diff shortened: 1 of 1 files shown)


dusdanig:
This fixes issue #2465

gregdek:
Thanks @dusdanig. @cove @joshuaconner @softzilla @smashwilson @zfil @ThomasSteinbach please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/docker/docker.py
```diff
@@ -1566,6 +1596,7 @@ def create_containers(self, count=1):
         else:
             params['host_config']['Memory'] = mem_limit
 
+        params['host_config']['ulimits'] = self.ulimits
```
smashwilson: `ulimit` support was added to [docker-py in 1.2.0](https://github.com/docker/docker-py/blob/60d814d742067f3e349119809935f8c0cb6c3b3b/docs/change_log.md#120) and [docker API version 1.18](https://docs.docker.com/engine/reference/api/docker_remote_api/#v1-18-api-changes), so we should gate ulimit support on an `ensure_capabilities` call. You can do this by adding an entry to [`_cap_ver_req`](https://github.com/dusdanig/ansible-modules-core/blob/devel/cloud/docker/docker.py#L654-L669) and then gating this call:

``` python
if self.ulimits is not None:
  self.ensure_capability('ulimits')
  params['host_config']['ulimits'] = self.ulimits
```

cloud/docker/docker.py
```diff
@@ -888,7 +908,7 @@ def get_start_params(self):
         optionals = {}
         for optional_param in ('devices', 'dns', 'volumes_from',
                 'restart_policy', 'restart_policy_retry', 'pid', 'extra_hosts',
-                'log_driver', 'cap_add', 'cap_drop', 'read_only', 'log_opt'):
+                'log_driver', 'cap_add', 'cap_drop', 'read_only', 'log_opt', 'ulimits'):
```
smashwilson: Reading through the source, I don't think this is actually used! It adds `ulimits` to `optionals`, but then it never makes its way into `params`.

cloud/docker/docker.py
```diff
@@ -370,6 +370,12 @@
     required: false
     default: 0
     version_added: "2.1"
+  ulimits:
+    description:
+      - ulimits, list ulimits with name, soft and optionally hard limit separated by colons. e.g. nofile:1024:2048
```
smashwilson: Would you mind noting the minimum docker-py and docker versions here?

> Requires docker-py >= 1.3.0 and docker >= 1.6.0

smashwilson:
Thanks for tackling this! This'll be handy.

Just a few comments, mostly on ensuring we fail correctly on older versions. It's looking good though.

needs_revision

dusdanig:
Thanks for the review. I really needed this feature ;)

I have updated the code and documentation accordingly.

smashwilson:
:+1: :sparkles: Perfect, thanks.

shipit

robynbergeron:
Thanks again to @dusdanig for this PR, and thanks @cove @joshuaconner @softzilla @smashwilson @zfil @ThomasSteinbach for reviewing. Marking for inclusion.

PR #3007, opened 2016-02-10
Enable env vars for vsphere_guest

jasonk:
This commit allows the connection information for
the vsphere_guest module to be provided as environment
variables, which makes it possible to use Cloud
Credentials from Ansible Tower in playbooks that utilize
vsphere_guest.

| ENV VAR | vsphere_guest param |
| --- | --- |
| VMWARE_HOST | vcenter_hostname |
| VMWARE_USER | username |
| VMWARE_PASSWORD | password |


juliedavila:
shipit

PR #3010, opened 2016-02-11
Filter exit commands when no other commands are included

grybak-arista:
In EOS, some commands will enter a configuration mode, where additional commands are not written until an exit command is given. When the exit command is included in the template, idempotency no longer works because exit is never found in the running-config. For example, the template

route-map map01 permit 10
   description A test map
   continue 20
   match as 22
   exit

will run successfully on the first run, but on the second run, the configuration clauses would be removed because they exist, and the final template will be

route-map map01 permit 10
   exit

which will run and indicate a change, although no change was performed.

The changes to eos_template.py filters the command list and removes any pair of commands where a non-exit command is followed immediately by an exit command that is indented by one level. The formatting implies we are entering a block and immediately exiting, so there is no need to enter the block at all.

So in the second run, the command list would be empty, and no change is made.

### DIFF
--- a/network/eos/eos_template.py
+++ b/network/eos/eos_template.py
@@ -109,6 +109,8 @@ responses:
   sample: ['...', '...']
 """
 
+import re
+
 def compare(this, other):
     parents = [item.text for item in this.parents]
     for entry in other:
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @grybak-arista. @privateip please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
@privateip This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

privateip:
+shipit

grybak-arista:
Sorry for the late reply.

Peter, we had discussed at one point just inserting the exit command into the NetworkModule.configure method in module_utils/eos.py.

From my testing, I think that solution would work just as well. I do not know if there may be a future case where the exit command may be required inside a template, but for now I think either method of handling the exit command will work.

privateip:
Yeah, I think thats the concern right now is putting that in without more exhaustive testing.  Lets get this change in now as we are both comfortable with it and we have tested it so we know it works and it addresses a real pain point.  We can discuss refactoring at a later date if that makes sense to do.

grybak-arista:
Sounds good. Thanks!

jimi-c:
Merged per the above discussion. Thanks!

PR #3014, opened 2016-02-11
fixes for uri module

bcoca:
- clarify docs on body_json behaviour
- only tranform into json if body input is not a string
  users keep passing json string and expecint it to not be jsonified again
- fixed issue with removes not handling path expansion correctly
- switched all path variables to 'type path' to handle expansions

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -63,12 +63,12 @@ options:
     default: null
   body:
     description:
-      - The body of the http request/response to the web service.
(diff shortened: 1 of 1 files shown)


robynbergeron:
Thanks @bcoca. @romeotheriault please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

Issue #2972, opened 2016-02-05
2.0.0.1 -- Yum module raises inaccurate failure messages


nan3333:
When using the yum module to install a package I had an issue due to the fact that my task was not provoking sudo. However, the error message generated by the module was "msg": "No Package matching 'xinetd' found available, installed or updated."

If I SSH to the box with the ansible user and run the same command (yum install xinetd) the error generated is "You need to be root to perform this command."

The error message specified by the yum module is inaccurate and misleading. Functionality should be corrected to return the error reported from the Yum tool rather than assuming that the package is not found.

Using ansible 2.0.0.1

jctanner:
In devel we have have the appropriate error message ...

```
TASK [yum] *********************************************************************
task path: /home/jtanner/workspace/issues/AMC-2972/site.yml:9
Using module file /home/jtanner/workspace/issues/AMC-2972/ansible/lib/ansible/modules/core/packaging/os/yum.py
<10.0.0.107> ESTABLISH SSH CONNECTION FOR USER: vagrant
<10.0.0.107> SSH: EXEC ssh -vvv -C -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=vagrant -o ConnectTimeout=10 -o ControlPath=/home/jtanner/.ansible/cp/ansible-ssh-%h-%p-%r 10.0.0.107 '/bin/sh -c '"'"'( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1466784232.64-119926675357730 `" && echo ansible-tmp-1466784232.64-119926675357730="` echo $HOME/.ansible/tmp/ansible-tmp-1466784232.64-119926675357730 `" ) && sleep 0'"'"''
<10.0.0.107> PUT /tmp/tmpm0khON TO /home/vagrant/.ansible/tmp/ansible-tmp-1466784232.64-119926675357730/yum.py
[...]

PR #3016, opened 2016-02-11
YUM: Catch 'as root' to attribute permissions error

nan3333:
Proposed fix for #2972.

If no packages are found by 'what provides' function check error message for permission problem before checking for previously-installed packages


robynbergeron:
Thanks @FatLASP for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

nan3333:
no problem. TBH a string comparison fix seems pretty rudimentary; but I've
checked it against several versions of RHEL and it seems to work. There may
be a more elegant fix that stands out to someone with a little more
experience with the code base

PR #3019, opened 2016-02-11
Fixed escaping issues with iam_policy policy_json template handling

sgnn7:
IAM doesn't accept preceeding spaces but Jinja2 templates require this
space to interpret the file correctly if you don't use `| to_json` - the
latter being highly unintuitive (as well as not represented in the docs)

Relevant links:
- https://stackoverflow.com/a/32014283
- https://groups.google.com/forum/#!topic/ansible-project/WUjiN9Wf32U

### DIFF
--- a/cloud/amazon/iam_policy.py
+++ b/cloud/amazon/iam_policy.py
@@ -307,6 +307,13 @@ def main():
           pdoc = json.dumps(pdoc)
         except Exception as e:
           module.fail_json(msg='Failed to convert the policy into valid JSON: %s' % str(e))
+
+      # XXX: IAM doesn't accept preceeding spaces but Jinja2 templates
+      #      require this space to interpret the file correctly if you don't
+      #      use `| to_json`:
+      #      - https://stackoverflow.com/a/32014283
+      #      - https://groups.google.com/forum/#!topic/ansible-project/WUjiN9Wf32U
+      pdoc = pdoc.strip()
   else:
     pdoc=None
 


robynbergeron:
Thanks @sgnn7. @defionscode please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
@defionscode This change is still pending your review; do you have time to take a look and comment? Please comment with text 'shipit' or 'needs_revision' as appropriate.

gregdek:
Thanks @sgnn7 for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

juliedavila:
@sgnn7 could you add more info to reproduce?

I tried the following and it worked without issue

``` yml
---
- name: test IAM
  hosts: localhost
  connection: local
  tasks:
    - name: Create IAM Role
      iam:
        iam_type: role
        name: testing3019
        region: us-east-1
        state: present

    - name: Add policy
      iam_policy:
        iam_type: role
        iam_name: testing3019
        region: us-east-1
        state: present
        policy_name: testing3019pol
        policy_json: '{"Version":"2012-10-17","Statement":[{"Action":["sns:*"],"Effect":"Allow","Resource":"*"}]}'
```

however, i did notice that it doesn't work with the file lookup, with or without your pull request. 

``` yml
---
- name: test IAM
  hosts: localhost
  connection: local
  tasks:
    - name: Create IAM Role
      iam:
        iam_type: role
        name: testing2684
        region: us-east-1
        state: present

    - name: Add policy
      iam_policy:
        iam_type: role
        iam_name: testing2684
        region: us-east-1
        state: present
        policy_name: testing3019
        policy_json: "{{ lookup('file', 'files/policy.json') | to_json }}" 
```

sgnn7:
@defionscode 

The error is when you use `lookup`s and to get the right content you have to use a preceeding space before the `{{` or [Jinja will mangle the escaping](https://stackoverflow.com/a/32014283) (this is still a real issue - even the [docs in the file](https://github.com/sgnn7/ansible-modules-core/blob/devel/cloud/amazon/iam_policy.py#L107) suggest it):
- Good

```
        policy_json: " {{ lookup('file', 'files/policy.json') | to_json }}" 
```
- Bad

```
        policy_json: "{{ lookup('file', 'files/policy.json') | to_json }}" 
```

So the actual bug here is that AWS doesn't accept that preceeding space so it rejects the change and thus you cannot use lookups at all. The fix here strips the string so that AWS can chew on it when you preceed the lookup with that space so that it doesn't explode.

juliedavila:
Ah! I see, thanks. I'll take a look again

juliedavila:
shipit

Dmitry1987:
Hi all,
I'm trying to use either "policy_json" or "policy_document", as suggested with all possible workarounds and methods to pass JSON to it ( load from template, file, pass into variable and use, escaping, spaces like suggested here, and finally simplifying my policy down to 1 line   _{ "Version": "2012-10-17" }_    )
but nothing works no matter what.

using -vvv debug I see it whether converting double quotes to single, or adding escape symbols and \n  (when it was multiline policy)  but always same result:  "Syntax errors in policy."

It seems like it doesn't work at all, I can't assign any custom policy using this module (only attaching managed policy works, but not creating inline policy).

Should I open detailed "issue" ticket somewhere?  I might be writing this into wrong thread

thank you

gregdek:
@Dmitry1987 yes, please open in another ticket.

sgnn7:
@Dmitry1987 Use this patch/PR and make note of my comment [here](https://github.com/ansible/ansible-modules-core/pull/3019#issuecomment-195570887) and you should be able to do this. Internally we do this by pulling the repo from Git and applying this patch through a script so I'd probably advise the same until this PR (or some other one that fixes this) gets merged.

I posted our Anisble-patching script here if you want to try it out: https://gist.github.com/sgnn7/56679b4246d4637d0fac994480ae6056

PR #3028, opened 2016-02-12
find module: don't follow symlinks

### LINKED ISSUE #3027: find module doesn't recognize symlinks correctly
##### Summary:

When using the "find" module, symlinks are falsely identified as regular files.
##### Steps To Reproduce:

How to reproduce:

```
mkdir /tmp/test
touch /tmp/file
ln -s /tmp/file /tmp/test/symlink
```

then run a playbook like this:

```

---
- hosts: localhost
  tasks:
  - find: paths="/tmp/test"
    register: find_result
  - debug: var=find_result
```
##### Expected Results:

```
        "files": [ 
            { 
                "islnk": true,  
                "isreg": false,  
                "path": "/tmp/test/symlink",  
            }
         ]
```
##### Actual Results:

In the output, you will see

```
        "files": [ 
            { 
                "islnk": false,  
                "isreg": true,  
                "path": "/tmp/test/symlink",  
            }
         ]
```

When running
 `- find: paths="/tmp/test" file_type=file`
to explicitly only match regular files, the result is the same (the symlink is in the results list).
##### Solution

The reason for this is that [os.stat](https://docs.python.org/2/library/os.html#os.stat) is used to get the file info: https://github.com/ansible/ansible-modules-core/blob/devel/files/find.py#L316, and `os.stat` follows symlinks.
Instead, [os.lstat](https://docs.python.org/2/library/os.html#os.lstat) should be used, which does not follow symlinks.
It would also be possible to add an additional option `follow_symlinks=yes/no` to determine behaviour for this, but IMO, not following symlinks should be the default.

Yannik:
fixes #3027

### DIFF
--- a/files/find.py
+++ b/files/find.py
@@ -313,7 +313,10 @@ def main():
                        continue
 
                     try:
-                        st = os.lstat(fsname)
+                        if params['follow']:
+                            st = os.lstat(fsname)
+                        else
+                            st = os.stat(fsname)
                     except:
                         msg+="%s was skipped as it does not seem to be a valid file or it cannot be accessed\n" % fsname
                         continue
--- a/files/find.py
+++ b/files/find.py
@@ -313,7 +313,7 @@ def main():
                        continue
 
                     try:
-                        st = os.stat(fsname)
+                        st = os.lstat(fsname)
                     except:
                         msg+="%s was skipped as it does not seem to be a valid file or it cannot be accessed\n" % fsname
                         continue
--- a/files/find.py
+++ b/files/find.py
@@ -314,9 +314,9 @@ def main():
 
                     try:
                         if params['follow']:
-                            st = os.lstat(fsname)
-                        else:
                             st = os.stat(fsname)
+                        else:
+                            st = os.lstat(fsname)
                     except:
                         msg+="%s was skipped as it does not seem to be a valid file or it cannot be accessed\n" % fsname
                         continue
--- a/files/find.py
+++ b/files/find.py
@@ -315,7 +315,7 @@ def main():
                     try:
                         if params['follow']:
                             st = os.lstat(fsname)
-                        else
+                        else:
                             st = os.stat(fsname)
                     except:
                         msg+="%s was skipped as it does not seem to be a valid file or it cannot be accessed\n" % fsname


sivel:
This should probably be a configuration.  The stat module uses a `follow` param, to determine how this should work.

Yannik:
Actually, there already is a parameter "follow" for the find module, which should determine whether to follow symlinks or not. It defaults to false, so this should not be happening. The parameter is only applied whether to follow directory-symlinks, but not file-symlinks.
Should I open a PR to make the "follow" paramter follow both directories/files?

files/find.py
```diff
@@ -313,7 +313,7 @@ def main():
                        continue
 
                     try:
-                        st = os.stat(fsname)
+                        st = os.lstat(fsname)
```
bcoca: ~~shouldn't this be conditioned on 'follow'? as this now breaks the opposite case (follow=yes)~~
follow explicitly state it only pertains to paths

Yannik: It depends. As path is always the actual symlink and not the file that it points to, I would expect the modes etc. to belong to that symlink (path) aswell. 
The other behaviour could be desired in some cases too, though. (Follow symlinked directories and list modes of the symlinked target - maybe change "path" too?)
Maybe two config settings, whether to follow symlinked directories and a second one, whether to follow symlinked files make sense?

bcoca: i'm thinking we emulate the find command, follow does both directory paths and files (make note of change in description).

another thing i was thinking was about expanding types (symlink, devices, fifos, etc) right now 'file' means 'regular file'

Yannik: I agree on expanding filetypes, this should also be a list, so you can select multiple file types.

Yannik:
@bcoca what do you think about this? now it uses os.stat/os.lstat depending on the "follow" option, as it has probably been intended.

Yannik:
When handling symlinks, it would be useful to have a option which models find's `-lname`.
(If `follow` models `find -L`)
As a reference: https://www.gnu.org/software/findutils/manual/html_node/find_html/Symbolic-Links.html

robynbergeron:
Thanks @Yannik for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

Yannik:
Hey @robynbergeron, can you give me an estimation how long it will take?

gregdek:
Thanks @Yannik for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregdek:
@Yannik A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

Yannik:
I do believe that this has been fixed in https://github.com/ansible/ansible/commit/85f6c26dfff53a0e71a0e983e89943781a0c4cd4.

PR #3030, opened 2016-02-12
Fix parameters where type is assumed.

Shrews:
A change is coming to Ansible where module params will default to str.
Many of our modules were taking advantage of this by not being explicit
about the type, so they will break when that change merges. This hopefully
catches those cases.


Shrews:
This was found when @abadger merged this change and our module tests in shade broke. It has since been reverted, but it will likely show up in the next release.

ping @emonty @abadger @j2sol

abadger:
Thanks @Shrews!  Merged to devel.  I'll talk with @jimi-c and @bcoca more about whether we want the underlying change in 2.0.2 or 2.1.0.  I think that contrary to what we've been telling people, the default actually isn't a string even in 1.x.  Instead it's a result of how we parse.  With yaml dicts, if you create an int value, then it's an int.  If you use key=value syntax to set a parameter (for instance, using `ansible -m wait_for -a 'timeout=5'`) then the value is a string.  So it might be something we decide should go introduce into 2.1.0 rather than being a bugfix to the 2.0.x series.

davidaah:
I tried running the os_server module from the devel branch and there seems to be an issue with `meta = dict(default=None, type='raw'),`. I don't believe 'raw' is a builtin type for python. Please let me know if I should open an issue for this since the branch has already been merged. The emitted error from ansible is:

```
failed: [127.0.0.1] => (item=1) => {"failed": true, "item": "1", "msg": "implementation error: unknown type raw requested for meta"}
```

Shrews:
@davidaah The Ansible 'raw' type is introduced in commit 62765858825173dd532827fe4c1dfb246eb8db47 which is also the change that breaks these modules.

davidaah:
Thanks for the clarification, I see the change! I missed this since it was in the base ansible project and also when I pulled this down I was running it against the latest 2.0.0.1 release

PR #3037, opened 2016-02-14
use expanduser() before testing user_data as file

immanetize:
I like to use ~/somepath instead of absolute paths because
that's more shareable.  Without expansion, the path wasn't
considered a file, and the resulting cloud-config user_data
contained a string for the file path instead of the file context.
So, expand it.

### DIFF
--- a/cloud/rackspace/rax.py
+++ b/cloud/rackspace/rax.py
@@ -285,7 +285,7 @@ def create(module, names=[], flavor=None, image=None, meta={}, key_name=None,
     if user_data:
         config_drive = True
 
-    if user_data and os.path.isfile(os.path.expanduser('user_data')):
+    if user_data and os.path.isfile(os.path.expanduser(user_data)):
         try:
             user_data = os.path.expanduser('user_data')
             f = open(user_data)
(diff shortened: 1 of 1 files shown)


cloud/rackspace/rax.py
```diff
@@ -285,7 +285,7 @@ def create(module, names=[], flavor=None, image=None, meta={}, key_name=None,
     if user_data:
         config_drive = True
 
-    if user_data and os.path.isfile(user_data):
+    if user_data and os.path.isfile(os.path.expanduser('user_data')):
```
sivel: The quotes around `user_data` would cause it to expand to `/home/username/user_data` as opposed the actual value of `user_data`.

immanetize:
fail, but fixed now, thanks @sivel.  Also tested with my playbook that was failing :)

abadger:
Note in 2.0 we added the path type for module parameters.  Path type is a
str that has expandvars and expanduser run on it.  In many cases you can
just switch a parameter to that you're instead of running those functions
manually.

sivel:
In this case the value is either a path or data.  Setting the type to path could cause problems if data was passed instead of a file.

robynbergeron:
Thanks @immanetize. @j2sol @sivel please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

omgjlk:
shipit

sivel:
Merged, thanks!

PR #3043, opened 2016-02-15
Add option to prevent gem from waiting for human input

techraf:
Now: when uninstalling a gem (`state=absent`) on which other gems are dependent, `gem` waits for human input with the following prompt:

```
`If you remove this gem, these dependencies will not be met.`
`Continue with Uninstall? [yN]`
```

Fix: add `--abort-on-dependent` argument for `gem` to fail in such case instead of waiting indefinitely.


robynbergeron:
Thanks @techraf for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #3061, opened 2016-02-18
Clarify that httplib2 is required on target host

hyperized:
Ansible wont install this requirement on target hosts, thus the documentation needs some clarification.

Related to:
- https://github.com/ansible/ansible/issues/14342
- https://github.com/ansible/ansible/issues/10343


abadger:
All requirements are required on the host that runs the module so it isn't
right to do this at the module level.  But we could do it in the code that
generates the docs and put it into the header.  So it would read something
like this on the html version:

```
<div><h3>Requirements</h3>(on machine that executed the module)<\div>
```

I don't have my computer out just now so I can't point you at the exact
code but it's probably in the module doc formatter in the ansible/ansible
repo.

hyperized:
Thank you for pointing out the right place!

PR #3075, opened 2016-02-19
Mismatch between documentaction and expected module variable names ...

bsmoo:
##### Issue Type:

Please pick one and delete the rest:
- Bugfix Pull Request
- Docs Pull Request
##### Plugin Name:

rax_cdb
##### Ansible Version:

```
$ ansible --version
ansible 2.0.0.2
```
##### Summary:

Please describe the change and the reason for it.

rax_cdb module was not working due to a mismatch with the documentation and the  expected module variable names. This has been fixed and tested on Ansible 1.9.4 and 2.0.0.2
##### Example output:

```
fatal: [localhost -> localhost]: FAILED! => {"changed": false, "failed": true, "msg": "version is required for the \"rax_cdb\" module"}
fatal: [localhost -> localhost]: FAILED! => {"changed": false, "failed": true, "msg": "type is required for the \"rax_cdb\" module"}
```

...(cdb_type/cdb_version)


sivel:
I do not believe this to be the correct fix.  The change should be to change the `module.params.get` lines to reference the correct value, `cdb_type` instead of `type`

bcoca:
@blakesmoore thanks for pointing this out and trying to fix it, the following commit took care of it 369806de25e282d03a05ceb90b1ead83d8005a17, I also added your argument rename as an alias so it will work both ways now.

bsmoo:
@sivel @bcoca The reason I went for the method of changing from `'cdb_type'` to `'type'`  as you've already loaded the cdb module so you shouldn't need to restate this fact.

As you can see in the [clb module](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/rackspace/rax_clb.py#L293) the argument is `'type'`.

bcoca:
understood, that is why i added it as an alias to allow for you to use naked `type` and `version`, but also kept the cdb_ ones as that is what original author intended.

PR #3091, opened 2016-02-22
Add Ceph RGW S3 compatibility

jmunhoz:
##### Issue Type:
- Feature Pull Request (minimal impact in the current code)
##### Plugin Name:

./cloud/amazon/s3.py
##### Ansible Version:

```
$ ansible --version
ansible 2.1.0 (devel 3ddcabee0e) last updated 2016/02/11 11:21:11 (GMT +200)
  lib/ansible/modules/core: (s3-ceph 1da965e039) last updated 2016/02/11 14:27:39 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 6aeb2ab6cf) last updated 2016/02/11 11:34:36 (GMT +200)
  config file = 

```
##### Summary:

Ansible supports S3 on AWS, Walrus and FakeS3 but it doesn't support Ceph RGW S3. This patch supports Ceph RGW S3.

More information at:

https://groups.google.com/forum/#!topic/ansible-devel/XxVeoHehpSM
##### Example output:

```
```

Ceph Object Gateway (Ceph RGW) is an object storage interface built on top of
librados to provide applications with a RESTful gateway to Ceph Storage
Clusters:

http://docs.ceph.com/docs/master/radosgw/

This patch adds the required bits to use the RGW S3 RESTful API properly.

Signed-off-by: Javier M. Mellid jmunhoz@igalia.com

### DIFF
--- a/cloud/amazon/s3.py
+++ b/cloud/amazon/s3.py
@@ -131,9 +131,13 @@ options:
     version_added: "2.0"
   s3_url:
     description:
-      - S3 URL endpoint for usage with Eucalypus, fakes3, etc.  Otherwise assumes AWS
+      - S3 URL endpoint for usage with Ceph, Eucalypus, fakes3, etc.  Otherwise assumes AWS
     default: null
     aliases: [ S3_URL ]
+  rgw:
+    description:
+      - Enable Ceph RGW S3 support
+    default: false
   src:
     description:
       - The source file path when performing a PUT operation.
@@ -151,6 +155,9 @@ extends_documentation_fragment: aws
 EXAMPLES = '''
 # Simple PUT operation
 - s3: bucket=mybucket object=/my/desired/key.txt src=/usr/local/myfile.txt mode=put
+
+# Simple PUT operation in Ceph RGW S3
+- s3: bucket=mybucket object=/my/desired/key.txt src=/usr/local/myfile.txt mode=put rgw=true s3_url=http://localhost:8000
 
 # Simple GET operation
 - s3: bucket=mybucket object=/my/desired/key.txt dest=/usr/local/myfile.txt mode=get
@@ -384,6 +391,7 @@ def main():
             prefix         = dict(default=None),
             retries        = dict(aliases=['retry'], type='int', default=0),
             s3_url         = dict(aliases=['S3_URL']),
+            rgw            = dict(default='no', type='bool'),
             src            = dict(),
         ),
     )
@@ -408,6 +416,7 @@ def main():
     prefix = module.params.get('prefix')
     retries = module.params.get('retries')
     s3_url = module.params.get('s3_url')
+    rgw = module.params.get('rgw')
     src = module.params.get('src')
 
     for acl in module.params.get('permission'):
@@ -437,6 +446,10 @@ def main():
     if not s3_url and 'S3_URL' in os.environ:
         s3_url = os.environ['S3_URL']
 
+    # rgw requires an explicit url
+    if rgw and not s3_url:
+        module.fail_json(msg='rgw flavour requires s3_url')
+
     # bucket names with .'s in them need to use the calling_format option,
     # otherwise the connection will fail. See https://github.com/boto/boto/issues/2836
     # for more details.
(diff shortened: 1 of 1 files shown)


thereallukl:
How about adding extra is_rgw method instead of new parameter (https://github.com/lleszczu/ansible-modules-core/commit/810cd9c2143777831a1d5fb09e48bfb8d7525e2f)? That way it's consistent with fakes3.

jmunhoz:
@lleszczu I would prefer being consistent with Walrus and AWS instead of fakes3. fakes3 scheme encoding/decoding looks a quick way to add/test with fakes3. It would be weird having walrus and aws using http/https and rgw using rgw/rgws.

Anyway, it would be trivial to adapt the patch if it make sense at the end. I guess you miss the mail I share on ansible-devel [1], as suggested by @bcoca, 12 days ago. Maybe you could reply that mail with this approach in order to boost the discussion. What do you think?

[1] https://groups.google.com/forum/#!topic/ansible-devel/XxVeoHehpSM

Thanks for commenting!

thereallukl:
>  would prefer being consistent with Walrus and AWS instead of fakes3. fakes3 scheme encoding/decoding looks a quick way to add/test with fakes3

Seems reasonable
+1

robynbergeron:
Thanks @jmunhoz for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

robynbergeron:
Hey, @leseb -- could you maybe eyeball this from the ceph angle and weigh in? :)

leseb:
@robynbergeron sure, will do :)

leseb:
From a Ceph angle, I don't see any issue, so :+1: for me.

cloud/amazon/s3.py
```diff
@@ -131,9 +131,13 @@
...
   s3_url:
     description:
-      - S3 URL endpoint for usage with Eucalypus, fakes3, etc.  Otherwise assumes AWS
+      - S3 URL endpoint for usage with Ceph, Eucalypus, fakes3, etc.  Otherwise assumes AWS
     default: null
     aliases: [ S3_URL ]
+  rgw:
+    description:
+      - Enable Ceph RGW S3 support
+    default: false
```
abadger, line 140: Needs to specify version_added: 2.2

cloud/amazon/s3.py
```diff
@@ -590,13 +608,21 @@ def main():
...
-        # If the bucket does not exist then bail out
-        if bucket_object is None:
-            module.fail_json(msg="Target bucket (%s) cannot be found"% bucket, failed=True)
-
-        list_keys(module, bucket_object, prefix, marker, max_keys)
+        if bucket == '/':
+            try:
+                all_buckets = s3.get_all_buckets()
+                s3_keys = [x.name for x in all_buckets]
+            except s3.provider.storage_response_error, e:
```
abadger: Need to make sure that these compile under python3 (we'll shoot for actually working under python3 later).  It looks like boto is a requirement here and boto requires python>=2.6 so we can just do:

``` python
except s3.provider.storage_response_error as e:
```

catching exceptions in files that do not have libraries which require python >= 2.6 would use something like this instead:

``` python
from ansible.module_utils.py24compat import get_exception
[...]
except s3.provider.storage_response_error:
    e = get_exception()
```

cloud/amazon/s3.py
abadger, line 140: From bcoca: If this option is specific to ceph then there needs to be an indication in the docs that it is ignored when used against other S3 compatible endpoints.

abadger:
made a few general comments.  Our main aws guy is on vacation right now so looking around for someone else to review this change from an S3 angle.

ryansb:
Other than Toshio's syntax comments, the code itself looks good here. I don't have a ceph cluster to test, but the S3 bucket functionality isn't affected by this change.

jmunhoz:
@abadger @ryansb Thanks for the reviewing. It is very appreciated. I added the mentioned changes on top of the original patches. Rebased to catch the head and tested with Ceph RGW. It works ok on my side.

cloud/amazon/s3.py
```diff
@@ -591,13 +610,21 @@ def main():
...
     # Support for listing a set of keys
     if mode == 'list':
-        bucket_object = get_bucket(module, s3, bucket)
-
-        # If the bucket does not exist then bail out
-        if bucket_object is None:
-            module.fail_json(msg="Target bucket (%s) cannot be found"% bucket, failed=True)
-
-        list_keys(module, bucket_object, prefix, marker, max_keys)
+        if bucket == '/':
```
ryansb: For clarity, can you make this only happen on RGW flavors? This code is specific to RGW, and I think it convolutes what we had before since there's no (apparent) reason for special-casing buckets named `/`.

jmunhoz: @ryansb I just removed the owned buckets listing operation instead of making it happen on RGW only. I think it could add unnecessary divergence against the AWS S3 interface in this moment.

jmunhoz:
@ryansb Pushed updated patchset tested with Ceph RGW.

ryansb:
LGTM

PR #3099, opened 2016-02-23
Fix: Incorrect parameters for module.set_fs_attributes_if_different in file.py

bassco:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

files/file.py
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Summary:

file module crashes with invalid number of parameters when set permissions or ownership
##### Example output:

```
File "/home/ubuntu/.ansible/tmp/ansible-tmp-1456176384.07-262519109202902/file, line 318, in main
  changed = module.set_fs_attributes_if_different(file_args, changed, diff)
  TypeError: set_fs_attributes_if_different() takes exactly 3 arguments (4 given)
"msg": "MODULE FAILURE"
```

### DIFF
--- a/files/file.py
+++ b/files/file.py
@@ -87,13 +87,6 @@ options:
       - 'force the creation of the symlinks in two cases: the source file does
         not exist (but will appear later); the destination exists and is a file (so, we need to unlink the
         "path" file and create symlink to the "src" file in place of it).'
-  follow:
-    required: false
-    default: "no"
-    choices: [ "yes", "no" ]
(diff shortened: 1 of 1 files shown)


bcoca:
you seem to be mixing codebases, this error only happens if you are using core that is older then when this feature was added, this is not the case in stable or current devel. Also you are reintroducing some old bugs.

bassco:
thanks for the feedback

PR #3112, opened 2016-02-24
added sql_log_bin setting to disable binary logging as option

Jmainguy:
##### Issue Type:

``````
 - Feature Pull Request

##### Plugin Name:

mysql_user

##### Ansible Version:

```
ansible 2.1.0
```

##### Summary:

Adds ability to disable sql_log_bin
``````

### DIFF
--- a/database/mysql/mysql_user.py
+++ b/database/mysql/mysql_user.py
@@ -69,6 +69,13 @@ options:
     choices: [ "yes", "no" ]
     default: "no"
     version_added: "1.4"
+  sql_log_bin:
+    description:
+      - Whether binary logging should be enabled or disabled for the connection.
+    required: false
+    choices: ["yes", "no" ]
+    default: "yes"
+    version_added: "2.1"
   state:
     description:
       - Whether the user should exist.  When C(absent), removes
@@ -138,6 +145,9 @@ mydb.*:INSERT,UPDATE/anotherdb.*:SELECT/yetanotherdb.*:ALL
 
 # Example using login_unix_socket to connect to server
 - mysql_user: name=root password=abc123 login_unix_socket=/var/run/mysqld/mysqld.sock
+
+# Example of skipping binary logging while adding user 'bob'
+- mysql_user: name=bob password=12345 priv=*.*:USAGE state=present sql_log_bin=no
 
 # Example .my.cnf file for setting the root password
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Jmainguy** [comment]
This is a rebase of #2639 I recommend closing that, and continuing the discussion here.
1. I have not tested this yet.
2. This appears to only be enabled/disabled for user_add, we should add this for user_mod, and user_delete right? @tersmitten @grypyrg

**Comment by gregdek** [comment]
Thanks @Jmainguy. Since you are a maintainer of this module, we are marking this PR for inclusion.

**Comment by grypyrg** [comment]
@Jmainguy : yes, any `user_mod`, `user_delete` should have this too. I wrote it originally to disable binary logging just after the connection is created.
The rebase that was pushed here does not, it only does it in `user_add`

Also, if I quickly check the code that you wrote, you basically disable binary logging by default, when it's set to `True`.
https://github.com/Jmainguy/ansible-modules-core/commit/367f8d0b4d08b88247fbdc0477c20b2eaf62f183#diff-b4dbfbaa641f4d515d58a88ce227a981R231

```
 +    if sql_log_bin:
 +        cursor.execute("SET SQL_LOG_BIN=0;")
 +
```

 It should be the other way around, if `sql_log_bin is false`, then we set `SQL_LOG_BIN=0`

Jmainguy:
@grypyrg Ty, sorry about that. I will fix that up

Jmainguy:
@grypyrg added changes, give it a look now.

grypyrg:
Thanks, looks better, but `privileges_grant` and `privileges_revoke` also need it. 

I still think this should be done directly after `mysql_connect`, then the rest of the session will skip binary logging completely. somewhere after https://github.com/Jmainguy/ansible-modules-core/blob/c9b87e3de44684a6009009e8dafbbf73d3eafb89/database/mysql/mysql_user.py#L540
No need to worry about it anymore.

Otherwise any other pull request in the future that adds certain functionality might easily break the code and have events not being skipped in the MySQL Binary log, and not only the code would be broken:
- it might cause unwanted binary log events in MySQL, which could easily cause replication to break across a replication tree (and cause production issues).
- It can cause errand GTID transactions in some environments: https://www.percona.com/blog/2014/05/19/errant-transactions-major-hurdle-for-gtid-based-failover-in-mysql-5-6/

Jmainguy:
@grypyrg well I feel a little foolish, that looks much cleaner and better. Take a look now.

grypyrg:
Sorry for being a bit late, but :+1: it looks good now. (didn't test though). 

Thanks!

Issue #3111, opened 2016-02-24
Fix Argument Specs for arguments without `type`


sivel:
In https://github.com/ansible/ansible/commit/62765858825173dd532827fe4c1dfb246eb8db47 we made a change to default all arguments without an explicit `type` to be `'str'` type.

This has been known to cause issues in numerous modules.

Here is a list of modules that do not supply `type`, and that have a default that is not an instance of `basestring`:
- _ec2_ami_search.py
- _glance_image.py
- _keystone_user.py
- _nova_compute.py
- _nova_keypair.py
- _quantum_floating_ip.py
- _quantum_floating_ip_associate.py
- _quantum_network.py
- _quantum_router.py
- _quantum_router_gateway.py
- _quantum_router_interface.py
- _quantum_subnet.py
- apache2_module.py
- apt.py
- apt_key.py
- apt_repository.py
- apt_rpm.py
- assemble.py
- async_status.py
- azure.py
- cloudformation.py
- command.py
- copy.py
- cron.py
- django_manage.py
- docker.py
- docker_image.py
- easy_install.py
- ec2.py
- ec2_ami.py
- ec2_ami_find.py
- ec2_asg.py
- ec2_eip.py
- ec2_elb.py
- ec2_elb_lb.py
- ec2_facts.py
- ec2_group.py
- ec2_key.py
- ec2_lc.py
- ec2_metric_alarm.py
- ec2_scaling_policy.py
- ec2_snapshot.py
- ec2_tag.py
- ec2_vol.py
- ec2_vpc.py
- ec2_vpc_net.py
- elasticache.py
- elasticache_subnet_group.py
- eos_command.py
- eos_config.py
- eos_eapi.py
- eos_template.py
- file.py
- gc_storage.py
- get_url.py
- git.py
- hg.py
- htpasswd.py
- iam.py
- iam_cert.py
- iam_policy.py
- ini_file.py
- ios_command.py
- ios_config.py
- ios_template.py
- iosxr_command.py
- iosxr_config.py
- iosxr_template.py
- junos_command.py
- junos_config.py
- junos_template.py
- lineinfile.py
- linode.py
- mount.py
- mysql_db.py
- mysql_user.py
- mysql_variables.py
- nxos_command.py
- nxos_config.py
- nxos_nxapi.py
- nxos_template.py
- ops_template.py
- os_auth.py
- os_floating_ip.py
- os_image.py
- os_image_facts.py
- os_ironic.py
[...]

PR #3113, opened 2016-02-24
cloudformation.py - Added explicit type for 'tags' option

Constantin07:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

cloud/amazon/cloudformation.py
##### Ansible Version:

```
ansible 2.1.0
```
##### Summary:

Fixes issue in #3111.


abadger:
Thanks for the fix!

PR #3134, opened 2016-02-28
Suggest using uri if curl is detected

mscherer:
##### Issue Type:
- Docs Pull Request
##### Plugin Name:

command
##### Summary:

Some people use the command module to run curl not to
download file, but to interact with remote HTTP api, and the
get_url suggestion is not suitable. So adding uri in the warning
would give a more meaningful warning.

Suggested by Tahvok on #ansible irc channel.


gregdek:

resmo:
makes sense

Issue #2863, opened 2016-01-22
win_stat returns md5, but the data is sha1


fxfitz:
##### Issue Type:
- Bug Report
##### Component Name:

win_stat module
##### Ansible Version:

ansible 1.9.3
(even though I'm on 1.9.3, I provide links to code in the devel branch)
##### Ansible Configuration:

N/A
##### Environment:

N/A
##### Summary:

The win_stat modules reportedly returns a checksum and md5, but in all reality it is a sha1.
##### Steps To Reproduce:

Run win_stat on a file and take note of 'md5' result (and the 'checksum' result if desired).
##### Expected Results:

I would expect 'md5' result to be the actual md5sum of the file.
Additionally, if `sha1` was returned in the result, I would expect it to be the actual sha1sum of the file.
##### Actual Results:

In actuality, the 'md5' result is the sha1sum of the file.
You can easily see the bug in the devel branch at these two links:
https://github.com/ansible/ansible-modules-core/blob/devel/windows/win_stat.ps1#L68
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/powershell.ps1#L213

sivel:
Seems the hashing function used to return MD5 in the past, but was renamed and made to return SHA1:

https://github.com/ansible/ansible/commit/8793308c39bf064106f08b74e5cb468c94bf1d83

Also, https://github.com/ansible/ansible-modules-core/commit/b5399d34464475a9aa87c6b928628cd262022cd5

jhawkesworth:
Yes that's right.  If I recall it was changed because MD5 can't be made FIPS 140 compliant.  
Probably best to rename 'checksum' eventually, although it might break existing playbooks, so the parameter should really be deprecated and changed after a couple of ansible versions have passed.
I guess just documenting the change for now would be the first step.

fxfitz:
I noticed this bug when I was looking at an old task of ours which had the following line:

```
when: "result.stat.md5 != {{ package_sha1 }}"
```

And obviously you shouldn't have that.  :-P

fxfitz:
I think the win_stat module should/could return different hashing algorithms, making it very similar to the [stat module](http://docs.ansible.com/ansible/stat_module.html).

Personally, I'm inclined to the following steps: 
1. Adding a new [optional] parameter `checksum_algorithm` (defaulting to SHA1, since that's what it is now)
[...]

PR #3146, opened 2016-02-29
Add get_checksum option to win_stat to work with new use of the stat module for checksumming

sivel:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

`win_stat`
##### Summary:

As of https://github.com/ansible/ansible/commit/b9d0662faf2910dea503c0c8c35ad9951a4c6e9c ansible now uses the stat module for checksumming instead of shell checksum code.

Unfortunately in the invocation of the stat module we set `get_md5=False` but `get_checksum=True`.  `win_stat` did not support `get_checksum` and as such never returned a checksum.

This PR effectively aliases `get_checksum` to `get_md5`, but in a way that should allow us to expand on allowing different algorithms and leaving `get_checksum` and `get_md5` independent of each other.

There are other issues with `win_stat` right now, where `get_md5` always returns a sha1 instead of an md5, however this PR does not try to solve that, but only adds guidance around the fact that `get_md5` returns a `sha1` since ansible v1.9.  See https://github.com/ansible/ansible-modules-core/issues/2863
##### Example:

N/A

### DIFF
--- a/windows/win_stat.ps1
+++ b/windows/win_stat.ps1
@@ -31,6 +31,7 @@ If ($path -eq $FALSE)
 }
 
 $get_md5 = Get-Attr $params "get_md5" $TRUE | ConvertTo-Bool;
+$get_checksum = Get-Attr $params "get_checksum" $TRUE | ConvertTo-Bool;
 
 $result = New-Object psobject @{
     stat = New-Object psobject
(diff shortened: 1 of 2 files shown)


sivel:
cc @jhawkesworth

PR #2844, opened 2016-01-19
Support tagging ELBs using boto 2.0.

welex91:
Patched ec2_elb_lb as per feedback from @jonhadfield and @Constantin07. I can close https://github.com/ansible/ansible-modules-extras/pull/1454 if the module owner agrees with this approach.

Issue #439, opened 2014-12-03
can't tag ELBs


smiller171:
##### Issue Type:

Feature Idea
##### Component Name:

ec2_elb_lb module
##### Ansible Version:

1.8.1
##### Environment:

Managing EC2 from generic.
##### Summary:

Currently there is no way to tag an ELB because it does not have a resource_id, so it can't be tagged by ec2_tag.
##### Steps To Reproduce:

I think that you should either have a "tags" parameter in the ec2_elb_lb module (like instance_tags in the ec2 module) or allow for the use of the resource name in ec2_tag.
##### Expected Results:

I expect to be able to add tags to ELBs by using either a parameter in the ec2_elb_lb module or using the ec2_tag module.
##### Actual Results:

Currently you can't tag ELBs from Ansible.

bobby:
+1, it would be nice to be able to uniformly tag all AWS resources provisioned via Ansible.

nicolas-g:
+1 , tagging has become really important for cost reporting, administration and automation ...

sebest:
+1 here

evitorbarros:
+1

ryanwalls:
+1

k3ypad:
Email sent to ansible devel list - should have something soon:
https://groups.google.com/forum/#!topic/ansible-devel/FJZ8eOqhfOc

eproxus:
What's the current status? Still missing boto support?

smiller171:
@eproxus yup, still missing. I'm tagging ELBs in my playbooks using the shell module to run AWSCLI commands

eproxus:
Ah, do you have a snippet you could share? :smile:

smiller171:
```
- name: tag ELB
  shell: "aws --region us-east-1 elb add-tags --load-balancer-names {{elbname}} --tags Key=stack,Value={{stack}} Key=layer,Value={{layer}} Key=owner,Value={{owner}}"
```

eproxus:
@smiller171 Awesome, thanks!

smiller171:
no problem @eproxus. I think Ansible 2.0 is supposed to switch to Boto3, which does have tagging support for ELBs

s0enke:
> no problem @eproxus. I think Ansible 2.0 is supposed to switch to Boto3, which does have tagging support for ELBs

@smiller171 Do you have any particular information about that? I could not find any hints that Ansible2 is going to be ported to Boto3.

[...]

PR #3163, opened 2016-03-01
add tagging functionality to elb module

nand0p:
##### Issue Type:
- Feature Pull Request
##### Plugin Name:

ec2_elb_lb
##### Summary:

Adds tagging to ec2_elb_lb module

this module will add/modify/delete tags by default. override this with `append_tags=True`


nand0p:
this PR is a result of conversation from
https://github.com/ansible/ansible-modules-extras/pull/1738

and originally reported in 
https://github.com/ansible/ansible-modules-core/issues/439

nand0p:
this PR has been moved to core (previously elb_tags.py in extras). @wimnat @tombamford @defionscode @smiller171

smiller171:
:+1:

juliedavila:
This should be using the boto3 connection mechanisms in module_utils/ec2.py (takes care of support for tokens, etc)

gregdek:
Thanks @nand0p. @jsdalton @rickmendes please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

cloud/amazon/ec2_elb_lb.py
```diff
@@ -25,6 +25,7 @@
 author:
   - "Jim Dalton (@jsdalton)"
   - "Rick Mendes (@rickmendes)"
+  - "Fernando Jose Pando (@oldmanlinux)"
```
wimnat: I would only do this if you have the blessing of the current authors and you want to take on responsibility for module maintenance

cloud/amazon/ec2_elb_lb.py
```diff
@@ -1000,7 +1033,8 @@ def main():
             stickiness={'default': None, 'required': False, 'type': 'dict'},
             access_logs={'default': None, 'required': False, 'type': 'dict'},
             wait={'default': False, 'type': 'bool', 'required': False},
-            wait_timeout={'default': 60, 'type': 'int', 'required': False}
+            wait_timeout={'default': 60, 'type': 'int', 'required': False},
+            tags={'default': None, 'required': False, 'type': 'list'}
```
wimnat: Default to empty list []

cloud/amazon/ec2_elb_lb.py
```diff
@@ -344,6 +351,13 @@
 except ImportError:
     HAS_BOTO = False
 
+try:
+    import boto3
+    import botocore
+    HAS_BOTO3 = True
+except ImportError:
+    HAS_BOTO3 = False
+
```
wimnat, line 415: The import statement should be nested with the rest of the tag code.  We don't want to break existing playbooks where people haven't got boto3 and they don't need the tag functionality.

manicminer:
Just FYI, in terms of consistency, the `ec2_asg` module replaces any tags currently set. I work with this in the `ec2_asg` module by retrieving the tags first and then adding my own (I have a co-authored [PR currently open](https://github.com/ansible/ansible-modules-extras/pull/1710) for an `ec2_asg_facts` module). I don't have a preference for either approach though.

cloud/amazon/ec2_elb_lb.py
nand0p, line 415: the module will not break when people are not declaring "tags" and do not have boto3.  if "tags" option is declared, and boto3 is not available, mod fails as fast as possible.

nand0p:
@defionscode updated mod to include ansible v2 boto_conn support (but commented out, as it would break for people using ansible v1).   is there a way to determine version of ansible running from inside the module, and make this conditional?

nand0p:
@jsdalton @rickmendes added myself to authors in this PR, as i feel i have a good understanding of this module after working closely thru it, and would be happy to help maintain.  please let me know if this is not appropriate and i should remove.

juliedavila:
@nand0p you're `version_added` bit for tags should suffice. There shouldn't be any expectation of support for older Ansible versions when it comes to brand new features (obviously for bug fixes there is).

manicminer:
@nand0p RE using the newer boto connection support: v1 users would not have this revision of the module shipped with their ansible? (docs by contrast are generated from the latest module)

ghost:
@nand0p Happy to have you as an additional author.

@gregdek I am happy to say shipit

gregdek:
Thanks again to @nand0p for this PR, and thanks @jsdalton @rickmendes for reviewing. Marking for inclusion.

mmochan:
@nand0p - Nice work on the new ELB module

I think I may have found 1 problem with the  _elb_tags function. It looks like you are missing the call to self.module on line 548

So this 
`client = boto3_conn(module, conn_type='client', resource='elb', region=region, endpoint=ec2_url, **aws_connect_kwargs)`

should become

`client = boto3_conn(self.module, conn_type='client', resource='elb', region=region, endpoint=ec2_url, **aws_connect_kwargs)`

Could I also make one suggestion? It would be nice if when adding tags you didn't need to explicitly add the "key" and "Value" for each tag.. 

This follows along with other ansible tagging modules e.g

```
    `tags:
      Environment: SIT
      Name: "Nginx-ELB-SIT"
`
```

My bad attempt at what this might look like in your code

```
   `formatted_tags = list()
    for tags in self.tags:
        for tag, value in tags.iteritems():
            formatted_tags.append({'Key': tag, 'Value': value})

    if not set(tags_new).issubset(set(tags_existing)):
        try:
            client.add_tags(LoadBalancerNames = [ self.name ], Tags = formatted_tags)`
```

I think you've done a fantastic job with this and it's great to see more Boto3 modules appearing. Thank you

wimnat:
@nand0p - I agree with @mmochan that you should specify tags in the same way as other modules:

```
tags:
  name: example
  env: prod
```

needs_revision

nand0p:
@mmochan @wimnat updated as per comments

mmochan:
@nand0p - I've tested those changes and can confirm they are working as expected. Thanks

One more request - It doesn't cater for the removal of existing tags if they have already been set.

nand0p:
@mmochan yes, only allowing for add/update is on purpose, as i would not want to delete tags, such as billing or janitor monkey, that may have been added manually, and their removal could cause downtime (ie. JM).  i was thinking a better way may be to add a new option for tag deletion? im not clear on what the best way to proceed with tag deletion is,..... if everyone feels that unspecified tags should be deleted, i can certainly add that (although i wouldnt want to run that code myself)

smiller171:
@nand0p There should probably be a flag (like purge_rules in the [ec2_group module](http://docs.ansible.com/ansible/ec2_group_module.html)) that determines whether it deletes missing tags or not.

smiller171:
The same flag should probably be added to the ec2_tags module

mmochan:
@nand0p - IMHO configuration management is about setting the desired state, and infrastructure as code in the form of ansible code is and should be the source of the desired state.  If I remove or add tags from my code I would expect the Tag entries to reflect this desired state. 

I would like to see this included in this module. 

Again thanks for this module.

smiller171:
@mmochan I think @nand0p is correct though in that there are many external tools you may decide to use at any given time that may need to add their own tags, and you don't want to have to rebuild their logic in your own configuration management to account for that. A true/false `purge_tags` parameter seems to make the most sense to make this behavior configurable. This is what was done in the ec2_group module and I think it was the correct approach.

mmochan:
@smiller171 Agreed in the ec2_group module there is a parameter that allows you to specify purge_rules, however I fail to see the relevance between Security Group/ Firewall  rules and resource tagging.

IMHO If you are making manual changes outside of your 'Infrastructure as Code single source of truth", where do you draw the line about what can be added? You might start with a couple of manually added tags but how many exceptions to the rule are you prepared to make? If you follow on down that track you are in for a whole world of pain.

If you have external tools that make changes to your Infra then you need to figure out a way to include those changes into your infra code base.  You need to be able to rebuild everything from your Infrastructure codebase otherwise you are doing it wrong :-)

juliedavila:
I side with the option of imperative statements, and side with @mmochan on this.

juliedavila:
That being said, I'm okay with a boolean flag that dictates whether to append tags or replace all

smiller171:
@mmochan Not everyone uses Ansible as their single source of truth, and it doesn't make sense for all use cases. Defaulting to removing tags that are not included, with a parameter to disable that functionality allows for all use-cases.

mmochan:
@smiller171 @defionscode I'm happy to give a thumbs up to  that boolean flag.

juliedavila:
Let's go with that. @nand0p are you good with that? make it like `explicit_tags` or smth like that.

nand0p:
@defionscode @mmochan @smiller171 ok great, i will go with boolean "purge_tags" option, that deletes all tags.  if used in combination with "tags" option, it will first delete all tags, and then add new ones specified.  also not sure how to deal with the travis CI failure that tests using python 2.6.  dictionary comprehension is only available in python >2.7..... EDIT: i can rewrite the dict comprehension to support 2.6

mmochan:
@nand0p  @smiller171 @defionscode - Is there a requirement for a new ec2_elb_facts boto3 module to support the new tagging functionality in this module?

juliedavila:
@mmochan not sure what you mean by that

juliedavila:
@nand0p you want to do the dict comprehension kinda like

`d = dict((key, value) for (key, value) in iterable)`

and then it will be 2.6 compatible

nand0p:
@defionscode @mmochan @smiller171 purge_tags option implemented and dictionary comprehension now backwards compatible with pre-python2.7

cloud/amazon/ec2_elb_lb.py
```diff
@@ -142,6 +143,21 @@
     required: false
     default: 60
     version_added: "2.1"
+  tags:
+    description: the tags dict to apply
+    type: dict
+    required: false
+    sample: { "Department": "Information Retrieval" }
+    version_added: "2.1"
+  purge_tags:
```
bcoca: append_tags might be a better name as it will keep consistent with other implementations of similar options (i.e. append in user module)

nand0p: "purge_tags" is able to be set True independent of "tags" being defined, so as to delete all existing tags from an ELB.  For this reason, "append_tags" may not be an accurate name?  It follows the pattern in ec2_group [purge_rules](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/amazon/ec2_group.py#L56) .... I can certainly rename "purge_tags" to "append_tags", and make it require "tags" to be defined, but im not sure this is the desired functionality as it would lose the tag purging ability.  @defionscode @mmochan @smiller171 @wimnat @tombamford @jsdalton @rickmendes thoughts?

bcoca: empty tag list with append being false would function as a purge

nand0p: ah yes. that makes perfect sense.  i will go ahead with the append_tags solution. thanks @bcoca

gregdek:
Thanks @nand0p for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

nand0p:
Sorry for the delay guys.  i am on vacation this week, but I will rebase this PR and update purge_tags -> append_tags on Monday. Cheers!

nand0p:
@gregdek @bcoca rebased and added append_tags option 

ready_for_review

mmochan:
@nand0p With these changes I now need to add both `tags:` and `append_tags: true`   to the playbook. If I don't then I will always receive `changed` when I run the playbook.  

It feels kind of messy to me.

nand0p:
@mmochan it appears this module always returns changed=true, regardless of this commit.  i will open a ticket, and add self.changer='method' to help debug and pinpoint where this is occurring.

Regarding the append_tags option, it is optional.  When it is not specified, it defaults to False, so as to ensure the ansible code is the source of truth and elb tags are added/updated/deleted accordingly.

gregdek:
Thanks @nand0p. To the current maintainers, @jsdalton, @rickmendes please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

mmochan:
@nand0p I am receiving the following error when I first create a new ELB. The module crashes although still creates the ELB. The second run of the playbook works fine and the ELB is marked as `changed`. However all  subsequent runs of the playbook are also marked as `changed`

If you need more detail let me know

```
  File "/Users/mmochan/.ansible/tmp/ansible-tmp-1459331170.92-130676148105316/ec2_elb_lb", line 3532, in <module>
    main()
  File "/Users/mmochan/.ansible/tmp/ansible-tmp-1459331170.92-130676148105316/ec2_elb_lb", line 1333, in main
    elb_man.ensure_ok()
  File "/Users/mmochan/.ansible/tmp/ansible-tmp-1459331170.92-130676148105316/ec2_elb_lb", line 474, in ensure_ok
    self._set_backend_policies()
  File "/Users/mmochan/.ansible/tmp/ansible-tmp-1459331170.92-130676148105316/ec2_elb_lb", line 1136, in _set_backend_policies
    backend_policies = self._get_backend_policies()
  File "/Users/mmochan/.ansible/tmp/ansible-tmp-1459331170.92-130676148105316/ec2_elb_lb", line 1128, in _get_backend_policies
    str(backend.instance_port) + ':' + policy.policy_name for backend in self.elb.backends
TypeError: 'NoneType' object is not iterable

failed: [localhost] => (item={u'connection_draining_timeout': 3600, u'health_check': {u'response_timeout': 5, u'ping_port': 22, u'ping_protocol': u'tcp', u'interval': 30, u'unhealthy_threshold': 2, u'healthy_threshold': 10}, u'name': u'xxx-vpc-21-nginx-blue-dev', u'tags': {u'Environment': u'dev', u'Color': u'blue', u'Role': u'nginx', u'Name': u'xxx-vpc-21-nginx-blue'}, u'listeners': [{u'instance_port': 80, u'load_balancer_port': 80, u'protocol': u'http'}], u'scheme': u'internal', u'stickiness': {u'type': u'application', u'enabled': False}}) => {"failed": true, "invocation": {"module_name": "ec2_elb_lb"}, "item": {"connection_draining_timeout": 3600, "health_check": {"healthy_threshold": 10, "interval": 30, "ping_port": 22, "ping_protocol": "tcp", "response_timeout": 5, "unhealthy_threshold": 2}, "listeners": [{"instance_port": 80, "load_balancer_port": 80, "protocol": "http"}], "name": "xxx-vpc-21-nginx-blue-dev", "scheme": "internal", "stickiness": {"enabled": false, "type": "application"}, "tags": {"Color": "blue", "Environment": "dev", "Name": "xxx-vpc-21-nginx-blue", "Role": "nginx"}}, "parsed": false}
```

nand0p:
@mmochan looks like the error you reported is due to the latest commit to this module a few days ago, and i have implemented a fix here.  it also appears, via the new "changer" variable returned, _set_health_check is returning changed=true when the specified health check already exists.  try running the module twice without a specified health check, so you can see tag manipulation is not responsible for changed=true.  I will submit a separate PR for the healthcheck fixup.

gregdek:
Thanks @nand0p for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

nand0p:
ready_for_review

mmochan:
Thanks @nand0p I'll test this today

mmochan:
@nand0p A few quick tests reveal that the `connection_draining_timeout` parameter is always returning `changed = True`

I can confirm that the latest commit does indeed fix the `elb.backends` issue.  Thanks for fixing that.

nand0p:
@mmochan i think the health_check and connection_draining always returning `changed = True` may be out-of-scope for this PR? I was planning on fixing that next.

mmochan:
@nand0p I'm not seeing the `health_check` returning `changed = True` except when I expect it to.

For `connection_draining_timeout` I can live with that for the time being.

I've been testing this all morning with multiple ELB configuration types and it's working really well.

Nice work and thank you.

gregdek:
Thanks @nand0p. To the current maintainers, @jsdalton, @rickmendes please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

nand0p:
nand0p> @bcoca regarding your comment about "new stuff should be boto3" i have a q about elb tagging.... i added this functionality via boto3 https://github.com/ansible/ansible-modules-core/pull/3163, but i just realized there is another PR that does the same with boto2, and has just been labeled "shipit" the other day https://github.com/ansible/ansible-modules-core/pull/2844 ... should i go ahead and close my boto3 PR?

bcoca> nand0p: in principle, yes close in favor of the other, but i would try to make it work with both boto/boto3 if possible (might be a followup PR)

i will wait for 2844 to be merged, and then rebase this PR and update it to use boto3 if it is available

bermudezjd:
@nand0p I think you want to be comparing keys instead of key/value sets when determining what to delete and add.  The docs on [CreateTags](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTags.html) - the method backing the `add_tags` boto batch method - say that it adds or overwrites existing tags, so I think you could just pass the contents of the `tags` dict along and get both effects.  Similarly, you could do a key comparison to determine the tags that exist on the ELB, but are not present in the `tags` dict passed to the module, and pass the difference along as what to delete.  I'd argue that `append_tags` isn't the best name for the latter functionality, since any new tags are appended to the ELB regardless of whether that flag is enabled or not.  I'd love to see a different name for that flag, perhaps `purge_tags` or something similar.

gregdek:
Thanks @nand0p for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

cloud/amazon/ec2_elb_lb.py
```diff
@@ -995,10 +1110,13 @@ def select_stickiness_policy(self):
 
     def _get_backend_policies(self):
         """Get a list of backend policies"""
-        return [
-            str(backend.instance_port) + ':' + policy.policy_name for backend in self.elb.backends
-            for policy in backend.policies
```
nand0p, line 1083: fixes required backend policies

nand0p:
ready_for_review

gregdek:
Thanks @nand0p. To the current maintainers, @jsdalton, @rickmendes please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

wimnat:
Can you squash to a single commit?

ghost:
@nand0p Thanks for hanging in there on this one and updating based on everyone's wishes. I can't see a reason to hold this back.

shipit

gregdek:
Thanks again to @nand0p for this PR, and thanks @jsdalton, @rickmendes for reviewing. Marking for inclusion.

gregdek:
Thanks @nand0p for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

nand0p:
merge conflict due to https://github.com/ansible/ansible-modules-core/commit/97f6daa85442a96d48c0c72e3438819e414e98b4 which duplicates functionality here in this PR, so im closing it.

smiller171:
@nand0p Looking at that PR, I'm not sure how it's a dupe. Can you explain how that PR adds tagging functionality to ELBs?

nand0p:
@smiller171 looks like tagging was added via boto2, and my other fixes have also been taken care of by other PRs.  i rebased anyways, but force push when squashing commits made this PR un-reopenable. 

lmk if there is still interest in this, and i can open PR from my rebased changes https://github.com/ansible/ansible-modules-core/compare/devel...nand0p:ec2-elb-lb-add-tagging?expand=1

smiller171:
@nand0p Unless we can confirm that this functionality has been added elsewhere, I'm still interested.

@defionscode you have thoughts on this?

PR #3166, opened 2016-03-01
Update pause.py

slimandslam:
##### Issue Type:
- Docs Pull Request
##### Plugin Name:

pause
##### Summary:

Pause MUST have standard input open regardless of whether you just want to wait some amount of time or if you want to prompt. If you want to run your playbook in the background using Linux "nohup", you will  need to do something like this to ensure that your pause statement does not fail.
##### Example:

$ nohup ansible-playbook myscript.yml 0</dev/null

It's fairly common to want to run ansible scripts using nohup. This info is necessary to ensure that pause doesn't bomb because of lack of STDIN. Please format my addition as you wish.

### DIFF
--- a/utilities/logic/pause.py
+++ b/utilities/logic/pause.py
@@ -52,4 +52,10 @@ EXAMPLES = '''
 
 # A helpful reminder of what to look out for post-update.
 - pause: prompt="Make sure org.foo.FooOverload exception is not present"
+
+# Pause MUST have standard input open regardless of whether you just want to wait some amount of time
+# or if you want to prompt. If you want to run your playbook in the background using Linux "nohup", you will
+# need to do something like this to ensure that your pause statement, which is inside
+# of myscript.yml, does not fail (this is a Linux command prompt example)
+- $ nohup ansible-playbook myscript.yml 0</dev/null
 '''


gregdek:
Thanks @slimandslam. @tbielawa please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

bcoca:
so pause has code to explicitly prevent prompting when stdin is not a tty, if this is affecting the 'pure wait' part, its a bug.

in any case pause does this for the entire host batch, not per host, for such a timeout use `wait_for` instead.

slimandslam:
So, instead, the pause docs should say something like: "Pause will always expect standard input to be available, so if you need to run your script in the background (not attached to a tty), use wait_for instead"

timlnx:
I can't tell exactly what's going on in the pause module now, it's changed in quite a few ways since I first wrote it so this may be incorrect:

Must the pause module **require** `stdin` regardless of the action type? I am thinking of a use case
- running a `pause` task for a pre-set amount of time in a `nohupped` scenario

I can't think of any reason to require any kind of input be available for this. The playbook runner should have prescient knowledge that if their playbook is running in the background then they have no way to `ctrl-c c` or `ctrl-c a` the pause task.

Or is that exactly what the change in https://github.com/ansible/ansible/pull/14755/files#diff-3fbd8ab6a6e5c90497a2d84142b3a767R145 (line >> 145) tries to implement?

bcoca:
@tbielawa a previous update detected if stdin is a tty or not and skipped the prompt processing, this worked well with redirected inputs or when running from cron.

the current problem is that nohup does not seem to offer an 'open stdin' causing the error reported above. The code you point at is my currently proposed fix.

timlnx:
Oh, sweet. Right on! Thanks for clearing that up for me. At a glance that
change seems to be doing the right thing. I give it a +1.

On Wed, Mar 2, 2016 at 1:38 PM, Brian Coca notifications@github.com wrote:

> @tbielawa https://github.com/tbielawa a previous update detected if
> stdin is a tty or not and skipped the prompt processing, this worked well
> with redirected inputs or when running from cron.
> 
> the current problem is that nohup does not seem to offer an 'open stdin'
> causing the error reported above. The code you point at is my currently
> proposed fix.
> 
> —
> https://github.com/ansible/ansible-modules-core/pull/3166#issuecomment-191364880
> .

## 

--Tim Bielawa
https://blog.lnx.cx

PR #3179, opened 2016-03-03
ec2_ami_find.py - Treat 'is_public' option as a bool

Constantin07:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

ansible/modules/core/cloud/amazon/ec2_ami_find.py
##### Summary:

The module doesn't return any AMIs if 'is_public' = no.
It seems the problem is that it's treated as a string.
##### Example:

```
TASK [backup : Get all AMIs created from instance] *****************************
task path: /home/******:109
ok: [localhost] => {"changed": false, "results": []}
```

### DIFF
--- a/cloud/amazon/ec2_ami_find.py
+++ b/cloud/amazon/ec2_ami_find.py
@@ -299,7 +299,7 @@ def main():
                 aliases = ['search_tags', 'image_tags']),
             architecture = dict(required=False),
             hypervisor = dict(required=False),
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @Constantin07. @tombamford please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

PR #3195, opened 2016-03-07
Unarchrefactor (2nd Try)

pileofrogs:
##### Issue Type:
- Feature Pull Request
##### Plugin Name:

unarchive
##### Summary:

This is the module side of a refactor to allow untar and unzip (and any future unarchivers) to have their own modules, thus allowing their own arguments. There is a corresponding PR for ansible. Don't accept one without the other!

This PR only refactors and does not add new arguments.

After doing it, I realized a lot of ways to make it better. I added a _copy method to the ActionBase class, thinking this would be good because other future modules could use it. However, it might be better to make an unarchive subclass of ActionBase and stick stuff in their. Also, I wound up duplicating a fair amount of code between the unarchive module and the untar and unzip modules. I would like to separate out some of the logic. Possibly into ansible.module_utils somewhere... sigh...

So, basically, if you aren't picky, take this PR. It should allow folks who want to submit PRs for new tar/zip features to be able to do so. If you're even remotely picky, fire it back with comments and I'll unscrew whatever parts you don't like.

There is a nearly identical sounding PR for ansible. One doesn't work without the other.


gregdek:
Thanks @pileofrogs for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

pileofrogs:
Lack of interest

PR #3202, opened 2016-03-08
fixes to assemble

bcoca:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

files/assemble
##### Summary:

It now uses atomic_move to avoid data corruption,
correclty cleans up temp files in every case and 
returns backup_file info if needed
validate validation happens upfront and avoids temp file creation
backup AFTER validating ... or you are just creating useless backups of same file

### DIFF
--- a/files/assemble.py
+++ b/files/assemble.py
@@ -20,7 +20,6 @@
 
 import os
 import os.path
-import shutil
 import tempfile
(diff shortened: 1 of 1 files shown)


abadger:
Looks good +1

PR #3214, opened 2016-03-09
add exponential backoff on wait

nand0p:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

ec2.py
##### Summary:

Adds exponential backoff on waiting for instance creation and termination.  

Instead of hammering AWS API _every second_ which commonly results in API rate throttling in high volume environments, this PR incorporates AWS best practices and mitigates issue with exponential backoff.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -258,6 +258,7 @@ author:
     - "Tim Gerla (@tgerla)"
     - "Lester Wade (@lwade)"
     - "Seth Vidal"
+    - "Fernando Jose Pando (@nand0p)"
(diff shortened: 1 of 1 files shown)


nand0p:
@tgerla @lwade added myself to authors in this PR, as i feel i have a good understanding of this module after working closely thru it, and would be happy to help maintain.

gregdek:
Thanks @nand0p for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

nand0p:
boto3 waiters may be best way to go. closing.

Issue #3217, opened 2016-03-09
ec2_vol: Add support for custom KMS keys to ec2_vol


nonshankus:
##### Issue Type:
- Feature Idea
##### Plugin Name:

cloud/amazon/ec2_vol.py
##### Ansible Version:

```
ansible 2.0.1.0
```
##### Ansible Configuration:

```
[defaults]
retry_files_enabled=False
host_key_checking=False
pipelining=True
```
##### Environment:

```
CentOS 7.2
python2-boto-2.39.0-1.el7.noarch.rpm (from EPEL testing)
```
##### Summary:

Currently the module `ec2_vol` (right along with most of the other modules concerned with encryption on AWS) doesn't support specifying a custom encryption key. One is only able to specify whether the volume is encrypted or not and thus the default encryption key (in this case here for EBS) is used, when encryption is enabled.

So this request is about adding sort of a quick fix to the `ec2_vol` module by adding a new parameter, but I guess for the future a more global review of the topic KMS might make sense.

nonshankus:
https://github.com/ansible/ansible-modules-core/pull/3064 also addresses this issue here.

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
This repository has been locked. All new issues and pull requests should be filed in https://github.com/ansible/ansible

Please read through the <a href="http://docs.ansible.com/ansible/dev_guide/repomerge.html#move-issues-and-prs-to-new-repo">repomerge</a> page in the dev guide. The guide contains links to tools which automatically move your issue or pull request to the ansible/ansible repo.

ansibot:
This issue was migrated to https://github.com/ansible/ansible/issues/29856

PR #3064, opened 2016-02-18
Added kms keys and tags to module.

kalpatel01:
Added 2 features to ec2_vol module:
- kms key support
- ability to add tags (Module exist to add and remove tags, adding here makes dynamic tagging easier)

Fixed bug where AnsibleAWSError doesn't exist.

PR #3218, opened 2016-03-09
Fixes: #3217 - Add a new parameter for the kms_key_id to the ec2_vol module

nonshankus:
##### Issue Type:
- Feature Pull Request
##### Plugin Name:

cloud/amazon/ec2_vol.py
##### Summary:

This pull request fixes/implements #3217. For more details please refer to the according issue.

It is important to point out, that this fix will require at **least boto v2.39.0**.
Here is the according boto change:
https://github.com/boto/boto/commit/2d7796a625f9596cbadb7d00c0198e5ed84631ed
And the according pull request for boto:
https://github.com/boto/boto/pull/3064
##### Example:

```
    - name: Created encrypted volume with custom key
      ec2_vol:
        device_name: /dev/sdb
        encrypted: yes
        kms_key_id: '{{ kms_key.KeyMetadata.Arn }}'
        instance: '{{ ec2.instance_ids }}'
        region: '{{global_region}}'
        volume_size: 20
        volume_type: gp2
        delete_on_termination: true
      register: ec2_vol
```


gregdek:
Thanks @nonshankus for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

nonshankus:
Thanks @gregdek for the heads-up. I sort of expected that. :)

cloud/amazon/ec2_vol.py
```diff
@@ -62,6 +62,11 @@
       - Enable encryption at rest for this volume.
     default: false
     version_added: "1.8"
+  kms_key_id:
+    description:
+      - Specify the id of the KMS key to use.
+    default: null
```
juliedavila, line 68: `default: null` can be omitted

cloud/amazon/ec2_vol.py
```diff
@@ -332,7 +338,10 @@ def create_volume(module, ec2, zone):
     if volume is None:
         try:
             if boto_supports_volume_encryption():
-                volume = ec2.create_volume(volume_size, zone, snapshot, volume_type, iops, encrypted)
+                if kms_key_id:
```
juliedavila, line 350: Can you add a check for boto version here? Something like 
Add a new import
`from distutils.version import StrictVersion`

``` python
if StrictVersion(boto.__version__) < StrictVersion('2.39.0'):
    module.fail_json(msg="A newer version of boto is required, %s detected but 2.39.0 is required for kms_key_id" % boto.__version__)
```

cloud/amazon/ec2_vol.py
nonshankus, line 68: Thanks @defionscode for the review. Committed an according change to remove that particular line.

cloud/amazon/ec2_vol.py
nonshankus, line 350: Thanks @defionscode for the feedback. That is indeed a good point. But is that consistent behaviour? On the one hand let the module fail, when a `kms_key_id` is given, and on the other gracefully ignore the `encryption` attribute, when the boto version is too old for it? Wouldn't it be more consistent to also ignore it?

juliedavila, line 350: For most things I'd prob be ok with what you're saying but the last thing I would want is for a user to think that a kms key was being used for encryption, as a result of a non-error during executing, when in fact it isnt.

kalpatel01:
https://github.com/ansible/ansible-modules-core/pull/3064 Have something already for this.  It does have one extra check.

juliedavila:
@kalpatel01 ah! thanks for pointing that out. I'd like for your PR to take priority as yours was first. I'll review it soon

cloud/amazon/ec2_vol.py
nonshankus: Done with 71cbec0f2722ff11449c604e71e18293e7fd1e74.

nonshankus, line 68: Accidentally removed the wrong line. What confused me was, that other parameters also have this line (`default: null`).

cloud/amazon/ec2_vol.py
nonshankus, line 350: Point taken. Added the according check, actually merged the check from https://github.com/ansible/ansible-modules-core/pull/3064 and corrected the naming of the method.

nonshankus:
@defionscode: As previously mentioned, merged the check from https://github.com/ansible/ansible-modules-core/pull/3064 while correcting the method naming. So this pull request here only implements the feature requested in the feature request #3217, whereas the pull request by @kalpatel01 also implements adding tags to volumes.

As the pull request of @kalpatel01 was way before mine I am perfectly fine with accepting it instead of mine.

Cheers

Issue #3232, opened 2016-03-12
Accelerate mode fails to initialize


ghost:
##### Issue Type:
- Bug Report
##### Plugin Name:

accelerate
##### Ansible Version:

```
$ ansible --version
ansible 2.0.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### Ansible Configuration:

N/A
##### Environment:

Debian Jessie
##### Summary:

Accelerate mode fails to setup remote daemon.

`from datetime import datetime` in the module gets overriden by `import datetime` when `lib/ansible/module_utils/basic.py` is pulled in, resulting in `AttributeError: 'module' object has no attribute 'now'`
##### Steps To Reproduce:

```
virtualenv .2
. ./.2/bin/activate
pip install ansible python-keyczar
```

test playbook

```

---

- hosts: localhost
  accelerate: true
  tasks:
  - command: "true"
```
##### Expected Results:

Accelerate daemon to be started and the play to continue in accelerated mode.
##### Actual Results:

output

```
attempting connection to localhost via the accelerated port 5099
connection to localhost failed, retrying...
connection to localhost failed, retrying...
connection to localhost failed, retrying...
Could not connect via the accelerated connection, exceeded # of tries
<localhost> ESTABLISH SSH CONNECTION FOR USER: None
<localhost> SSH: EXEC ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o ConnectTimeout=10 -o ControlPath=/home/geoff/.ansible/cp/ansible-ssh-%h-%p-%r -tt localhost '/bin/sh -c '"'"'( umask 22 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1457775753.88-47181511787438 `" && echo "` echo $HOME/.ansible/tmp/ansible-tmp-1457775753.88-47181511787438 `" )'"'"''
<localhost> PUT /tmp/tmp55AOg8 TO /home/geoff/.ansible/tmp/ansible-tmp-1457775753.88-47181511787438/accelerate
[...]

PR #3233, opened 2016-03-12
Fixes: #3232. Use of datetime consistent with module_utils/basic.py

ghost:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

accelerate
##### Summary:

Make use of `datetime` consistent with `lib/ansible/module_utils/basic.py` to prevent the import of the latter stopping the module from initializing the accelerate daemon

### DIFF
--- a/utilities/helper/accelerate.py
+++ b/utilities/helper/accelerate.py
@@ -98,7 +98,7 @@ import traceback
 
 import SocketServer
 
-from datetime import datetime
+import datetime
 from threading import Thread, Lock
 
 # import module snippets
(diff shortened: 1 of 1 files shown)


bcoca:
I'm not sure I understand what the problem is here, can you elaborate on what the error is?

ghost:
Sorry, I should have added a gist of the generated module script. https://gist.github.com/geoffwright/b34dd361b0a1a7dd71b6

The accelerate module imports datetime with `from datetime import datetime`, shortly after `module_utils/basic.py` gets pulled in and imports datetime with `import datetime`. After `module_utils/basic.py` boilerplate is done we go back to the accelerate module and it attempts `datetime.now()` and fails.

mattclay:
Tested on Ubuntu 15.10, fixes the issue for me.

abadger:
Merged to both devel and stable-2.0.   Thanks @geoffwright for the fix and @mattclay for the review

abadger:
Note -- once we have ziploader in 2.1.0 we should be able to remove `the import *` portion of the import from basic and then we won't have to worry about this.

PR #3239, opened 2016-03-14
Add diff and check mode to cron module

towolf:
##### Issue Type:
- Feature Pull Request
##### Plugin Name:

cron
##### Summary:

The cron module was missing both check and diff modes. 

Diff mode is a major new feature on the 2.1 Roadmap.
##### Example:

example screenshot

### DIFF
--- a/system/cron.py
+++ b/system/cron.py
@@ -538,7 +538,7 @@ def main():
             insertafter=dict(required=False),
             insertbefore=dict(required=False),
         ),
-        supports_check_mode = False,
+        supports_check_mode = True,
         mutually_exclusive=[
                 ['reboot', 'special_time'],
                 ['insertafter', 'insertbefore'],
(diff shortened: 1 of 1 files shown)


towolf:
Notice that there are two logic errors in the cron module.

First, the default user is stated to be `root`. This is not reflected in the code. The default user is whoever the task is run as. Not setting `user` leaves it as `None`. (Only left this in the screenshot for illustration. I removed the spurious output in the final PR.) (docs issue)

The second is that a cron.d file is only unlinked when `name` is omitted. Removing the last named item in a cron.d snippet with leave behind an empty file. (docs issue)

resmo:
Thanks @towolf. To the current maintainers, @dsummersl, @EvanK please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

system/cron.py
```diff
@@ -593,14 +604,22 @@ def main():
         special_time = "reboot"
 
     # if requested make a backup before making a change
-    if backup:
+    if backup and not module.check_mode:
```
dsummersl, line 607: nice catch...

EvanK:
shipit

resmo:
Thanks again to @towolf for this PR, and thanks @dsummersl, @EvanK for reviewing. Marking for inclusion.

Issue #527, opened 2014-12-16
git: Doesn't pick up new tags for already installed revision


pfalcon:
If git module sees that revision to install matches that of already installed, it stop there. However, it may be the case that this revision was tagged since the previous install, and with the current behavior, git module won't fetch this new tag. This may in turn affect software which uses "git describe" command to show version information - instead the latest of tag it shows something else (raw revision or version based on previous tag).

pfalcon:
```
$ ansible-playbook --version
ansible-playbook 1.7.1
```

abadger:
If you can give examples of ansible playbooks or command lines that misbehave and show what hte desired outcome vs the current outcome is, that would help us look into this.  Thanks!

cjunge-work:
I've just run into this issue but for branches. Below is the steps on how it happened for us.

Ansible 1.8.4

1) Create playbook with task that checks out master branch:

```
-   name: Clone/update repo
    git:
        dest: "{{ cache_dir }}"
        repo: "git@github.com:{{ repo }}"
        version: "master"
        update: yes
        force: yes
        recursive: yes
```

2) Add new branch to repo (eg. production), do not add new commit. Push branch to origin.

3) Run playbook again with new branch name

```
-   name: Clone/update repo
    git:
        dest: "{{ cache_dir }}"
        repo: "git@github.com:{{ repo }}"
        version: "production"
        update: yes
        force: yes
        recursive: yes
```

error: `msg: Failed to checkout production`

Doing a manual git fetch origin fixes it. I assume this would apply to tags as well.

fredsted:
I'm running ansible 1.9.4 and new tags aren't recognised. Getting `msg: Failed to checkout tagname`.

Am I specifying the tag wrong?

robinro:
I can't reproduce this with either 2.0 or 1.9.4.

I wrote a testcase that is the case described above in https://github.com/robinro/ansible/blob/expand_test_git/test/integration/roles/test_git/tasks/main.yml#L513-L555
It does
- checkout a repo
- make sure the _newtag_ does not exist
- add the tag to the remote repo
- update the checkout with `version=newtag`
- make sure the new repo has _newtag_

@fredsted @cjunge-work 
Please have a look at the testcase and give a similar complete set of tasks that fails for you if possible.

ansibot:
@ansible, ping. This issue is still waiting on your response.
[...]

PR #3254, opened 2016-03-15
make git updates respect depth

### LINKED ISSUE #14954: git update with depth limit creates too big local repo
##### Issue Type:
- Feature Idea
##### Ansible Version:

```
ansible 1.9.4
```

and 

```
ansible 2.0.1.0
  config file = /home/XXX/ansible.cfg
```
##### Ansible Configuration:

```
[defaults]
nocows = 1
```
##### Environment:

Fedora 23, git version 2.5.0
##### Summary:

When running playbook with git statement that has update=yes and depth=1, behaviour has changed with version 2.0. 

Now updated repository will become much bigger.
##### Steps To Reproduce:

Run this playbook with ansible 1.9 and 2.0 twice: 

```

---
- hosts: 127.0.0.1
  connection: local
  become: no
  tasks:
  - name: check out pma from git
    git: dest=./pma depth=1 update=yes repo=https://github.com/phpmyadmin/phpmyadmin.git
```
##### Expected Results:

With Ansible 1.9, in both runs the resulting size of pma directory is 76M. Would expect that in 2.0 result would be same, so updated repo does not grow without any updates.
##### Actual Results:

WIth Ansible 2.0, first run result ./pma dir 76M size but second run 
(updating cloned git repo) is resulting 603M directory.

robinro:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

source_control/git
##### Summary:

make git updates respect argument `depth`

until now `fetch` gets all tags and heads at full depth, this change
- uses `depth` argument for `fetch`
- only get the specified `version` in `fetch`
- Fixes ansible/ansible#14954


robinro:
I'm not sure something breaks by removing the fetch of `tags/*`. I don't think it should because the checked out version is specified, but maybe someone assumes all tags are fetched.

resmo:

robinro:
This might affect #527, I'll check this and add tests.

abadger:
At today's Ansible Core meeting we reviewed the code and thought it looked good to merge.  We want integration tests so that we don't accidentally break it with other git fixes and we'll have to keep an eye out for any regressions with tag handling.  robinro is going to work on this and then we can merge.

https://meetbot.fedoraproject.org/ansible-meeting/2016-04-12/ansible_core_meeting.2016-04-12-15.08.log.html#l-64

robinro:
@abadger integration tests are available at ansible/ansible#15391
I also added a change to be more conservative and always do a full clone when `depth` is not specified. This is what more what users might assume.

#527 is independent as far as I see.

abadger:
Decided at today's meeting to merge @robinro will keep an eye out for bugs that could be related to this.

PR #3259, opened 2016-03-16
Feature/iam cert not found

### LINKED ISSUE #2547: iam_cert module occasionally returns BotoServerError 404 NotFound

joelthompson:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

iam_cert
##### Summary:

Fixes #2547. There's an existing PR in #2549 but it is buggy, so this is based off #2549 and fixes it (and also rebases based off current devel).

### DIFF
--- a/cloud/amazon/iam_cert.py
+++ b/cloud/amazon/iam_cert.py
@@ -273,10 +273,18 @@ def main():
                                                     iam.get_all_server_certs().\
                                                     list_server_certificates_result.\
                                                     server_certificate_metadata_list]
-    orig_bodies = [iam.get_server_certificate(thing).\
-                  get_server_certificate_result.\
-                  certificate_body \
-                  for thing in orig_certs]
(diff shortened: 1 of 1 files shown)


resmo:
Thanks @joelthompson for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

Issue #3153, opened 2016-03-01
fstab parameter of mount module won't work


64er:
##### Issue Type:
- Bug Report
##### Ansible Version:

ansible 2.0.0.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = /usr/share/ansible
##### Environment:

Ubuntu 14.04 64Bit
##### Summary:

set fstab=**/tmp/fstab** in mount module result in 'can't find mount point in **/etc/fstab**'
##### Steps To Reproduce:

```
./test-module -m ../lib/ansible/modules/core/system/mount.py -a "src=/dev/sda9 name=/tmp/mnt fstype=ext3 state=mounted **fstab=/tmp/fstab**"
```
##### Expected Results:

```
changed: ok
```

and a line in /tmp/fstab and 'mount -a' should show the mounted device
##### Actual Results:

```
***********************************
PARSED OUTPUT
{
    "failed": true, 
    "invocation": {
        "module_args": {
            "dump": null, 
            "fstab": **"/tmp/fstab"**, 
            "fstype": "ext3", 
            "name": "/tmp/mnt", 
            "opts": null, 
            "passno": null, 
            "src": "/dev/sda9", 
            "state": "mounted"
        }
    }, 
    "msg": "Error mounting /tmp/mnt: mount: /tmp/mnt konnte nicht in **/etc/fstab oder /etc/mtab** gefunden werden\n"
}
```

mscherer:
Yep, it doesn't work on Linux, it seems to be only on freebsd for the current code. There is the option -T for Linux, could you give a try to the commit I propose ?

64er:
I tried your commit. I got an error like "no Option -T for mount". I looked into the manpage of mount and couldn't find -T on my Ubuntu 14.04 system.

mscherer:
Ok so that's only for newer coreutils. I guess we can add that to the doc and maybe some code to verify that, but I do not see any good way.

alikins:
Looks like the fstab option is only applied on freebsd, so needs docs updated.

ansibot:
@grossws, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

jesse-welch:
This change appears to break mount tasks on RHEL6 / CentOS6.

[...]

PR #3271, opened 2016-03-17
Add support for setting fstab location on Linux too for mount

mscherer:
##### Issue Type:
- Bugfix Pull Request
##### Plugin Name:

mount
##### Summary:

The mount module support using a different fstab. While that's a advanced feature and seldomly used, it wasn't working on Linux as seen on #3153 for mounting new FS. 

I am not sure since when the -T option was supported, but it work on RHEL 7, but not on RHEL 6.

### DIFF
--- a/system/mount.py
+++ b/system/mount.py
@@ -240,6 +240,9 @@ def mount(module, **kwargs):
     if get_platform().lower() == 'freebsd':
         cmd += [ '-F', args['fstab'], ]
 
+    if get_platform().lower() == 'linux':
+        cmd += [ '-T', args['fstab'], ]
+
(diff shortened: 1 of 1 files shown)


gregdek:

alikins:
The '-T' option doesn't seem to exist in RHEL6 or ubuntu14.

mscherer:
RHEL 6, I knew, but Ubuntu 14 is a bit surprising. 

It does exist on latest Debian, and according to git, it was committed in Jan 2012: https://github.com/karelzak/util-linux/commit/64b6bc4ff0ae182318dc71ed605a81203775003c

PR #2644, opened 2015-12-07
Correct problem with force:no

daveres:
I have been testing win_get_url script and I have detected that when I use the parameter force:no, the script fails for several reasons.

I have modified the script and here is the modified block (I hope it could be helpful)

PR #3273, opened 2016-03-18
Win_get_url improvements.

### LINKED ISSUE #2146: win_get_url with force=no appears broken.
##### Ansible Version:

ansible 1.9.1
  configured module search path = /etc/ansible/library

 (but likely same problem with 2.0 dev)
##### Ansible Configuration:

Probably not relevant
##### Environment:

controller: Centos 6.5
target: windows 
##### Summary:

Fails if file already exists.

This line is the cause of the immediate failure:
$stream = New-Object System.IO.StreamReader($response.GetResponseStream())

There is no variable called $response, it should probably be $webResponse.

I think this is because the code has been copied verbatim from this stackoverflow answer, which also has the same bug: http://stackoverflow.com/questions/26533819/how-can-i-download-a-file-only-if-it-has-changed-using-powershell

Second problem is that it doesn't actually work, at least for downloading binary files.

I think the code needs to change to request the headers and then compare the date of the last-modified header with the file last modified.

If the source date time is older than the file modified then I think it should fetch the file using the same $client.DownloadFile call - if you use on a zip file, the following lines create a corrupted zip file which is a different size from the file fetched with  $client.DownloadFile
##### Steps To Reproduce:

download a zip file using win_get_url
download the same file again.
##### Expected Results:

1st run: changed should be set to true
2nd and subsequent runs: changed should be set to false
##### Actual Results:

1st run is successful.
2nd run fails with 
"You cannot call a method on a null-valued expression."

jhawkesworth:
##### Issue Type:

- Bugfix Pull Request
##### Plugin Name:

win_get_url
##### Summary:

Fixes https://github.com/ansible/ansible-modules-core/issues/2146 (my original bug report)
Fixes https://github.com/ansible/ansible-modules-core/pull/2644 (existing PR, stalled)

System.Net.WebClient.DownloadFile method is used to retrieve file regardless of force yes/no (previously a different method of downloading the file was used if Force=no).

Force=no now retrieves the last mod time via http HEAD request, to save unnecesary downloading.

More specific "Error when requesting Last-Modified date from $url" exception message if the HEAD request fails.

Existing integration tests pass against Server 2008 R2 (Powershell 3.0) and Server 2012 R2 (Powershell 4.0)
##### Example:

cd test/integration
export TEST_FLAGS='-t test_win_get_url -vv -i /etc/ansible/hosts'
make test_winrm

### DIFF
--- a/windows/win_get_url.ps1
+++ b/windows/win_get_url.ps1
@@ -54,7 +54,7 @@ if($skip_certificate_validation){
 
 $force = Get-Attr -obj $params -name "force" "yes" | ConvertTo-Bool
 
-If ($force -or -not (Test-Path $dest)) {
+Function Download-File($result, $url, $dest, $username, $password, $proxy_url, $proxy_username, $proxy_password) {
     $webClient = New-Object System.Net.WebClient
     if($proxy_url) {
         $proxy_server = New-Object System.Net.WebProxy($proxy_url, $true)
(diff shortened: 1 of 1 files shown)


resmo:
Thanks @jhawkesworth. Since you are a maintainer of this module, we are marking this PR for inclusion.

resmo:
@jhawkesworth should this fix be backported to stable-2.0?

jhawkesworth:
Only functional difference is more specific error message on failure so I think there is no need to backport.

PR #3280, opened 2016-03-19
Fix shadowing of datetime in accelerate helper.

mattclay:
##### Issue Type:

Bugfix Pull Request
##### Plugin Name:

accelerate
##### Summary:

Commit https://github.com/ansible/ansible/commit/0b92abaf67de53349bb4d2733f49750d9a4d8277 added a datetime import to ansible.module_utils.basic which shadows the datetime import in the accelerate helper. This causes calls to datetime.now() in the accelerate helper to fail.
##### Example:

Attempting to run a playbook which uses accelerate mode will fail:

```
fatal: [localhost]: FAILED! => {"failed": true, "msg": "Failed to connect to 127.0.0.1 on the accelerated port 5099"}
```

Checking to see if the accelerate helper is running on the target host will show that it is not. Running manually the helper which was uploaded to the remote host shows the reason:

```
AttributeError: 'module' object has no attribute 'now'
```

This fix renames the datetime import in the accelerate helper so it is no longer shadowed.

Tested on Ubuntu 15.10.


mattclay:
Per @bcoca, a more appropriate fix for this issue is PR #3233.

PR #3311, opened 2016-03-23
Yum module always downloads remote rpms. fixes #1452

### LINKED ISSUE #1452: yum module install URL reports changed in --check

kustodian:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

yum module
##### ANSIBLE VERSION

```
ansible 2.1.0
```
##### SUMMARY

Yum module now always downloads a remote RPM file and checks if it's really installed or not.

This pull request fixes #1452

### DIFF
--- a/packaging/os/yum.py
+++ b/packaging/os/yum.py
@@ -482,6 +482,19 @@ def local_nvra(module, path):
                             header[rpm.RPMTAG_RELEASE],
                             header[rpm.RPMTAG_ARCH])
     
+def local_name(module, path):
+    """return package name of a local rpm passed in"""
+
+    ts = rpm.TransactionSet()
+    ts.setVSFlags(rpm._RPMVSF_NOSIGNATURES)
+    fd = os.open(path, os.O_RDONLY)
+    try:
+        header = ts.hdrFromFdno(fd)
+    finally:
+        os.close(fd)
+
+    return header[rpm.RPMTAG_NAME]
+    
 def pkg_to_dict(pkgstr):
 
     if pkgstr.strip():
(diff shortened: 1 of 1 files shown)


kustodian:
@abadger this pull request will finally fix the annoying bug where yum always reports changed in check mode. It works as we agreed in one of the previous pull requests, slower (it downloads the package), but it's accurate.

One thing I would like that the yum module could do in the future is that `state=latest` works with URLs as well, but that can be done some other time.

gregdek:
Thanks @kustodian for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

resmo:
What about using a request with ETAG If we got an "not modified HTTP 304", it will skip the download and we can just return a changed=False

ETAG http://serverfault.com/questions/690341/algorithm-behind-nginx-etag-generation

kustodian:
@resmo I don't know how does that solve anything, since you still don't know if the package is installed or not?

packaging/os/yum.py
```diff
@@ -556,33 +569,34 @@ def install(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
                 res['msg'] += "No Package file matching '%s' found on system" % spec
                 module.fail_json(**res)
 
-            nvra = local_nvra(module, spec)
+            pkg_name = local_name(module, spec)
```
kustodian, line 572: We have to pass just the package name here and not the full package version and arch, so that we can check if the package is installed. We don't care if the versions don't match, only if the package is installed.

abadger:
Thanks @kustodian!  Looks good.  Merged to devel.

PR #3316, opened 2016-03-24
service: allow user to query service status

ritzk:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

system/service.py
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

Add the ability to query for service status

```
Before 
$ ansible -i "127.0.0.1," all -m service -a"name=atd state=status"
127.0.0.1 | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "value of state must be one of: running,started,stopped,restarted,reloaded, got: status"


After
$ ansible -i "127.0.0.1," all -m service -a"name=atd state=status"
127.0.0.1 | SUCCESS => {
    "changed": false, 
    "state": true
}

```

Based of request from : https://www.reddit.com/r/ansible/comments/4bpvry/service_module_to_return_the_status_of_a_service/


resmo:
IMHO status would rather go into a facts module, status is not a state.

ritzk:
@resmo, does not make any sense to me, as the service module already has this implemented ( with get_service_status )

```
class Service(object):
    """
    This is the generic Service manipulation class that is subclassed
    based on platform.

    A subclass should override the following action methods:-
      - get_service_tools
      - service_enable
      - get_service_status
      - service_control

```

We could always return "running: true" over "state". This allows us to query only for a particular service status.

Are we looking to query for all the service status, if this is pushed into facts ( which could be slower) ?

resmo:
Thanks @ritzk for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

freeseacher:
@resmo any news?

resmo:
@freeseacher thanks for the ping, coming back to it shortly

rdmarsh:
Any chance this will be merged soon?

cornfeedhobo:
@ritzk did you move this to ansible/ansible yet?

piahoo:
will it be merged?

PR #3320, opened 2016-03-25
Added restart functionality to ec2.py

hyperized:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

cloud/amazon/ec2.pyc
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

Added the 'rebooted' state to EC2, so servers can be soft (OS level) rebooted without a full power-off

```
- hosts: localhost
  tasks:
    - name: reboot EC2 instance
      ec2:
        region: "{{ region }}"
        instance_id: i-********
        state: rebooted
```

```
PLAY ***************************************************************************

TASK [reboot EC2 instance] *****************************************************
changed: [localhost]

```

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -20,6 +20,7 @@ module: ec2
 short_description: create, terminate, start or stop an instance in ec2
 description:
     - Creates or terminates ec2 instances.
+    - C(state=restarted) was added in 2.2
 version_added: "0.9"
 options:
   key_name:
@@ -212,7 +213,7 @@ options:
     required: false
     default: 'present'
     aliases: []
-    choices: ['present', 'absent', 'running', 'stopped']
+    choices: ['present', 'absent', 'running', 'restarted', 'stopped']
   volumes:
     version_added: "1.5"
     description:
@@ -499,6 +500,15 @@ EXAMPLES = '''
     state: running
 
 #
+# Restart instances specified by tag
+#
+- local_action:
+    module: ec2
+    instance_tags:
+        Name: ExtraPower
+    state: restarted
+
+#
 # Enforce that 5 instances with a tag "foo" are running
 # (Highly recommended!)
 #
@@ -734,8 +744,8 @@ def create_block_device(module, ec2, volume):
     # http://aws.amazon.com/about-aws/whats-new/2013/10/09/ebs-provisioned-iops-maximum-iops-gb-ratio-increased-to-30-1/
     MAX_IOPS_TO_SIZE_RATIO = 30
 
-    # device_type has been used historically to represent volume_type, 
-    # however ec2_vol uses volume_type, as does the BlockDeviceType, so 
+    # device_type has been used historically to represent volume_type,
+    # however ec2_vol uses volume_type, as does the BlockDeviceType, so
     # we add handling for either/or but not both
     if all(key in volume for key in ['device_type','volume_type']):
         module.fail_json(msg = 'device_type is a deprecated name for volume_type. Do not use both device_type and volume_type')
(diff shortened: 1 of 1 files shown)


Caroga:
+1 very much needed here

resmo:
Thanks @hyperized for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

cloud/amazon/ec2.py
```diff
@@ -212,7 +212,7 @@
     required: false
     default: 'present'
     aliases: []
-    choices: ['present', 'absent', 'running', 'stopped']
+    choices: ['present', 'absent', 'running', 'rebooted', 'stopped']
```
ryansb: The keyword we usually use is "restarted"

hyperized: I would argue not to use it in this case because restarting implies a stop / start, which for AWS means the VM is stopped entirely and started on potentially new hardware. A restart can also invoke the termination sequence. Rebooted imho is the best keyword here as it implies a soft (OS) restart.

ryansb: Hrm, that's tough. In my experience, I rarely need to do a stop/start of an instance in one go, so maybe we could use "restarted" for this, and then document that for a stop/start users should do:

```
- ec2: <some options> state=stopped
- ec2: <some options> state=running
```

I worry about using such close keywords (rebooted & restarted) because I don't think all users will get the distinction and could trip up.

hyperized: I thought initially that the AWS documentation was clear on the distinction between reboot and restart, however I just realize they made our discussion worse: http://docs.aws.amazon.com/cli/latest/reference/ec2/reboot-instances.html

"If a Linux/Unix instance does not cleanly shut down within four minutes, Amazon EC2 performs a hard reboot."

I'm in favor of calling it 'restarted' and having people who want a hard stop / start do this via the method described above. I'll get the fix out on monday.

hyperized:
@ryansb your comments have been incorporated.

hyperized:
I see I forgot some changes to the function names, i'll redo those tomorrow :).

ryansb:
needs_revision

hyperized:
Revised, rebased and ready to go :)

cloud/amazon/ec2.py
```diff
@@ -1326,6 +1336,78 @@ def startstop_instances(module, ec2, instance_ids, state, instance_tags):
...
+            # Check "termination_protection" attribute
+            if inst.get_attribute('disableApiTermination')['disableApiTermination'] != termination_protection:
+                inst.modify_attribute('disableApiTermination', termination_protection)
+                changed = True
+
+            # Check instance state
+            if inst.state != state:
+                instance_dict_array.append(get_instance_info(inst))
+                try:
+                    inst.restart()
```
ryansb: In a fit of find-and-replace, it looks like you renamed `inst.reboot` to `.restart` which isn't actually a method. Changing it works locally for me.

cloud/amazon/ec2.py
```diff
@@ -499,6 +500,15 @@
     state: running
 
 #
+# Restart instances specified by tag
+#
+- local_action:
+    module: ec2
```
ryansb, line 506: Technically this also requires a `region: <somewhere>` option, otherwise the EC2 library can't connect. You can do this in a followup patch, no problem.

hyperized, line 506: region is not commonly used in other examples like terminate and start, I think a follow up PR should be done including those everywhere required.

ryansb, line 506: Fair enough.

ryansb:
Tested locally, and looks ready to :ship:. shipit

gregdek:
Thanks again to @hyperized for this PR, and thanks @ansible for reviewing. Marking for inclusion.

Caroga:
Any idea when this will be shipped, just curious?

ryansb:
Oops, I can just go ahead and do that. Sorry

PR #3324, opened 2016-03-25
Update postgresql_user.py

caylorme:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

database/postgresql/postgresql_user
##### ANSIBLE VERSION

devel
##### SUMMARY

Added a parameter to allow users to specify synchronous_commit for creation, alteration, or deletion of postgresql users. This will allow users to set up users before the replication nodes have been implemented.

A common use case for this is when users want to set up a replication user before standing up the replication nodes. If the master is already configured to wait for synchronous commits, the transaction will never take place.

adding the ability to override synchronous_commit on the transactions


database/postgresql/postgresql_user.py
```diff
@@ -120,6 +120,11 @@
     default: 'no'
     choices: [ "yes", "no" ]
     version_added: '2.0'
+  synchronous_commit:
+    description:
+      - Specifies whether transaction commit will wait for WAL records to be written to disk before the command returns a "success" indication to the client. Valid values are on, remote_write, local, and off. The default, and safe, setting is on.
+    required: false
+    default: true
```
resmo, line 127: please add a `version_added: "2.1"` to this new arg

resmo:
Thanks @caylorme for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

caylorme:
@resmo : I'm going to close this so we can use PR 3326. This PR contains the changes you've requested, but contains the changes for both modules.

https://github.com/ansible/ansible-modules-core/pull/3326

PR #3358, opened 2016-04-02
Allow async to not need an args file since new-style modules have args embedded

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

async_wrapper
##### SUMMARY

async_wrapper was trying to cover all the bases by passing an arguments file all the time.  This is not necessary for new-style modules.  Removing hte need for the argument file means we can stop sending an arguments file always which will make async slightly faster.  (and will let the new ziploader debug stuff work as well).


nitzmahone:
:+1:

PR #3361, opened 2016-04-04
Small documentation and comment style fixes

djmattyg007:
Fixes some indentation and creates something of a standard order for
attributes of options.


bcoca:
we don't normally take style fixes as this breaks attribution.

PR #2375, opened 2015-10-28
made port default depending on db engine

bcoca:
should fix #452

PR #3365, opened 2016-04-04
Fix read replica create BC break

michaelmoussa:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

rds
##### ANSIBLE VERSION

```
ansible 2.0.1.0
```
##### SUMMARY

This PR fixes a BC break introduced in #2375 when creating read replicas. The `db_engine` param is not a required param (in fact, the docs still say it is only used when `command=create`). If `db_engine` is not provided, Ansible fails when executing [this line](https://github.com/ansible/ansible-modules-core/blob/e277a9b1b96bc1aa65aa92f60563b8262fdaa882/cloud/amazon/rds.py#L1084).

This change will skip that entire block of logic if the `db_engine` is not set, which is fine for `command=replicate`. In the case of `command=create`, the `db_engine` should be provided anyway, so the existing functionality would be preserved in that case.

```
BEFORE:
  File "/Users/mmoussa/.ansible/tmp/ansible-tmp-1459801584.73-40074538159080/rds", line 3286, in <module>
    main()
  File "/Users/mmoussa/.ansible/tmp/ansible-tmp-1459801584.73-40074538159080/rds", line 1085, in main
    if '-' in module.params['db_engine']:
TypeError: argument of type 'NoneType' is not iterable

fatal: [127.0.0.1]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_name": "rds"}, "parsed": false}


AFTER:
ok: [127.0.0.1]
```


gregdek:
Thanks @michaelmoussa. To the current maintainers, @bpennypacker, @willthames please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

willthames:
`db_engine` is not a parameter that can be passed to boto [1][2] when creating a replica, so it shouldn't be a parameter here when `command=replicate`

I think we should just change it to 

```
if module.params['port'] is None and module.params['command'] == 'create':
```

[1] http://boto.cloudhackers.com/en/latest/ref/rds2.html#boto.rds2.layer1.RDSConnection.create_db_instance_read_replica
[2] http://boto.cloudhackers.com/en/latest/ref/rds.html#boto.rds.RDSConnection.create_dbinstance_read_replica

willthames:
needs_revision

michaelmoussa:
@willthames I updated the PR, but kept the check for `and module.params['db_engine'] is not None`. The reason is because if someone runs it with `command == 'create'` but forgets to specify the `db_engine`, they'll see the same error I was getting before, which isn't terribly helpful:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: argument of type 'NoneType' is not iterable
fatal: [127.0.0.1]: FAILED! => {"changed": false, "failed": true, "parsed": false}
```

With the `and module.params['db_engine'] is not None` check, the error will be:

```
fatal: [127.0.0.1]: FAILED! => {"changed": false, "failed": true, "msg": "Parameter db_engine required for create command"}
```

... which tells them very clearly what is wrong.

willthames:
@michaelmoussa yep, makes sense. Thanks for the fix.

michaelmoussa:
Thank you @willthames . Is there anything I need to do in order to get this merged?

willthames:
shipit @gregdek (I haven't personally tested it but the change is simple and an improvement - and this particular code path has clearly had more testing in this PR than ever before).

gregdek:
Thanks again to @michaelmoussa for this PR, and thanks @bpennypacker, @willthames for reviewing. Marking for inclusion.

PR #3374, opened 2016-04-06
Wrap calls to main() with if check

Shrews:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

cloud/openstack/os_client_config.py
cloud/openstack/os_image.py
cloud/openstack/os_ironic.py
cloud/openstack/os_ironic_node.py
cloud/openstack/os_object.py
cloud/openstack/os_security_group.py
##### ANSIBLE VERSION

2.1
##### SUMMARY

A change was merged to the main Ansible core code that can cause
a potential hang if any libraries are called that use threading.
This change was:

  https://github.com/ansible/ansible/commit/4b0aa1214ca989c71bb87f7a805b1e3673faefa9

This affected the os_object module by causing a hang on the shade
create_object() API call (which in turn calls swiftclient which
uses threading). The fix is to make sure all modules have a main()
that is wrapped with an `if __name__ == '__main__'` check.


abadger:
Looks good.  Thanks @Shrews!

PR #3379, opened 2016-04-07
Add exponential backoff retries to ec2_elb_lb

joelthompson:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ec2_elb_lb
##### ANSIBLE VERSION

Current devel branch
##### SUMMARY

The ec2_elb_lb module currently fails when hitting AWS API throttling errors.

This PR adds retry logic with exponential backoff (and random jitter) to some of the operations that are likely to hit such errors. This is implemented via a decorator that can be easily added to other methods in the future.

In addition, there are two spots in the current module that use `time.sleep` (`_wait_for_elb_removed` and `_wait_for_elb_interface_removed`) but `time` is never imported, so they would likely throw errors in those situations. This imports time properly.

### DIFF
--- a/cloud/amazon/ec2_elb_lb.py
+++ b/cloud/amazon/ec2_elb_lb.py
@@ -360,6 +360,29 @@ try:
 except ImportError:
     HAS_BOTO = False
 
+import time
+import random
+
+def _throttleable_operation(max_retries):
+    def _operation_wrapper(op):
+        def _do_op(*args, **kwargs):
+            retry = 0
+            while True:
+                try:
+                    return op(*args, **kwargs)
+                except boto.exception.BotoServerError, e:
+                    if retry < max_retries and e.code in \
+                            ("Throttling", "RequestLimitExceeded"):
+                        retry = retry + 1
+                        time.sleep(min(random.random() * (2 ** retry), 300))
+                        continue
+                    else:
+                        raise
+        return _do_op
+    return _operation_wrapper
+
+
+_THROTTLING_RETRIES = 5
 
 class ElbManager(object):
     """Handles ELB creation and destruction"""
@@ -403,6 +426,7 @@ class ElbManager(object):
         self.elb = self._get_elb()
         self.ec2_conn = self._get_ec2_connection()
 
+    @_throttleable_operation(_THROTTLING_RETRIES)
     def ensure_ok(self):
         """Create the ELB"""
         if not self.elb:
@@ -546,6 +570,7 @@ class ElbManager(object):
 
         return info
 
+    @_throttleable_operation(_THROTTLING_RETRIES)
     def _wait_for_elb_removed(self):
         polling_increment_secs = 15
         max_retries = (self.wait_timeout / polling_increment_secs)
@@ -563,6 +588,7 @@ class ElbManager(object):
 
         return status_achieved
 
+    @_throttleable_operation(_THROTTLING_RETRIES)
     def _wait_for_elb_interface_removed(self):
         polling_increment_secs = 15
         max_retries = (self.wait_timeout / polling_increment_secs)
@@ -590,6 +616,7 @@ class ElbManager(object):
 
         return status_achieved
 
+    @_throttleable_operation(_THROTTLING_RETRIES)
     def _get_elb(self):
         elbs = self.elb_conn.get_all_load_balancers()
         for elb in elbs:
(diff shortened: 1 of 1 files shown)


cloud/amazon/ec2_elb_lb.py
```diff
@@ -360,6 +360,29 @@
...
+                try:
+                    return op(*args, **kwargs)
+                except boto.exception.BotoServerError, e:
+                    if retry < max_retries and e.code in \
+                            ("Throttling", "RequestLimitExceeded"):
+                        retry = retry + 1
+                        time.sleep(min(random.random() * (2 ** retry), 300))
+                        continue
+                    else:
+                        raise
```
ryansydnor, line 380: do you want to raise 'e' here to propagate the original exception?

cloud/amazon/ec2_elb_lb.py
```diff
@@ -360,6 +360,29 @@
...
+import random
+
+def _throttleable_operation(max_retries):
+    def _operation_wrapper(op):
+        def _do_op(*args, **kwargs):
+            retry = 0
+            while True:
+                try:
+                    return op(*args, **kwargs)
+                except boto.exception.BotoServerError, e:
```
ryansydnor, line 373: should use the 'as' syntax to be forward compatible (if that time every comes for ansible!)

cloud/amazon/ec2_elb_lb.py
joelthompson, line 380: A bare "raise" like this in an exception handler should re-raise the original exception, as I recall: https://docs.python.org/2/tutorial/errors.html#raising-exceptions

cloud/amazon/ec2_elb_lb.py
```diff
@@ -360,6 +360,29 @@
 except ImportError:
     HAS_BOTO = False
 
+import time
+import random
+
+def _throttleable_operation(max_retries):
```
ryansydnor, line 366: wow. this seems like an enormously valuable helper function. seems like it would be more useful farther down in the library where many modules could leverage it. i see [this](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/api.py#L82) and wonder if we could leverage it, or if we should just have a boto specific implementation in the [ec2 module utils](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/ec2.py).

either way, i wouldn't block this PR going through (as it seems useful) but would suggest something like this deeper in the library.

cloud/amazon/ec2_elb_lb.py
joelthompson, line 373: Actually... no. That syntax isn't Python 2.4 compatible, and while boto requires 2.6 and thus would support the "as" syntax, the Ansible team would prefer to use consistent style across all their projects. Either use the "," syntax, or there's some other methods in module_utils.basic that try to do this in an independent way, but the "," syntax is generally pretty standard.

cloud/amazon/ec2_elb_lb.py
joelthompson, line 366: Agree :) See my comment at the end of #2968 where this sort of thing was also discussed.

cloud/amazon/ec2_elb_lb.py
ryansydnor, line 373: Yep. [You're right](http://docs.ansible.com/ansible/intro_installation.html#managed-node-requirements). Sorry!

cloud/amazon/ec2_elb_lb.py
ryansydnor, line 380: Just did a little test. Checks out.

ryansydnor:
LGTM :shipit:

gregdek:
Thanks @joelthompson. To the current maintainers, @jsdalton, @rickmendes please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

ghost:
@joelthompson Nice way to handle the throttling. Thanks for the add. shipit

linuxdynasty:
:shipit:

gregdek:
Thanks again to @joelthompson for this PR, and thanks @jsdalton, @rickmendes for reviewing. Marking for inclusion.

joelthompson:
FYI @rickmendes -- I updated the PR with a few additional uses of the retry decorator, nothing material.

@gregdek -- is there a chance of getting this included in the 2.1 code freeze? As is, it'll throw an uncaught exception if there's a throttling error.

Thanks!

ghost:
@joelthompson Looks good. Thanks.

joelthompson:
And one more tweak to work around a boto/boto#3526

Basically, that boto bug causes the retry decorator to not recover under certain circumstances.

joelthompson:
Thanks @resmo! :)

PR #3400, opened 2016-04-10
Improve doc accuracy of template module

gaomd:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

template
##### ANSIBLE VERSION

latest version
##### SUMMARY

Inaccurate words was used in template module: _local server_, now follows the writing style of `copy` module.

### DIFF
--- a/files/template.py
+++ b/files/template.py
@@ -37,7 +37,7 @@ description:
 options:
   src:
     description:
-      - Path of a Jinja2 formatted template on the local server. This can be a relative or absolute path.
+      - Path of a Jinja2 formatted template on the Ansible controller. This can be a relative or absolute path.
     required: true
   dest:
     description:


gregdek:
Thanks @gaomd for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

docschick:
@gaomd The original text seems okay to me.  Is there something about it that seems confusing or unclear? What problem are you trying to solve with the wordsmithing? I want to make sure I understand so that we can be like-minded here. :)

gaomd:
@docschick "Local **server**" is inaccurate, local maybe an MacBook, a Linux desktop, "local path" seems more reasonable.

docschick:
I guess I don't understand why "local server" is inaccurate here. The language used in the short description of "Templates a file out to a remote server" seems to match, to me.  
(@bcoca This may be where my limited knowledge of Ansible falls short. )

Also, "Local path to" versus "Local path of"  changes the meaning a bit as well. I prefer "of" in this case, since we're not talking about navigating to the file itself but the navigation of the file. 

Looking at files/copy.py -- if we need to update the template.py docs, then there may be more to do. The short description, for example, says "Copies files to remote locations".   Does the short description here need to be updated to reflect "remote locations" instead of "remote server"?

May this would work too:     "Path of a Jinja2 formatted template on the local system"  ??

bcoca:
probably using 'local path of the Ansible "master"' would be a clearer way to describe this, many people consider the machine executing ansible a 'server' for the purpose of management, others would decry that it is their 'workstation' ... not a semantic war i'm going to get into, 'master' would work for this case no matter what the underlying machine is.

gregdek:
We should probably have one canonical term for "the machine you run Ansible
from," as this comes up quite a lot. "master"? "controller"? It would be
nice to have an answer that everyone agrees on and accepts.

docschick:
FWIW, "ansible master" come up in more google search terms than "ansible controller". In some of the intro documentation, we use controller (like: http://docs.ansible.com/ansible/intro_windows.html).

gregdek:
@docschick You are in the enviable position of choosing once and for all. :)

docschick:
I'm leaning strongly toward "controller" based on current docs usage and feedback from some core team members. 

I think this would be acceptable:
"Path of a Jinja2 formatted template on the local Ansible controller. This can be a relative or absolute path"

gaomd:
@docschick Commit updated, with the nice choice: Ansible controller.

---

_For the record, the previous commit we've been discussing here is:_

``` diff
- Path of a Jinja2 formatted template on the local server. ...
+ Local path to a Jinja2 formatted template. ...
```

gaomd:
Commit updated again, to remove **local** prefix of Ansible controller, since it can live on local (workstation) or on remote (a cloud server running ansible as a controller).

docschick:
Thanks @gaomd! We appreciate your patience as well as the fix!

PR #3413, opened 2016-04-12
Fix mount's handling of passno

jkugler:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

mount module
##### ANSIBLE VERSION

Master
##### SUMMARY

Mount was not handling passno correctly, and would try to run .replace() on an integer input. This tells Ansible that the input is an integer, so it converts correctly.

```
Before: AttributeError: 'int' object has no attribute 'replace'
```


jimi-c:
Closing per further discussion in IRC. This does not resolve the original issue.

Issue #2895, opened 2016-01-27
mysql_user won't create user if password is empty


walro:
**Bug Report**

**Ansible version**: 2.0.0.2

**Ansible Configuration**: See below

**Environment**: Running OS X 10.11.3, managing MySQL 5.5.46 on FreeBSD 10.2.

**Summary**: It seems like the `mysql_user` module won't create a user if the password is empty, this worked fine in 1.9.4.

**Steps To Reproduce**:

`ANSIBLE_CONFIG=ansible.vagrant.cfg ansible freebsd-mysql -m mysql_user -a "name='ha-check' password='' host='%'" --become`

**Expected results**: User is created.

**Actual results**: User is not created.

---

Contents of `ansible.vagrant.cfg`:

```
ask_sudo_pass=False
hostfile=.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory
log_path=logs/ansible.vagrant.log
host_key_checking=False
remote_user=vagrant
private_key_file=~/.vagrant.d/insecure_private_key
retry_files_enabled=False
```

bob-rove:
+1

Jmainguy:
Whats the point in having a user where you don't know the password?

walro:
> Whats the point in having a user where you don't know the password?

I do know the password, it's empty/not set/blank. HAProxy needs such a user to perform its [MySQL health check](https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4-option%20mysql-check).

Jmainguy:
@walro @bob-rove checkout #3368

walro:
hey @Jmainguy as of Ansible 2.1 this seems to be working again, thanks!

I'll let you decide what to do with the issue, but from my point of view it may be closed.

PR #3445, opened 2016-04-15
Allow an empty password when adding a new MySQL user.

jgeusebroek:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

database/mysql/mysql_user
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file =
```
##### SUMMARY

It is currently impossible to add a mysql user without a password. This small patch fixes this.

While generally it is unwise not to use a password, in my opinion Ansible should not force a password to be used. For example, the haproxy mysql-check, requires an empty password.

### DIFF
--- a/database/mysql/mysql_user.py
+++ b/database/mysql/mysql_user.py
@@ -32,7 +32,7 @@ options:
     required: true
   password:
     description:
-      - set the user's password. (Required when adding a user)
+      - set the user's password.
     required: false
     default: null
   encrypted:
@@ -232,6 +232,8 @@ def user_add(cursor, user, host, host_all, password, encrypted, new_priv, check_
         cursor.execute("CREATE USER %s@%s IDENTIFIED BY PASSWORD %s", (user,host,password))
     elif password and not encrypted:
         cursor.execute("CREATE USER %s@%s IDENTIFIED BY %s", (user,host,password))
+    else:
+        cursor.execute("CREATE USER %s@%s", (user,host))
 
     if new_priv is not None:
         for db_table, priv in new_priv.iteritems():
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @jgeusebroek. To the current maintainers, @Jmainguy please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit' or 'needs_revision' as appropriate.

Jmainguy:
This is designed to fix #2895

Jmainguy:
Works in my testing for new users.

@mscherer @jgeusebroek Do yall think this should be expanded to change the users password to blank if not set? Or leave it as is, if blank and new user, create new user with blank password, if blank and user exists, do not change password to blank.

What do yall think?

jgeusebroek:
I didn't even see #2895 :) Looks like he ran into the same problem.

Good point. To prevent user error, I think it might be better to leave as is.
If for some reason a password variable might suddenly not be set, it would prevent leaving a gaping security hole.

On the other hand, you can't take into account all user errors, and it might be better to have consistent behaviour, and hence also reset the password to blank if omitted.

mscherer:
IIRC, mysql can use more than password to connect. So mandating passwords when creating the user would prevent that. So I would be in favor of the change. And the code do look fine from a quick look.

Jmainguy:
Awesome, thanks guys

:shipit:

gregdek:
Thanks again to @jgeusebroek for this PR, and thanks @Jmainguy for reviewing. Marking for inclusion.

resmo:
No security concerns from me either. shipit

PR #2635, opened 2015-12-06
add vlidate_certs option for unarchive module

vmindru:
missing thsis option, needed when fetching file from HTTPS

PR #3452, opened 2016-04-18
Fix unarchive permission issue to tempdir

### LINKED ISSUE #2708: unarchive permission issues with become and http src

gregsidelinger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

unarchive
##### ANSIBLE VERSION

```
ansible 2.0.0.2
  config file =
```
##### SUMMARY

Fixes #2708

Updated the unarchive module to use a different temp directory to download http files when copy=False and it can't write to the default temp directory.  The download file is also cleaned up before the module exists.


gregdek:
Thanks @gregsidelinger. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

dagwieers:
Personally I am not convinced that download functionality should be part of the unarchive module. I'd prefer to deprecate this functionality, rather than extending it. Leave downloading to more specialized modules.

This discussion is related to #2635

gregsidelinger:
Personally I think the fetch_url functionally should be part of AnsibleModule so that any module can use it just like the file options are.  I have a handful of custom modules that make use of pulling files from urls and I find it very useful instead of having to always copy/get_url the file before hand.

Also without the fetch support you turn an unarchive into three steps, not two.
fetch
unarchive
clearnup

dagwieers:
Correct, but the same goes for something as simple as **template.** I would like to template a remote file from somewhere. Should we add this capability to template because there are use-cases ? And we can turn 3 steps into one ? Sure it is convenient, but I am not convinced.

In the end the **unarchive**, **template**, and other modules will have all options to handle certificates, encryption, authentication, and whatnot...

dagwieers:
@gregsidelinger Let it be clear that my personal opinions don't matter here ;-)

gregdek:
Thanks @gregsidelinger for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregsidelinger:
ready_for_review

I've fixed the conflict created after @dagwieers latest changes to the module.

gregdek:
Thanks @gregsidelinger. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

Issue #3450, opened 2016-04-17
ansible-pull Failing to checkout non-HEAD commit hashes


srkiNZ84:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

Git module 
##### ANSIBLE VERSION

```
ansible 2.0.0.2
  config file =
  configured module search path = Default w/o overrides
```
##### OS / ENVIRONMENT

N/A
##### SUMMARY

When doing an "ansible-pull" command the git module fails to checkout if the commit hash specified is not the current "HEAD"
##### STEPS TO REPRODUCE

Create a repository with a simple playbook e.g.

```
- hosts: all
  tasks:
    - name: List files in directory
      connection: local
      command: ls -al
      register: filelisting

    - name: Print out listing
      connection: local
      debug: msg="{{ filelisting.stdout_lines }}"
```

After the initial commit, note the commit hash.

Then add another commit after the initial one (e.g. add a "dummy" file to the repo).

The repository should have two commits now, each with their own hash (lets refer to them as COMMIT1 and COMMIT2).

Run:

```
ansible-pull listfiles.yml -U git@bitbucket.org:username/ansible-pull-test.git -i hosts -C COMMIT1 -vvvvvv
```
##### EXPECTED RESULTS

Expected the command to clone the repository at that commit hash and execute the "listfiles.yml" playbook.
##### ACTUAL RESULTS

```
$ ansible-pull listfiles.yml -U git@bitbucket.org:username/ansible-pull-test.git -i hosts -C COMMIT1 -vvvvvv
No config file found; using defaults
Starting Ansible Pull at 2016-04-17 12:38:42
/usr/local/bin/ansible-pull listfiles.yml -U git@bitbucket.org:username/ansible-pull-test.git -i hosts -C COMMIT1 -vvvvvv
EXEC: /usr/local/bin/ansible -i "hosts" -c local  -vvvvvv -m git -a "name=git@bitbucket.org:username/ansible-pull-test.git dest=/Users/srdan/.ansible/pull/srdans-macbook-pro.local version=COMMIT1 depth=1" "localhost,MacBook-Pro.local,127.0.0.1"
No config file found; using defaults
Loaded callback minimal of type stdout, v2.0
ESTABLISH LOCAL CONNECTION FOR USER: username
[...]

PR #3466, opened 2016-04-19
git: fall back to full clone if version is hash

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

source_control/git.py
##### ANSIBLE VERSION

devel, 2.1.0
##### SUMMARY

This is based on the discussion at https://github.com/ansible/ansible-modules-core/issues/3450#issuecomment-211829562 and is an actual fix for #3450

If version is not a branch or tag (i.e. a hash), we need a full checkout to be able to switch to it, so the depth argument is ignored.

While strictly speaking, it is a conflicting of arguments, I would prioritize the checkout of the right `version` over the savings of `depth`, and always fall back to a full clone/fetch when needed. 

Especially with ansible-pull choosing a hash like in #3450 is broken otherwise.

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -321,14 +321,18 @@ def clone(git_path, module, repo, dest, remote, depth, version, bare,
     except:
         pass
     cmd = [ git_path, 'clone' ]
+
+    branch_or_tag = is_remote_branch(git_path, module, dest, repo, version) \
+        or is_remote_tag(git_path, module, dest, repo, version)
+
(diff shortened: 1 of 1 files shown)


robinro:
This also works with the additional tests in ansible/ansible#15471.

abadger:
Thanks @robinro and @srkiNZ84 for the diagnosis, report, discussion, and fix ;-)

PR #3464, opened 2016-04-19
git: with depth only fetch actual version, don't use refs/heads/*

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

source_control/git.py
##### ANSIBLE VERSION

devel, 2.1.0
##### SUMMARY
- don't use refs/heads/branchname for branches
- for tags it's needed though
- fixes #3456

PR #3492, opened 2016-04-20
fix git switch branches in combination with depth on old git versions

robinro:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

source_control/git.py
##### ANSIBLE VERSION

devel, 2.1.0
##### SUMMARY

Fixes a bug that makes the module fail when depth is used in combination with switching to a different  (non-local) branch.
The bug showed up in centos6/7 when https://github.com/ansible/ansible/pull/15471 was merged.
- the old implementation is correct, but relies on git working correctly
  (as only newer versions do)
- with some older git version (e.g. 1.8.3), git fetch remote branch does
  not work, if the branch does not yet exist locally
- this patch works around that without explicitly checking the git version
- it specifies the requires git versions required explicitly in the docs, also the ones to get full depth functionality
- a function call non-existent in git 1.7.1 is replaced with the corresponding older formulation
- future refactoring is needed to reduce the number of fetches to a
  minimum; but this patch makes the code "correct" and allows to enable the test in https://github.com/ansible/ansible/pull/15490 again


robinro:
This partially reverts #3464.
It currently works with centos7, but another bug showed up in the git version of centos6...

gregdek:
This PR is a work-in-progress. @robinro, when you're ready to have this PR reviewed, please remove WIP from the title of the PR.

robinro:
@abadger please have a look and merge, this should be in 2.1 as the bug fixed breaks git when changing version in combination with depth (which is default in ansible-pull).
The corresponding revert of the revert of the test is https://github.com/ansible/ansible/pull/15490, which passes once this is merged.

2496d77 fixes the problem with centos6, the old code (#934) used a git feature that is not available in git version 1.7.1; this now implements a workaround for what the new git feature does.
I ran the full testsuite on centos6. Centos7 gives me some unrelated ssh problem. On OpenSuSE 42.1 and ubuntu 12.04 I only checked the git role, which passed.

I'm aware this change increases the code complexity and does more fetches than absolutely necessary, but the idea is to get the code correct first and then performance-optimize/refactor once it's working and the tests are there. I feel we make good progress on the test front and are at a point that allows a large refactor to address some of the open issues soon.

gregdek:

abadger:
Thanks @robinro!

PR #3561, opened 2016-04-27
fixes docstring issue for junos_config arguments

privateip:
The docstring was incorrectly merged with some old entries that will
cause the documentation not to build correclty.  It is now fixed.


abadger:
Thanks for the quick response @privateip !

PR #3575, opened 2016-05-01
unarchive: use Python's tarfile module for tar listing

### LINKED ISSUE #11348: bug in unarchive module
Hi

I have found a problem with the unarchive module when the file to uncompress contains "weird" characters. With this playbook the problem can be reproduced:

```

---

- name: reproduce the unarchive bug
  hosts: 127.0.0.1
  connection: local

  tasks:
  - name: Download LMOD
    get_url:
      url=https://github.com/TACC/Lmod/archive/6.0.4.tar.gz
      dest=/tmp/
      mode=0644
      validate_certs=no

  - name: Uncompress LMOD
    unarchive: src=/tmp/Lmod-6.0.4.tar.gz dest=/tmp copy=no mode=0644
```

And this is the output when executing it:

```
root@mgmt05 /tmp # ansible-playbook unarchive-bug.yml

PLAY [reproduce the unarchive bug] ********************************************

ok: [127.0.0.1]

TASK: [Download LMOD] *********************************************************
ok: [127.0.0.1]

TASK: [Uncompress LMOD] *******************************************************
  File "/root/.ansible/tmp/ansible-tmp-1434998674.55-80965098747038/unarchive", line 1895, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1434998674.55-80965098747038/unarchive", line 289, in main
    res_args['changed'] = module.set_fs_attributes_if_different(file_args, res_args['changed'])
  File "/root/.ansible/tmp/ansible-tmp-1434998674.55-80965098747038/unarchive", line 1090, in set_fs_attributes_if_different
    file_args['path'], file_args['mode'], changed
  File "/root/.ansible/tmp/ansible-tmp-1434998674.55-80965098747038/unarchive", line 928, in set_mode_if_different
    path_stat = os.lstat(path)
OSError: [Errno 2] No such file or directory: '/tmp/Lmod-6.0.4/contrib/Presentations/Keynote_Lmod_SC14-Calcul_Que\\314\\201bec_experience.pdf'


           to retry, use: --limit @/root/unarchive-bug.retry

127.0.0.1                  : ok=2    changed=0    unreachable=0    failed=1
```

the file can be uncompressed without issues using "tar" directly in the command line or using the ansible shell module to call "tar". This is the file name which triggers the error:

```
Keynote_Lmod_SC14-Calcul_Québec_experience.pdf
```

I have found this issue with ansible 1.9.1

ghost:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

unarchive
##### ANSIBLE VERSION

```
ansible-playbook 2.2.0 (devel 1d6608e84f) last updated 2016/04/29 16:43:56 (GMT -400)
  lib/ansible/modules/core: (fix-unarchive-encoding 40ee859114) last updated 2016/04/30 21:23:35 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD ca310f3d15) last updated 2016/04/29 16:53:29 (GMT -400)
  config file =
```
##### SUMMARY

fixes https://github.com/ansible/ansible/issues/11348

Depending on the current active locale, `tar`'s file listing can end up
spitting backslash-escaped characters. Unfortunately, when that happens,
we end up with double-escaped backslashes, giving us a wrong path,
making our action fail.

We could try un-double-escaping our paths, but that would be complicated
and, I think, error-prone. The easiest way forward seemed to simply use
the `tarfile` module.

Why use it only for listing? Because the `unarchive` option also
supports the `extra_opts` option, and that supporting this would require
us to mimick `tar`'s interface.

For listing files, however, I don't think that the loss of `extra_opts`
support causes problems (well, I hope so).

Before the fix:

```
$ cat foo.yml
- hosts: localhost
  tasks:
    - unarchive: src=foo.tar copy=no dest=/tmp

$ cd /tmp
$ mkdir foo
$ touch foo/é.txt
$ tar -cf foo.tar foo
$ LANG=C ansible-playbook -c local foo.yml
[WARNING]: Host file not found: /etc/ansible/hosts

 [WARNING]: provided hosts list is empty, only localhost is available


ok: [localhost]

fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Unexpected error when accessing exploded file: [Errno 2] No such file or directory: '/tmp/foo/\\\\303\\\\251.txt'"}

        to retry, use: --limit @foo.retry

```

After the fix:

```
$ LANG=C ansible-playbook -c local foo.yml
 [WARNING]: Host file not found: /etc/ansible/hosts

 [WARNING]: provided hosts list is empty, only localhost is available


ok: [localhost]

localhost                  : ok=1    changed=0    unreachable=0    failed=0
```


gregdek:
Thanks @hsoft. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

abadger:
This looks pretty good but it doesn't handle xz compressed tar archives.  If we go this route you'll probably have to write a new function in the TarXzArchive class to handle that case.  Maybe you can use one of the streaming modes: https://docs.python.org/2/library/tarfile.html#tarfile.open  along with invoking the xz program via subproces to do that.

dagwieers:
It is still my intention to create a pure-python unarchive. But I think we may have to consider calling it differently. Because it cannot be drop-in replacement:
- extra_opts would not be possible
- it may require additional python modules for e.g. xz support
- we will have to implement the whole logic of gtar ourselves (which is needed for check-mode and diff-mode anyway)

If we also want to support .gz. .bzip2, .xz files, we could go for module "extract" or something similar ?

PS it would supersede this change though...

ghost:
@dagwieers yeah, I saw that in the module's todo. But here, my goal is to fix an immediate bug. That's why I limit my scope to file listing.

@abadger you're right, thanks for pointing that out, I'll work on it and come back with something.

dagwieers:
@hsoft, why don't we set the locale for gtar ? The change would be minimal and at least the support would be identical ?

ghost:
@dagwieers because it's dependent on the encoding the archive was created in, and even more than that: the encoding under which the _file_ was created. Nothing stops a tar archive from having a latin-1 encoded filename next to a utf-8 encoded one. In this case, you're necessarily going to get backslash-escaped filenames, regardless of your current locale.

ghost:
@abadger I've overloaded `TarXzArchive` so that it pre-decompresses the archive before sending it to `TarArchive`. If I reproduce the same situation as is described in my first message, but with `xz` compression, I get a successful playbook run.

dagwieers:
@hsoft Ok, I understand.

abadger:
Thanks @hsoft!  Merged to devel and stable-2.1 branches (for release with 2.1.0)

dagwieers:
This fix broke the "exclude" implementation. It does reimplement excluding files, but `gtar --exclude` works on directory-basis as well, so the behaviour is different :-/

dagwieers:
And it is practically impossible to implement the `gtar --exclude` logic ourselves (if you read gtar's exclude implementation, which is a combination of per-component globbing combined with full-path matching. That is why I was basically against mixing 2 different routines for handling an archive (tarfile and gtar).

ghost:
Ooh, sorry, I didn't think about that. What should we do, add a flag to let the user select between `tarfile` and `gtar` implementations? A separate module?

In my original message, I wrote:

> We could try un-double-escaping our paths, but that would be complicated
> and, I think, error-prone.

Maybe that this is the best approach after all...

ghost:
Oh, I might have have a better idea! Essentially, what is `files_in_archive` (the only place where we use `tarfile`) used for?
1. Performing `chown` on the resulting files
2. Returning a list of extracted files
3. Determining if we can handle the archive

Except for number 3, these are all post-extract actions. So, what we could do is ignore the exclude flag entirely and only return files which are listed in the archive _and_ which exists in the dest path.

As for determining if we can handle the archive, we should use something else than `files_in_archive`.

What do you think about that?

dagwieers:
What about check-mode ? It would always be empty, if the file was never extracted (and the destination directory is empty).

I would like to evaluate the original problem, can you send me a file that has the original problem you were trying to solve ?

dagwieers:
Maybe this is easier:

```
string = "foo/\\303\\251.txt"
print string.decode("string_escape")
```

I am going to integrate this to see if it works in all cases, and we can go back to the original code.

ghost:
@dagwieers the original problem is that we _don't_ get a string like `"foo/\303\251.txt"` from stdout, but rather `"foo/\\303\\251.txt"`.

Also: we don't actually need a unicode string from that `"foo/\303\251.txt"`. We only need that path to actually refer to the right file. In fact, we should _not_ try to decode that path. We should keep it as a byte string.

dagwieers:
No, it works. You see a double slash because it is represented like that.

The output of gtar only has a single backslash. I reverted your change and implemented this and it works fine for the reproducer you shared. Please test !

ghost:
I confirm: it works! Sorry for having brought forward a solution that was way too complicated. It's much better this way.

It really bugged me not to understand why it works though, so I read a bit about `string_escape`, and it's indeed the right thing to do, but the example you gave above is the wrong one. The backslash given by gtar's stdout _are_ double-escaped (and re-doubled-doubled in the repr `'/tmp/foo/\\\\303\\\\251.txt'`), but that's precisely what `string_escape` allows us to fix:

```
>>> string = "foo/\\303\\251.txt"
>>> string.decode('string_escape')
'foo/\xc3\xa9.txt'
```

_This_ is the path were actually looking for.

dagwieers:
@hsoft Strange, if I run the `LANG=C gtar -tvf foo.tar` command on the command line, I don't see any escaping, or double-escaping. I still think the escaping is because of Ansible (representation) and not the actual string we get from gtar itself. Please verify, it might be that gtar on your system behaves differently, but it shouldn't.

```
[dag@moria ansible.testing]$ LANG=C gtar -tvf files/foo.tar 
drwxrwxr-x dag/dag           0 2016-05-27 14:28 foo/
-rw-rw-r-- dag/dag           0 2016-05-27 14:28 foo/\303\251.txt
```

ghost:
@dagwieers Yes, I know. `gtar`'s output is correct. It's ansible processing of that output that escapes the backslash. That's what we have to undo with `string_escape`.

ghost:
Thinking further, we _might_ want to unescape all commands output, not just the one from `unarchive`, but that's a much broader question with many implications.

dagwieers:
No, in python the string is NOT escaped at all. Ansible's JSON representation of the string is escaped. Internally the filename is not escape, or double-escaped. Internally the output is exactly as we get it from gtar.

PS At least not on my versions of Ansible, I noticed your example was from the master branch (2.2.0), and it was double-escape, I have not seen this specific behaviour before and I cannot reproduce it.

ghost:
If that was the case, we wouldn't have this bug here, wouldn't we?

dagwieers:
Yes we would,  because gtar returns an octal representation of 2 bytes as a string, and we have to convert it back into 2 bytes. That's what we do using `decode('string_escape')`.

The reason why gtar returns an octal representation of unicode characters is because it cannot do unicode when `LANG=C`. It is as simple as that.

ghost:
Yes, I see it now. Sorry for the noise.

dagwieers:
What may be confusing in all this is that the octal representation of 2 bytes in a string happens to be `\\303\\251` in python. But that translates to the string `\303\251` as is the output from gtar.

dagwieers:
Arghl, GitHub eats my escaped slashes :-)

Issue #1321, opened 2015-05-13
Teach ec2_lc to automatically generate names from a base name


jsternberg:
The ec2_lc module gives us a little bit of difficulty right now because it can't modify a launch configuration. We would like it if the module allowed an option to automatically generate a hash of the module parameters and include it in the launch configuration name. Right now, we have to manually update the uuid like this:

```
- ec2_lc:
    name: myservice-<hash1>
    state: present

- ec2_asg:
    launch_config_name: myservice-<hash1>
    replace_all_instances: yes
    state: present
```

We'd have to go and replace hash1 with hash2 manually every time we make a change. Sometimes, changes are made implicitly by a variable getting updated somewhere in the system (like a new global AMI).

I would like to be able to use `register` with `ec2_lc` and then pass the result to `ec2_asg` and have `ec2_lc` automatically determine the name using something like an md5 hash. So I would do:

```
- ec2_lc:
    name: myservice
    hash_method: md5
    state: present
```

Then have this create a launch configuration with `myservice-<md5hashofparams>`.

I'm not sure what the desired method is nor do I know what the argument should be named.

jsternberg:
Updated the above to make it more clear that the launch config name should be a hash and not a uuid.

shawnsi:
@jsternberg How does my change look?  I'm using it locally to solve some problems.

jsternberg:
Looks fine to me.

akvadrako:
It's rather confusing that "name" is changed by the module

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibot:
This repository has been locked. All new issues and pull requests should be filed in https://github.com/ansible/ansible

[...]

PR #3581, opened 2016-05-02
Adding append_hash parameter to ec2_lc module for #1321

shawnsi:
Before now I've used select parameters (instance size and AMI) in my launch configuration names.  At some point AWS stopped allowing deletion of a launch configuration which is attached to an autoscaling group.  That makes it very difficult to deal with changes that didn't explicitly update the launch configuration name.  This hash based approach should guarantee a consistent and distinct launch configuration name for every possible set of parameters.

This implements the issue described in #1321 as well.

### DIFF
--- a/cloud/amazon/ec2_lc.py
+++ b/cloud/amazon/ec2_lc.py
@@ -106,6 +106,12 @@ options:
       - A list of security group id's with which to associate the ClassicLink VPC instances.
     required: false
     version_added: "2.0"
+  append_hash:
+    description:
+      - Appends an md5 hash of launch configuration properties to the name.
+    required: false
+    default: false
+    version_added: "2.1"
 extends_documentation_fragment:
     - aws
     - ec2
@@ -133,6 +139,8 @@ EXAMPLES = '''
 
 from ansible.module_utils.basic import *
 from ansible.module_utils.ec2 import *
+
+import hashlib
 
 try:
     from boto.ec2.blockdevicemapping import BlockDeviceType, BlockDeviceMapping
@@ -182,6 +190,7 @@ def create_launch_config(connection, module):
     ebs_optimized = module.params.get('ebs_optimized')
     classic_link_vpc_id = module.params.get('classic_link_vpc_id')
     classic_link_vpc_security_groups = module.params.get('classic_link_vpc_security_groups')
+    append_hash = module.params.get('append_hash')
     bdm = BlockDeviceMapping()
 
     if volumes:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @shawnsi for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

akvadrako:
This seems quite important to have, since otherwise launch configurations are not idempotent. I would however prefer another approach because this doesn't behave how one would expect ansible modules to behave. Instead of an `append_hash` I would use `name_prefix`, which replaces `name` (since it's no longer accurate) and appends a hash.

Or, even better, allow a general magic variable in tasks called `task_hash` then we can say `name=mylc-{{ task_hash }}`.

shawnsi:
@akvadrako Are you proposing that I then set `name` from the concatenated `name_prefix` variable and computed hash?  I currently register the result of the task calling the module and use the `name` variable in any downstream launch configuration references.

akvadrako:
I think that would preserve the semantics of **name** better.

cloud/amazon/ec2_lc.py
```diff
@@ -212,6 +221,15 @@ def create_launch_config(connection, module):
         classic_link_vpc_id=classic_link_vpc_id,
     )
 
+    if append_hash:
+        # MD5 of launch configuration properties
+        h = hashlib.md5()
+        h.update(str(frozenset(lc.__dict__)))
+
+        # Update name variables with md5 hash
+        name = '-'.join((name, h.hexdigest()))
```
ryansb, line 230: This is mostly an aesthetic request, but the AWS UI doesn't handle super-long launchconfig names that well. Could you maybe truncate the hash to 8 characters? For the hash-collision-concerned: 8 hex characters encodes 32 bits of the hash, and to reach a probability of 50% collision, you'd need to have (on average) 77164 hashes. Of course, it's unlikely someone will have 77,000 launchconfigs, let alone 77,000 at the same time. 

```
hash_size = 2**32 # 32 bits
probUnique = 1.0
for k in xrange(1, 200000):
    probUnique = probUnique * (hash_size - (k - 1)) / hash_size
    if probUnique <= 0.5: 
        print "greater than 50% probability of collision at {} hashes, probability {}".format(k, probUnique * 100)
        break
```

```
```

shawnsi: @ryansb thanks for the feedback.  I think aligning with API capabilities and lower probability of collisions makes sense for automation.  Is there a specific operation in the console this complicates?  I've been using it in production for some time and haven't hit any issues.

My vote is for more bits.  According to https://en.wikipedia.org/wiki/Birthday_attack 32 bit collisions would creep into the realm of possibility without exceeding launch configuration limits.  It probably won't happen to either of us but why make it more likely?

shawnsi:
@ryansb do you have any opinion on @akvadrako's comments regarding the handling of the `name` variable?

PR #1909, opened 2015-08-13
Fix service.enabled for masked services

evverx:
**Issue Type:** Bugfix Pull Request

**Ansible Version:** ansible 2.0.0 (devel c6733cc1a7) last updated 2015/08/13 12:30:03 (GMT +000)

**Environment:**

``` sh
Description:    Ubuntu 15.04
Release:    15.04
```

**Steps To Reproduce:**

``` sh
root# systemctl enable ssh
root# systemctl mask ssh
root# ansible all -m service -a 'name=ssh enabled=true' --sudo
```

**Expected Results:**

``` sh
127.0.0.1 | SUCCESS => {
    "changed": true,
    "enabled": true,
    "name": "ssh"
}
root# systemctl is-enabled ssh
enabled
```

**Actual Results:**

``` sh
127.0.0.1 | SUCCESS => {
    "changed": false,
    "enabled": true,
    "name": "ssh"
}
root# systemctl is-enabled ssh
masked
```

PR #3660, opened 2016-05-16
first draft systemd service plugin

bcoca:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

systemd
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

New systemd module to manage systemd services.

### DIFF
--- a/system/systemd.py
+++ b/system/systemd.py
@@ -247,10 +247,28 @@ def main():
         module.fail_json(msg='failure %d running systemctl show for %r: %s' % (rc, unit, err))
 
     # load return of systemctl show into dictionary for easy access and return
-    for line in out.split('\n'):
-        if line.strip() and '=' in line:
-            k,v = line.split('=', 1)
-            result['status'][k] = v.strip()
+    k = None
+    multival = []
+    for line in out.split('\n'): # systemd can have multiline values delimited with {}
+        if line.strip():
+            if k is None:
+                if '=' in line:
+                    k,v = line.split('=', 1)
+                    if v.lstrip().startswith('{'):
+                        if not v.rstrip().endswith('}'):
+                            multival.append(line)
+                            continue
+                    result['status'][k] = v.strip()
+                    k = None
+            else:
+                if line.rstrip().endswith('}'):
+                    result['status'][k] = '\n'.join(multival).strip()
+                    multival = []
+                    k = None
+                else:
+                    multival.append(line)
+
+
 
     if 'LoadState' in result['status'] and result['status']['LoadState'] == 'not-found':
         module.fail_json(msg='Could not find the requested service "%r": %s' % (unit, err))
--- a/system/systemd.py
+++ b/system/systemd.py
@@ -1,0 +1,336 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2016, Brian Coca <bcoca@ansible.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+module: systemd
+author:
+    - "Ansible Core Team"
+version_added: "2.2"
+short_description:  Manage services.
+description:
+    - Controls systemd services on remote hosts.
+options:
+    name:
+        required: true
+        description:
+            - Name of the service.
+        aliases: ['unit', 'service']
+    state:
+        required: false
+        default: null
+        choices: [ 'started', 'stopped', 'restarted', 'reloaded' ]
+        description:
+            - C(started)/C(stopped) are idempotent actions that will not run commands unless necessary.
+              C(restarted) will always bounce the service. C(reloaded) will always reload.
+    enabled:
+        required: false
+        choices: [ "yes", "no" ]
+        default: null
+        description:
+            - Whether the service should start on boot. B(At least one of state and enabled are required.)
+    masked:
+        required: false
+        choices: [ "yes", "no" ]
+        default: null
+        description:
+            - Whether the unit should be masked or not, a masked unit is impossible to start.
+    daemon_reload:
+        required: false
+        default: no
+        choices: [ "yes", "no" ]
+        description:
+            - run daemon-reload before doing any other operations, to make sure systemd has read any changes.
+        aliases: ['daemon-reload']
+notes:
+    - One option other than name is required.
+requirements:
+    - A system managed by systemd
+'''
+
+EXAMPLES = '''
+# Example action to start service httpd, if not running
+- systemd: state=started name=httpd
+# Example action to stop service cron on debian, if running
+- systemd: name=cron state=stopped
+# Example action to restart service cron on centos, in all cases, also issue deamon-reload to pick up config changes
+- systemd: state=restarted daemon_reload: yes name=crond
+# Example action to reload service httpd, in all cases
+- systemd: name=httpd state=reloaded
+# Example action to enable service httpd and ensure it is not masked
+- systemd:
+    name: httpd
+    enabled: yes
+    mased: no
+'''
+
+RETURN = '''
+status:
+    description: A dictionary with the key=value pairs returned from `systemctl show`
+    returned: success
+    type: complex
+    sample: {
+            "ActiveEnterTimestamp": "Sun 2016-05-15 18:28:49 EDT",
+            "ActiveEnterTimestampMonotonic": "8135942",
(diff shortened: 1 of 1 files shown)


gregdek:

mscherer:
Should masked be a state rather that a specific verb ?

bcoca:
no, as you can also want to stop it at the same time:

`systemd: name=httpd3 masked=yes state=stopped`

gregdek:
Thanks @bcoca for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

nemesifier:
Hi, is there a way to try this new module?

bcoca:
like all modules, download into a 'library/' dir adjacent to your play or inside a role, be aware that this will only work with Ansible >=2.1. Also, based on feedback, I'll possibly be changing it soon to use shared code which will bump that up to 'current devel'.

system/systemd.py
```diff
@@ -0,0 +1,329 @@
...
+
+    # Enable/disable service startup at boot if requested
+    if module.params['enabled'] is not None:
+        # do we need to enable the service?
+        enabled = False
+        (rc, out, err) = module.run_command("%s is-enabled '%s'" % (systemctl, unit))
+
+        # check systemctl result or if it is a init script
+        initscript = '/etc/init.d/' + unit
+        if rc == 0 or (os.access(initscript, os.X_OK) and bool(glob.glob('/etc/rc?.d/S??' + unit))):
```
evverx, line 309: This doesn't work properly.
See https://github.com/ansible/ansible-modules-core/pull/1909#issue-100800154

bcoca, line 309: it 'works as expected' as it will fail to enable masked services, this module has a specific option do deal with masking.

This is now an explicit option, so the fix for the service module does not really carry over.

nemesifier:
Does this have any relation with the `service` module?

Do I have to alter my playbooks/roles in order to try this? Or is the `daemon-reload` case handled automatically?

system/systemd.py
evverx, line 309: > it will fail to enable masked services

``` sh
$ systemctl is-enabled ssh
masked

$ ls -l /etc/init.d/ssh
-rwxr-xr-x 1 root root 4077 Mar 21 12:06 /etc/init.d/ssh

$ find /etc/rc[0-9]* -name '*ssh*'
/etc/rc2.d/S02ssh
/etc/rc3.d/S02ssh
/etc/rc4.d/S02ssh
/etc/rc5.d/S02ssh
```

So, `if rc == 0 or (os.access(initscript, os.X_OK) and bool(glob.glob('/etc/rc?.d/S??' + unit))):` returns `True`. No?

evverx:
> is the daemon-reload case handled automatically?

```
    daemon_reload:
        required: false
        default: no
        choices: [ "yes", "no" ]
        description:
            - run daemon-reload before doing any other operations, to make sure systemd has read any changes.
        aliases: ['daemon-reload']
```

I think `default: no` is the good choice. See https://youtu.be/wVk-NWtiIZY?t=6m23s

abadger:
We should be doing as little as possible at the toplevel of a module.  The reason is that when we start to unittest modules, the toplevel will be nearly impossible to test.  We can test main(), although not well (it's almost equivalent to treating the module as a blackbox.  But we're able to import the code and monkey patch/mock it if needed).  Even better is to push each unit of functionality into its own function.

Doing that lets us test, for instance, enabling and disabling a service separate from parsing the parameters for the module.

gregdek:

gregdek:

system/systemd.py
```diff
@@ -0,0 +1,329 @@
+#!/usr/bin/python
```
alikins, line 1: ^^^ 
I'd rather not use the 'systemd' name space yet. (Maybe... 'systemctl' or 'systemctl_unit' or 'systemd_service' or 'systemd_unit')

system/systemd.py
```diff
@@ -0,0 +1,329 @@
...
+        default: null
+        description:
+            - Whether the service should start on boot. B(At least one of state and enabled are required.)
+    masked:
+        required: false
+        choices: [ "yes", "no" ]
+        default: null
+        description:
+            - Whether the unit should be masked or not, a masked unit is impossible to start.
+    daemon_reload:
```
alikins, line 53: Could daemon_reload be a separate module? Seems like it could potentially be needed by itself as a task (and/or, for use in a notify handler)

system/systemd.py
```diff
@@ -0,0 +1,329 @@
...
+# Example action to start service httpd, if not running
+- systemd: name=httpd state=started
+# Example action to stop service httpd, if running
+- systemd: name=httpd state=stopped
+# Example action to restart service httpd, in all cases, also issue deamon-reload to pick up config changes
+- systemd: name=httpd state=restarted daemon_reload: yes
+# Example action to reload service httpd, in all cases
+- systemd: name=httpd state=reloaded
+# Example action to enable service httpd
+- systemd: name=httpd enabled=yes
```
alikins: Might as well make the examples parallel the systemctl command, ie 'systemd: state=started name=httpd'.
(The systemctl cli ordering is still weird to me, but probably worth matching it for the examples).

system/systemd.py
```diff
@@ -0,0 +1,329 @@
...
+DOCUMENTATION = '''
+module: systemd
+author:
+    - "Ansible Core Team"
+version_added: "2.2"
+short_description:  Manage services.
+description:
+    - Controls systemd services on remote hosts.
+options:
+    name:
```
alikins, line 29: add alias for 'unit' ?

bcoca, line 29: yes, will add an alias as 'unit', name= is what we use generically in the service module.

system/systemd.py
```diff
@@ -0,0 +1,329 @@
...
+    rc = 0
+    out = err = ''
+    result = {
+        'name':  unit,
+        'changed': False,
+        'status': {},
+    }
+
+    #TODO: check if service exists
+    (rc, out, err) = module.run_command("%s show '%s'" % (systemctl, unit))
```
alikins, line 245: If we are running the cli tools, it may be useful to add a systemd aware 'module.run_command' analog. 
- could be useful to be able to grab the journal entries related to running of the command/unit/pid to potentially include in the return code or debugging info
- single place to specify host/machine/user for --user/--host/--machine options

system/systemd.py
bcoca, line 1: it needs to match the service_mgr fact, which already is systemd, we might still have systemd/systemctl_properties and other modules, but the main perception of systemd is still as a service manager.

system/systemd.py
bcoca, line 53: I don't think it merits a module as it would only ever do 'the one thing', it is here for the case in which the units have been modified (template/copy/etc tasks) and you want to execute the service action on the updated (reloaded) config and it seems to be a logical fit.

system/systemd.py
bcoca: I really have no opinion on this, I should also probably add pure YAML invocations as others really don't like the k=v syntax ...

system/systemd.py
bcoca, line 245: I'll leave that to future refactoring, I don't really see a use for other than --user.

gregdek:

grahamrhay:
@bcoca @evverx while I don't want daemon-reload to run more often than necessary, I would also prefer not to have to explicitly ask for it either. systemd does tell you when it's necessary:

```
Warning: Unit file changed on disk, 'systemctl daemon-reload' recommended.
```

bcoca:
@grahamrhay yes, and I even had code to do that, but it is still far from optimal. There are a few scenarios in which you want to avoid this, specially if you are changing multiple unit files and don't want to incur in the performance penalty of constantly reloading. 

Other CM tools had this by default and have been backtracking as not only is there a performance cost, the system is unresponsive to events while reloading. Check the video above, even if they fix many of the issues at the systemd level, we will keep this default as not all distros will be able upgrade to a version of systemd that mitigates these issues.

grahamrhay:
hmm. you may be right, but without that you're not really giving me anything that I can't do already with a handler (i.e. the current workaround). I still have to keep track of which unit files I have changed myself.

gregdek:
Thanks @bcoca for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

xrow:
Hi,

can someone test if this test case inside docker will work for 2.2.0 Milestone? We had issues with earlier versions.

https://gitlab.com/dennisxrow/ansible-service-testcase

linusgg:
Hi,
I am having issue on ansible 2.2.0 (devel 0430923647) last updated 2016/07/05 14:46:17 (GMT +1000)
The daemon-reload directive didn't seem to run before the actual action.
After I added a new service file, run this play
`- name: test | make sure activemq starts on boot
      systemd:
        daemon_reload: yes
        name: unipay-activemq
        enabled: yes
        state: started`
It gives me {"changed": false, "failed": true, "msg": "Could not find the requested service \"'unipay-activemq'\": "}

But if I do a work around to make sure daemon-reload is called before actual action,
`- name: test | reload daemon
      systemd:
        daemon_reload: yes
        name: sshd

    - name: test | make sure activemq starts on boot
      systemd:
        daemon_reload: yes
        name: unipay-activemq
        enabled: yes
        state: started`

Then it works. 
Could anyone have  a look at this?

runcom:
@bocca there's no way to get the systemd module to work with "systemd --user" services,unit,timers etc etc

PR #3667, opened 2016-05-16
Add several modules to py3 compile test since they are already py3 ready

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

group, selinux, setup
##### SUMMARY

Since we test python3 syntax for ping after being ported, we should as well do it for the module who are already ok being compiled on python3.


gregdek:

mscherer:
i guess this should also be of @abadger interest

abadger:
Thanks @mscherer!

Issue #864, opened 2015-02-27
script module crashes with "Exception while splitting arguments"


bitmax1:
When calling a windows powershell script, that sends backslashes to std_err script module crashes with exception, caused by the call of vvvv because split_args can't handle the input.

Sample windows output:
Der Pfad 
"HKLM:\SOFTWARE\Policies\Microsoft\Windows\RemovableStorageDevices" kann nicht 
gefunden werden, da er nicht vorhanden ist.
In C:\Users\Administrator\AppData\Local\Temp\ansible-tmp-1425043462.98-71560949
861849\enableRDPcdromAccess.ps1:7 Zeichen:1
- set-ItemProperty -Path 
  'HKLM:\SOFTWARE\Policies\Microsoft\Windows\RemovableStora ...
- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
    + CategoryInfo          : ObjectNotFound: (HKLM:\SOFTWARE\...eStorageDevic 
   es:String) [Set-ItemProperty], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.SetIt 
   emPropertyCommand

```

cchurch:
Could you provide an example playbook and PowerShell script to help us reproduce?

bitmax1:
Sorry, can't reproduce it anymore by myself either.
Having changed line 153 of ansible/lib/ansible/runner/connection_plugins/winrm.py to 
<code>return (result.status_code, '', result.std_out.decode('cp437').encode('utf-8'), result.std_err.decode('cp437').encode('utf-8'))</code>
to come around the problems with the german umlauts, the problem with the output above mentioned seems to vanished again.
Next time something like this happens, i'll freeze my machine (i promise)

bitmax1:
Got the error again, this time reproducable.

In conjunction with a windows machine:

On the windows machine place file test.txt in c:\
Content:

<pre>
C:\Users\Administrator\AppData\Local\Temp\ansible-tmp-1426092743.23-120099741664982
C:\Users\Administrator\AppData\Local\Temp\ansible-tmp-1426092743.23-120099741664982
 "choco.exe install stunnel -source https://chocolatey.org/api/v2/",
[...]

PR #3693, opened 2016-05-18
ops_template.py: Change integer key into stirng

keithnoguchi:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ops_template
##### ANSIBLE VERSION

```
ansible-playbook 2.2.0
  config file = /usr/local/git/dc-on-docker/ansible.cfg
```
##### SUMMARY

ops.dc.read(), which is called by openswitch.py, returns the current OVSDB state in Python dictionary
format.  Due to some keys in our schema defined as integer, we use "integer" type for those of the
keys. e.g. BGP AS number as shown in our schema, line number #864:

https://github.com/open-switch/ops/blob/master/schema/vswitch.extschema#L864

Without this fix, we get exception inside ops_template, when we convert python dictionary into
JSON format, as JSON's keys are string.  We can fix the issue in ops.dc() side to convert all
the key index to string, but fixing inside ops_template.py will minimize the impact at this stage.

We can come back if we need to fix it either openswitch.py or ops.dc() itself, later.

Here is the simple test, with the latest ansible with success:

https://travis-ci.org/keinohguchi/ops-bgp-role/jobs/131008821

This fix passing the update variable to the str()
so that it avoids the exception when ops.dc.read()
returns a dictionary which contains non-string keys.

This is due to the fact that some of the key types in
OpenSwitch schema are actually defined as integer
and ops.dc declerative config module encode those
in integer inside the dictionary.  This could be
the right encoding from the schema point of view
but someone needs to convert it to the string
somewhere, as JSON key should be string.

### DIFF
--- a/network/openswitch/ops_template.py
+++ b/network/openswitch/ops_template.py
@@ -195,7 +195,7 @@ def main():
         updates = dict()
         for path, key, new_value, old_value in changeset:
             path = '%s.%s' % ('.'.join(path), key)
-            updates[path] = new_value
+            updates[path] = str(new_value)
         result['updates'] = updates
 
         if changeset:


keithnoguchi:
Thank you @privateip for your support!  As we chat over IRC, here is the PR for the fix in ops_template.py.

Thank you for your support!

Cheers,

PR #3697, opened 2016-05-18
Make sure users know where to find the full docker picture.

chouseknecht:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

docker.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (doc-updates 22de8a9a21) last updated 2016/05/18 14:31:59 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 92bf802cb8) last updated 2016/05/18 14:11:02 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD e710dc47fe) last updated 2016/05/18 06:08:06 (GMT -400)
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### SUMMARY

Adding a link to Getting Started with Docker guide.

PR #3698, opened 2016-05-18
Py3 exclude list

abadger:
##### ISSUE TYPE
- Feature Pull Request
- Bugfix Pull Request
##### COMPONENT NAME

azure modules, linnode module, and a couple windows modules.  travis test configuratoin.
##### ANSIBLE VERSION

```
2.1
```
##### SUMMARY

A few changes to the azure, linnode, and windows (stub) modules to make them work under python3.  Also change the travis test configuriation from a whitelist of modules to test under python3 to a blacklist.  This was a prerequisite for adding guidelines that all new modules need to be syntax compatible with python3 (running under python3 will come at some later date).


willthames:
I note from the azure module that you're using

`except AzureException as e`

which is fine for Azure as it's never going to run with python 2.4, but wouldn't be something that could be done with modules that do need to run on 2.4. 

Do we have any guidelines for what modules need to work with python 2.4 and what modules need to work with python3 (it's a lot easier to make python 2.7 modules compatible with python 3)

sivel:
Something like:

```
from ansible.module_utils.basic import get_exception
...
try:
    ...
except IOError:
    e = get_exception()
```

abadger:
I think eventually all modules are going to need to be syntax compatible with Python3.  However, not all of them will run with python3.  I think we're still absorbing what that might mean so there's not a definitive answer to what that means.  One possiblitiy is that a module like yum where the underlying python module is never going to be ported to python3 will still run via python3 up to the point where it looks for the yum library and calls fail_json because the library does not exist for python3.

The end goal is for every module whose libraries run on python3 to also run on python3.  However, that goal is a ways off.  For the next release (2.2) we're definitely not going to be that advanced but we should have enough done that we can kinda estimate what timeframe we're looking at for the rest of the code to work.

The ruling about python2.4 is that as long as an ansible module's dependent libraries support python2.4 then the module needs to support python-2.4 as well.

In terms of how to support python2.4 and python3 for exceptions, sivel's recipe will work.  I've been trying to get people to use get_exception from pycompat24 instead of from basic so that we'll have an easier time getting rid of it later, when we are to drop python2.4 compatibility.  The recipe for that would look like:

``` python
from ansible.module_utils.pycompat24 import get_exception
[...]
try:
    raise ValueError('Example exceptin')
except ValueError:
    e = get_exception()
   module.fail_json(msg=str(e))
```

I added a document where we can put information on porting code to work on both python2.4 and python3 here: https://github.com/ansible/ansible/blob/devel/docsite/rst/developing_modules_python3.rst

gregdek:
Thanks @abadger for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

willthames:
Looks like #3697 broke this.

PR #3702, opened 2016-05-19
Convert packaging/* to py3/py2.4 syntax

mscherer:
##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/packaging/os/rhn_register.py
+++ b/packaging/os/rhn_register.py
@@ -120,7 +120,7 @@ try:
     import up2date_client
     import up2date_client.config
     HAS_UP2DATE_CLIENT = True
-except ImportError, e:
+except ImportError:
     HAS_UP2DATE_CLIENT = False
 
 # INSERT REDHAT SNIPPETS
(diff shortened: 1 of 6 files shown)


gregdek:

packaging/os/rhn_register.py
```diff
@@ -119,7 +119,8 @@
 try:
     import up2date_client
     import up2date_client.config
-except ImportError, e:
+except ImportError:
+    e = get_exception()
```
abadger: I think we may have to move the ansible.module_utils imports up above this otherwise get_exception won't be defined yet.  Could convert the ansible.module_utils imports to not be wildcard imports at the same time.

abadger:
One comment for a thing that I think will throw an error if not changed a little bit.  I'm also going to be merging the travis.yml change from a whitelist to a blacklist soon https://github.com/ansible/ansible-modules-core/pull/3698  so you can convert the travis portion of htis change to using that as well.

packaging/os/rhn_register.py
mscherer: Yeah, I was wondering if it would work or not when I wrote the patch, but didn't dig much. Another solution is just to not get the exception in a variable and just show a simple generic error message.

mscherer:
Since sivel fixed the rhn_register code in https://github.com/ansible/ansible-modules-core/commit/1d87213de17e94c793e2850203aada5c339d850b,  the issue of order is also fixed, so I went ahead and rebased on top of the blacklist changes.

abadger:
Thanks @mscherer!  merged.

PR #3707, opened 2016-05-19
Fixing module description.

chouseknecht:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

docker.py 
##### ANSIBLE VERSION

```
ansible 2.1.0.0 (stable-2.1 eae26891ea) last updated 2016/05/18 15:12:42 (GMT -400)
  lib/ansible/modules/core: (stable-2.1 742ac22507) last updated 2016/05/18 15:16:44 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD d25274a958) last updated 2016/05/18 15:13:03 (GMT -400)
```
##### SUMMARY

Fixing module description.


sivel:
Just to provide some context here.  This error seems to be coming from calling `ansible.utils.module_docs.get_docstring`.  One main thing to note is that we prefer the C YAML parser if available, and on travis it is available.  So the parsing error seems to be a more strict parsing of this portion of the docs by that C YAML parser, as on my mac I only have the python YAML parser available, and it passes there.

chouseknecht:
Tried to reproduce on a Centos 7 vagrant box with libyaml and pyyaml installed. It passes fine. Not sure how to reproduce.

sivel:
I cannot either.  In fact I reverted all the way back to before your changes, and it still fails.

abadger:
We figured this out.  Needed to update the submodule refs in ansible/ansible because ansible-module-validate was parsing the yaml from modules in the submodule refs for ansible/ansible to tell if any new parameters had been added.  Since that old version of hte modules had the error, that parsing was failing.

PR #3713, opened 2016-05-20
Detect if systemd is active on Gentoo Linux

### LINKED ISSUE #3360: Hostname not set correctly on Gentoo Linux with systemd
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

hostname
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = 
```
##### OS / ENVIRONMENT

Host OS: N/A
Target OS: Gentoo Linux
##### SUMMARY

The hostname task only updates /etc/conf.d/hostname, no difference what init are using: OpenRC or systemd.

This pull request can be relevant to this bug https://github.com/ansible/ansible-modules-core/pull/2341

See also issue https://github.com/ansible/ansible-modules-core/issues/573
##### STEPS TO REPRODUCE
1. Set hostname using the hostname task on target Gentoo Linux with systemd.
2. Check /etc/hostname
3. Check /etc/conf.d/hostname
##### EXPECTED RESULTS

If target OS is Gentoo Linux with systemd then SystemdStrategy should be used, not OpenRCStrategy. Module should works as expected and should checks if systemd is active, like in module service.py
##### ACTUAL RESULTS

Actually only /etc/conf.d/hostname changed.

anonymous09:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

hostname
##### ANSIBLE VERSION

```
ansible 2.0.1.0
  config file = 
```
##### SUMMARY

Fixes #3360 
Read comment in code.


bcoca:
Your detection is very version dependent, for code that works across most systemd versions look at: 
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts.py#L512

anonymous09:
Thank you for answer.

I think that older versions of systemd we can ignore because the oldest version that supported by Gentoo is systemd-218.

You tell about changes https://github.com/systemd/systemd/commit/3a90ae0 in systemd-28 (2011 year!). This version is abandoned and absolutely unsupported now. What actual distributives use systemd of version 27 or earlier now?

Are we really need old unsupported crap?

bcoca:
sadly, unsupported != not in use, we arrived to the current detection function mostly from user feedback.

anonymous09:
Well, so can I directly use _check_systemd() (https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/facts.py#L512)?

Whether it is possible to use the functions that begin with an underscore? They are only for internal using, aren't they?

bcoca:
I'll change that as we already need it from other places

anonymous09:
Ok, let me know when you change it, please.

bcoca:
done in http://github.com/ansible/ansible/commit/c8f0cdbdfd1754219885287666d594f6fcd1993d

anonymous09:
Unfortunately, I can't understand how properly call is_systemd_managed(). Can you help me, please?
It's doesn't work:
`Facts(Hostname, load_on_init = False).is_systemd_managed()`

anonymous09:
ping?

bcoca:
can you define "doesn't work"? seems to be fine for my tests.

anonymous09:
It does not matter now, because we have another pretty nice solution that already merged by upstream.

https://github.com/ansible/ansible-modules-core/pull/4382

PR #3717, opened 2016-05-20
Add new OpenStack module for changing metadata of compute instances

ruizink:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

cloud/openstack/os_server_metadata.py
##### ANSIBLE VERSION

```
ansible 2.0.0.0
```
##### SUMMARY

This change adds a new OpenStack module to allow the update of the metadata on compute instances.

Without this module, the only way to change the metadata of a compute instance on OpenStack is during creation using the `os_server` ansible module. This implies that we can no longer update the metadata of a running instance.

This module depends on a feature currently PR'ed on [OpenStack shade repo](https://review.openstack.org/#/c/319395)

### DIFF
--- a/cloud/openstack/os_server_metadata.py
+++ b/cloud/openstack/os_server_metadata.py
@@ -104,6 +104,7 @@ metadata:
     sample: {'key1': 'value1', 'key2': 'value2'}
 '''
 
+
 def _needs_update(server_metadata={}, metadata={}):
     return len(set(metadata.items()) - set(server_metadata.items())) != 0
 
@@ -150,15 +151,14 @@ def main():
             # check if it needs update
             if _needs_update(server_metadata=server.metadata,
                              metadata=meta_param):
-                cloud.server_set_metadata(server_param,
-                                          meta_param)
+                cloud.set_server_metadata(server_param, meta_param)
                 changed = True
         elif state == 'absent':
             # remove from params the keys that do not exist in the server
             keys_to_delete = _get_keys_to_delete(server.metadata.keys(),
                                                  meta_param.keys())
             if len(keys_to_delete) > 0:
-                cloud.server_delete_metadata(server_param, keys_to_delete)
+                cloud.delete_server_metadata(server_param, keys_to_delete)
                 changed = True
 
         if changed:
--- a/cloud/openstack/os_server_metadata.py
+++ b/cloud/openstack/os_server_metadata.py
@@ -31,7 +31,7 @@ module: os_server_metadata
 short_description: Add/Update/Delete Metadata in Compute Instances from
 OpenStack extends_documentation_fragment: openstack
 version_added: "2.2"
-author: "Mario Santos (@_RuiZinK_)"
+author: "Mario Santos (@ruizink)"
 description:
    - Add, Update or Remove metadata in compute instances from OpenStack.
 options:
@@ -91,6 +91,18 @@ EXAMPLES = '''
             group:
 '''
 
+RETURN = '''
+server_id:
+    description: The compute instance id where the change was made
+    returned: success
+    type: string
+    sample: "324c4e91-3e03-4f62-9a4d-06119a8a8d16"
+metadata:
+    description: The metadata of compute instance after the change
+    returned: success
+    type: dict
+    sample: {'key1': 'value1', 'key2': 'value2'}
+'''
 
 def _needs_update(server_metadata={}, metadata={}):
     return len(set(metadata.items()) - set(server_metadata.items())) != 0
--- a/cloud/openstack/os_server_metadata.py
+++ b/cloud/openstack/os_server_metadata.py
@@ -1,0 +1,167 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+# Copyright (c) 2016, Mario Santos <mario.rf.santos@gmail.com>
+#
+# This module is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This software is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @ruizink for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

ruizink:
ready_for_review

ruizink:
Yay, green!
It's now ready_for_review
Thanks

gregdek:
Thanks @ruizink. To the current maintainers, @emonty, @shrews, @juliakreger, @j2sol, @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

rcarrillocruz:
Honestly, the fact that metadata is not idempotent on os_server is a bug to me.
I rather fix that on os_server, rather than having an specific module just for this.

I'd like to hear @Shrews and @emonty opinions.

ruizink:
I really understand what you're saying @rcarrillocruz, I guess I wouldn't felt the need to write this module if I was able to do it with the os_server in the first place. But even if the os_server was idempotent, changing the metadata would be done as a whole set, risking overriding metadata set externally. You would also need to know some details of the instance prior to calling the module (all of the required parameters basically - image, flavor, nics). Imagine the case where a given instance was created by an external system. It would get a bit more complex to gather all the instance details and probably would force us to use dynamic inventory.  
This module is a bit more flexible, because you can provide the key/values to set or delete, rather than the whole set.
This module is more user friendly as well, because you only need to provide the instance name, and a small subset of key/values.
I hope you understand my point.
Thanks

jvale:
I completely agree @rcarrillocruz. However I think @ruizink's feature also makes perfect sense in addition to os_server's idempotency as it covers a different and valid use case.

rcarrillocruz:
I agree this is useful if a user wants to manage metadata without managing the whole server resource, but I don't think we should have such specific use cases modules.

We have two patterns in the OpenStack modules:

os_&lt;resource name&gt;
os_&lt;resource name&gt;_actions

The primer is to create/update resources, the latter for one-off actions on resources.
The use case you are addressing with this PR makes sense to me to put it on os_server_actions:

http://docs.ansible.com/ansible/os_server_actions_module.html

You could create a new 'verb' on that module called like 'set_metadata' and pass on the key/values you want to set.

Of course, I still think metadata should be idempotent on os_server which I'll try to fix myself in other PR.

Thanks

ruizink:
Looking at the `os_server_actions` code, we can clearly see that is completely targeted to manage the server state. It wouldn't surprise me that the module was called `os_server_state` rather than `os_server_actions`.
Although I understand and agree with the two pattern approach, l think that in order to allow for expandability the original module had to be less specific to system state and let room for other actions to fit in.
Of course this can be done in several ways, but going for the reuse of the current `os_server_actions` module to support other actions will have a negative impact on: module usability, code readability as well as code maintainability.
Thanks

Shrews:
Thanks for the discussion!

Here are my thoughts (for whatever they are worth)...

I totally get the need for this module. And, like @rcarrillocruz, I too have concerns about setting a standard for having one module to update a single value of a resource. I REALLY don't want to see that become a precedent. Ideally, the os_server module would be able to handle this since that makes the most sense. But having looked at making that module handle updates myself, I know that is a herculean task wrought with perils and dragons.

My initial thoughts aligned with Ricky to move this to the actions module, but I also agree with @ruizink that doing so would be an awkward fit as it would then move beyond its purpose of managing server state.

How about this for a crazy idea... How about we make a new 'action' value for os_server, maybe 'update', that required only the server name and the field(s) to update. Obviously we could only update certain fields (like, we couldn't update flavor w/o deleting the server). I worry a bit about idempotency with this as I only just came up with it and haven't fully thought it through. What are the community's thoughts here? cc: @emonty and @j2sol

bcoca:
I would leave that as state=present and infer that you need an update depending on other options supplied.

nicolasbarbe:
> I know that is a herculean task wrought with perils and dragons.

@Shrews could you please elaborate on this? is it a matter of effort or risk (regression)? 

I would love to see `os_server` working as expected, from my point of view, its implementation is not complete. Looking at the code, the case of updating a server is not managed ... so I expect to have some similar issues with the other properties, is that right?

Shrews:
@nicolasbarbe Been a LOOONG time since I looked at this, but, as I recall, making os_server understand an update to an existing server means comparing MANY things about the existing server and the options supplied, and seeing which values are different and updating the server (or any of it's associated attributes) appropriately. Obviously, some things could not be updated (e.g., flavor or ram). And if something like the list of volumes changes, that adds more complexity to os_server.

Not saying it's impossible, it will just be very difficult to get right. I'm not undertaking that task, but certainly won't deter anyone from the community from doing it.

ruizink:
If you want to follow @rcarrillocruz 's first suggestion above, that is to extend the `os_server_actions` module, I have a [possible implementation](https://github.com/ruizink/ansible-modules-core/blob/feature_os_server_actions/cloud/openstack/os_server_actions.py) in my own fork that adds new action verbs to the module. `_metadata_actions = ['set_metadata', 'delete_metadata']`

I can easily create a new PR with that solution and close this one if you all agree.

nicolasbarbe:
@Shrews, I understand now better the kind of complexity you were referring. Can we just implement this step by step, property per property, starting with the metadata, and updating the documentation accordingly. The task will be easier to tackle and less subject to regression. It can only improve the current situation, no?

@ruizink I am not a big fan of adding this to `os_server_action` ... It does not feel super natural to perform a _server action_ (similar to start / stop) to change metadata. It seems to me not really consistent and I would prefer to fix os_server in the first place, but this is just my understanding :smile:

Shrews:
@nicolasbarbe I don't think property-by-property is a good idea. It could leave the module in an incomplete state if no one commits to finishing it. You might could do the most common properties folks would want to update in a single pass, then the less common ones could be added later. But we'd have to see what that looks like then solicit feedback from the community.

ruizink:
Migrated to https://github.com/ansible/ansible/pull/19113

Issue #3734, opened 2016-05-25
File module: Allow 'compress' state


bendoh:
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

File
##### ANSIBLE VERSION

```
ansible 2.0.0.2
  config file =
  configured module search path = Default w/o overrides
```
##### OS / ENVIRONMENT

N/A
##### SUMMARY

There is no obvious way to ensure a file is compressed. For my particular use case, I want to keep a sql dump file on hand, but make sure it's always compressed. 

The `file` module seems like a good place to add this, by adding a 'compress' state, which would be semantically equivalent to the 'file' state, except the file would be kept compressed using an additional `compression` argument, which would default to 'gzip'. The `compression` argument could be influenced by an environment variable or top-level variable, whatever is felt to be appropriate.
##### STEPS TO REPRODUCE

``` yaml
- name: Ensure file is compressed
  file: path=/the/target-file state=compress
```
##### EXPECTED RESULTS

The file at `/the/target-file` would be compressed and become `/the/target-file.{gz,zip}`, depending on the compression used.

abadger:
No need to have both an issue and a PR so going to go ahead and close this in favor of #3735

PR #3735, opened 2016-05-25
Accept 'compress' state to file module

bendoh:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

file
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file =
```
##### SUMMARY

There is no obvious way to ensure a file is compressed. For my particular use case, I want to keep a sql dump file on hand, but make sure it's always compressed because they are eminently compressible and should be maintained that way.

Addresses #3734, filed earlier.

Sample playbook:

``` yaml

---
- hosts: all
  tasks:
  - name: Test file compress
    file: path=/Users/bendoh/vc/ansible-tests/myfile state=compress
```

``` sh
% ls -l ~/vc/ansible-tests/files
total 16
-rw-r--r-- 1 bendoh staff 12324 May 24 22:44 myfile

% ansible-playbook -i localhost, playbook.yml

PLAY [all] *********************************************************************

ok: [localhost]

TASK [Test file compress] ******************************************************
changed: [localhost]

localhost                  : ok=2    changed=1    unreachable=0    failed=0

% ls -l ~/vc/ansible-tests/files
total 4
-rw-r--r-- 1 bendoh staff 119 May 24 22:56 myfile.gz

% ansible-playbook -i localhost, playbook.yml

PLAY [all] *********************************************************************

ok: [localhost]

TASK [Test file compress] ******************************************************
ok: [localhost]

localhost                  : ok=2    changed=0    unreachable=0    failed=0

```

### DIFF
--- a/files/file.py
+++ b/files/file.py
@@ -300,8 +300,11 @@ def main():
                 # Decompress
                 if compression == 'gzip':
                     try:
-                        with gzip.open(compress_path) as f_in, open(path, 'w') as f_out:
-                            shutil.copyfileobj(f_in, f_out) 
+                        f_in = gzip.open(compress_path)
+                        f_out = open(path, 'w')
+                        shutil.copyfileobj(f_in, f_out)
+                        f_in.close()
+                        f_out.close()
                     except OSError:
                         e = get_exception
                         module.fail_json(path=path, msg='Unable to write to compressed file: %s' % str(e))
@@ -484,8 +487,11 @@ def main():
                 # Compress the file
                 if compression == 'gzip':
                     try:
-                        with open(path) as f_in, gzip.open(compress_path, 'w') as f_out:
-                            shutil.copyfileobj(f_in, f_out) 
+                        f_in = open(path)
+                        f_out = gzip.open(compress_path, 'w')
+                        shutil.copyfileobj(f_in, f_out)
+                        f_in.close()
+                        f_out.close()
                     except OSError:
                         e = get_exception()
                         module.fail_json(path=path, msg='Error, could not write compressed file: %s' % str(e))
--- a/files/file.py
+++ b/files/file.py
@@ -22,6 +22,7 @@ import errno
 import shutil
 import stat
 import grp
+import gzip
 import pwd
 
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @bendoh for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

abadger:
I'm not sure we want this, or at least, not sure we want it as part of file.  file is largely about metadata surrounding the file rather than contents (there's a few subcommands that do a little with content but we frequently end up debating whether those were mistakes or not).

Some of the issues this particular implementation raises are what the filename should be (I think it we did implement it, the filename should be what the user specifies in the playbook, no suffix should be appended).  Should one be able to compress a compressed file (this is more of an issue once multiple compression types are implemented).  Do we need to read file magic to tell if a file has already been compressed (since we can't rely on suffix)?  What should the behaviour be if state=compress is specified with a directory?  With a directory and recurse?

I'm not going to close this immediately but if another core committer doesn't step forward to defend this idea we probably will end up closing it later.  it might be better to create a new module that compresses files idempotently for extras rather than adding it to ifle.

bendoh:
You raise some very valid points, and after sleeping on this, I still think that this functionality could be useful, but also that it belongs more in an 'archive' or 'compress' module of its own, which could be more feature-complete and indeed support directories. Having a module of its own would also avoid the weird logic associated with what happens when the previous state of a file is 'hard' or 'link': the module would only operate on plain files (and, probably, trees,) and be able to deal with its own corner cases more effectively.

Is there any particular reason 'unarchive' doesn't have an 'archive' equivalent? That seems like a natural place to put this sort of logic, and it seems like a pretty common use case where users would otherwise turn to a shell command to achieve the same thing.

bcoca:
The reason there is no 'archive' yet is that no one has written one. There has not been much demand for one either, though a couple of times people have been talking about submitting one.

As it stands I'm going to close the ticket as I think we can all agree file is not the correct place for this feature.

PR #3767, opened 2016-05-27
Revert PR #3575 since it causes problems related to exclude patterns

### LINKED ISSUE #11348: bug in unarchive module

dagwieers:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

unarchive
##### SUMMARY

By using a different method for getting archive filelists, and extracting we introduced new problems related to excluding based on gtar patterns.

As a result files that would be excluded by gtar, would still be in the filelist. Implementing our own gtar compatible pattern exclusion mechanism is near to impossible (believe me, we looked at it...). The best way is to look at the original problem and deal with that, and ensure that extraction and filelists are done with the exact same tool and exact same options.

The solution is to decode the octal unicode representation in gtar's output back to unicode. Since gtar has no problem extracting these files in LANG=C, we simply has to compensate for it.

This reverts #3575 and fixes ansible/ansible#11348.


dagwieers:
@hsoft Please test. It works for the original reproducer that was shared in ansible/ansible#11348.

ghost:
As I write in https://github.com/ansible/ansible-modules-core/pull/3575#issuecomment-222148311 , I confirm that your approach works. It's much better this way!

dagwieers:
@hsoft Thanks for verifying !

shipit

gregdek:
Thanks @dagwieers. Since you are one of the maintainers (@dagwieers, @pileofrogs) of this module and you commented with "shipit", we are marking this PR for inclusion.

files/unarchive.py
```diff
@@ -539,23 +536,24 @@ def __init__(self, src, dest, file_args, module):
...
+            cmd += ' --exclude="' + '" --exclude="'.join(self.excludes) + '"'
+        cmd += ' -f "%s"' % self.src
+        rc, out, err = self.module.run_command(cmd)
+        if rc != 0:
             raise UnarchiveError('Unable to list files in the archive')
 
-        for filename in tf.getnames():
+        for filename in out.splitlines():
+            # Compensate for locale-related problems in gtar output (octal unicode representation) #11348
+            filename = filename.decode('string_escape')
```
abadger: We can't use string_escape as there's no way to make it compatible with python3.  There may also be encoding issues as string_escape always uses latin1 as its encoding even though the string will usually be utf-8 but I'm not certain about that.  I started looking into this and finally found the "unicode_escape doesn't work in general" answer to http://stackoverflow.com/questions/4020539/process-escape-sequences-in-a-string-in-python  We had made a variation on that work in the controller code.  I struggled to adapt it to this situation for a bit before #3575 was submitted.  That seemed like an easier to understand solution than fighting with a finicky regex so I abandoned that effort when we merged the tarfile solution.

abadger:
needs_revision  See the inline note about string_escape.

dagwieers:
It is quite ironic that you have concerns about unicode support while you
are forcing everything to the C locale anyway.

dagwieers:
We can use `codecs.escape_decode()` instead of `decode('string_decode')`, which would work on python3 as well. Or simply test the availability and use either `string_decode` or `unicode_decode`.

Strange that it passed all Travis checks if it doesn't work on Python 3 ?

abadger:
I remember that unicode_decode doesn't work.  In fact, I think that's the examples that the author of "unicode_decode" provides is showing.  I'm not sure if codecs.escape_decode() didn't work... I know that I tried that but I can't remember if there was something wrong with it (maybe about it taking bytes but we have text?  or vice versa?) or if it was just getting very complex and then #3575 was presented.

abadger:
Also... have to be sure to test with filenames that do not have a representation in latin-1 (For instance,  くらとみ)  I beleive those will get mangled by codecs.escape_decode().   Unmangling them is probably where things started getting complex.  (I think they'll need to be unmangled in order to handle creates and exclude).

ghost:
Since our problem is strictly with unescaping octal representations and has nothing to do with unicode, an option could be to write (or steal from Python's code) that function ourselves and apply it. Here's me messing up with the interpreter, coming to the conclusion that writing such a function, when strictly restricted to octal escape parsing, is rather simple and safe:

```
>>> s = '\\303\\251.txt'
>>> import re
>>> re.findall(r'\\[0-7]{3}', s)
['\\303', '\\251']
>>> 0303
195
>>> int('303', base=8)
195
>>> f = lambda match: chr(int(match.group(0)[1:], base=8))
>>> re.sub(r'\\[0-7]{3}', f, s)
'\xc3\xa9.txt'
```

dagwieers:
@hsoft That's fine for me as well, but that's basically what I assume the `string_escape` codec did. Things that were not escaped, would not be decoded either. I don't see how any other byte would be decoded by `string_escape` or `unicode_escape`. But then again, I am not familiar with python 3 and not a great expert on python internals. (Got lost looking at cpython code to find out what it really does...)

dagwieers:
BTW @abadger  on Python 2.6 everything looks fine using `string_escape` or `escape_decode`.

```
[dag@moria ~]$ python -c 'print "\123èé\303\251くらとみ".decode("string_escape")'
Sèééくらとみ
[dag@moria ~]$ python -c 'import codecs; print codecs.escape_decode("\123èé\303\251くらとみ")[0]'
Sèééくらとみ
```

ghost:
For reference, the actual implementation of `escape_decode()` is at https://github.com/python/cpython/blob/9f2531ccbcc014f4bd1c41e51b9a06fc4a612065/Objects/bytesobject.c#L1108 (on python 3).

abadger:
@dag <nod>  I'm pretty sure it's python3 where thing fall apart... Note, though, that I think your test is off... We're getting octal escaped values.  So it should be more like this:

```
path='/var/tmp/S\303\250\303\251\303\251\343\201\217\343\202\211\343\201\250\343\201\277'
import codecs
de_escaped = codecs.escape_decode(path)[0]
print(repr(de_escaped))
print(de_escaped)
final = de_escaped.decode('utf-8')
print(repr(final))
print(final)
```

On python2, this is fine and yields the string "/var/tmp/Sèééくらとみ"

On python3 this messes up and gives us mangled output: "/var/tmp/SÃ¨Ã©Ã©ããã¨ã¿"

abadger:
Actually, my test is messed up too.... We have to use double backslashes in the path string in order to simulate this properly.

abadger:
```
path=b'/var/tmp/S\\303\\250\\303\\251\\303\\251\\343\\201\\217\\343\\202\\211\\343\\201\\250\\343\\201\\277'
import codecs
de_escaped = codecs.escape_decode(path)[0]
print(repr(de_escaped))
print(de_escaped)
final = de_escaped.decode('utf-8')
print(repr(final))
print(final)
```

codecs.escape_decode seems to work there.  So the only problem with using it is that it may go away in the future (upstream doesn't think that conversions from bytes => bytes or text => text are okay codecs.  escape_decode is undocumented because they're thinking of getting rid of it.)

abadger:
@dag I'm going to be on a couple days of vacation (Thursday and Friday).  Once string_escape is switched out with escape_decode, you can let any of the other committers know that this change looks fine to me.

abadger:
Looks good.  +1 to merge.

abadger:
Thanks @dag.  Merged to devel and stable-2.1

PR #3777, opened 2016-05-28
win_setup: Add missing properties

dagwieers:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

windows setup.ps1
##### ANSIBLE VERSION

v2.1
##### SUMMARY

We add some basic properties for Windows that are available on other platforms.

This improves ansible/ansible#15506 a lot.

### DIFF
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -108,7 +108,7 @@ Set-Attr $result.ansible_facts "ansible_system_description" ([string] $win32_os.
 Set-Attr $result.ansible_facts "ansible_system_vendor" $win32_cs.Manufacturer
 Set-Attr $result.ansible_facts "ansible_os_family" "Windows"
 Set-Attr $result.ansible_facts "ansible_os_name" ($win32_os.Name.Split('|')[0]).Trim()
-Set-Attr $result.ansible_facts "ansible_distribution" $osversion.VersionString
+Set-Attr $result.ansible_facts "ansible_distribution" $win32_os.Caption
 Set-Attr $result.ansible_facts "ansible_distribution_version" $osversion.Version.ToString()
 Set-Attr $result.ansible_facts "ansible_distribution_major_version" $osversion.Version.Major.ToString()
 Set-Attr $result.ansible_facts "ansible_kernel" $osversion.Version.ToString()
@@ -127,8 +127,6 @@ Set-Attr $result.ansible_facts "ansible_user_id" $env:username
 Set-Attr $result.ansible_facts "ansible_user_uid" ([int] $user.User.Value.Substring(42))
 Set-Attr $result.ansible_facts "ansible_user_sid" $user.User.Value
 
-# Use English locale
-$culture = New-Object System.Globalization.CultureInfo('en-US')
 $date = New-Object psobject
 $datetime = (Get-Date)
 $datetime_utc = $datetime.ToUniversalTime()
@@ -146,7 +144,8 @@ Set-Attr $date "second" $datetime.ToString("ss")
 Set-Attr $date "time" $datetime.ToString("HH:mm:ss")
 Set-Attr $date "tz_offset" $datetime.ToString("zzzz")
 Set-Attr $date "tz" ([System.TimeZoneInfo]::Local.Id)
-Set-Attr $date "weekday" $datetime.ToString("dddd", $culture)
+# Ensure that the weekday is in English
+Set-Attr $date "weekday" $datetime.ToString("dddd", [System.Globalization.CultureInfo]::InvariantCulture)
 Set-Attr $date "weekday_number" (Get-Date -UFormat "%w")
 Set-Attr $date "weeknumber" (Get-Date -UFormat "%W")
 Set-Attr $date "year" $datetime.ToString("yyyy")
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -83,7 +83,7 @@ Set-Attr $result.ansible_facts "ansible_interfaces" $formattednetcfg
 Set-Attr $result.ansible_facts "ansible_architecture" $win32_os.OSArchitecture
 
 # FIXME: Reformat the date to MM/DD/YEAR format
-Set-Attr $result.ansible_facts "ansible_bios_date" $win32_bios.ReleaseDate
+Set-Attr $result.ansible_facts "ansible_bios_date" $win32_bios.ReleaseDate.ToString("MM/dd/yyyy")
 Set-Attr $result.ansible_facts "ansible_bios_version" $win32_bios.SMBIOSBIOSVersion
 Set-Attr $result.ansible_facts "ansible_hostname" $env:COMPUTERNAME
 Set-Attr $result.ansible_facts "ansible_fqdn" "$([System.Net.Dns]::GetHostByName((hostname)).HostName)"
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -54,6 +54,7 @@ if ($factpath -ne $null) {
 
 $win32_os = Get-CimInstance Win32_OperatingSystem
 $win32_cs = Get-CimInstance Win32_ComputerSystem
+$win32_bios = Get-CimInstance Win32_Bios
 $osversion = [Environment]::OSVersion
 $capacity = $win32_cs.TotalPhysicalMemory # Win32_PhysicalMemory is empty on some virtual platforms
 $netcfg = Get-WmiObject win32_NetworkAdapterConfiguration
@@ -81,13 +82,21 @@ Set-Attr $result.ansible_facts "ansible_interfaces" $formattednetcfg
 
 Set-Attr $result.ansible_facts "ansible_architecture" $win32_os.OSArchitecture
 
-Set-Attr $result.ansible_facts "ansible_hostname" $env:COMPUTERNAME;
+# FIXME: Reformat the date to MM/DD/YEAR format
+Set-Attr $result.ansible_facts "ansible_bios_date" $win32_bios.ReleaseDate
+Set-Attr $result.ansible_facts "ansible_bios_version" $win32_bios.SMBIOSBIOSVersion
+Set-Attr $result.ansible_facts "ansible_hostname" $env:COMPUTERNAME
 Set-Attr $result.ansible_facts "ansible_fqdn" "$([System.Net.Dns]::GetHostByName((hostname)).HostName)"
+Set-Attr $result.ansible_facts "ansible_product_name" $win32_cs.Model
+Set-Attr $result.ansible_facts "ansible_product_serial" $win32_bios.SerialNumber
+Set-Attr $result.ansible_facts "ansible_product_version" $win32_cs.Name
 Set-Attr $result.ansible_facts "ansible_system" $osversion.Platform.ToString()
+Set-Attr $result.ansible_facts "ansible_system_vendor" $win32_cs.Manufacturer
 Set-Attr $result.ansible_facts "ansible_os_family" "Windows"
 Set-Attr $result.ansible_facts "ansible_os_name" ($win32_os.Name.Split('|')[0]).Trim()
 Set-Attr $result.ansible_facts "ansible_distribution" $osversion.VersionString
 Set-Attr $result.ansible_facts "ansible_distribution_version" $osversion.Version.ToString()
+Set-Attr $result.ansible_facts "ansible_windows_domain" $win32_cs.Domain
 
 $date = New-Object psobject
 Set-Attr $date "date" (Get-Date -format d)
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -152,6 +152,7 @@ Set-Attr $date "weeknumber" (Get-Date -UFormat "%W")
 Set-Attr $date "year" $datetime.ToString("yyyy")
 Set-Attr $result.ansible_facts "ansible_date_time" $date
 
+# Win32_PhysicalMemory is empty on some virtual platforms
 Set-Attr $result.ansible_facts "ansible_memtotal_mb" ([math]::round($win32_cs.TotalPhysicalMemory / 1024 / 1024))
 Set-Attr $result.ansible_facts "ansible_swaptotal_mb" ([math]::round($win32_os.TotalSwapSpaceSize / 1024 / 1024))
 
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @dagwieers for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

dagwieers:
Broadcasting @bcoca @jctanner @jimi-c @gregdek Need feedback, this isn't exactly rocket-science, but it _is_ Windows ;-)

windows/setup.ps1
```diff
@@ -81,25 +83,66 @@ Set-Attr $result.ansible_facts "ansible_interfaces" $formattednetcfg
...
-Set-Attr $date "minute" (Get-Date -format mm)
-Set-Attr $date "iso8601" (Get-Date -format s)
+$datetime = (Get-Date)
+$datetime_utc = $datetime.ToUniversalTime()
+Set-Attr $date "date" $datetime.ToString("yyyy-MM-dd")
+Set-Attr $date "day" $datetime.ToString("dd")
+Set-Attr $date "epoch" (Get-Date -UFormat "%s")
+Set-Attr $date "hour" $datetime.ToString("HH")
+Set-Attr $date "iso8601" $datetime_utc.ToString("yyyy-MM-ddTHH:mm:ssZ")
+Set-Attr $date "iso8601_basic" $datetime.ToString("yyyyMMddTHHmmssffffff")
```
jhawkesworth, line 138: is this meant to be $datetime or $datetime_utc - also the same for L130?

windows/setup.ps1
```diff
@@ -155,6 +198,7 @@ if ($winrm_cert_expiry)
 
 $PendingReboot = Get-PendingRebootStatus
 Set-Attr $result.ansible_facts "ansible_reboot_pending" $PendingReboot
+Set-Attr $result.ansible_facts "module_setup" $true
```
jhawkesworth, line 213: I'm curious, what is the purpose of module_setup.  If it is allways $true isn't it a bit redundant?

windows/setup.ps1
dagwieers, line 138: It is meant to mimic the exact behaviour of Linux:

``` json
        "ansible_date_time": {
            "date": "2016-06-09", 
            "day": "09", 
            "epoch": "1465423812", 
            "hour": "00", 
            "iso8601": "2016-06-08T22:10:12Z", 
            "iso8601_basic": "20160609T001012022369", 
            "iso8601_basic_short": "20160609T001012", 
            "iso8601_micro": "2016-06-08T22:10:12.022502Z", 
            "minute": "10", 
            "month": "06", 
            "second": "12", 
            "time": "00:10:12", 
            "tz": "CEST", 
            "tz_offset": "+0200", 
            "weekday": "Thursday", 
            "weekday_number": "4", 
            "weeknumber": "23", 
            "year": "2016"
        }, 
```

jhawkesworth:
cc @nitzmahone  with your windows hat on :-)

windows/setup.ps1
dagwieers, line 213: The aim of this was (as for other facts modules) to test if the module was ran at least once. Just as the python setup module does.

windows/setup.ps1
```diff
@@ -77,29 +81,79 @@ foreach ($adapter in $ActiveNetcfg)
...
+Set-Attr $result.ansible_facts "ansible_owner_contact" ([string] $win32_cs.PrimaryOwnerContact)
+
+Set-Attr $result.ansible_facts "ansible_user_dir" $env:userprofile
+Set-Attr $result.ansible_facts "ansible_user_gecos" ([string] $user.Label)
+Set-Attr $result.ansible_facts "ansible_user_id" $env:username
+Set-Attr $result.ansible_facts "ansible_user_uid" ([int] $user.User.Value.Substring(42))
+Set-Attr $result.ansible_facts "ansible_user_sid" $user.User.Value
+
+# Use English locale
+[System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::GetCultureInfo('en-US')
```
nitzmahone: This should probably be using [System.Globalization.CultureInfo]::InvariantCulture. Makes me a bit nervous to change the thread culture out from under PS midway through a script (who the hell knows what it's caching under the covers)... If you're just worried about getting culture-invariant date formatting, you can pass InvariantCulture as the second arg to ToString() on these and leave the thread culture alone.

windows/setup.ps1
```diff
@@ -77,29 +81,79 @@ foreach ($adapter in $ActiveNetcfg)
...
+Set-Attr $date "time" $datetime.ToString("HH:mm:ss")
+Set-Attr $date "tz_offset" $datetime.ToString("zzzz")
+Set-Attr $date "tz" ([System.TimeZoneInfo]::Local.Id)
+Set-Attr $date "weekday" $datetime.ToString("dddd")
+Set-Attr $date "weekday_number" (Get-Date -UFormat "%w")
+Set-Attr $date "weeknumber" (Get-Date -UFormat "%W")
+Set-Attr $date "year" $datetime.ToString("yyyy")
 Set-Attr $result.ansible_facts "ansible_date_time" $date
 
-Set-Attr $result.ansible_facts "ansible_totalmem" $capacity
```
nitzmahone, line 102: I have mixed feelings about removing totalmem (instead of leaving it alongside memtotal_mb). Bugs the crap out of me that nobody made sure the keys on this matched *nix setup, and we haven't documented these keys, so... I guess I could go either way.

windows/setup.ps1
```diff
@@ -77,29 +81,79 @@ foreach ($adapter in $ActiveNetcfg)
...
+Set-Attr $result.ansible_facts "ansible_memtotal_mb" ([math]::round($win32_cs.TotalPhysicalMemory / 1024 / 1024))
```
nitzmahone, line 155: This one should probably be in a try/catch or have a defensive check around the value if the old comment at the top about TotalPhysicalMemory being empty is accurate (I've not experienced that myself).

windows/setup.ps1
dagwieers: The problem was that on a Flemish Windows, the day-of-week was in Flemish. Your suggestion works fine. Expect a fix !

windows/setup.ps1
dagwieers, line 155: The old comment was about Win32_PhysicalMemory being empty.
I assume we were using Win32_ComputerSystem because of this instead.

windows/setup.ps1
dagwieers, line 102: Right, my goal was to make it identical to *nix, disregarding backward compatibility.
Let me know what you want me to do, it is hard for me to estimate the impact.

windows/setup.ps1
nitzmahone, line 155: Yep, you're right- I misread

windows/setup.ps1
nitzmahone, line 102: Lemme check with a couple others- since we haven't written it down yet (I was hoping to get to windows setup return docs for 2.2), it's probably ok...
@trondhindenes @jhawkesworth @bcoca any opinion?

windows/setup.ps1
dagwieers, line 155: Maybe I should retain that comment nonetheless. Expect an update !

windows/setup.ps1
```diff
@@ -77,29 +81,80 @@ foreach ($adapter in $ActiveNetcfg)
...
+Set-Attr $result.ansible_facts "ansible_processor_vcpus" ($win32_cpu.NumberOfLogicalProcessors / $win32_cs.NumberOfProcessors)
+Set-Attr $result.ansible_facts "ansible_product_name" $win32_cs.Model.Trim()
+Set-Attr $result.ansible_facts "ansible_product_serial" $win32_bios.SerialNumber
+Set-Attr $result.ansible_facts "ansible_product_version" ([string] $win32_cs.SystemFamily)
 Set-Attr $result.ansible_facts "ansible_system" $osversion.Platform.ToString()
+Set-Attr $result.ansible_facts "ansible_system_description" ([string] $win32_os.Description)
+Set-Attr $result.ansible_facts "ansible_system_vendor" $win32_cs.Manufacturer
 Set-Attr $result.ansible_facts "ansible_os_family" "Windows"
 Set-Attr $result.ansible_facts "ansible_os_name" ($win32_os.Name.Split('|')[0]).Trim()
 Set-Attr $result.ansible_facts "ansible_distribution" $osversion.VersionString
```
nitzmahone: The value on this one bugs me- doesn't match typical values for ansible_distribution (same base name of Microsoft Windows NT for everything, and includes a version which no other platform does). Closest correct thing I could think of would be the value for $win32_os.Caption (which is basically $win32_os.Name after nuking the install location)... Any thoughts?

trondhindenes:
I would be very careful with removing existing properties, at least outside of major version bumps. That said, nice job starting to bring windows facts in line with nix!

windows/setup.ps1
```diff
@@ -128,7 +128,7 @@ Set-Attr $result.ansible_facts "ansible_user_uid" ([int] $user.User.Value.Substr
 Set-Attr $result.ansible_facts "ansible_user_sid" $user.User.Value
 
 # Use English locale
-[System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::GetCultureInfo('en-US')
+$culture = New-Object System.Globalization.CultureInfo('en-US')
```
nitzmahone: I can't imagine a Windows system without en-US, but  [System.Globalization.CultureInfo]::InvariantCulture is the documented right thing to use here (and saves us an object creation, wee)...

windows/setup.ps1
dagwieers: Sounds reasonable, I didn't want to touch it lacking the rationale (and not using the value). But I am happy to add more consistency.

It is however important to test on other Windows versions. I only have Windows 10 (desktops) at my disposal.

nitzmahone:
@trondhindenes this would definitely be for 2.2 - I hate having duplicate properties... I don't imagine there's been a ton of use on those values, but tough to say.

windows/setup.ps1
dagwieers: BTW there is no specific attribute to store the flavor (e.g. Home, Pro, etc...). I think it's important to have that too.

nitzmahone: Agreed- I know there's a way to get it (but can't remember where off the top of my head)... Lemme look.

windows/setup.ps1
```diff
@@ -77,29 +81,80 @@ foreach ($adapter in $ActiveNetcfg)
...
+Set-Attr $result.ansible_facts "ansible_hostname" $env:COMPUTERNAME
+Set-Attr $result.ansible_facts "ansible_fqdn" ($ip_props.Hostname + "." + $ip_props.DomainName)
+Set-Attr $result.ansible_facts "ansible_processor" $cpu_list
+Set-Attr $result.ansible_facts "ansible_processor_cores" $win32_cpu.NumberOfCores
+Set-Attr $result.ansible_facts "ansible_processor_count" $win32_cs.NumberOfProcessors
+Set-Attr $result.ansible_facts "ansible_processor_threads_per_core" ($win32_cpu.NumberOfLogicalProcessors / $win32_cs.NumberOfProcessors / $win32_cpu.NumberOfCores)
+Set-Attr $result.ansible_facts "ansible_processor_vcpus" ($win32_cpu.NumberOfLogicalProcessors / $win32_cs.NumberOfProcessors)
+Set-Attr $result.ansible_facts "ansible_product_name" $win32_cs.Model.Trim()
+Set-Attr $result.ansible_facts "ansible_product_serial" $win32_bios.SerialNumber
+Set-Attr $result.ansible_facts "ansible_product_version" ([string] $win32_cs.SystemFamily)
```
nitzmahone: SystemFamily is new to 10, missing on <=2012R2 - returns "" on my Win10 box too....

trondhindenes:
well as long as we havent' documented anything yet, I guess it's not a huge thing.
As for windows OS (pro/home/etc), use the OperatingSystemSKU attribute on win32_operatingsystem. The return value is numerical, with value mappings documented here: https://msdn.microsoft.com/en-us/library/aa394239(v=vs.85).aspx (search for OperatingSystemSKU )

windows/setup.ps1
nitzmahone: It's included on $win32_os.Caption and Name- only other thing would be to build a table lookup on $win32_os.OperatingSystemSKU against the table at https://msdn.microsoft.com/en-us/library/windows/desktop/ms724358(v=vs.85).aspx (bleh)

nitzmahone:
Heh, I was just looking at the same thing. Surprises me they don't have that table baked in somewhere.

jhawkesworth:
Right now I use very little from the windows fact gathering, and agree we should where possible have the same key names across OSes.
Might need to fix up the tests though if totalmem gets removed - I think from memory there's a check in there.

windows/setup.ps1
dagwieers: I expected that to come from the BIOS info (actually DMI), but it wasn't available there. Will look this evening for other places with the same content.

dagwieers:
@jhawkesworth I have a small collect.yml playbook that creates an asset database (CSV file) based on (a lot of) these facts. That's the main reason I've tried to implement as much as I could from the *nix setup. The network/disk/mount information however is largely missing.

https://github.com/crombeen/ansible

dagwieers:
Alright, what's needed to get this merged as soon as possible ?

jhawkesworth:
For something core like this, personally I'd like to see a few people have tested it against various different windows versions, both physical and virtualized.

And also update the integration tests.

jhawkesworth:
Also I'd like to try feeding the output into 
https://github.com/fboender/ansible-cmdb
just to see if there's anything else that could usefully be added to bring the facts closer to linux ones.

dagwieers:
I'd prefer to close this one and leave it up to someone else to add the remaining facts that other people use. Merging early avoids things like #3910 and makes the improvements available to users in the devel branch much sooner. Which means more people will be testing this.

nitzmahone:
I'll try to test today on everything I have laying around; if there's nothing showstopping, I'll just merge and fix inline.

nitzmahone:
Once we figure out what to do with ansible_product_version (SystemFamily fails on everything pre-10/2016), this tests fine for me on 7, 10, 2008, 2008R2, 2012, 2012R2 (I commented out the offending line), and I'm good to merge to devel.

dagwieers:
shipit !

nitzmahone:
sold- thanks!

richardcs:
This broke working with PS 3.0, all of the instances of $win32_cpu need to be changed to work with only the Count property and line 125, Label needs to be changed to Name. If this is backwards compatible. Please make these changes.

PR #3761, opened 2016-05-26
Fix unarchive screenscraping by selecting C locale

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

files/unarchive.py
##### ANSIBLE VERSION

```
2.1 and devel
```
##### SUMMARY

Fixes #3706

PR #3778, opened 2016-05-28
fix for #3706

### LINKED ISSUE #3706: unarchive issue with ansible 2.1 rc3
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

unarchive
##### ANSIBLE VERSION

```
ansible-playbook 2.1.0.0
  configured module search path = ['/home/yannig/dev/ansible-conf/ansible/playbooks/library']
```
##### CONFIGURATION

None
##### OS / ENVIRONMENT

Nothing special
##### SUMMARY

When using unarchive with Ansible 2.1 rc3, I get an error. Using ansible devel or ansible 2.0 work.
##### STEPS TO REPRODUCE

Get unarchive.yml at the following location: https://github.com/Yannig/yannig-ansible-playbooks/blob/master/unarchive/unarchive.yml

And run it.

```
ansible-playbook unarchive.yml
```
##### EXPECTED RESULTS

```
PLAY [Unarchive problem] *******************************************************

TASK [file] ********************************************************************
changed: [localhost]

TASK [file] ********************************************************************
changed: [localhost]

TASK [get_url] *****************************************************************
changed: [localhost]

changed: [localhost]

localhost                  : ok=4    changed=4    unreachable=0    failed=0
```
##### ACTUAL RESULTS

```
PLAY [Unarchive problem] *******************************************************

TASK [file] ********************************************************************
changed: [localhost]

TASK [file] ********************************************************************
changed: [localhost]

TASK [get_url] *****************************************************************
ok: [localhost]

fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Unexpected error when accessing exploded file: [Errno 2] Aucun fichier ou dossier de ce type: '/tmp/unarchive/apache-tomee-plus-1.7.4/webapps'"}

        to retry, use: --limit @unarchive.retry

localhost                  : ok=3    changed=2    unreachable=0    failed=1
```

Note: the devel version is not concerned about this issue.

cambid:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

files/unarchive.py
##### ANSIBLE VERSION

  ansible 2.1.0.0
##### SUMMARY

added C-locale to fix #3706, should be a better solution than #3761


gregdek:
Thanks @cambid. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

abadger:
This should fix the issue reported just as well as #3761.  Note, though, that we don't screenscrape every command that we run so that's why I only added the environment override for some of the invocations of run_command in #3761.  This one has the virtue of not needing to remember to add the relevant environment override whenever a command is screenscraped. The other one has the benefit of only changing the environment variables when we know that we're screenscraping; leaving the environment the same for all other calls.  @dagwieers which one is your preference?

abadger:
@dagwieers replied in #3761 that he'd rather see us set the locale for predictable message string when screenscraping for all commands instead of just the ones that would be a problem now.

Merged this to devel and stable-2.1 branches.

Yannig:
:+1:

dagwieers:
@abadger Thanks ! I think this is a better approach overall. Maybe even something to reuse for other modules as well ? (If we want to avoid translated command output)

ferrants:
I only experienced this when running from my mac. When running from a centos instance, I had no issue. It is because `gtar` was not available on my mac. The docs for unarchive specify that `gtar` is needed. I installed `gtar` through `brew`, `brew install gnu-tar` and it began working fine.

PR #3795, opened 2016-05-30
new module - win_pendingreboot

trondhindenes:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

win_pendingreboot
##### ANSIBLE VERSION

```
2.2.0
```
##### SUMMARY

New module to check for pending reboot on windows hosts. Discussed with @nitzmahone

### DIFF
--- a/windows/win_pendingreboot.ps1
+++ b/windows/win_pendingreboot.ps1
@@ -1,0 +1,37 @@
+#!powershell
+# (c) 2015, Trond Hindenes <trond@hindenes.com>, and others
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+# WANT_JSON
+# POWERSHELL_COMMON
+
+$params = Parse-Args $args
+$result = New-Object psobject
+Set-Attr $result "changed" $false
+Set-Attr -obj $result -name "pending_reboot" -value $false
+
+# Check if reboot is required, if so notify CA. The MSFT_ServerManagerTasks provider is missing on client SKUs
+$featureData = invoke-wmimethod -EA Ignore -Name GetServerFeature -namespace root\microsoft\windows\servermanager -Class MSFT_ServerManagerTasks
+$regData = Get-ItemProperty "HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager" "PendingFileRenameOperations" -EA Ignore
+if(($featureData -and $featureData.RequiresReboot) -or $regData)
+{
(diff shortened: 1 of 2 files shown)


gregdek:
Thanks @trondhindenes for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

trondhindenes:
Looks like Travis fails on parsing options. This module simply doesn't take any, how do I properly reflect that in docs?

jhawkesworth:
I don't know how to get round the travis fail, but it occurs to me that the place for this might actually be the setup module.  I know you are most likely to want this particular piece of information halfway through a playbook, but its easy enough to run setup again (I am less bothered by how long it takes to gather facts now that winrm is so much faster).  Its true that running setup would be slower than running just this module but you would still have a time saving over going through an unnecessary reboot.

I sort of like the idea of it being in setup as that way there's just one (existing) way of finding out facts about a host and users don't need to go searching for this.  As facts about a host go 'pending reboot' is definitely ephemeral and I can see why you'd want to just discover whether or not a reboot was pending.

sivel:
A couple of things.  I need to fix a bug in `ansible-testing` to handle the docs validation error you are seeing.

Additionally, new modules are supposed to be submitted to -extras, not -core.

trondhindenes:
Thanks @sivel. As for choosing core for this module, @nitzmahone asked me to put it here. Tho I certainly see @jhawkesworth 's point about putting it in the setup script as well. That might be a better place for this.

sivel:
Ok, it is now passing due to no `options`, however the PR is still failing, due to lack of missing (required) `RETURN` docs.  See http://docs.ansible.com/ansible/developing_modules.html#example for information on `RETURN`.

nitzmahone:
@trondhindenes and I discussed similar things ahead of time (setup or discrete module); we agreed that discrete module is probably better in general, but that we could easily add this fact to setup as well (at which point I'd probably add an ansible_facts return to this module to override the setup fact value). 

The reason for adding to core directly is due to the pending ship-split of extras in 2.2- we were thinking it'd be nice to add a reboot-only-if-needed behavior to win_reboot that would call this module, and I didn't want to have a core action depending on an extras module that might not be there.

jhawkesworth:
Wouldn't want to stand in the way of this module, but the more I think about it the more I'd like it in setup.
It would be nice to be able to run a remediation playbook and immediately reboot any boxes that needed a reboot as first task in the play.
Looks like travis is failing because of ./cloud/google/gc_storage.py

nitzmahone:
I'm leaning toward doing both (so long as it's pretty quick to run)- might mean factoring this code into powershell.ps1 and just calling it from the module.

On the gc_storage thing, it's not a legitimate failure (the original docs had the wrong arg names). The way the validator works, it should correct itself when we do a 2.1.1 release.

trondhindenes:
@jhawkesworth I totally see your point about putting it in setup. It shouldn't drastically increase script execution time for the script. @nitzmahone, putting it in powershell.ps1 would still require us to reference it from setup.ps1, agree?

nitzmahone:
@trondhindenes Right- that'd be what would allow us to reference it from both setup and a win_pendingreboot module... I can definitely see wanting a lighter-weight way to check for reboot needed during a play run (instead of refreshing facts). If we're all agreed on that course of action, I can just move the code into powershell.ps1 or @trondhindenes can do it in a couple of PRs if he's got the time...

trondhindenes:
no prob, I'll get that done tomorrow.

nitzmahone:
Noice- ping me on the PR and I'll review ASAP

trondhindenes:
Closing, moved to #16161

PR #3856, opened 2016-06-03
Fix #147 automatically expose published ports

### LINKED ISSUE #147: docker module: Default network behavior does not mimic docker run
##### Issue Type:

Bug Report
##### Ansible Version:

1.8
##### Environment:

CentOS 6.5
##### Summary:

I'm filing this issue, though this may be a corner case that a majority of users may never see.

When starting a container with docker run from an image that does not expose ports, docker run will automatically expose the ports defined in -p option passed to the docker run command. With Ansible, just defining the ports for the container does not properly set up networking. The expose option must also be used to list what ports should be exposed. 
##### Steps To Reproduce:

Networking does not get setup.

```

---
- hosts: all
  tasks:
  - name: Start Network container
    docker: image=kubernetes/pause
            name=network
            ports=8080:8080,8443:8443
```

Docker run command automatically exposes the listed ports and sets up networking properly.

```
docker run -d -p 8080:8080 -p 8443:8443 --name network kubernetes/pause
```
##### Expected Results:

Container should have the port exposed and be listed in iptables showing routes for those ports to the container that was started.
##### Actual Results:

Container does not have networking routing setup unless 'expose' option is explicitly defined in the playbook.

chouseknecht:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

docker_container.py
##### ANSIBLE VERSION

```
```
##### SUMMARY

Fix issue #147. Any published port options should automatically be exposed, otherwise, the ports are not actually exposed on the network, leaving the user to add the expose option. The _docker run_ command automatically handles exposing the ports, and so the module should do the same.

### DIFF
--- a/cloud/docker/docker_container.py
+++ b/cloud/docker/docker_container.py
@@ -727,8 +727,10 @@ class TaskParameters(DockerBaseClass):
                 except ValueError as exc:
                     self.fail("Failed to convert %s to bytes: %s" % (param_name, exc))
 
-        self.ports = self._parse_exposed_ports()
         self.published_ports = self._parse_publish_ports()
+        self.ports = self._parse_exposed_ports(self.published_ports)
+        self.log("expose ports:")
+        self.log(self.ports, pretty_print=True)
         self.publish_all_ports = None
         if self.published_ports == 'all':
             self.publish_all_ports = True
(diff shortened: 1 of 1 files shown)


gregdek:

PR #3879, opened 2016-06-06
Return reboot pending as true/false

trondhindenes:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

setup.ps1 now returns ansible_reboot_pending (true/false). Using new module function from #16161 (https://github.com/ansible/ansible/pull/16161)
This PR should be merged after #16161 because of its dependency
##### ANSIBLE VERSION

2.2.0
##### SUMMARY

Discussed in https://github.com/ansible/ansible-modules-core/pull/3795
Sample gather_facts output:
`"ansible_reboot_pending":  false`

### DIFF
--- a/windows/setup.ps1
+++ b/windows/setup.ps1
@@ -153,6 +153,9 @@ if ($winrm_cert_expiry)
     Set-Attr $result.ansible_facts "ansible_winrm_certificate_expires" $winrm_cert_expiry.NotAfter.ToString("yyyy-MM-dd HH:mm:ss")
 }
 
+$PendingReboot = Get-PendingRebootStatus
+Set-Attr $result.ansible_facts "ansible_reboot_pending" $PendingReboot
+
 # See if Facter is on the System Path
 Try {
     $facter_exe = Get-Command facter -ErrorAction Stop


trondhindenes:
ping @jhawkesworth and @nitzmahone

jhawkesworth:
Great, thank you for sorting this out.

nitzmahone:
@trondhindenes were you going to submit the original module as well (just calling the shared func)? If so, it should probably return both a direct value as well as an ansible_facts dictionary with the ansible_reboot_pending fact value (to override the gathered fact).

trondhindenes:
hm not sure there's a point doing that. Anyone can just call the setup module at any time if they want super-up-to-date values. THat said, it's only about 3 lines of code so I'll happily add it if theres a need for it.

nitzmahone:
OK by me either way- we can just leave in setup. At least today, the roundtrips to run any module dwarf the execution time difference between setup and a bespoke pendingreboot module.

dagwieers:
Hmm, tested this today on a Windows 10 laptop, and I get:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: +                  ~~~~~~~~~~~~~~~~~~~~~~~
fatal: [windows1]: FAILED! => {"changed": false, "failed": true, "msg": "The term 'Get-PendingRebootStatus' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again."}
```

nitzmahone:
Strange- working fine for me against 2012R2 and Win10... Sure you've got a matching ansible/ansible and modules-core repo?

PR #3926, opened 2016-06-10
fix unarchive doesn't extract changed tar file

neowu:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

unarchive module
ansible-modules-core/files/unarchive.py
##### ANSIBLE VERSION

2.1.0.0-1ppa~trusty
##### CONFIGURATION

none
##### OS / ENVIRONMENT

N/A
##### SUMMARY

unarchive is not updating dest (extract the tar) if tar content changed and still keep old file structures, (means if you run "tar -d", it reports Mod Time changed)

by checking the  ansible-modules-core/files/unarchive.py, it seems TgzArchive is not handling 
"Mod time differs" condition in is_unarchived method
##### STEPS TO REPRODUCE

1) prepare tar file from folder, make it demo.tar 

2) run this command in playbook
- unarchive: src=/opt/packages/demo.tar dest=/opt/app/ copy=no

3) update content of any file, make newer demo.tar
4) run the same playbook
##### EXPECTED RESULTS

the dest folder should be updated,
##### ACTUAL RESULTS

ansible ubuntu skipped it, even though the tar file is changed

ok: [localhost]

### DIFF
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -132,6 +132,7 @@ import subprocess
 OWNER_DIFF_RE = re.compile(r': Uid differs$')
 GROUP_DIFF_RE = re.compile(r': Gid differs$')
 MODE_DIFF_RE = re.compile(r': Mode differs$')
+MOD_TIME_DIFF_RE = re.compile(r': Mod time differs$')
 #NEWER_DIFF_RE = re.compile(r' is newer or same age.$')
 MISSING_FILE_RE = re.compile(r': Warning: Cannot stat: No such file or directory$')
 ZIP_FILE_MODE_RE = re.compile(r'([r-][w-][stx-]){3}')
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @neowu. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

dagwieers:
Seems to work fine. Thanks !

shipit

gregdek:
Thanks again to @neowu for this PR, and thanks @dagwieers, @pileofrogs for reviewing. Marking for inclusion.

panuhorsmalahti:
I had this same issue, and applying this PR fixed it.

I'm not sure, but I think this is a regression from 2.0 -> 2.1.

dagwieers:
Yes, the zip support was rewritten between 2.0 and 2.1. But there were also changes to gtar support (to avoid unarchiving on every run). Unfortunately the unarchive module is not idempotent and does not support check-mode and diff-mode at this point. A second native rewrite will be needed to have a proper idempotent unarchive module.

Avoiding regressions is only possible if we can engage our users to test their setup against newer Ansible/modules, which is currently very limited (hence the regressions in released versions).

dagwieers:
@abadger Can you please merge this one as well ?

jameshartig:
Looks like this is still stuck in devel. Can this make it into a release? This is breaking our deployments since files are not being overwritten.

dagwieers:
@abadger @alikins Can we get this fixed released in stable-2.1 ASAP, we missed release v2.1.1.0 now :-(

Issue #3760, opened 2016-05-26
ansible docker_image can't find a locally available image on build


jordy25519:
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

docker_image
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /home/user/scratch/deployment/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

OS: N/A
docker-py 1.8.1
docker version

```
Client:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Tue Apr 26 23:30:23 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Tue Apr 26 23:30:23 2016
 OS/Arch:      linux/amd64
```
##### SUMMARY

When building a docker image locally Ansible fails to find a locally available base image specified in the  `FROM` directive of the dockerfile. 
Resulting Error:

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Error building localhost:5000/image - code: None message: Error: image library/base_image not found"}
```

[See this similar issue with docker compose](https://github.com/docker/compose/issues/1568)
##### STEPS TO REPRODUCE

Create docker registry

```
-   name: Creating docker registry
    docker:
        name: registry
        image: "registry:2"
        state: started
        restart_policy: always
        ports:
            - "5000:5000"
        env:
            REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY: /var/lib/registry
        volumes:
            "{{host_registry_dir}}:/var/lib/registry"
```

Build, tag, push base image

```
-   name: Create base_image locally
    docker_image:
        name: "{{private_docker_registry}}/base_image"
        path: "{{build_folder}}/dockerfiles/base_image"
        dockerfile: base_image.Dockerfile
```

Build,tag,push child image

```
-   name: Create image locally
    docker_image:
        name: "{{private_docker_registry}}/image"
        path: "{{build_folder}}/dockerfiles/image"
        dockerfile: image.Dockerfile
```

Simplified `image.Dockerfile` using the locally built base_image as its base:

```
FROM base_image

EXPOSE 5555

ENTRYPOINT ["./entrypoint.sh"]
```
[...]

PR #3505, opened 2016-04-20
Adding refactored docker_image module.

chouseknecht:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

docker_image.py 
##### ANSIBLE VERSION

``````
ansible 2.1.0 (docker_testing 9bb069f873) last updated 2016/04/18 12:36:03 (GMT -400)
  lib/ansible/modules/core: (detached HEAD cf01087a30) last updated 2016/04/05 16:31:53 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 204b4bab56) last updated 2016/04/05 16:31:56 (GMT -400)
  config file =
  configured module search path = Default w/o overrides```

##### SUMMARY
The new docker_image module, completely replacing the current docker_image.
``````

PR #3961, opened 2016-06-16
Removes check for authentication data in Docker client configuration file

gotgenes:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

docker_image
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

This change addresses the issue where users of docker_image are unable to push images to local/private Docker registries. This bug was introduced unwittingly in #3505. The bug was first described in ansible/ansible-modules-extras#2312, re-posted as #3760. The result of this bug is that when pushing to insecure registries — which do not require a configuration file containing authentication credentials — the push will fail with a message like the following

```
"Error: configuration for localhost:5000 not found. Try logging into localhost:5000 first."
```

The bug centers around the assumption that in order to push an image to the registry, the user will need to be able to authenticate with the registry. While this assumption holds for the public Docker Hub and holds true for some private registries, for other private registries including those following the Docker documentation on running insecure registries, this does not hold true. That is, for these private registries, no authentication is needed to push images to or pull images from them.

For registries that do not require authentication, there is no way to circumvent the check in docker_image for credentials to use for authentication, and so use of the docker_image module in these circumstances will always fail any playbook or any role using docker_image to push to insecure registries. So while the check is meant to be helpful to alert the user that the push will not succeed, it is actually premature and overbearing.

Additionally, although the code goes through the hassle of trying to obtain authentication credentials, it disposes of them immediately. It never passes the configuration to the client doing the push of the image, meaning that the docker-py client will need to do its own management of finding credentials, anyway—an implementation detail that is subject to change and could easily diverge from the check in docker_image. In these circumstances, docker-py will raise its own error and issue its own error message anyway, so again, the check for credentials in docker_image is premature and duplicated effort.

Given this check has broken docker_image for a principle use case of pushing to insecure registries, and en light of the Pythonic principle of [EAFP](https://docs.python.org/3/glossary.html#term-eafp), this change removes the check for a Docker configuration file. In circumstances where it is needed, the client continues to be responsible for finding and using it. In circumstances where it is not needed, the push no longer fails and proceeds as expected.

cc @chouseknecht


gregdek:
Thanks @gotgenes for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

cloud/docker/docker_image.py
```diff
@@ -371,12 +371,6 @@ def push_image(self, name, tag=None):
         registry, repo_name = auth.resolve_repository_name(repository)
 
         if re.search('/', repository):
-            if registry:
```
bcoca, line 374: i would not eliminate this block, but just the error on load, keep a flag and then return it if/when it fails to connect.

gotgenes, line 374: Thanks for your feedback. Would you mind providing more information here, or a code snippet? Unfortunately I don't understand what is meant by, "keep a flag and then return it if/when it fails to connect." I am extremely new to Ansible's codebase, so I don't know yet the idea of returning values and what that does to the flow.

Would returning such a value cause the task to be marked as a failure? I want to avoid the task being marked as a failure in cases where no credentials were needed in the first place.

My thinking here is that an exception raised by docker-py should handle the case of terminating the play when no credentials were found but they were indeed necessary. I would be amenable to keeping this logic and downgrading `self.fail` to `self.warn` (if there is such a method?), however, I refer back to this portion of the PR's original description (emphasis added):

> the docker-py client will need to do its own management of finding credentials, anyway—an implementation detail that is subject to change and could easily diverge from the check in docker_image.

Thus, docker_image could end up issuing warnings when they are unwarranted because a different scheme is used to collect credentials by the docker-py client.

bcoca, line 374: I was thinking on `found_credentials = True|False` , you can then use when the connection fails with 'unauthorized' and then issue this fail vs the other one that indicates that the repo is missing.

chouseknecht:
PR #4171 replaces this. It removes registry auth checking altogether, letting the API handle it. And, it adds a little bit better API error parsing.

PR #3965, opened 2016-06-16
Do not copy file if running on check mode

ewigkeit1204:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

files/copy.py
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

When running a copy task on check mode, ansible copies if dest differs from src.

This PR do not copy file if running on check mode.

### DIFF
--- a/files/copy.py
+++ b/files/copy.py
@@ -301,33 +301,34 @@ def main():
 
     backup_file = None
     if checksum_src != checksum_dest or os.path.islink(dest):
-        try:
-            if backup:
-                if os.path.exists(dest):
-                    backup_file = module.backup_local(dest)
-            # allow for conversion from symlink.
-            if os.path.islink(dest):
-                os.unlink(dest)
-                open(dest, 'w').close()
-            if validate:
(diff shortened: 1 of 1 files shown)


nitzmahone:
I'm unable to reproduce the original problem (copy doesn't copy during check mode for me in 2.1.0)- the check_mode support is handled on the host [in the copy action](https://github.com/ansible/ansible/blob/5900571273da45474ed2e1868194cd24abf371d8/lib/ansible/plugins/action/copy.py#L211-215)\- by the time we call the module, it's assumed that we're actually going to do the copy.

Can you provide a reproduction of what you're trying to fix?

nitzmahone:
@bcoca points out that the action _will_ run the module unconditionally when remote_src is specified- good to have the check_mode stuff actually implemented on the module side as well, in case anything else calls it, so we'll just merge and assume that was the issue.

nitzmahone:
cherry-picked to stable-2.1

Issue #4007, opened 2016-06-22
junos_config.py delete statement not working


stianvi:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

network/junos/junos_config.py
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

N/A
##### OS / ENVIRONMENT

--- JUNOS 15.1R3.6 built 2016-03-24 18:40:35 UTC
##### SUMMARY

delete statements don't work due to bug in if-loop
##### STEPS TO REPRODUCE

add a statement beginning with delete under the lines and run the playbook. diff_config if-loop will never match when action is delete

Example requires already configured description on the ge-0/0/0 interface

```
- name: delete interface ge-0/0/0 description 
  hosts: all
  roles:
  - Juniper.junos
  connection: local
  gather_facts: no

  tasks:
    - name: Delete vlan on Juniper Junos 
      junos_config:
        lines:
        - delete interfaces ge-0/0/0 description
        replace: True 
        host: "{{ inventory_hostname }}"
```

##### EXPECTED RESULTS

The configuration lines would be removed from configuration and will return to default configuration
##### ACTUAL RESULTS

No configuration change. Ansible report change=false

```
Using /etc/ansible/ansible.cfg as config file

PLAYBOOK: debug-delete-vlan.yml ************************************************
1 plays in debug-delete-vlan.yml

PLAY [Create vlan] *************************************************************

TASK [Delete vlan on Juniper Junos] ********************************************
task path: /etc/ansible/playbooks/debug-delete-vlan.yml:11
Using module file /usr/lib/python2.7/site-packages/ansible/modules/core/network/junos/junos_config.py
<netlab-sw01-a.osl.basefarm.net> ESTABLISH LOCAL CONNECTION FOR USER: stianv
<netlab-sw01-a.osl.basefarm.net> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1466596354.01-95243515745333 `" && echo ansible-tmp-1466596354.01-95243515745333="` echo $HOME/.ansible/tmp/ansible-tmp-1466596354.01-95243515745333 `" ) && sleep 0'
[...]

PR #4008, opened 2016-06-22
Fixed delete statement for juniper_config.py #4007

stianvi:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

network/junos/junos_config.py
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

Changed if-loop so it checks towards current configuration correctly.

Before:

```
PLAY [Create vlan] *************************************************************

TASK [Delete vlan on Juniper Junos] ********************************************
ok: [netlab-sw01-a.osl.basefarm.net]

netlab-sw01-a.osl.basefarm.net : ok=1    changed=0    unreachable=0    failed=0   
```

After:

```
PLAY [Create vlan] *************************************************************

TASK [Delete vlan on Juniper Junos] ********************************************
changed: [netlab-sw01-a.osl.basefarm.net]

netlab-sw01-a.osl.basefarm.net : ok=1    changed=1    unreachable=0    failed=0  
```

### DIFF
--- a/packaging/os/apt.py
+++ b/packaging/os/apt.py
@@ -551,7 +551,7 @@ def install_deb(m, debs, cache, force, install_recommends, allow_unauthenticated
     else:
         m.exit_json(changed=changed, stdout=retvals.get('stdout',''), stderr=retvals.get('stderr',''), diff=retvals.get('diff', ''))
 
-def remove(m, pkgspec, cache, purge=False,
+def remove(m, pkgspec, cache, purge=False, force=False,
            dpkg_options=expand_dpkg_options(DPKG_OPTIONS), autoremove=False):
     pkg_list = []
     pkgspec = expand_pkgspec_from_fnmatches(m, pkgspec, cache)
@@ -565,6 +565,11 @@ def remove(m, pkgspec, cache, purge=False,
     if len(packages) == 0:
         m.exit_json(changed=False)
     else:
+        if force:
+            force_yes = '--force-yes'
+        else:
+            force_yes = ''
+
         if purge:
             purge = '--purge'
         else:
@@ -580,7 +585,7 @@ def remove(m, pkgspec, cache, purge=False,
         else:
             check_arg = ''
 
-        cmd = "%s -q -y %s %s %s %s remove %s" % (APT_GET_CMD, dpkg_options, purge, autoremove, check_arg, packages)
+        cmd = "%s -q -y %s %s %s %s %s remove %s" % (APT_GET_CMD, dpkg_options, purge, force_yes ,autoremove, check_arg, packages)
 
         rc, out, err = m.run_command(cmd)
         if m._diff:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @stianvi. To the current maintainers, @privateip please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

privateip:
This doesn't quite solve the bug.   Delete statements are slightly different from set statements in that they should account for partials lines.  Consider the following stanza:

```
system {
    name-server {
        8.8.8.8;
        8.8.4.4;
    }
```

Handling of the delete line must satisfy the each of the following (all valid):

```
delete system name-server 8.8.8.8
delete system name-server
```

The first deleting a single name server, the second deleting all name servers.   This fix does not address the second use case.

stianvi:
My test it did accept partial lines. I only check if the content match some parts of the lines, not the beginning or the start. 

`
lines:
- delete system name-server 8.8.8.8
- delete system name-server
  `

Before:

```
{master:0}
stianv@netlab-sw01-a> show configuration | match name-server | display set
set system name-server 8.8.8.8
set system name-server 8.8.4.4
```

After:

```
{master:0}
stianv@netlab-sw01-a> show configuration | match name-server | display set
```

If you want to use startswith then you'll need to remove "set " away from cfg. I can't see a nice way to do this..

```
-if cfg[4:].startswith(cfgline):
-    updates.add(cfgline)
```

privateip:
Partial tests have to be done from the beginning of the the line otherwise you will have false positives.  Add the change as above to remove the "set " and the fix should be valid then.

Royce92:
You are using ansible tower 2.2, this bug is still being reported on ansible v2.1

rromanyak:
I'm running ansible 2.1.2.0 and delete command wasn't working for me. I looked a the code an saw the typo in 'delete' elif line. After correcting it, delete commands work:

Before:

```
    elif action == 'delete' and not config:
        updates.add(line)
```

After:

```
    elif action == 'delete' and cfgline not in config:
        updates.add(line)
```

stianvi:
@rromanyak could you give me an example on what you are trying to delete?
Seems to me the elif isn't needed, since if it isn't in the config - then what is the point of adding the delete statement?
I'll look into this when I get more info - or you could create a PR and the ansible team would probably push the change pretty quick.

rromanyak:
I was just trying to delete a static route or anything else (tried also to delete a term in a policy). The playbook didn't give any error but also didn't change anything. 

This is the playbook:

---
- hosts: test_routers
  connection: local
  gather_facts: no
  
  tasks:
  - name: Run set command
    junos_config:
      host: "{{ inventory_hostname }}"
      username: automation
      confirm: "{{confirm}}"
      lines: ["{{CLI}}"]

PR #4014, opened 2016-06-22
fix 2nd example in package module

lloydbenson:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

package
##### ANSIBLE VERSION

N/A : http://docs.ansible.com/ansible/package_module.html
##### SUMMARY

Just fixing a couple of typos so it looks right on the main site.

### DIFF
--- a/packaging/os/package.py
+++ b/packaging/os/package.py
@@ -57,5 +57,5 @@ EXAMPLES = '''
 
 # This uses a variable as this changes per distribution.
 - name: remove the apache package
-  package : name={{apache}} state=absent
+  package: name=apache state=absent
 '''


bcoca:
the example is a variable on purpose, as the module name for apache changes depending on distribution.

lloydbenson:
Is the semi colon also on purpose?

lloydbenson:
If the variable is on purpose wouldn't it make sense to have the other example with ntpdate also in brackets to be consistent?

bcoca:
i fixed the colon directly, ntpdate is one of the few packages that seems to have the same name across distros. This is noted in the comments above the examples.

PR #4017, opened 2016-06-22
Attempt to track actual wall time in async_wrapper

cboylan:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

async_wrapper
##### SUMMARY

Previously the async wrapper had two issues with its timeout tracking.
The first was that it did one sleep without accounting for that sleep in
the remaining time calculations so it would potentially sleep one step
too long. The second is that it assumes that time.sleep() always sleeps
for exactly the amount of time passed to it, but it may sleep for a
shorter time if any signals are received by the process.

Address the first issue by only sleeping in the while loop so that every
sleep is accounted for in the remaining time calculations. Address the
second issue by checking the current time.time() on each iteration to
determine the actual remaining wall time instead of the assumed value
based on step.

Note this is not tested yet, but I noticed that the older code was potentially problematic especially with short timeouts (since the first step is unaccounted for), or with long poll intervals where wrapper could exit sooner (due to step being assumed as accurate for wall time delta) and not be noticed for a while because poll is not due for a while.

### DIFF
--- a/utilities/logic/async_wrapper.py
+++ b/utilities/logic/async_wrapper.py
@@ -170,23 +170,25 @@ if __name__ == '__main__':
             sub_pid = os.fork()
             if sub_pid:
                 # the parent stops the process after the time limit
-                remaining = int(time_limit)
+                time_limit = int(time_limit)
 
                 # set the child process group id to kill all children
                 os.setpgid(sub_pid, sub_pid)
 
-                notice("Start watching %s (%s)"%(sub_pid, remaining))
-                time.sleep(step)
+                notice("Start watching %s (%s)"%(sub_pid, time_limit))
+                start_time = time.time()
                 while os.waitpid(sub_pid, os.WNOHANG) == (0, 0):
-                    notice("%s still running (%s)"%(sub_pid, remaining))
                     time.sleep(step)
-                    remaining = remaining - step
+                    # Calculate remaining time against actual elapsed
+                    # wall time.
+                    remaining = time_limit - (time.time() - start_time)
                     if remaining <= 0:
                         notice("Now killing %s"%(sub_pid))
                         os.killpg(sub_pid, signal.SIGKILL)
                         notice("Sent kill to group %s"%sub_pid)
                         time.sleep(1)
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @cboylan for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

utilities/logic/async_wrapper.py
```diff
@@ -170,23 +170,25 @@ def _run_module(wrapped_cmd, jid, job_path):
...
             if sub_pid:
                 # the parent stops the process after the time limit
-                remaining = int(time_limit)
+                time_limit = int(time_limit)
 
                 # set the child process group id to kill all children
                 os.setpgid(sub_pid, sub_pid)
 
-                notice("Start watching %s (%s)"%(sub_pid, remaining))
-                time.sleep(step)
```
bcoca, line 179: Do not remove this sleep, it is here to allow for the fs to ensure that the required files are in place.
In some cases it will report back to the process everything is OK but then the fork won't see the files to execute yet.

cboylan, line 179: Is it the mkdir on line 138 that we need to make sure is present? Sleep is a poor way to ensure that file systems have synced, maybe we can solve that more directly? (I am not 100% sure what the required files are in this case so not sure how we can do that yet).

That said, all we did is rely on the sleep inside of the while loop to do all the sleeping. This means that we have gone from: notice, sleep, waitpid, notice, sleep to notice, waitpid, sleep. The waitpid call should only care about the running process which is already running at this point and not care about files on disk. So I am not sure why this changes anything with respect to files.

There is a possible race if sub_pid has already exited and we will not enter the loop and sleep at all, but at that point we just exit and don't write anything to files.

All this to say I don't quite understand the concern so more details would be nice, but also take this as an argument for why I think this is fine as is (and maybe it isn't I just need more infos).

bcoca, line 179: i'm looking for the original issue that made this required , going from memory here.

PR #4042, opened 2016-06-25
service module: use sysrc on FreeBSD

### LINKED ISSUE #2664: Bug: service module: "unable to set rcvar" in Check Mode due to mishandled SystemExit exception
**Issue Type**: Bug report
**Ansible version**: 1.9.4 (but seems to be present in HEAD as well)
**Environment**: FreeBSD 10.1 target

**Summary**:
The service module fails with message `unable to set rcvar` in Check Mode against FreeBSD (and probably any rc.conf-based system) without any error actually occurring

**Steps to reproduce**:

```
$ ansible -C foo -m service -a "name=ntpd state=started enabled=yes"
[...]
foo | FAILED >> {
    "failed": true, 
    "msg": "{\"msg\": \"changing service enablement\", \"changed\": true}\r\n{\"msg\": \"unable to set rcvar\", \"failed\": true}[...]
```

**Cause**:
- service_enable_rcconf() is run inside a try: block
- if changed is True and we are in check_mode, exit_json() is called (L993)
- exit_json() uses sys.exit(), which raises a SystemExit exception
- the calling try: block mishandles the SystemExit exception as the above error

While the issue could be easily fixed like this ...

```
    try:
        return self.service_enable_rcconf()
    except SystemExit:
        pass
    except:
        self.module.fail_json(msg='unable to set rcvar')
```

... I think the larger issue is that every service routine (not just _rcconf()) uses this scheme for an early exit from check_mode and thus would misbehave inside a try: block.

But the above change certainly is an interim solution.

problame:
sysrc(8) is the designated userland program to edit rc files on FreeBSD.

It first appeared in FreeBSD 9.2, hence is available on all supported
versions of FreeBSD.

Side effect: fixes #2664


gregdek:
Thanks @problame. To the current maintainers, @grossws please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

problame:
@gregdek So far, nobody has reviewed the patch. What is the normal response time to patches of this (small) size?

grossws:
@problame, usual time for _core_review_ was 1-3 _months_ even for bugfix patches. I volunteered to make reviews and am trying to do reviews in timeframe of a few weeks.

Patch size is not the only factor that affects review time. Presence of spare time is other important factor and familiarity with target OS is another one. To reduce review time you may provide test FreeBSD VM and/or provide some additional info like actual FreeBSD version distribution among users.

Be patient, we're not a paid support team but a community volunteers.

system/service.py
```diff
@@ -941,6 +941,7 @@ class FreeBsdService(Service):
 
     platform = 'FreeBSD'
     distribution = None
+    sysrc_cmd = "/usr/sbin/sysrc"
```
bcoca: I would use the module's find_bin function to locate sysrc as you cannot always trust it to be in the default location

Another point is that this is new in v9 (iirc) so this will break the module for those using v8 and below. I know they are ancient and unsupported ... but people still run these systems. So i would fallback to the current system if sysrc is not available.

And this is just an informative note, I'm trying to deprecate this module in favor of 'per system' ones, so I might end up pulling these changes into a bsdrc module. But systemctl is only one I've had time to separate.

problame:
I incorporated @bcoca's suggestions in the last commit.

@grossws As for the test VMs, I found the following Vagrant boxes to be quite usable (after `vagrant mutate`ing them to libvirt)

(I ran my non-automated tests against these boxes)

**FreeBSD 8.4 (no sysrc, should use legacy code path**
http://dlang.dawg.eu/vagrant/FreeBSD-8.4-i386.box
http://dlang.dawg.eu/vagrant/FreeBSD-8.4-amd64.box

**FreeBSD 9.2 (earliest sysrc available, should use new code path)**
http://iris.hosting.lv/freebsd-9.2-i386.box

**FreeBSD 10.2 (sysrc available, should use new code path)**
https://petar.xyz/files/freebsd-10.2-amd64-wunki.box

jbergstroem:
I think this broke new services. See https://github.com/ansible/ansible/issues/17892.

PR #4045, opened 2016-06-27
Fix inconsistent default value for network mode when `net` unspecified

PikachuEXE:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

cloud/docker
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = /users/pikachuexe/projects/spacious/spacious-rails/ansible.cfg
  configured module search path = ./ansible/library
```

This applied to all versions with docker module supporting `net` and `bridge`, which should be 2.0+
##### SUMMARY

The default value of network mode being used is inconsistent when `net` is unspecified
For actual container it uses `None` (in python), which translates to `default` in docker.
But in comparison it uses `bridge`.

```
Before:
"reload_reasons": "net (default => bridge)"

After:
"reload_reasons": null
```

### DIFF
--- a/cloud/docker/_docker.py
+++ b/cloud/docker/_docker.py
@@ -949,7 +949,7 @@ class DockerManager(object):
             'publish_all_ports': self.module.params.get('publish_all_ports'),
             'privileged': self.module.params.get('privileged'),
             'links': self.links,
-            'network_mode': self.module.params.get('net'),
+            'network_mode': self.module.params.get('net') or 'bridge',
         }
 
         optionals = {}


gregdek:

sfrique:
@PikachuEXE Have you tested with 2.1 ? 
I had the same problem with 2.0 but I think it was fixed.

oryband:
Wow I just burned half an hour banging my head against this issue. Right now I resolved this by manually adding `net: bridge` which should be the sane default.

PR #4054, opened 2016-06-27
Make launch config module convergent

mpappas86:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

ec2_lc module
##### ANSIBLE VERSION

```
> ansible --version
ansible 2.0.1.0
  config file = 
```
##### SUMMARY

AWS launch configs are immutable. ec2_lc currently handles this by simply not changing the launch config even when asked. This is somewhat opaque to the user (as the module returns with "OK"), and also just a frustrating pattern.

My changes allow the ec2_lc module to treat launch configs as mutable, using the following pattern:
1. Identify if the launch config is already exactly as desired. If it is, leave it alone. Otherwise, proceed.
2. Create a "Blue" launch config with the desired new setup.
3. Swap any ASG's using the original launch config onto the blue one.
4. Delete the original launch config.
5. Create a new launch config with the originally desired name with the desired new setup.
6. Swap any ASG's using the blue launch config onto the new one.
7. Delete the blue launch config.
This is non-destructive, in that ultimately the only thing that is changed is exactly what the user requested. 

It's worth noting that there's still one slightly opaque behavior here - the ASG's are _not_ forced to update their instances, so the new launch config only takes effect next time new instances need to be launched. This is intentional, since the risk of service outages should be managed in an individual way, but if anyone has any ideas about how to integrate nice behavior around this I'm all ears.

### DIFF
--- a/cloud/amazon/ec2_lc.py
+++ b/cloud/amazon/ec2_lc.py
@@ -109,7 +109,7 @@ options:
 extends_documentation_fragment:
     - aws
     - ec2
-requires: 
+requires:
     - "boto >= 2.39.0"
 """
 
@@ -142,6 +142,8 @@ try:
     HAS_BOTO = True
 except ImportError:
     HAS_BOTO = False
+
+import copy
 
 
 def create_block_device(module, volume):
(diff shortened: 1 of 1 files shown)


cloud/amazon/ec2_lc.py
```diff
@@ -298,14 +385,15 @@ def main():
 
     try:
         connection = connect_to_aws(boto.ec2.autoscale, region, **aws_connect_params)
-    except (boto.exception.NoAuthHandlerFound, AnsibleAWSError) as e:
+    except (boto.exception.NoAuthHandlerFound, AnsibleAWSError), e:
```
mpappas86: Woops, this was an accidental overwrite from a previous version, will fix

gregdek:
Thanks @mpappas86 for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #4122, opened 2016-07-07
Add new module for manipulating XML files #4112

### LINKED ISSUE #4112: New core module submission: XML
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

new module
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.1.0.0
```
##### CONFIGURATION

_N/A_
##### OS / ENVIRONMENT

_N/A_
##### SUMMARY

This is a new module submission for consideration in **core**: _xml_

> A module to manage various properties of XML documents

Presently hosted at https://github.com/cmprescott/ansible-xml with an open issue to track progress of acceptance as an upstream module in https://github.com/cmprescott/ansible-xml/issues/52
##### HISTORY

This module was originally submitted for consideration in https://github.com/ansible/ansible/issues/16147 wherein we were instructed to open an issue on this tracker instead.

cmprescott:
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

files/xml
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = 
```
##### SUMMARY

New module for manipulating xml files. Multiple asks to have this included with ansible instead of in its own repo.

Fixes #4112 

```
ansible 2.1.0.0
  config file = 
```
- Used python from https://github.com/cmprescott/ansible-xml/tree/0.3.4
- Did not include unit tests because does not seem like a natural place for them

### DIFF
--- a/files/xml.py
+++ b/files/xml.py
@@ -1,0 +1,470 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# xml - Manage bits and pieces of XML files
+#
+# Copyright 2014, Red Hat, Inc.
+# Tim Bielawa <tbielawa@redhat.com>
+# Magnus Hedemark <mhedemar@redhat.com>
+#
+# This software may be freely redistributed under the terms of the GNU
+# general public license version 2.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+
+DOCUMENTATION = '''
+---
+module: xml
+short_description: Manage bits and pieces of XML files or strings
+description:
+  - A CRUD-like interface to managing bits of XML flies. You might also be interested in a brief tutorial, U(http://www.w3schools.com/xpath/). Note that module this does not handle complicated xpath expressions. So limit xpath selectors to simple expressions.
+version_added: "1.0"
+options:
+  file:
+    description:
+      - Path to the file to operate on. File must exist ahead of time.
+    required: true unless xmlstring is given
+    default: null
+    choices: []
+  xmlstring:
+    description:
+      - A string containing XML on which to operate.
+    required: true unless file is given
+    default: null
+    choices: []
+  xpath:
+    description:
+      - A valid XPath expression describing the item(s) you want to manipulate. Operates on the document root, C(/), by default.
+    required: false
+    default: /
+    choices: []
+  namespaces:
+    description:
+      - The namespace prefix:uri mapping for the XPath expression. Needs to be a *map*, not a list of items.
+    required: false
+    default: null
+    choices: []
+  ensure:
+    description:
+      - Set or remove an xpath selection (node(s), attribute(s))
+    required: false
+    default: present
+    choices:
+      - "absent"
+      - "present"
+  value:
+    description:
+      - Desired state of the selected attribute. Either a string, or to unset a value, the Python C(None) keyword (YAML Equivalent, C(null)).
+    required: false
+    default: Elements default to no value (but present). Attributes default to an empty string.
+    choices: []
+  add_children:
+    description:
+      - 'Add additional child-element(s) to a selected element. Child elements must be given in a list and each item may be either a string (ex: C(children=ansible) to add an empty C(<ansible/>) child element), or a hash where the key is an element name and the value is the element value.'
+    required: false
+    default: null
+    choices: []
+  set_children:
+    description:
+      - 'Set the the child-element(s) of a selected element. Removes any existing children. Child elements must be specified as in C(add_children).'
+    required: false
+    default: null
+    choices: []
+  count:
+    description:
+      - "Search for a given C(xpath) and provide the count of any matches"
+    required: false
+    default: null
+    choices: []
+  print_match:
+    description:
+      - "Search for a given C(xpath) and print out any matches"
+    required: false
+    default: null
+    choices: []
+  pretty_print:
+    description:
+      - "Pretty print output XML"
+    required: false
+    default: false
+    choices: []
+  content:
+    description:
+      - "Search for a given C(xpath) and get content"
+    required: false
+    default: false
+    choices:
+      - "attribute"
+      - "text"
+requirements:
+    - The remote end must have the Python C(lxml) library installed
+author: Tim Bielawa, Magnus Hedemark
+'''
+
+
+from io import BytesIO
+from lxml import etree
+try:
+    import json
+except:
+    import simplejson as json
+import lxml
+import os
+
+def print_match(module, tree, xpath, namespaces):
+    match = tree.xpath(xpath, namespaces=namespaces)
+    match_xpaths = []
+    for m in match:
+        match_xpaths.append(tree.getpath(m))
+    match_str = json.dumps(match_xpaths)
+    msg = "selector '%s' match: %s" % (xpath, match_str)
+    finish(module, tree, xpath, namespaces, changed=False, msg=msg)
+
+def count(module, tree, xpath, namespaces):
+    """ Return the count of nodes matching the xpath """
+    hits = tree.xpath("count(/%s)" % xpath, namespaces=namespaces)
+    finish(module, tree, xpath, namespaces, changed=False, msg=int(hits), hitcount=int(hits))
+
+def is_node(tree, xpath, namespaces):
+    """ Test if a given xpath matches anything and if that match is a node.
+
+    For now we just assume you're only searching for one specific thing."""
+    if xpath_matches(tree, xpath, namespaces):
+        # OK, it found something
+        match = tree.xpath(xpath, namespaces=namespaces)
+        if type(match[0]) == lxml.etree._Element:
+            return True
+
+    return False
+
+def is_attribute(tree, xpath, namespaces):
+    """ Test if a given xpath matches and that match is an attribute
+
+An xpath attribute search will only match one item"""
+    if xpath_matches(tree, xpath, namespaces):
+        match = tree.xpath(xpath, namespaces=namespaces)
+        if type(match[0]) == lxml.etree._ElementStringResult:
+            return True
+    return False
+
+def xpath_matches(tree, xpath, namespaces):
+    """ Test if a node exists """
+    if tree.xpath(xpath, namespaces=namespaces):
+        return True
+    else:
+        return False
+
+def delete_xpath_target(module, tree, xpath, namespaces):
+    """ Delete an attribute or element from a tree """
+    try:
+        for result in tree.xpath(xpath, namespaces=namespaces):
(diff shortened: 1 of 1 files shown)


abadger:
This probably needs to go to ansible-modules-extras first and then be brought into core if it proves to be a module that is robust and widely needed.

I'll post some comments against the code here since I'm going away for a few days of vacation and might not see it when you repost it in ansible-modules-extras.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+
+DOCUMENTATION = '''
+---
+module: xml
+short_description: Manage bits and pieces of XML files or strings
+description:
+  - A CRUD-like interface to managing bits of XML flies. You might also be interested in a brief tutorial, U(http://www.w3schools.com/xpath/). Note that module this does not handle complicated xpath expressions. So limit xpath selectors to simple expressions.
+version_added: "1.0"
```
abadger, line 23: version_added is the version it's first shipped with ansible.  At the moment we're working on version 2.2.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+# -*- coding: utf-8 -*-
+#
+# xml - Manage bits and pieces of XML files
+#
+# Copyright 2014, Red Hat, Inc.
+# Tim Bielawa <tbielawa@redhat.com>
+# Magnus Hedemark <mhedemar@redhat.com>
+#
+# This software may be freely redistributed under the terms of the GNU
+# general public license version 2.
```
abadger, line 11: I believe that all modules shipped with ansible itself need to be licensed GPLv3 or later.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+module: xml
+short_description: Manage bits and pieces of XML files or strings
+description:
+  - A CRUD-like interface to managing bits of XML flies. You might also be interested in a brief tutorial, U(http://www.w3schools.com/xpath/). Note that module this does not handle complicated xpath expressions. So limit xpath selectors to simple expressions.
+version_added: "1.0"
+options:
+  file:
+    description:
+      - Path to the file to operate on. File must exist ahead of time.
+    required: true unless xmlstring is given
```
abadger, line 28: I think that the doc generator parses this field as a boolean value  So the explanation of when this is optinal and when it is required needs to go in the parameter's description.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+    description:
+      - "Search for a given C(xpath) and get content"
+    required: false
+    default: false
+    choices:
+      - "attribute"
+      - "text"
+requirements:
+    - The remote end must have the Python C(lxml) library installed
+author: Tim Bielawa, Magnus Hedemark
```
abadger, line 103: This is a list field so it's better written as:

```
author:
  - Tim Bielawa
  - Magnus Hedemark
```

Also, it's good to put github IDs in here so that the right people can be notified when bugs or PR reviews against this module are needed.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+'''
+
+
+from io import BytesIO
```
abadger, line 107: Currently, unless there's a library dependency which requires a more recent version of python, we make modules compatible with python-2.4 and above.  If there is a dependency on a library which needs a more recent version of python then we need to list that in the documentation's requirements section.

lxml is a hard one to evaluate.  It's old enough that there is a version for python2.4 but the current version does not support python2.4.  Are you using any of the newer lxml features or will this work fine with the lxml versions that run on pyhton2.4?  (If you need a newer version thn that needs to be listed in the requirements documentation as well).

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+  content:
+    description:
+      - "Search for a given C(xpath) and get content"
+    required: false
+    default: false
+    choices:
+      - "attribute"
+      - "text"
+requirements:
+    - The remote end must have the Python C(lxml) library installed
```
abadger, line 102: Normally we list python libraries in the requirements section in a way that's easy and familiar for people who have worked with pip before.  So something more like this:

```
requirements:
  - lxml
```

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+author: Tim Bielawa, Magnus Hedemark
+'''
+
+
+from io import BytesIO
+from lxml import etree
```
abadger, line 108: For this and the other lxml import we need to make sure that the module won't traceback but instead will return a nice error message if the library is not installed.  To do that we follow a standard idiom that looks like this:

``` python
HAS_LXML = True
try:
    from lxml import etree
    import lxml
except ImportError:
    HAS_LXML = False
```

and then in the main() function, once we have an instance of the module we do something like this:

``` python

if not HAS_LXML:
    module.fail_json(msg='The xml ansible module needs to have the lxml python library installed on the managed machine')
```

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+    finish(module, tree, xpath, namespaces, changed=False, msg=int(hits), hitcount=int(hits))
+
+def is_node(tree, xpath, namespaces):
+    """ Test if a given xpath matches anything and if that match is a node.
+
+    For now we just assume you're only searching for one specific thing."""
+    if xpath_matches(tree, xpath, namespaces):
+        # OK, it found something
+        match = tree.xpath(xpath, namespaces=namespaces)
+        if type(match[0]) == lxml.etree._Element:
```
abadger, line 137: Instead of using type and comparing, use isinstance.  For example:

``` python
if isinstance(match[0], lxml.etree._Element):
    [...]
```

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+                    # Delete an attribute
+                    parent = result.getparent()
+                    # Pop this attribute match out of the parent
+                    # node's 'attrib' dict by using this match's
+                    # 'attrname' attribute for the key
+                    parent.attrib.pop(result.attrname)
+                elif is_node(tree, xpath, namespaces):
+                    # Delete an element
+                    result.getparent().remove(result)
+    except Exception, e:
```
abadger, line 175: making modules compatible with python3 is something that we're slowly working towards.  Part of that effort requires that exceptions be valid python 3 syntax.  This can be troublesome when a module also has to be compatible with pyhton-2.4.  So we have a helper function that can be used:

```
[[Somewhere at the top, with your imports]]
from ansible.module_utils.pycompat24 import get_exception

[[ Here, where you're catching the exception:]]
except Exception:
    e = get_excetion()
```

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+def set_target_children(module, tree, xpath, namespaces, children):
+    matches = tree.xpath(xpath, namespaces=namespaces)
+
+    # Create a list of our new children
+    children = children_to_nodes(module=module, children=children)
+    children_as_string = [lxml.etree.tostring(c) for c in children]
+
+    changed = False
+
+    def replace_children_of(match):
```
abadger, line 189: If there's no need to make this module a closure or similar, it's better to make these private toplevel functions instead of nested functions.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+
+    for element in tree.xpath(xpath, namespaces=namespaces):
+        if not attribute:
+            changed = changed or (element.text != value)
+            if not module.check_mode and (element.text != value): element.text = value
+        else:
+            changed = changed or (element.get(attribute) != value)
+            if ":" in attribute:
+                attr_ns, attr_name = attribute.split(":")
+                attribute = "{{{0}}}{1}".format(namespaces[attr_ns], attr_name)
```
abadger, line 234: format is another python-2.6+ ism.  If there's a need to make this compatible with python-2.4 it will have to change to percent formatting instead.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+def set_target(module, tree, xpath, namespaces, attribute, value):
+    changed = False
+
+    if not is_node(tree, xpath, namespaces):
+        abort(module, "Xpath " + xpath + " does not reference a node!")
+
+    for element in tree.xpath(xpath, namespaces=namespaces):
+        if not attribute:
+            changed = changed or (element.text != value)
+            if not module.check_mode and (element.text != value): element.text = value
```
abadger, line 229: Two lines is preferred here.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+            if not module.check_mode and (element.get(attribute) != value): element.set(attribute, value)
+
+    finish(module, tree, xpath, namespaces, changed)
+
+def pretty(module, tree):
+    xml_string = etree.tostring(tree, xml_declaration=True, encoding='UTF-8', pretty_print=module.params['pretty_print'])
+    changed = False
+
+    if module.params['file']:
+        xml_file = os.path.expanduser(module.params['file'])
```
abadger, line 244: Ansible's argument_spec now has a type='path' which does expanduser and expandvars for you.  That's likely a better thing to use than doing this manually.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+        return node
+    else:
+        abort(module, "Invalid child type: %s. Children must be either strings or hashes." % str(ch_type))
+
+def children_to_nodes(module=None, children=[]):
+    """turn a str/hash/list of str&hash into a list of elements"""
+    return [child_to_element(module, child) for child in children]
+
+def abort(m, msg):
+    m.fail_json(msg=msg)
```
abadger, line 318: Most Ansible modules (all the ones I can think of) are written so that functions are defined before they are called.  So this helper function and finish should likely be near the top of the module file.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+
+    ##################################################################
+    # Format the xml only?
+    if module.params['pretty_print']:
+        pretty(module, x)
+
+    abort(module, "don't know what to do")
+
+######################################################################
+from ansible.module_utils.basic import *
```
abadger, line 469: This can now be at the top of the file and no longer needs to be a wildcard import for Ansible-2.1 or later.

files/xml.py
```diff
@@ -0,0 +1,470 @@
...
+main()
```
abadger, line 470: this should be protected by the standard python if **name** == **main** conditional:

``` python
if __name__ == "__main__":
    main()
```

abadger:
@cmprescott Okay, I've gone through the code and added some basic style and guidelines compliance notes.  I haven't run the module or checked that it functions well so when other people review they may find more things to change.

Since I'm rusty on my lxml API... is this module idempotent?  If I ask to add <subnode/> to <node> in my playbook and then run the playbook twice, will I end up with (idempotent):

```
<node><subnode/></node>
```

or (nonidempotent):

```
<node><subnode/><subnode/></node>
```

?

PR #4141, opened 2016-07-08
Correct the handling of state=latest for yum groups.

### LINKED ISSUE #4119: yum module doesn't install group if 'state=latest' but does if 'state=present'
I've seen #2510 and it seems to address the _name_ of the group, not whether the state is present or latest.  Neither the note nor the Redhat docs that @abadger linked to address the issue of "state", just the group name.

##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

yum module
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### CONFIGURATION

No changes.
##### OS / ENVIRONMENT

Running ansible built via homebrew on a Mac running 10.10.5.
Targets are Digital Ocean Droplets (for these tests) running Digital Ocean's CentOS 7 image.
##### SUMMARY

Using state=latest with the yum module to install groups (e.g. the Development Group) does not actually install anything, but using state=present does.

This playbook will install the Development Group

```
- hosts: all
  tasks:
  - name: "install development group"
    yum: name="{{ item }}" state=present
    with_items:
      - '@development'
```

This one does not:

```
- hosts: all
  tasks:
  - name: "install development group"
    yum: name="{{ item }}" state=latest
    with_items:
      - '@development'
```
##### STEPS TO REPRODUCE

Create a "droplet" with Digital Ocean's CentOS 7 x64 image (a 512mb droplet is fine).

Use

```
ansible-playbook -i <droplet_ip_addr>, -u root --private-key=<path to private key> playbook.yml
```

to run either of the playbooks included above.

Log into the machine, e.g. with

```
ssh -i <path to private key> root@<droplet ip addr>
```

and execute `cc` and you should see this (when state=latest):

```
[root@yumbug ~]# cc
-bash: cc: command not found
```

or this (when state=present):

```
[root@yumbug ~]# cc
cc: fatal error: no input files
compilation terminated.
```
##### EXPECTED RESULTS

See above.
##### ACTUAL RESULTS

See above.

jctanner:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

yum.py
##### ANSIBLE VERSION

```
2.2 
```
##### SUMMARY

Fixes #4119


gregdek:
Thanks @jctanner for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

packaging/os/yum.py
```diff
@@ -909,14 +910,16 @@ def latest(module, items, repoq, yum_basecmd, conf_file, en_repos, dis_repos):
         if len(pkgs['install']) > 0:    # install missing
             cmd = yum_basecmd + ['install'] + pkgs['install']
             rc, out, err = module.run_command(cmd)
-            res['changed'] = True
+            if not out.strip().lower().endswith("no packages marked for update"):
+                res['changed'] = True
```
abadger, line 914: When screen scraping output we have to make sure that the program returns strings in the language we expect.  Some modules set the locale in one place to do that.  The yum module (probably because it calls several different programs and doesn't screenscrape all of them) sets the environment on each external call.  You can do that with something like the following:

``` python
         lang_env = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C')
         rc, out, err = module.run_command(cmd, environ_update=lang_env)
         if not out.strip().lower().endswith("no packages marked for update"):
             res['changed'] = True
```

And the same for the other place we screenscrape.

abadger:
Made one note about setting the locale before screenscraping.  Other than that, this looks good.  Feel free to merge once that's corrected.

PR #3887, opened 2016-06-08
Don't forcibly push loaded images to registry.

Jorge-Rodriguez:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

docker_image
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

There are use cases where the `docker_image` module is used to load packaged images on the target machine but there is no need to push that image to a registry, let alone forcibly and opaquely upload it to docker hub, which could happen if the target machine has docker.io configured credentials.

This PR removes the forcible push and requires an explicit request on the playbook task to do that by specifying `push: True`

Fixes #3763

PR #4171, opened 2016-07-12
Add push parameter and let API handle registry auth checking.

chouseknecht:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

docker_image.py
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 7a078317a3) last updated 2016/07/12 12:06:58 (GMT -400)
  lib/ansible/modules/core: (fix_3961 dcce0dfdfa) last updated 2016/07/12 16:32:46 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 482b1a640e) last updated 2016/07/11 17:13:47 (GMT -400)
```
##### SUMMARY

Follow up to PR #3887. The module no longer decides when to push an image, so there needs to be a `push=yes|no` parameter.

Replacement PR #3961. Removes registry auth checking, letting the API handle it. Adds better API error handling, letting the user know when registry authentication is required to fix the issue.

### DIFF
--- a/cloud/docker/docker_image.py
+++ b/cloud/docker/docker_image.py
@@ -32,26 +32,26 @@ description:
 options:
   archive_path:
     description:
-      - Use with state 'present' to archive an image to a .tar file.
+      - Use with state C(presen) to archive an image to a .tar file.
     required: false
-    default: null
     version_added: "2.1"
   dockerfile:
     description:
-      - Use with state 'present' to provide an alternate name for the Dockerfile to use when building an image.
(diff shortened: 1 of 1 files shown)


cloud/docker/docker_image.py
```diff
@@ -32,26 +32,26 @@
 options:
   archive_path:
     description:
-      - Use with state 'present' to archive an image to a .tar file.
+      - Use with state C(presen) to archive an image to a .tar file.
```
gotgenes, line 35: Typo on "presen" (should be "present").

PR #4190, opened 2016-07-14
Add states for stopping and starting machines

jbaublitz:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

gce
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

This will allow nodes to be started and stopped iteratively and synchronously. I have added two helper functions modeled after the deletion helper function using libcloud functionality to start and stop one or more machines in a loop. I hope to eventually work towards asynchronous iterative starting, stopping, creation, and destruction in a later pull request as this will potentially require libcloud contributions as well.

```
    "changed": true, 
    "instance_names": [
        "gce-test0", 
        "gce-test1"
    ], 
    "invocation": {
        "module_args": {
            "credentials_file": null, 
            "disk_auto_delete": true, 
            "disks": null, 
            "external_ip": "ephemeral", 
            "image": "debian-7", 
            "instance_names": "gce-test0,gce-test1", 
            "ip_forward": false, 
            "machine_type": "n1-standard-1", 
            "metadata": null, 
            "name": null, 
            "network": "default", 
            "pem_file": null, 
            "persistent_boot_disk": false, 
            "preemptible": null, 
            "project_id": null, 
            "service_account_email": null, 
            "service_account_permissions": null, 
            "state": "absent", 
            "subnetwork": null, 
            "tags": null, 
            "zone": "us-central1-a"
        }
    }, 
    "state": "absent", 
    "zone": "us-central1-a"
```

/cc @ryansb @erjohnso

### DIFF
--- a/cloud/google/gce.py
+++ b/cloud/google/gce.py
@@ -121,7 +121,7 @@ options:
       - desired state of the resource
     required: false
     default: "present"
-    choices: ["active", "present", "absent", "deleted"]
+    choices: ["active", "present", "absent", "deleted", "started", "stopped", "terminated"]
   tags:
     description:
       - a comma-separated list of tags to associate with the instance
@@ -502,6 +502,59 @@ def create_instances(module, gce, instance_names):
 
     return (changed, instance_json_data, instance_names)
 
+def start_instances(module, gce, instance_names, zone_name):
+    """Starts a list of stopped instances.
+
+    module: Ansible module object
+    gce: authenticated GCE connection object
+    instance_names: a list of instance names to start
+    zone_name: the zone where the instances reside prior to termination
+
+    Returns a dictionary of instance names that were started.
+    """
+    changed = False
+    started_instance_names = []
+    for name in instance_names:
+        inst = None
+        try:
+            inst = gce.ex_get_node(name, zone_name)
+        except ResourceNotFoundError:
+            pass
+        except Exception as e:
+            module.fail_json(msg=unexpected_error_msg(e), changed=False)
+        if inst and inst.state == libcloud.compute.types.NodeState.STOPPED:
+            gce.ex_start_node(inst)
+            started_instance_names.append(inst.name)
+            changed = True
+
+    return (changed, started_instance_names)
+
+def stop_instances(module, gce, instance_names, zone_name):
+    """Stops a list of instances.
+
+    module: Ansible module object
+    gce: authenticated GCE connection object
+    instance_names: a list of instance names to stop
+    zone_name: the zone where the instances reside prior to termination
+
+    Returns a dictionary of instance names that were stopped.
+    """
+    changed = False
+    stopped_instance_names = []
+    for name in instance_names:
+        inst = None
+        try:
+            inst = gce.ex_get_node(name, zone_name)
+        except ResourceNotFoundError:
+            pass
+        except Exception as e:
+            module.fail_json(msg=unexpected_error_msg(e), changed=False)
+        if inst and inst.state == libcloud.compute.types.NodeState.RUNNING:
+            gce.ex_stop_node(inst)
+            stopped_instance_names.append(inst.name)
+            changed = True
+
+    return (changed, stopped_instance_names)
 
 def terminate_instances(module, gce, instance_names, zone_name):
     """Terminates a list of instances.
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @jbaublitz. To the current maintainers, @erjohnso please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jbaublitz:
/cc @erjohnso @supertom @ryansb 

So the libcloud changes that I mentioned in the pull request originally are almost done and have taken much less time than expected. I want to start working on asynchronous iterative creation, destruction, starting and stopping of nodes in ansible. Currently provisioning multiple nodes is blocking on each node to complete and I would like to use the existing libcloud functions and changes I have made to speed this up. Two questions:
1. Would this be better suited for another pull request?
2. Because it will probably take some time for the libcloud pull request to be reviewed and incorporated into trunk for libcloud, should I wait until it is officially approved to submit changes to ansible? This seems like the most reasonable solution but because I'm looking to have these changes in 2.2, I would like to submit them as soon as possible. While this may not be ideal, I can also just link to the pull request in libcloud once I create it even if it has not been accepted yet so that anyone looking to use it can patch custom builds for now.

Thoughts?

supertom:
Hi @jbaublitz,

Thanks for the PR!  I'll have a look.

My thoughts on your questions:
1. I would say so.  My gut tells me there will be some discussion over it's implementation and what feedback is provided to the user.  Plus, what you have now will provide new, needed functionality now for users, so I'm inclined to do it in two phases.
2. I agree with you, submit it as soon as there is a PR to point to in libcloud.  I suspect we'll need to had a version guard at some point to your Ansible change when we know which version it's included in.

Thanks again!

cloud/google/gce.py
```diff
@@ -502,6 +502,59 @@ def create_instances(module, gce, instance_names):
 
     return (changed, instance_json_data, instance_names)
 
+def start_instances(module, gce, instance_names, zone_name):
```
supertom: Think we can get it into one function?  The two functions are kinda boilerplate.  I don't see a proliferation of states anytime soon but one function would be easier to maintain.

jbaublitz: Ok I was wondering about that. I will update.

jbaublitz:
@supertom I have collapsed all of the repetitive functions into a `change_instance_state` function (which was previously `terminate_instances`) that I tested on the following states:

```
['absent', 'deleted', 'started', 'stopped', 'terminated']
```

According to my tests it works for everything but creation which of course needs a separate function. Does this look okay?

supertom:
Thanks for the concise function, @jbaublitz!

I'm a little concerned about using the libcloud NodeState here.  Which version of libcloud are you using for testing?

I ask because in version [0.20.1](https://github.com/apache/libcloud/blob/87deb04498ea8bb5068e362810c3d2ab89359b48/libcloud/compute/types.py#L207), NodeState is an Int, and in [1.0+](https://github.com/apache/libcloud/blob/a47a51e14e5300f3fd1b21533c407086e92a3cd1/libcloud/compute/types.py#L262), it is a String.

It looks like it should do the [right thing](https://github.com/apache/libcloud/blob/a47a51e14e5300f3fd1b21533c407086e92a3cd1/libcloud/compute/drivers/gce.py#L1018) but I think we should confirm before merging.  SG?

jbaublitz:
I have been testing in version 1.0+. Let me know if you need me to test an earlier version because I have access to version 0.20 through CentOS I believe.

supertom:
I'd appreciate it if you could try it out.  If you use a virtualenv, you can then install it with pip install apache-libcloud==0.20.1.

I could do it next week, but if there is a problem, you'd be setting up an environment to test it anyway. ;-)

jbaublitz:
Okay absolutely! I'll test this weekend and will report back.

jbaublitz:
@supertom I'm having some trouble with both libcloud 0.20 and 1.0 unrelated to my changes. It is telling me I have insufficient permissions on the Google Cloud node I'm running ansible on but it works perfectly fine on my local machine. Not sure what the problem is because I am using the same invocation and credentials but I will have to keep debugging before I can give you an answer. I am currently able to do what I need to do through gcloud commands so it seems to be a libcloud problem.

supertom:
@jbaublitz I appreciate you working this.  

Sounds to me like it might be a scopes/role issue.  What scopes does the VM have?
`gcloud compute instances describe <instance_name>`
(you can also look in the UI for this info.)

You can also check the role permissions on your Service Account, but since it worked for you locally, I don't think that's the issue from what you describe.

jbaublitz:
@supertom Thank you for the tip! I recreated the VM with service account permissions and it is now working. I don't usually create VMs from the console so it seems that that was the issue. I can now confirm that this works in libcloud 0.20 on Ubuntu 16.04. It both changes state and correctly reports compliance check results for `started` and `stopped`.

supertom:
LGTM - shipit!

Thanks @jbaublitz for contributing!

jbaublitz:
@ryansb Could you take a look at this? Thanks!

PR #4216, opened 2016-07-20
Allow Cloudformation use file contents directly

benbarclay:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

cloudformation
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

This allows a user to give the cloudformation module a string that contains the contents of a cloudformation template.
This allows a user to use a lookup plugin to ensure that a cloudformation role can be distributed as a plugin without having to hardcode the full path, or upload to a remote s3 bucket.

This was discussed in [1460](https://github.com/ansible/ansible-modules-core/issues/1460)

To be entirely honest, not really sure how to test this....

### DIFF
--- a/cloud/amazon/cloudformation.py
+++ b/cloud/amazon/cloudformation.py
@@ -78,6 +78,11 @@ options:
       - Location of file containing the template body. The URL must point to a template (max size 307,200 bytes) located in an S3 bucket in the same region as the stack. This parameter is mutually exclusive with 'template'. Either one of them is required if "state" parameter is "present"
     required: false
     version_added: "2.0"
+  template_contents:
+    description:
+      - Contents of a template. This is intended to be used with a lookup plugin to ensure role portability.
+    required: false
+    version_added: "2.0"
   template_format:
     description:
     - For local templates, allows specification of json or yaml format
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @benbarclay. To the current maintainers, @jsmartin please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

cloud/amazon/cloudformation.py
```diff
@@ -143,6 +148,21 @@
       ClusterSize: 3
     tags:
       Stack: ansible-cloudformation
+
+# Use a template with a lookup
+- name: launch ansible cloudformation example
+  cloudformation:
+    stack_name="ansible-cloudformation" state=present
+    region=us-east-1 disable_rollback=true
+    template_contents="{{ lookup('file', 'cloudformation.json') }}"
```
ryansb, line 157: This example would actually fail because the JSON file contents would be parsed into a Python type.

ryansb:
I'm not sure adding this option is the way to go. inline JSON is usually a pain (see https://github.com/ansible/ansible/issues/7005#issuecomment-42894813 ) and the `iam_policy` module's `policy_json` option. 

Maybe a better way would be to use a variable to store the template path, so the existing option could still be used. You could even use the `template` lookup to make conditional decisions about the file path.

benbarclay:
Makes sense to me. I'll close this ticket.

PR #4233, opened 2016-07-21
Remove git and synchronize modules from list of commands we want user…

abadger:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

command.py
##### ANSIBLE VERSION

```
devel
```
##### SUMMARY

Some of the modules we tell people to use cannot approach the full functionality of the command line tool so don't tell people that they should be using the modules instead of the tool.  Also, synchronize doesn't work the way most other modules do wrt delegate so we shouldn't encourage people to use it if they alrady know what they're doing with rsync.


abadger:
@jimi-c gave a plus one to the concept here already.   Once tests pass I'll merge this.

PR #4243, opened 2016-07-23
Git: Determine if remote URL is being changed

### LINKED ISSUE #4006: git module changes repo URL and returns "changed": false when run with --check
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

git module
##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.1.0.0
  config file = 
```
##### CONFIGURATION

```
[defaults]
nocows = yes
vault_password_file = .vault-password
allow_world_readable_tmpfiles = yes
```
##### OS / ENVIRONMENT
- control host: OS X 10.11.5
- managed host: Debian 7, Debian 8
##### SUMMARY

git module changes repo URL and returns "changed": false when run with --check
##### STEPS TO REPRODUCE

``` yaml
# git-clone.yml

- hosts: localhost
  gather_facts: no
  tasks:
  - local_action: git repo={{url}} dest=ansible version=v1.0
```

```
$ ansible-playbook git-clone.yml -e url=https://github.com/ansible/ansible.git -v

TASK [git] *********************************************************************
changed: [localhost -> localhost] => {"after": "6a64e9f0248d402aa60faf33bc720acbcd56e50b", "before": null, "changed": true}

$ git -C ansible remote -v
origin  https://github.com/ansible/ansible.git (fetch)
origin  https://github.com/ansible/ansible.git (push)

$ ansible-playbook git-clone.yml -e url=https://github.com/famousgarkin/ansible.git -v --check

TASK [git] *********************************************************************
ok: [localhost -> localhost] => {"after": "6a64e9f0248d402aa60faf33bc720acbcd56e50b", "before": "6a64e9f0248d402aa60faf33bc720acbcd56e50b", "changed": false}

localhost                  : ok=1    changed=0    unreachable=0    failed=0   

$ git -C ansible remote -v
origin  https://github.com/famousgarkin/ansible.git (fetch)
origin  https://github.com/famousgarkin/ansible.git (push)
```
##### EXPECTED RESULTS
- returns `"changed": true`
- repo URL is not changed
##### ACTUAL RESULTS
- returns `"changed": false`
- repo URL is changed

rcutmore:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

Git
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 134d70e7b9) last updated 2016/07/18 07:21:22 (GMT -400)
  lib/ansible/modules/core: (issue-4006 74a98e6e6f) last updated 2016/07/22 08:01:42 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD 68ca157f3b) last updated 2016/07/18 07:22:17 (GMT -400)
  config file = 
```
##### SUMMARY

Previously the remote source was set regardless of whether it needed to be changed or not. It was also set even if the command was run in check mode. If the remote source was the only thing being changed the reported _changed_ result was incorrect as well. This update determines if the remote source is actually changing and reports that correctly. If the command is run in check mode the remote source is no longer changed.

Example changing remote source from https://github.com/ansible/ansible.git to https://github.com/rcutmore/ansible.git, without any other changes to repo.

Before:

```
ok: [localhost -> localhost] => {"after": "6a64e9f0248d402aa60faf33bc720acbcd56e50b", "before": "6a64e9f0248d402aa60faf33bc720acbcd56e50b", "changed": false, "warnings": []}
```

After:

```
changed: [localhost -> localhost] => {"after": "6a64e9f0248d402aa60faf33bc720acbcd56e50b", "before": "6a64e9f0248d402aa60faf33bc720acbcd56e50b", "changed": true, "warnings": []}
```

Ansible reported there were no changes when only the remote URL for a
repo was changed. This properly tracks and reports when the remote URL
for a repo changes.

Fixes #4006


gregdek:
Thanks @rcutmore for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

robinro:
@rcutmore 
Thanks for your work to make the git module better. It appears those break the existing integration tests. On a quick look it seems your change does not handle local repo paths correctly and always reports changed on those.

Also it would be nice to have integration tests for the issue you are fixing here. 
You can make use of per-task checkmode to write integration tests, see https://github.com/ansible/ansible/pull/16056

I'd be happy to test your changes and help if you want.

rcutmore:
@robinro Thanks for taking a look and the suggestions, I appreciate it.

It looks like the tests are failing on [this](https://github.com/ansible/ansible/blob/devel/test/integration/roles/test_git/tasks/main.yml#L565) assertion. I get the same failure when running the integration tests locally, but I think I'm missing something or reading the test results incorrectly. The test is asserting that changed is false and that appears to be the result, but it still fails (see below). 

`fatal: [testhost]: FAILED! => {"assertion": "not update_new_tag|changed", "changed": false, "evaluated_to": false, "failed": true}`

I get the correct _changed_ result (either changed or not changed) for local repo paths when manually running commands outside the test suite. I haven't been able to track down what's causing the test failure but I'll keep working on it.

robinro:
@rcutmore 

The actual problem is not in the assertion task, but in the task before that registers update_new_tag. 

> 2016-07-23 13:58:05 TASK [test_git : update copy with new tag] *************************************
> 2016-07-23 13:58:06 changed: [testhost] => {"after": "4f2148c62783a6df42a2ae7e5436b8b2ce4deb1f", "before": "4f2148c62783a6df42a2ae7e5436b8b2ce4deb1f", "changed": true, "warnings": []}

This should be `changed: false`, since it updates the repo and fetches a tag, but doesn't change the checkout. With your PR this is `changed: true`. See for example in some previous git run.

rcutmore:
Thanks, that makes sense and helps me.

rcutmore:
I pushed a change to fix the handling of local repo paths but there's still an issue with the CentOS 6 tests. I believe this is because `git ls-remote --get-url` isn't available in that version of Git, so I'll work on a different approach for fetching the URL.

robinro:
Thanks for the fixes @rcutmore 

Does `git remote get-url REMOTE` work on the old version? `ls-remote` is worse for performance, since it connects to the remote, which is not needed as far as I understand the code.

Instead of importing the os module, you could also use https://github.com/ansible/ansible/blob/3afe50dfe2d58586b119e2bd390c0c422a1bcda0/lib/ansible/utils/path.py#L27, which covers a few more exotic cases for the path.

rcutmore:
@robinro Thanks, when I'm able to I'll push a change to use `unfrackpath` instead of importing the os module.

No, it looks like `git remote get-url REMOTE` was added fairly recently so it won't work. I'm considering parsing the output of `git remote show -n REMOTE` instead. This should be supported and using the -n switch avoids connecting to the remote. Does that sound reasonable?

robinro:
@rcutmore 
I in general don't like output-parsing. It's fairly fragile (stdout content changes more often then API, and we already have a problem with the API). Furthermore it increases the complexity of the module, by re-implementing features available in newer git versions.

Personally I'd leave the bug unfixed for old git versions if there is no easy way to do it and exclude those versions from the test.

rcutmore:
@robinro Agreed, I just didn't think that would be an option.

I'll work on the aforementioned change as well as using `git remote get-url REMOTE` to get the remote URL (unless I can find a better alternative). Then I'll submit some integration tests to the Ansible repo, excluding unsupported versions in the appropriate tests.

rcutmore:
I updated this PR with the required changes.

The `git remote get-url REMOTE` command is now used to fetch the remote URL. When this command isn't available (Git versions prior to 2.7) then `None` is returned as the URL, causing the remote URL to be set and the repo to be reported as changed.

I also submitted a PR to the Ansible repo to update and add Git integration tests [here](https://github.com/ansible/ansible/pull/16893).

I believe this is now ready_for_review.

robinro:
`git remote get-url` was only added in git 2.7? Oh my...

According to https://git-scm.com/docs/git-ls-remote/1.7.12.2 `git ls-remote --get-url` is available down to 1.7, so I should revert my earlier comment and suggest to use `ls-remote`. Sorry for the confusion.

rcutmore:
No problem, I'll make that change.

rcutmore:
This is now updated to use `ls-remote --get-url`, so I think it should be good to go.

gregdek:
Thanks @rcutmore for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

rcutmore:
@gregdek The tests for this fail because there's a related PR ([here](https://github.com/ansible/ansible/pull/16893)) needed to fix the tests as well. Is there anything else I should be doing?

robinro:
The old test failing is not really necessary:
You can change line 520 to
`if remote_url and (remote_url == repo or remote_url == unfrackpath(repo)):`
Then False is only returned if `remote_url` is not empty. By doing so you can run the original test again and can probably remove all version tests for get_url in your new tests.

rcutmore:
@robinro I think that would only change the functionality if `repo` happens to be None. From what I can tell this parameter can't be None so adding the `remote_url` check before the equality tests wouldn't change anything.

This gives me an idea though. Instead of returning True on [this line](https://github.com/rcutmore/ansible-modules-core/blob/a705bb7348315c66cf016937455bd3c84c926fa5/source_control/git.py#L528) we can check whether or not `remote_url` is None. If it's None then we return False (maintain original behavior for Git versions prior to 1.7.5), otherwise True is returned. Then we should be able to remove the version checks in the tests.

I'll knock this out when I have time this weekend.

rcutmore:
ready_for_review

robinro:
looks good to me

gregdek:
Thanks @rcutmore for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

jctanner:
@rcutmore a few merges earlier today caused merged conflicts in your PR. I fixed them, but I'm not really sure I can squash+merge without loosing attribution to you. Could you rebase and fix your branch?

My diff is here if you need a reference for whatever reason: https://github.com/ansible/ansible-modules-core/pull/4394

gregdek:
Thanks @rcutmore for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

rcutmore:
@jctanner No problem, I think this should take care of it.

ready_for_review

gregdek:
Thanks @rcutmore for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

likg:
This PR introduced new bug which prevents work of git module on remote hosts caused by [L209: from ansible.utils.path import unfrackpath](https://github.com/ansible/ansible-modules-core/blob/devel/source_control/git.py#L209) because ansible.utils.path is not available on remote hosts.

mattclay:
@likg Please open an issue to report the bug and reference this PR.

jctanner:
@mattclay & @likg I reproduced the error and opened an issue https://github.com/ansible/ansible-modules-core/issues/4425

PR #4254, opened 2016-07-26
Add rule revocation capabilities to ec2_group

### LINKED ISSUE #4251: ec2_group should allow for security group revocations
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME

ec2_group
##### ANSIBLE VERSION

ansible 2.1.0.0
##### OS / ENVIRONMENT

N/A
##### SUMMARY

ec2_group allows us to create security groups, to remove security groups and to add new rules to security groups. What it doesn't allow us to do is to remove rules, so we can't easily revoke rules that have been set. This feature is already available in AWS and boto3 ([relevant documentation](https://boto3.readthedocs.io/en/latest/reference/services/ec2.html#EC2.SecurityGroup.revoke_ingress).
##### STEPS TO REPRODUCE

Let's say you have a playbook that launches a new EC2 instance. That role also adds a new rule to an existing ELB, so that the new EC2 instance can access it. If you now have a `cleanup` playbook that destroys that EC2 instance, you ideally want to remove the rule you just added to the ELB. A simple revoke would be ideal, but that feature is not available.

We could have two new options to the module, `revoke_rules` and `revoke_rules_egress`, to take care of this.

Sodki:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

ec2_group
##### ANSIBLE VERSION

```
ansible 2.1.0.0
```
##### SUMMARY

ec2_group allows us to create security groups, to remove security groups and to add new rules to security groups. What it doesn't allow us to do is to remove rules, so we can't easily revoke rules that have been set. With this change we can.

Fixes #4251

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -67,6 +67,18 @@ options:
     required: false
     default: 'true'
     aliases: []
+  revoke_rules:
+    version_added: "2.2"
+    description:
+      - List of firewall inbound rules to revoke in this group (see example). If none are supplied, nothing will change.
+    required: false
+    default: []
+  revoke_rules_egress:
+    version_added: "2.2"
+    description:
+      - List of firewall outbound rules to revoke in this group (see example). If none are supplied, nothing will change.
+    required: false
+    default: []
 
 extends_documentation_fragment:
     - aws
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @Sodki for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

ryansb:
Is there a reason not to list all the rules in the task, then set the `purge_rules` flag so it removes all rules that are nonexistent in your spec?

Sodki:
The reason is that we might not know all the rules. Consider a resource that might be accessed by different roles and that you want the roles to be able to clean up after themselves. So role A adds a new rule, but knows nothing about the requirements for role B or C. When role A is doing the clean up, it just wants to remove the rule that is specific to role A, with zero knowledge about the rest.

PR #4262, opened 2016-07-27
get_url: always download on local checksum mismatch

mmatuska:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

network/basics/get_url
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel aece03312f) last updated 2016/07/27 15:28:05 (GMT +200)
  lib/ansible/modules/core: (detached HEAD e09ab5bda3) last updated 2016/07/27 15:00:34 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 24812da6e8) last updated 2016/07/27 15:00:34 (GMT +200)
```
##### SUMMARY

When the code path for setting last_mod_time is reached, it is either force or we have a mismatching checksum.

a) mismatching checksum
setting last_mod_time causes serious misbehavior. The clients sents the last-mod-time to the server and this will usually be newer than the one on the server so no download happens. We ALWAYS need to download when the checksum does not match.

b) force
setting last_mod_time makes no sense if we force the download anyway

Therefore I recommend removing this code part altogether. Please merge this to 2.1!


gregdek:
Thanks @mmatuska. To the current maintainers, @jpmens please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jpmens:
shipit

gregdek:
Thanks again to @mmatuska for this PR, and thanks @jpmens for reviewing. Marking for inclusion.

abadger:
@jimi-c asked for input on this.  It looks good to me.  It's a little confusing as we're loading two meanings onto force here.  get_url.py has a force parameter so that the user can tell the module to force download regardless of any other factors.  Then fetch_url() has a force parameter that means, do the download instead of trying to be smart about timestamps.  If checksums mismatch we have to send force=True to fetch_url() to force it to download and changing force here as this PR does is one way to do that.

shipit

PR #4314, opened 2016-08-03
unsafe writes!

### LINKED ISSUE #4188: Redeploying with template/copy doesn't update file inside docker container
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

template module
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = 
```
##### CONFIGURATION

No changes
##### OS / ENVIRONMENT

Debian GNU/Linux 8.5 (jessie)
##### SUMMARY

The suggestion is related to https://github.com/docker/docker/issues/15793 where a file mounted in a container doesn't update due to how Ansible replaces the file.
In short, when Ansible replaces a file it creates a new inode and replaces the old file with that. Docker keeps the previous inode as the mountpoint which means that the content of the file in the container never gets updated.
##### STEPS TO REPRODUCE

Create a playbook with the tasks below.
Create a file with some content named host_file.j2
Run the playbook.
Verify that the file in the container have the correct content.
$ docker exec test cat /tmp/container_file
Modify the content of the file host_file.j2
Run the playbook again.
See that the file in the container doesn't have the new content.
$ docker exec test cat /tmp/container_file

```
- name: "Create file"
  template:
    src=file.j2
    dest=/tmp/host_file

- name: "Start container"
  docker_container:
    image: debian
    name: test
    state: started
    volumes:
      - /tmp/host_file:/tmp/container_file
    entrypoint: [bash,-c,'sleep infinity']
```

##### EXPECTED RESULTS

The docker mounted file is updated with the content of the newly deployed file.
This can be accomplished by first deploy to a intermediate file and then use a cp command (cp intermediate_file mounted_file) which replaces the content and does not create a new inode.
I'd guess that we would get the same result if Ansible transfers the file to the host as a temporary file then copies it to the proper destination and finally removes the temporary file. It's not as atomic as what I think is happening now (mv) but it will work with files mounted in a docker container.

Maybe an option in template/copy module to use cp+rm instead of mv is a way forward without changing old behavior and breaking existing playbooks?

```
- name: "Create file"
  template:
    src=file.j2
    dest=/tmp/host_file.tmp

- name: "Copy file"
  command:
    cp /tmp/host_file.tmp /tmp/host_file

- name: "Start container"
  docker_container:
    image: debian
    name: test
    state: started
    volumes:
      - /tmp/host_file:/tmp/container_file
    entrypoint: [bash,-c,'sleep infinity']
```
##### ACTUAL RESULTS

The docker mounted file is NOT updated with the content of the newly deployed file. This is because Ansible renames(?) the file which means that the file reference is pointing to a new inode, but the old inode is the one still mounted in the docker container.

bcoca:
##### ISSUE TYPE
- Feature Pull Request
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

allow for using bad methods for writing to files, when we have no other option

fixes #4188, https://github.com/ansible/ansible/issues/13652

depends on https://github.com/ansible/ansible/pull/17016

### DIFF
--- a/files/assemble.py
+++ b/files/assemble.py
@@ -235,7 +235,7 @@ def main():
         if backup and dest_hash is not None:
             result['backup_file'] = module.backup_local(dest)
 
-        module.atomic_move(path, dest)
+        module.atomic_move(path, dest, unsafe_writes=module.params['unsafe_writes'])
         changed = True
 
     cleanup(path, result)
--- a/files/copy.py
+++ b/files/copy.py
@@ -92,6 +92,17 @@ options:
     version_added: "1.8"
     description:
       - 'This flag indicates that filesystem links, if they exist, should be followed.'
+  unsafe_writes:
+    description:
+     -  Normally this module uses atomic operations to prevent data corruption or inconsistent reads from the target files,
+        sometimes systems are configured or just broken in ways that prevent this. One example are docker mounted files,
+        they cannot be updated atomically and can only be done in an unsafe manner.
+     -  This option allows ansible to fall back to unsafe methods of updating files for those cases in which you do not have any other choice.
+        Be aware that this is subject to race conditions and can lead to data corruption.
+    choices: [ "yes", "no" ]
+    required: false
+    default: "no"
+    version_added: "2.2"
 extends_documentation_fragment:
     - files
     - validate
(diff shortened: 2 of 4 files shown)


abadger:
Cautious +1.  Since the user has to specify they want unsafe writes in a parameter I think this is okay.

gregdek:
Thanks again to @bcoca for this PR, and thanks @ansible for reviewing. Marking for inclusion.

gregdek:

PR #4326, opened 2016-08-04
junos: Docs update

gundalow:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

junos_
##### ANSIBLE VERSION

devel
##### SUMMARY

More docs updates

`O(...)` isn't used anywhere else

---

This change is [Reviewable](https://reviewable.io/reviews/ansible/ansible-modules-core/4326)

### DIFF
--- a/network/junos/junos_config.py
+++ b/network/junos/junos_config.py
@@ -5,7 +5,7 @@
 # Ansible is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+# (at your option) any later version.choices
 #
 # Ansible is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
@@ -56,9 +56,6 @@ options:
         argument will effectively remove all current configuration
         statements on the remote device.
     required: false
-    choices:
-        - yes
-        - no
     default: null
   confirm:
     description:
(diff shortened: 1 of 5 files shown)


gregdek:
Thanks again to @gundalow. This module is going into community review.

network/junos/junos_config.py
```diff
@@ -77,9 +80,12 @@
     description:
       - The C(replace) argument will instruct the remote device to
         replace the current configuration hierarchy with the one specified
-        in the corresponding hierarchy of the source configuraiton loaded
+        in the corresponding hierarchy of the source configuration loaded
         from this module.
     required: true
+    choices:
```
privateip, line 83: have we standardized on true, false vs yes, no for bool choices?

bcoca, line 83: We normally don't add 'choices' to boolean arguments, just state that it is a boolean.

YAML itself uses a list ([y, yes, on, 1, true] and case variants)  to convert into True/False, we try to match that in jinja2 and ansible module processing.

gundalow:
> We normally don't add 'choices' to boolean arguments, just state that it is a boolean.
> 
> YAML itself uses a list ([y, yes, on, 1, true] and case variants) to convert into True/False, we try to match that in jinja2 and ansible module processing.

ergh, this is inconsistent all all over the modules in both repos, including some things listed as saying they support `B, O, O, L, E, A, N, A` _headdesk_, though I think I've fixed all of those.

So I'll remove `choices:` that I edited.

I'd be tempted to raise a PR to make this consistent across all modules, but that's a discussion we need to have in the Core Team.

PR #4332, opened 2016-08-04
Docker service plugin

chouseknecht:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

docker_service.py 
##### ANSIBLE VERSION

```
ansible 2.2.0 (docker_service_plugin 127446b460) last updated 2016/08/04 12:58:37 (GMT -400)
  lib/ansible/modules/core: (detached HEAD 24db4de245) last updated 2016/08/02 04:22:23 (GMT -400)
  lib/ansible/modules/extras: (detached HEAD db979dde74) last updated 2016/08/02 04:22:23 (GMT -400)
```
##### SUMMARY

Relies on [PR #16954](https://github.com/ansible/ansible/pull/16954). Adds _dest_ parameter to support copying _project_src_ directory from the control host to _dest_ on target host.

From commit message:

> Use when deploying a project from a remote host. Requires docker_service plugin. When project_src and dest are provided, contents of project_src will be recursively copied from the control host to dest on the target host. The module is then executed with project_src set to the value of dest.
> 
> Only contents of project_src are copied. If the docker-compose.yml file (other other project files) reference files outside of project_src, those files will not be copied.
> 
> Expects dest to be the final directory name. For example, if project_src basename is _foo_, and the desired basename of dest is _foo_, include _foo_ in dest value.

### DIFF
--- a/cloud/docker/docker_service.py
+++ b/cloud/docker/docker_service.py
@@ -42,6 +42,16 @@ options:
         - Required when no C(definition) is provided.
       type: path
       required: false
+  dest:
+      description:
+        - Use when running docker_service on a remote host to recursively copy the contents of C(project_src) from
+          the control host into C(dest) on the remote host. Only files found in C(project_src) are copied. Anything
+          referenced in docker-compose.yml that may exist outside of C(project_src) will not be copied.
+        - Supply the final directory path. For example, if the basename of C(project_src) is I(foo), and the
+          desired basename of C(dest) is also C(foo), then include C(foo) in C(dest).
+      type: path
+      required: false
+      version_added: "2.2"
   project_name:
       description:
         - Provide a project name. If not provided, the project name is taken from the basename of C(project_src).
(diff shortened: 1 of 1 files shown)


cloud/docker/docker_service.py
```diff
@@ -529,6 +535,9 @@ def __init__(self, client):
         if self.files:
             self.options[u'--file'] = self.files
 
+        if self.remote_src and not self.project_src:
+            self.client.fail("Parameter error: project_src required when specifying remote_src.")
```
bcoca: just use required_if argspec function

cloud/docker/docker_service.py
```diff
@@ -42,6 +42,12 @@
         - Required when no C(definition) is provided.
       type: path
       required: false
+  remote_src:
+      description:
+        - Use when running docker_service on a remote host to copy C(project_src) from the control host to
+          C(remote_src) on the remote host.
+      type: path
+      required: false
```
bcoca, line 53: version_added?

PR #4037, opened 2016-06-24
Detect gnu versus bsd tar and act accordingly

jctanner:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

unarchive.py
##### ANSIBLE VERSION

```
2.2
```
##### SUMMARY

Detect if tar is gnu or bsd and then set the appropriate commands.

PR #4352, opened 2016-08-06
Detect tar type (bsd, gnu) and only use gnu tar. 

mattclay:
##### ISSUE TYPE

Bugfix Pull Request
##### COMPONENT NAME

unarchive
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 72cca01cd4) last updated 2016/08/05 18:43:01 (GMT -700)
  lib/ansible/modules/core: (detached HEAD c0d373f258) last updated 2016/08/05 11:41:13 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 33716b1837) last updated 2016/08/05 17:19:53 (GMT -700)
  config file = 
```
##### SUMMARY

In the unarchive module, the is_unarchived method in TgzArchive does not check the return code of the command it executes before processing the results. This causes the unarchive module to fail on some platforms, such as OS X, when it would otherwise succeed.

For example, run the following command on OS X 10.11.6 (El Capitan):

```
 ansible -m unarchive -a 'src=test.zip dest=/tmp' localhost
```

Before the fix it results in an error:

```
localhost | FAILED! => {
    "changed": false,
    "dest": "/tmp",
    "failed": true,
    "gid": 0,
    "group": "wheel",
    "handler": "TgzArchive",
    "mode": "0755",
    "msg": "Unexpected error when accessing exploded file: [Errno 2] No such file or directory: '/tmp/test.txt'",
    "owner": "root",
    "size": 11,
    "src": "/Users/mclay/.ansible/tmp/ansible-tmp-1470454495.94-225990864770004/source",
    "state": "link",
    "uid": 0
}
```

With the fix it results in success:

```
localhost | SUCCESS => {
    "changed": true,
    "dest": "/tmp",
    "extract_results": {
        "cmd": "/usr/bin/tar -C \"/tmp\" -xz -f \"/Users/mclay/.ansible/tmp/ansible-tmp-1470454587.94-69945395103370/source\"",
        "err": "",
        "out": "",
        "rc": 0
    },
    "gid": 0,
    "group": "wheel",
    "handler": "TgzArchive",
    "mode": "0755",
    "owner": "root",
    "size": 11,
    "src": "/Users/mclay/.ansible/tmp/ansible-tmp-1470454587.94-69945395103370/source",
    "state": "link",
    "uid": 0
}
```

The limitation of this fix is that change detection does not work on OS X due to tar not supporting the arguments used when detecting changes. The module will always return changed in this case.

Additionally, if extraction fails due to unsupported arguments passed to tar, this will be reported instead of the misleading `No such file or directory` error:

```
localhost | FAILED! => {
    "changed": false, 
    "dest": "/tmp", 
    "extract_results": {
        "cmd": "/usr/bin/tar -C \"/tmp\" -xz --mode=\"0755\" -f \"/Users/mclay/.ansible/tmp/ansible-tmp-1470456165.19-119094952494889/source\"", 
        "err": "tar: Option --mode=0755 is not supported\nUsage:\n  List:    tar -tf <archive-filename>\n  Extract: tar -xf <archive-filename>\n  Create:  tar -cf <archive-filename> [filenames...]\n  Help:    tar --help\n", 
        "out": "", 
        "rc": 1
    }, 
    "failed": true, 
    "gid": 0, 
    "group": "wheel", 
    "handler": "TgzArchive", 
    "mode": "0755", 
    "msg": "failed to unpack /Users/mclay/.ansible/tmp/ansible-tmp-1470456165.19-119094952494889/source to /tmp", 
    "owner": "root", 
    "size": 11, 
    "src": "/Users/mclay/.ansible/tmp/ansible-tmp-1470456165.19-119094952494889/source", 
    "state": "link", 
    "uid": 0
}
```


gregdek:
Thanks @mattclay. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

dagwieers:
While I understand that the change does make it work, I don't think this is a change we want to add.
it assumes that an non-zero return code can be ignored and forces an unarchive operation.

Why does it fail on MacOSX ? What option is incompatible with gtar on MacOSX ?

PS The current unarchive implementation is broken by design and requires a pure-python re-implementation to get it right.

mattclay:
@dag It fails because it's bsdtar, which doesn't support the -C or -d options.

dagwieers:
@mattclay So the main problem is that we allow the **tar** command, while we should be using **gtar** specifically. Can you try to install GNU tar (gtar) instead ?

mattclay:
@dag Installing gtar may make it work, but if that's our stance, we should return an appropriate error instead of `No such file or directory`.

dagwieers:
@mattclay I agree.

@abadger What do you think ? Can we remove support for the **tar** command ?

abadger:
@dagwieers I agree with documenting that non-gnu tar probably will not work.  We should continue to support systems where the "tar" command is gnu tar, though.

dagwieers:
@abadger That's what I was afraid of.

BTW non-gnu tar will never work the way it is implemented now. We shouldn't give anyone any false hope ;-) Maybe add a default warning when **tar** is found as the command to use ?

abadger:
@dagwieers yeah -- or maybe something like if we only find "tar" then run "tar --version" in an attempt to find out if it's GNU tar?

mattclay:
Non-GNU tar does work, at least for some limited functionality. I don't see why we can't allow use of non-GNU tar when it gets the job done.

abadger:
@mattclay, it doesn't properly detect changed vs unchanged (what our users often call "idempotence").  That's a large bug.  Maybe a warning and documentation that the tar command will not report changed correctly if GNU tar is not used would be fine but I'm on the fence... after all, there's not much that separates command: tar -xzvf TARFILE.tar.gz and unarchive that always reports changed.

mattclay:
@abadger You make a good point that, without it being idempotent, it isn't much different from running the tar command. I'm OK with only supporting GNU tar, but in that case the TgzArchive class should detect the lack of GNU tar in can_handle_archive and return False.

dagwieers:
Thanks @mattclay for persisting !
I need to test this, but won't have the time until next week :-/

mattclay:
I made two additional changes based on our discussion:
- Added detection for the type of tar command (bsd, gnu) using logic similar to what @jctanner was using in PR #4037. If gnu tar isn't detected, the TgzArchive class will return False from can_handle_archive. This prevents lack of GNU tar from blocking use of ZipArchive, as otherwise TgzArchive will return True for ZIP archives, but then fail to extract them.
- Added reasons for returning False to the can_handle_archive methods. These reasons are shown as part of the "Failed to find handler" message to indicate why no handler was found.

The additional details on "Failed to find handler" should avoid confusion from users who do not have GNU tar installed, as it will now be much more clear what caused the failure. Here is what it looks like on OS X when using unarchive on a TGZ when GNU tar is not installed:

```
localhost | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "Failed to find handler for \"/Users/mclay/.ansible/tmp/ansible-tmp-1470958095.46-247060186073906/source\". Make sure the required command to extract the file is installed. Command \"/usr/bin/tar\" detected as tar type: bsd. GNU tar required. Command \"/usr/bin/unzip\" could not handle archive."
}
```

ilovezfs:
I think a better approach here would be to find a way to equally support BSD style tar as well. It's highly unusual in any software to have gnu-tar-only as a requirement. Out of the 3,595 formulae we have in Homebrew core, there are only three formulae which have a hard gnu-tar dependency, and only two of those three have it as a non-build-time-only dependency.

This really shouldn't be required.

gregdek:
Thanks @mattclay for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

mattclay:
I've rebased and squashed commits.

ready_for_review

gregdek:
Thanks @mattclay. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

files/unarchive.py
```diff
@@ -627,7 +642,10 @@ def is_unarchived(self):
         # setting anyway
 
         # What is different
-        unarchived = True
+        if rc == 0:
```
dagwieers: I am not convinced that this is correct. A non-zero return code means nothing here. And certainly should not cause unextracting the archive.

mattclay:
@dagwieers I've removed the return code checking, restoring the previous behavior.

mattclay:
@dagwieers Any further thoughts on this? Is it ready to merge?

inferiorhumanorgans:
The other problem is that because ansible is accepting BSD tar as a usable tar, and BSD tar supports other non-tar formats, unarchiving other file formats (ex: zip) is broken because tar takes precedence over unzip.

dagwieers:
@inferiorhumanorgans Nope, zip comes before tar nowadays.

inferiorhumanorgans:
@dagwieers Unfortunately not in 2.1.x

Issue #4178, opened 2016-07-13
Reseting all locale environment variables causes checkout / update fail on repos that contain unicode filenames.


mullnerz:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

subversion
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
##### CONFIGURATION

default
##### OS / ENVIRONMENT

Ubuntu 15.10 or 14.04
##### SUMMARY

Reseting all locale environment variables causes checkout / update failure on repos that contain unicode filenames.
This bad behavior was introduced with commit 7020c8dcbeb6d08761c014730ee61558793ac00f fixing the issue "source_control/subversion.py needs to reset LC_MESSAGES #3255"
##### STEPS TO REPRODUCE

```
test.yml:

---
- hosts: localhost
  tasks:
    - subversion: repo="https://subversion.assembla.com/svn/test-utf8-files/" dest="test-utf8-files"

ansible-playbook test.yml         

```
##### EXPECTED RESULTS

```
$ ansible-playbook test.yml         

PLAY [localhost] ***************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [subversion] **************************************************************
changed: [localhost]

PLAY RECAP *********************************************************************
localhost                  : ok=2    changed=1    unreachable=0    failed=0   
```
##### ACTUAL RESULTS

```
$ ansible-playbook test.yml         

PLAY [localhost] ***************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [subversion] **************************************************************
[...]

PR #4358, opened 2016-08-08
Revert "Reset all locale environment variables before running svn commands"

mullnerz:
This reverts commit 99456fbc9851209f4482cdf6c246806ba1e9fbff.
#4178

### DIFF
--- a/source_control/subversion.py
+++ b/source_control/subversion.py
@@ -217,10 +217,7 @@ def main():
     export = module.params['export']
     switch = module.params['switch']
 
-    # We screenscrape a huge amount of svn commands so use C locale anytime we
-    # call run_command()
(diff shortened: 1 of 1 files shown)


bcoca:
So this just brings back the other bug, is there a solution that fixes both? like just setting LC_MESSAGES instead of all the vars?

mullnerz:
I can confirm that setting only LANG and LC_MESSAGES to 'C' doesn't break subversion with UTF-8 files in a repository. So if only setting these variables fixes the original bug, then this could be the deal:

`module.run_command_environ_update = dict(LANG='C', LC_MESSAGES='C')`

bcoca:
update this PR with that and I'll merge

mullnerz:
Ok, thank you!

ghost:
Please also merge this fix to stable-2.1.

PR #3647, opened 2016-05-13
Use integers as ip-version in os_subnet

ZZelle:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

cloud/openstack/os_subnet
##### ANSIBLE VERSION

```
ansible 2.0.2.0
  config file = 
  configured module search path = Default w/o overrides
```
##### SUMMARY

os_subnet module expects "4" or "6" as ip-version in playbooks BUT OpenStack Network API expects integers (4 or 6) as ip-version[1] which is a bit confusing. Moreover os_subnet uses these strings in subnet create http requests so it doesn't respect Network API 

Remark: it's currently working only because Neutron implementation is more permissive (it transforms string into intergers) than OpenStack Network API.

[1] http://developer.openstack.org/api-ref-networking-v2.html#createSubnet

PR #4435, opened 2016-08-16
Change ip_version type from str to int to follow Neutron specification

yuta-hono:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

cloud/openstack/os_subnet
##### ANSIBLE VERSION

```
ansible 2.1.0.0
  config file =
```
##### SUMMARY

This is the re-submited pull request based on pull request #3647 since that request seems to be not active anymore.
Current ansible call OpenStack neutron subnet related API with parameter ip_version as str but it should be int. Currently Neutron allows this from implementation, but there is no guarantee which cause problem since this is not follows API design.
Followings are original description.

```
os_subnet module expects "4" or "6" as ip-version in playbooks BUT OpenStack Network API expects integers (4 or 6) as ip-version[1] which is a bit confusing. Moreover os_subnet uses these strings in subnet create http requests so it doesn't respect Network API

Remark: it's currently working only because Neutron implementation is more permissive (it transforms string into intergers) than OpenStack Network API.

[1] http://developer.openstack.org/api-ref-networking-v2.html#createSubnet
```


gregdek:
Thanks @yuta-hono. To the current maintainers, @emonty, @shrews, @juliakreger, @j2sol, @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

yuta-hono:
ready_for_review

Shrews:
Thanks for this, but the shade library should be handling the burden of hiding details like this and making things user friendly. Also, this change would break any playbooks currently sending a string value, and we don't want people mad at us.  :)

I proposed a change to shade to handle this:

   https://review.openstack.org/363846

yuta-hono:
Got it, Then I close this ticket.

PR #2923, opened 2016-01-30
Fix Mount state=mounted opts=bind mounts on every run #1942

piotrplenik:

Issue #1942, opened 2015-08-20
Mount state=mounted opts=bind mounts on every run


rbu:
##### Ansible Configuration:

```
[ssh_connection]
pipelining = True

[defaults]
hash_behaviour = merge
```
##### Environment:

Running on Linux / Python 2.7.8. Target is CentOS 7.
##### Summary:

Running a bind mount task will mount the directory every time it is run, creating _two_ mtab entries.
##### Steps To Reproduce:

Given the following task:

```
- mount: opts=bind fstype=none state=mounted name=/dest src=/src
```

Run it twice.
##### Expected Results:
1. There is one entry in fstab:
   `/src /dest none bind 0 0`
2. There is one `mount`:
   `/dev/sda1 on /dest type btrfs (rw,relatime,ssd,space_cache)`
##### Actual Results:
1. There is one entry in fstab:
   `/src /dest none bind 0 0`
2. There are three `mount`s:

```
/dev/sda1 on /dest type btrfs (rw,relatime,ssd,space_cache)
/dev/sda1 on /dest type btrfs (rw,relatime,ssd,space_cache)
/dev/sda1 on /src type btrfs (rw,relatime,ssd,space_cache)
```
##### Notes:

This is a re-occurrence of #86. Note that [https://github.com/ansible/ansible-modules-core/blob/6a872469fc1bda54126dfa91338cf7a26bde701c/system/mount.py#L359](the special case exactly for this) that was introduced in #49 expects `mount -l` to output the source path. This does not reliably work with all distributions. Some versions/configurations of util-linux do not keep the source path in /etc/mtab, as is seen in the `mount` output above.

Note that `findmnt` lists this:

```
TARGET                                                          SOURCE             FSTYPE      OPTIONS
...
/                                                               /dev/sda1          btrfs       rw,relatime,ssd,space_cache
...
├─/dest                                                         /dev/sda1[/src]    btrfs       rw,relatime,ssd,space_cache
│ └─/dest                                                       /dev/sda1[/src]    btrfs       rw,relatime,ssd,space_cache
└─/src                                                          /dev/sda1[/src]    btrfs       rw,relatime,ssd,space_cache
```

emielmolenaar:
Same problem here, running Ansible 1.9.2 on Debian Jessie. Targets are all Debian Jessie.

samsun387:
having the exact same problem on Ubuntu 14.04LTS. Ansible version 1.9.4

luto:
[...]

PR #4439, opened 2016-08-16
Fixing bind mount on Linux (#1942)

jtyr:
This PR replaces PR #2923 as that contribution is idle for long time and the solution available there is  insufficient anyway.


gregdek:
Thanks @jtyr. To the current maintainers, @grossws please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jtyr:
Perhaps @bcoca would like to have a look at this PR as he commented on the proposed solution in #2923.

jtyr:
@abadger I have modified the code to work correctly for both the `mounted` and the `unmounted` state. I have also touched the rest of the code to make it PEP8 and FLAKE8 compatible which improves readability and maintainability of the code. Please feel free to review it.

For the bind mount testing, you can use the following commands (run as root):

```
$ mkdir /tmp/{aaa,bbb}
$ ansible localhost --connection local -i localhost, -m mount -a 'name=/tmp/bbb src=/tmp/aaa fstype=none opts=bind state=mounted'
$ ansible localhost --connection local -i localhost, -m mount -a 'name=/tmp/bbb src=/tmp/aaa fstype=none opts=bind state=unmounted'
```

jtyr:
@abadger I have just noticed that you have done some work in #4980. Some of it overlaps with my today's morning work. My code, apart of some aesthetic improvements, simplifies the workflow in the `main()` function. Please review it.

abadger:
@jtyr Cool.  I'll take a look.  There's also an integration test here: https://github.com/ansible/ansible/pull/17718

alikins:
@jtyr I didn't test #4439 extensively, but it did seem to pass the integration tests in https://github.com/ansible/ansible/pull/17718 with similar results to https://github.com/ansible/ansible-modules-core/pull/4980

jtyr:
@alikins: As @abadger mentioned in #17718, we shall address the fstab issue in another PR. Is that the only problem you have found with this PR?

abadger:
@jtyr Nice job.  There's a lot I like in your additional cleanups:
- Consolidating on get_platform and normalizing case with .lower()
- Making use of elif is great
- In main(), consolidating on a single exit_json() and making fail_json() the way to exit early.

So if you're available this morning, let's continue on with your version.

The one major thing I like better in my version is the API and calling of bind_is_mount() [your version] vs is_bind_mount() [my version].
- is_bind_mount answers the question of whether the destination is a bind mount whereas bind_is_mount answers the question of whether ansible needs to perform a change.  My version is closer to what the current ismount() function answers and is more generic.
  - When I made that change, the first step was reversing True and False as return values/expected return values.
  - The next step was refactoring to return early from the function when we detect a bind mount instead of using the "changed" variable.  We can do that because this is now a function instead of 
- I dislike how all the existing functions in mount.py make heavy use of *args.  I didn't refactor the existing functions but I did write is_bind_mount() to do without those.
- src and fstype need to be optional because they aren't needed when implementing state=unmounted and state=absent.  However, we should still take them as they allow state=mounted to differentiate between the bind mount existing and something else being mounted at the destination location.
- It would also be good to keep the docstring and licensing note from my version.

Other things I saw:
- state='absent' also needs to check bind mounts.  it's essentially first removing an entry from fstab and then performing unmount.

alikins:
@jtyr 

> Is that the only problem you have found with this PR?

Nothing critical. A few tiny unimportant note's I'll add inline...

system/mount.py
```diff
@@ -250,87 +344,138 @@ def unset_mount(module, **kwargs):
...
-    """ mount up a path or remount if needed """
+    """Mount up a path or remount if needed."""
 
-    # solaris kwargs: name, src, fstype, opts, boot, passno, state, fstab=/etc/vfstab
-    # linux kwargs: name, src, fstype, opts, dump, passno, state, fstab=/etc/fstab
+    # solaris kwargs:
+    #   name, src, fstype, opts, boot, passno, state, fstab=/etc/vfstab
+    # linux kwargs:
+    #   name, src, fstype, opts, dump, passno, state, fstab=/etc/fstab
     if get_platform() == 'SunOS':
```
alikins: Given the multiple calls to get_platform, I'd probably just do a 'platform = get_platform()' and use 'platform' var through out.

And given there are multiple methods that need the platform info, I'd probably set it in the main and pass it to the methods. Especially if the methods are switched to use regular args and not passed a args/kwargs.

alikins: Maybe the same for 'bin_path'

system/mount.py
```diff
@@ -250,87 +344,138 @@ def unset_mount(module, **kwargs):
...
+        bin_path = module.get_bin_path('findmnt', required=True)
+        cmd = '%s -nr %s' % (bin_path, kwargs['name'])
+
+    rc, out, err = module.run_command(cmd)
+    mounts = []
+
+    if len(out):
+        mounts = out.strip().split('\n')
+
+    pattern = re.compile('\[(.*)\]')
```
alikins: Would like to see a comment about what the regex is for, and a sample of what it is trying to match. And a better var name (ie, something like 'bind_mount_pattern' or 'bind_mount_regex'.)

system/mount.py
```diff
@@ -250,87 +344,138 @@ def unset_mount(module, **kwargs):
...
+        if get_platform().lower() == 'linux':
+            source = arguments[1]
+            result = pattern.search(arguments[1])
+
+            if len(result.groups()) > 0:
+                source = result.group(1)
+
+            if arguments[0] == kwargs['name'] and source == kwargs['src']:
+                changed = False
+        elif (
```
alikins, line 413: Personally not a fan of the open paren like this, but I'll leave that up to the module maintainers ;->

alikins:
Looks good to me.

abadger:
This is a good bugfix from what we have now.  After testing it: https://app.shippable.com/runs/57e95a50ebfefd0e00d1a805/summary  it fixes freebsd but still has problems on ubuntu12 and ubuntu14.  The original bug where the bind mount is not detected so running the task a second time bind mounts over the top of the previous bind mount is still there on those platforms.  I think this means that those platforms do not have the original source directory in the output from findmnt or we aren't properly parsing its output to get it.

I'm going to merge this into devel and stable-2.1 as it is an improvement.  I won't enable the tests that I wrote and merged earlier as they won't pass until the problems on older ubuntu are fixed.

PR #4481, opened 2016-08-20
fixes issue with duplicated commands in CommandRunner

privateip:
The CommandRunner will not allow duplicate commands to be added to the
command stack.  This fix will now catch the exception and continue if
a duplicate command is attempting to be added to the runner instance.

### DIFF
--- a/network/ios/ios_facts.py
+++ b/network/ios/ios_facts.py
@@ -118,9 +118,17 @@ ansible_net_memtotal_mb:
 import re
 
 from ansible.module_utils.basic import get_exception
-from ansible.module_utils.netcmd import CommandRunner
+from ansible.module_utils.netcli import CommandRunner, AddCommandError
 from ansible.module_utils.ios import NetworkModule
 
+
+def add_command(runner, command):
+    try:
+        runner.add_command(command)
+    except AddCommandError:
+        # AddCommandError is raised for any issue adding a command to
+        # the runner.  Silently ignore the exception in this case
+        pass
 
 class FactsBase(object):
 
@@ -133,7 +141,7 @@ class FactsBase(object):
 class Default(FactsBase):
 
     def commands(self):
-        self.runner.add_command('show version')
+        add_command(self.runner, 'show version')
 
     def populate(self):
         data = self.runner.get_command('show version')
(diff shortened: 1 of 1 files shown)


gregdek:

network/ios/ios_facts.py
```diff
@@ -118,9 +118,14 @@
...
 from ansible.module_utils.basic import get_exception
-from ansible.module_utils.netcmd import CommandRunner
+from ansible.module_utils.netcli import CommandRunner, AddCommandError
 from ansible.module_utils.ios import NetworkModule
 
+def add_command(runner, command):
+    try:
+        runner.add_command(command)
+    except AddCommandError:
+        pass
```
gundalow, line 131: It might be worth adding a comment to explain to future maintainers that we do this do duplicate commands don't return error.

I know currently we only return `AddCommandError` when there is a duplicate, but that could change.

Should we have a specific exception for this case, or check that the exception message starts with `'duplicated command detected'`

network/ios/ios_facts.py
```diff
@@ -118,9 +118,14 @@
 import re
 
 from ansible.module_utils.basic import get_exception
-from ansible.module_utils.netcmd import CommandRunner
+from ansible.module_utils.netcli import CommandRunner, AddCommandError
 from ansible.module_utils.ios import NetworkModule
 
+def add_command(runner, command):
```
gundalow, line 125: Could you please use two blank lines before `def` will make it easier to enforce pep8ish later on.

gundalow:
Does the `Interfaces` class also need changing to use `add_command`
https://github.com/ansible/ansible-modules-core/pull/4481/files#diff-5289c564bee1835cfb860fbb4cf5f2fdR210

revision_needed - improve docs & optional other improvements

network/ios/ios_facts.py
privateip, line 131: agree we should open an issue to track the change to module_utils/netcli.py

privateip:
ready_for_review

network/ios/ios_facts.py
gundalow, line 131: Raised in Trello

gundalow:
shipit

PR #4541, opened 2016-08-25
GCE Managed Instance Group Support

supertom:
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

gce_mig
##### ANSIBLE VERSION

```
ansible 2.2.0 (gce_mig_tests c78f74155e) last updated 2016/08/25 18:00:37 (GMT +000)
  lib/ansible/modules/core: (gce-mig 54ce81eae0) last updated 2016/08/25 18:17:29 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 101f9f5f46) last updated 2016/08/15 21:29:47 (GMT +000)
  config file = /home/supertom/.ansible.cfg
```
##### SUMMARY

Support for GCE Managed Instance Groups and Autoscalers.

This module enables support for GCE [Managed Instance Groups (MIG)](https://cloud.google.com/compute/docs/instance-groups/), including create, delete, resize and recreate.  In addition, an [Autoscaler](https://cloud.google.com/compute/docs/autoscaler/) can be configured for the MIG.

A PR for the Integration Tests exists at ansible/ansible#17249.
###### Create MIG

```
- name: Managed Instance Group Example
  hosts: localhost
  gather_facts: False
  tasks:
    - name: Create MIG
      gce_mig:
        name: ansible-mig-example
        zone: us-central1-c
        state: present
        size: 1
        template: my-instance-template-1
```
###### Configure Autoscaler

```
    - name: Update MIG with Autoscaler
      gce_mig:
        name: ansible-mig-example
        zone: us-central1-c
        state: present
        autoscaling:
          enabled: yes
          name: my-autoscaler
          policy:
            min_instances: 2
            max_instances: 5
            cool_down_period: 37
            cpu_utilization:
              target: .39
            load_balancing_utilization:
              target: 0.4
```

### DIFF
--- a/cloud/google/gce_mig.py
+++ b/cloud/google/gce_mig.py
@@ -1,0 +1,763 @@
+#!/usr/bin/python
+# Copyright 2016 Google Inc.
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: gce_mig
+version_added: "2.2"
+short_description: Create, Update or Destroy a Managed Instance Group (MIG).
+description:
+    - Create, Update or Destroy a Managed Instance Group (MIG).  See
+      U(https://cloud.google.com/compute/docs/instance-groups) for an overview.
+      Full install/configuration instructions for the gce* modules can
+      be found in the comments of ansible/test/gce_tests.py.
+requirements:
+  - "python >= 2.6"
+  - "apache-libcloud >= 1.2.0"
+notes:
+  - Resizing and Recreating VM are also supported.
+  - An existing instance template is required in order to create a
+    Managed Instance Group.
+author:
+  - "Tom Melendez (@supertom) <tom@supertom.com>"
+options:
+  name:
+    description:
+       - Name of the Managed Instance Group.
+    required: true
+  template:
+    description:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @supertom for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

cloud/google/gce_mig.py
```diff
@@ -0,0 +1,763 @@
...
+        # Do not set a default size here.  For Create and some update
+        # operations, it is required and should be explicitly set.
+        # Below, we set it to the existing value if it has not been set.
+        size=dict(type='int'),
+        state=dict(choices=['absent', 'present'], default='present'),
+        zone=dict(required=True),
+        autoscaling=dict(type='dict', default=None),
+        service_account_email=dict(),
+        service_account_permissions=dict(type='list'),
+        pem_file=dict(),
```
ryansb, line 613: It looks like this is never used anywhere. Leftover from copypaste?

cloud/google/gce_mig.py
```diff
@@ -0,0 +1,763 @@
...
+        {'name': 'min_instances', 'required': False, 'type': int},
+        {'name': 'cool_down_period', 'required': False, 'type': int}
+    ] # yapf: disable
+
+    (as_policy_valid, as_policy_msg) = _check_params(
+        params['autoscaling']['policy'], as_policy_fields)
+    if not as_policy_valid:
+        return (False, as_policy_msg)
+
+    # TODO(supertom): check utilization fields
```
ryansb, line 322: This only impacts validation, right? So the actual utilization functionality should work just fine.

ryansb:
Looks good overall, can you please add a check for libcloud v1.2 being available? It's quite possible people will have much older versions.

cloud/google/gce_mig.py
supertom: It needs to be there, at least for the time-being.  The gce_connect function, defined in module_utils will look for these params (we also currently use the secrets file, a convention of libcloud, to authenticate, which is why they aren't required).  I suspect it's possible to define these once and then have all the modules leverage it.

cloud/google/gce_mig.py
supertom: Yes.  The utilization will be be passed through as-is and the API will accept/reject.  The TODO is for us to one day, save the API call and validate in the module.

PR #4570, opened 2016-08-26
Add support for boolean substitution

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

seboolean
##### SUMMARY

SELinux since 2012 use a configuration file to
convert boolean names from a old name to a new name,
for preserving backward compatibility.

However, this has to be done explicitely when using the python
bindings, and the module was not doing it.

Openshift ansible script use this construct to detect if
a boolean exist or not:

```
- name: Check for existence of virt_sandbox_use_nfs seboolean
  command: getsebool virt_sandbox_use_nfs
  register: virt_sandbox_use_nfs_output
  failed_when: false
  changed_when: false

- name: Set seboolean to allow nfs storage plugin access from containers(sandbox)
  seboolean:
    name: virt_sandbox_use_nfs
    state: yes
    persistent: yes
  when: virt_sandbox_use_nfs_output.rc == 0
```

On a system where virt_sandbox_use_nfs do not exist, this work. But
on a system where virt_sandbox_use_nfs is a alias to virt_use_nfs (like
Fedora 24), this fail because the seboolean is not aware of the alias.

### DIFF
--- a/system/seboolean.py
+++ b/system/seboolean.py
@@ -182,6 +182,11 @@ def main():
     result = {}
     result['name'] = name
 
+    if hasattr(selinux, 'selinux_boolean_sub'):
+        # selinux_boolean_sub allows sites to rename a boolean and alias the old name
+        # Feature only available in selinux library since 2012.
+        name = selinux.selinux_boolean_sub(name)
+
     if not has_boolean_value(module, name):
         module.fail_json(msg="SELinux boolean %s does not exist." % name)
 


gregdek:

detiber:
@mscherer I'm wondering if it would also make sense to provide something in the failure to differentiate between the boolean not being present and an issue setting the boolean? 

This isn't a current issue that we face with openshift-ansible, but previously we had to also test for the existence of the boolean because of differences in policy versions (we hit this when virt_use_nfs was first introduced)

mscherer:
Something like "ignore if it doesn't exist" ? I need to see the bindings, and try to provide a patch.

detiber:
@mscherer I was actually thinking of just the addition of another key to the exit_json and fail_json calls, but I like the approach you took in https://github.com/ansible/ansible-modules-core/pull/4571 better.

system/seboolean.py
```diff
@@ -182,6 +182,9 @@ def main():
     result = {}
     result['name'] = name
 
+    if 'selinux_boolean_sub' in selinux:
```
abadger:
``` python
>>> 'selinux_boolean_sub' in selinux
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'module' is not iterable
```

Maybe one of the following would work:

``` python
try:
    name = selinux.selinux_boolean_sub(name)
except AttributeError:
    # selinux_boolean_sub allows sites to rename a boolean and alias the old name
    # Feature only available in selinux library since 2012.
    pass
```

``` python
if hasattr(selinux, 'selinux_boolean_sub'):
    # selinux_boolean_sub allows sites to rename a boolean and alias the old name
    # Feature only available in selinux library since 2012.
    name = selinux.selinux_boolean_sub(name)
```

abadger:
Traceback found with this code.  Gave two alternatives that can be used (and a comment so that casual readers of the code know why it's okay if this function isn't found).

mscherer:
Took the 2nd, I do not like using and abusing exceptions, thanks for the suggestion.

abadger:
Thanks!  merged to devel.

PR #4571, opened 2016-08-26
Add a way to ignore failure when boolean do not exist

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

seboolean
##### SUMMARY

Ansible is used by various groups as a installer, and to be multi
platform, they sometime need to set selinux boolean
who may or may not be present in the current policy.
Testing them is cumbersome, so ignoring that specific error
permit to reduce code.

See https://github.com/ansible/ansible-modules-core/pull/4570#issuecomment-242858439

I am not 100% sure this is the right design for that problem (for one, this is a bit not generic enough), but since the code was fast enough to write, this permit to discuss the exact code of a proposed solution.

### DIFF
--- a/system/seboolean.py
+++ b/system/seboolean.py
@@ -42,6 +42,13 @@ options:
     required: true
     default: null
     choices: [ 'yes', 'no' ]
+  ignore_missing:
+    description:
+      - Ignore errors if the boolean do not exist in the policy
+    required: false
+    default: no
+    choices: ['yes', 'no' ]
+    version_added: "2.2"
 notes:
    - Not tested on any debian based system
 requirements: [ ]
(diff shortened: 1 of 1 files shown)


gregdek:

abadger:
+0 here.  Pros and Cons as I see it:

Con:
- Specifying this parameter would mean that typos wouldn't be caught (otoh, it is a parameter so the user is choosing to turn it on or off).
- We already have ignore_errors in playbooks.  That can be used to do this

Pro:
- ignore_errors by itself does not differentiate between a key that doesn't exist and a different error setting the bool.  Would have to register the result and write some tasks just to decide if non-existence is what's going on  (And maybe change the return value from the module so that the playbook can find this out)

**edit:** Separated the Pro and Con headings for easier reading

mscherer:
Ok so I think we do not need exactly that PR, I can recreate it if needed. So closing

PR #4602, opened 2016-08-30
Adding nxos_bgp.py

GGabriele:
##### ISSUE TYPE

- New Module Pull Request
##### COMPONENT NAME

nxos_bgp
##### ANSIBLE VERSION

```
ansible 2.2.0
  configured module search path = ['/usr/share/ansible/']

```
##### SUMMARY

New module to manages BGP configuration on NXOS devices

### DIFF
--- a/network/nxos/nxos_bgp.py
+++ b/network/nxos/nxos_bgp.py
@@ -1,0 +1,1194 @@
+#!/usr/bin/python
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+DOCUMENTATION = '''
+---
+module: nxos_bgp
+version_added: "2.2"
+short_description: Manages BGP configuration
+description:
+    - Manages BGP configurations on NX-OS switches
+author: Jason Edelman (@jedelman8), Gabriele Gerbino (@GGabriele)
+extends_documentation_fragment: nxos
+notes:
+    - State 'absent' removes the whole BGP ASN configuration when VRF is
+      'default' or the whole VRF instance within the BGP process when using
+      a different VRF.
+    - 'default', when supported, restores params default value.
+    - Configuring global parmas is only permitted if VRF is 'default'.
+options:
+    asn:
+        description:
+            - BGP autonomous system number. Valid values are String,
+              Integer in ASPLAIN or ASDOT notation.
+        required: true
+    vrf:
+        description:
+            - Name of the VRF. The name 'default' is a valid VRF representing the global bgp.
+        required: false
+        default: null
+    bestpath_always_compare_med:
+        description:
+            - Enable/Disable MED comparison on paths from different autonomous systems.
+        required: false
+        choices: ['true','false', 'default']
+        default: null
+    bestpath_aspath_multipath_relax:
+        description:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks again to @GGabriele. This module is going into community review.

We notice that you are one of the maintainers (@privateip, @GGabriele, @gundalow) of this module, so when you think it's ready, please comment "shipit" and we will consider it for merging.

network/nxos/nxos_bgp.py
```diff
@@ -0,0 +1,1265 @@
...
+        result['changed'] = True
+        result['updates'] = commands
+
+    return result
+# END OF COMMON CODE
+
+import re
+
+WARNINGS = []
+BOOLEANS_TRUE = ['yes', 'on', '1', 'true', 'True', 1, True]
```
gundalow: Why are you redefining `BOOLEANS_TRUE`, `BOOLEANS_FALSE`?

You've imported them by doing
`from ansible.module_utils.basic import BOOLEANS_TRUE, BOOLEANS_FALSE`

gundalow:
If someone does
`event_history_cli= 'size_small'`
then later on
`event_history_cli='default'`
Will this actually be set to default or will that information have been lost?

From:
['size_small', 'size_medium', 'size_large', 'size_disable', 'default']

network/nxos/nxos_bgp.py
```diff
@@ -0,0 +1,1265 @@
...
+    - Configuring global parmas is only permitted if VRF is C(default).
+options:
+    asn:
+        description:
+            - BGP autonomous system number. Valid values are String,
+              Integer in ASPLAIN or ASDOT notation.
+        required: true
+    vrf:
+        description:
+            - Name of the VRF. The name 'default' is a valid VRF representing the global bgp.
```
gundalow: BGP

network/nxos/nxos_bgp.py
GGabriele: I added also "True" and "False" (uppercase strings) since I experienced strange behaviors based on user inputs.

GGabriele:
The behavior will be the same of CLI. From my lab, using `event_history_cli= 'size_small'` will have no effect actually on default. 

Using `event_history_cli= 'size_medium'` first and then `event_history_cli= 'default'` will set the param to its default.

network/nxos/nxos_bgp.py
gundalow: I believe if you use `type='bool'` then case will not be an problem.

I think the issue is because you are trying to use bools & 'default', which is a different concept to how we've done things with other modules.

Is their a realistic use-case of someone going from an explicit yes/no case back to default?

GGabriele: I developed it this way because some "boolean" parameters have default values (yes/no) when starting a new BGP process and also because this is something that other configuration management systems support. 

Anyway, if this doesn't fit with your design I'm okay to fix it. Just let me know, since I have many more modules ready to be pushed before 09/05 :)

network/nxos/nxos_bgp.py
```diff
@@ -0,0 +1,1265 @@
...
+            "confederation_id": "", "confederation_peers": "",
+            "graceful_restart": true, "graceful_restart_helper": false,
+            "graceful_restart_timers_restart": "120",
+            "graceful_restart_timers_stalepath_time": "300", "local_as": "",
+            "log_neighbor_changes": false, "maxas_limit": "",
+            "neighbor_down_fib_accelerate": false, "reconnect_interval": "60",
+            "router_id": "1.1.1.1",  "suppress_fib_pending": false,
+            "timer_bestpath_limit": "", "timer_bgp_hold": "180",
+            "timer_bgp_keepalive": "60", "vrf": "test"}
+state:
```
privateip: why return the state that was input... that is already available if doing a -vvv (or just look in the playbook) ... i dont understand the value of doing this?

privateip:
- mixing bool and str args is not a good idea .... need to rework the args to support either bool or string or separate into multiple args
- imports should now be explicit instead of wildcard.
- why have you embedded netcfg into the module?  I dont see what changes have occurred that necessitate that?

GGabriele:
@privateip Regarding your last point, as we discussed we're embedding that code to make the module work with `Ansible 2.1` too.

gundalow:
@GGabriele I've sent you some message son networktocode Slack

PR #4603, opened 2016-08-30
redhat_subcription module accepts id in the pool argument [#3898]

### LINKED ISSUE #3898: redhat_subscription attaches all pools of the same name
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

redhat_subscription
##### ANSIBLE VERSION

```
[cloud-user@node8 ~]$ ansible --version
ansible 1.9.4
```
##### CONFIGURATION

None
##### OS / ENVIRONMENT

N/A
##### SUMMARY

Based on the current implementation, you can specify pool like: pool='^OpenShift Enterprise$'. This will return all of the available subscriptions that match and in turn grab the pool_id and attach them.  The issue is that if you have multiple unique subscriptions of the exact same product (due to different dates), every pool that matches will get attached no matter what.

If you had 5 unique 'OpenShift Enterprise' subscriptions, you would be oversubscribing by 4 every time you provisioned a server.

There should be a way to define how many pools and or how many subscriptions you want to actually consume through the registration.  Additionally, given the regex is only used to obtain the pool_id, it would also be helpful if you could simply specify the pool_id upfront and avoid the RegEx all together.
##### STEPS TO REPRODUCE

Use 'redhat_subscription' to subscribe and attach to a pool where there are multiple pools with the exact same name.
##### EXPECTED RESULTS

Should only consume a single subscription.
##### ACTUAL RESULTS

Consumes N subscriptions where N is the number of subscriptions with the same name.

afunix:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

redhat_subscription
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 1c33b5a9f0) last updated 2016/08/30 13:36:06 (GMT +300)
  lib/ansible/modules/core: (detached HEAD 5310bab12f) last updated 2016/08/30 13:38:34 (GMT +300)
  lib/ansible/modules/extras: (detached HEAD 2ef4a34eee) last updated 2016/08/30 13:38:34 (GMT +300)
  config file = 
```
##### SUMMARY

Fixes #3898

Currently `redhat_subcription` module accepts only pool names in `pool` argument.
Unfortunately it is possible that different pools has the same name.
What even worse that `redhat_subscription` module can choose unsupported pools by name (for example pool for virtual machine on hardware host) which leads to module/playbook failure.

The change applies the regexp via `pool` argument to all available pool ids and lets user choose the specific required pool by id.

### DIFF
--- a/packaging/os/redhat_subscription.py
+++ b/packaging/os/redhat_subscription.py
@@ -444,6 +444,8 @@ class RhsmPools(object):
         for product in self.products:
             if r.search(product._name):
                 yield product
+            if r.search(product.PoolID):
+                yield product
 
 
 def main():


gregdek:
Thanks @afunix. To the current maintainers, @barnabycourt, @alikins please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

PR #4640, opened 2016-09-01
Validate rule fails when rule is an empty string.

kwoodson:
##### ISSUE TYPE

Bugfix Pull Request
##### COMPONENT NAME

cloud/amazon/ec2_group.py
##### ANSIBLE VERSION

```
ansible 2.2.0
```

I don't think this is dependent on ansible version.
##### SUMMARY

```
Using module file /usr/lib/python2.7/site-packages/ansible/modules/core/cloud/amazon/ec2_group.py
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: kwoodson
<127.0.0.1> EXEC /bin/sh -c '/usr/bin/python2 && sleep 0'
  File "/tmp/ansible_Noe2He/ansible_module_ec2_group.py", line 474, in <module>
    main()
  File "/tmp/ansible_Noe2He/ansible_module_ec2_group.py", line 348, in main
    if rule['proto'] in ('all', '-1', -1):
TypeError: string indices must be integers, not str

fatal: [localhost]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "invocation": {
        "module_name": "ec2_group"
    }, 
    "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_Noe2He/ansible_module_ec2_group.py\", line 474, in <module>\n    main()\n  File \"/tmp/ansible_Noe2He/ansible_module_ec2_group.py\", line 348, in main\n    if rule['proto'] in ('all', '-1', -1):\nTypeError: string indices must be integers, not str\n", 
    "module_stdout": "", 
    "msg": "MODULE FAILURE", 
    "parsed": false
}
```

The rule parameter being passed in is an empty string.  It needs to be validated as a dict and an exception thrown if it does not have the attributes expected.

### DIFF
--- a/cloud/amazon/ec2_group.py
+++ b/cloud/amazon/ec2_group.py
@@ -163,6 +163,10 @@ def validate_rule(module, rule):
     VALID_PARAMS = ('cidr_ip',
                     'group_id', 'group_name', 'group_desc',
                     'proto', 'from_port', 'to_port')
+
+    if not isinstance(rule, dict):
+        module.fail_json(msg='Invalid rule parameter type [%s].' % type(rule))
+
     for k in rule:
         if k not in VALID_PARAMS:
             module.fail_json(msg='Invalid rule parameter \'{}\''.format(k))


gregdek:
Thanks @kwoodson for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

cloud/amazon/ec2_group.py
```diff
@@ -163,6 +163,10 @@ def validate_rule(module, rule):
     VALID_PARAMS = ('cidr_ip',
                     'group_id', 'group_name', 'group_desc',
                     'proto', 'from_port', 'to_port')
+
+    if not isinstance(rule, dict):
```
abadger, line 167: Since this module requires python-2.6 or higher, we should use the ABCs to check whether something is "dict-like".

``` python
from collections import Mapping
[...]
    if not isinstance(rule, Mapping):
```

note that there's also a MutableMapping ABC if this interface needs to write to the dict as well as read from it.

abadger:
ryansb is doing a more thorough review of this.  I don't know ec2, just python ;-)

cloud/amazon/ec2_group.py
kwoodson: @abadger,

Do you want to check all 3? dict, Mapping, and MutableMapping?

I see another isinstance call on line 142.  I put this here to match the other ones.  If you'd prefer that I change this in both locations to whatever is decided in the above check I'm fine with that.

ryansb:
Looks good to me.

cloud/amazon/ec2_group.py
ryansb: I'm not sure checking for Mapping/MutableMapping is valuable here - these params are always coming from inside the module, which is parsing them as a dict.

PR #4741, opened 2016-09-08
Fix spot instance creation by ignoring instance_initiated_shutdown_behavior

amenonsen:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ec2.py
##### ANSIBLE VERSION

```
HEAD
```
##### SUMMARY

Before this, all spot instance requests would fail because the code
_always_ called module.fail_json when the parameter was set (which it
always was, because the module parameter's default was set to 'stop').

As the comment said, this parameter doesn't make sense for spot
instances at all, so the error message was also misleading.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -1051,8 +1051,9 @@ def create_instances(module, ec2, vpc, override_count=None):
                       private_ip_address = private_ip,
                     ))
 
-                # Spot instances do not support start/stop thereby not having the option to change shutdown behavior
-                params['instance_initiated_shutdown_behavior'] = instance_initiated_shutdown_behavior
+                # For ordinary (not spot) instances, we can select 'stop'
+                # (the default) or 'terminate' here.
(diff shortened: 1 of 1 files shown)


amenonsen:
This was broken in @c8ca1a62.

@lanzagar can you confirm that this solves the problem for you?

lanzagar:
My problem is indeed solved with this PR.

amenonsen:
OK, @lanzagar and I have both tested spot instance creation, and I've tested non-spot instance creation to ensure that nothing's regressed for that case. I'm inclined to merge this bugfix soonish—because spot instance creation is currently broken—and will do so later today unless anyone objects.

cc: @radamsfc, @ryansb (as author and committer of the earlier code)

gregdek:
Thanks @amenonsen for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

ryansb:
Works for me. :ship: dat

radamsfc:
Thanks for the fix 👍

PR #4751, opened 2016-09-08
minor bug fix to pass path to difference() in ios_config

privateip:
The ios_config module needs to pass the path kwarg to difference when
specifying match=exact or strict.


gregdek:

Issue #4869, opened 2016-09-15
eos_eapi: TypeError: load_config() got an unexpected keyword argument 'session'\n"


gundalow:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

eos_eapi
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 70e63ddf6c) last updated 2016/09/15 10:17:19 (GMT +100)
  lib/ansible/modules/core: (devel 683e5e4d1a) last updated 2016/09/15 10:17:22 (GMT +100)
  lib/ansible/modules/extras: (devel 170adf16bd) last updated 2016/09/15 10:17:23 (GMT +100)

```
##### SUMMARY

I believe the fix is

```
diff --git a/network/eos/eos_eapi.py b/network/eos/eos_eapi.py
index 7edbe3f..c09cddb 100644
--- a/network/eos/eos_eapi.py
+++ b/network/eos/eos_eapi.py
@@ -190,8 +190,10 @@ urls:
 import re
 import time

+import ansible.module_utils.eos
+
+from ansible.module_utils.network import NetworkModule, NetworkError
 from ansible.module_utils.netcfg import NetworkConfig, dumps
-from ansible.module_utils.eos import NetworkModule, NetworkError
 from ansible.module_utils.basic import get_exception

 PRIVATE_KEYS_RE = re.compile('__.+__')
@@ -273,7 +275,7 @@ def load_config(module, commands, result):
     session = 'ansible_%s' % int(time.time())
     commit = not module.check_mode

-    diff = module.config.load_config(commands, session=session, commit=commit)
+    diff = module.config.load_config(commands, commit=commit)

     # once the configuration is done, remove the config session and
     # remove the session name from the result
```

Based on https://github.com/ansible/ansible-modules-core/pull/4804/files

However there may be other things that need changing
##### STEPS TO REPRODUCE

```

```

gundalow:
Confirmed this is now fixed, thanks

PR #4892, opened 2016-09-17
fix up eos_eapi module

privateip:
- fixes TypeError: load_config() got an unexpected keyword argument 'session'\n"
- removes qos argument

ref #4869


gundalow:
Confirmed `TypeError: load_config() got an unexpected keyword argument 'session'\n"` is now fixed

PR #4935, opened 2016-09-20
Found a small bug related to excluding files

dagwieers:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

unarchive
##### SUMMARY

Due to the change (includes list to excludes list) one specific case was
missing from the exclude list.

This is a must-have fix for devel and stable-2.1.


dagwieers:
shipit

dagwieers:
Skip this, when rebasing this fix was already pulled.

PR #4946, opened 2016-09-20
fixes junos_config to be idempotent when confirming a commit

privateip:
The junos_config module would always return true when confirming a commit
This changes the module to now check first making the feature idempontent


gregdek:

PR #4980, opened 2016-09-23
Fix bind mount

### LINKED ISSUE #1942: Mount state=mounted opts=bind mounts on every run

### LINKED ISSUE #2924: Checking is mount does not work correctly with "bind" option
**Issue Type:**

Bug Report

**Ansible Version:**

ansible-modules-core devel

**Environment:**
Ansible is running from Ubuntu 15.04
I am managing an Debian Wheezy instance

**Summary:**

When using the mount module and specifying an option of "bind", checking is mountpoint exist does not work properly. This has been described in #86. 

In this issue, I replaced `os.path.ismount(path)` into separated method, that can recognize bind too.

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

system/mount.py
##### ANSIBLE VERSION

```
devel, stable-2.1
```
##### SUMMARY

**important** This includes @jtyr 's changes in #4439  While writing tests for that I found that bind unmount was also broken.  Since I needed jtyr's fixes in order to fix unmount I ended up building on top of his PR.

```
Fix unmounting bind mounts and cleanups

* No longer use wildcard imports
* Pull bind_mount detection into its own function
* Use bind_mount detection for unmount as well.
* Specify that the name and src are path type so tilde is expanded.
* Note that is_bind_mounted() needs license okay before anyone tries to
  put it into module_utils

Fixes #1942
Fixes #2924
Fixes #4439
```

### DIFF
--- a/system/mount.py
+++ b/system/mount.py
@@ -20,6 +20,10 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
+
+import platform
+
+
 DOCUMENTATION = '''
 ---
 module: mount
@@ -270,7 +274,7 @@ def mount(module, **kwargs):
         )
     args.update(kwargs)
 
-    mount_bin = module.get_bin_path('mount')
+    mount_bin = module.get_bin_path('mount', required=True)
 
     name = kwargs['name']
 
@@ -296,7 +300,7 @@ def mount(module, **kwargs):
 def umount(module, **kwargs):
     """ unmount a path """
 
-    umount_bin = module.get_bin_path('umount')
+    umount_bin = module.get_bin_path('umount', required=True)
     name = kwargs['name']
     cmd = [umount_bin, name]
 
@@ -399,20 +403,38 @@ def main():
         name, changed = set_mount(module, **args)
         if state == 'mounted':
             res = 0
+
             if ismount(name):
                 if changed and not module.check_mode:
-                    res,msg = mount(module, **args)
+                    res, msg = mount(module, **args)
             elif 'bind' in args.get('opts', []):
                 changed = True
-                cmd = 'mount -l'
+                bin_path = module.get_bin_path('mount', required=True)
+                cmd = '%s -l' % bin_path
+
+                if platform.system() == 'Linux':
+                    bin_path = module.get_bin_path('findmnt', required=True)
+                    cmd = '%s -nr %s' % (bin_path, args['name'])
+
                 rc, out, err = module.run_command(cmd)
                 allmounts = out.split('\n')
+
                 for mounts in allmounts[:-1]:
                     arguments = mounts.split()
-                    if arguments[0] == args['src'] and arguments[2] == args['name'] and arguments[4] == args['fstype']:
+
+                    if platform.system() == 'Linux':
+                        source = re.compile('\[(.*)\]').search(arguments[1]).group(1)
+
+                        if arguments[0] == args['name'] and source == args['src']:
+                            changed = False
+                    elif arguments[0] == args['src'] and arguments[2] == args['name'] and arguments[4] == args['fstype']:
                         changed = False
-                if changed:
-                    res,msg = mount(module, **args)
+
+                    if not changed:
+                        break
+
+                if changed and not module.check_mode:
+                    res, msg = mount(module, **args)
             else:
                 changed = True
                 if not module.check_mode:
--- a/system/mount.py
+++ b/system/mount.py
@@ -21,9 +21,6 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 
-import platform
-
-
 DOCUMENTATION = '''
 ---
 module: mount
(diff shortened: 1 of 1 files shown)


abadger:
@jtyr Since I was writing test for your change and found that I needed a working bind unmount I ended up needing to refactor based on your code.  Can you take a look at this?  @jimi-c is planning on cutting 2.1.2rc4 tomorrow morning and if possible I'd like to have your seal of approval on this before then.  (I know it's short notice so if not, no worries.  we'll get it in for 2.2.0 and if jimi-c or someone else think it's a showstopper for 2.1.2 we'll delay the rc4 to get it in there).

gregdek:
Thanks @abadger. To the current maintainers, @grossws please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

system/mount.py
```diff
@@ -378,7 +435,7 @@ def main():
         module.exit_json(changed=changed, **args)
 
     if state == 'unmounted':
-        if ismount(name):
+        if ismount(name) or is_bind_mounted(module, name):
```
alikins, line 438: (minor non-blocking comment on pre-existing code) The ordering of the 'if check_mode' and 'if ismount' statements is different between the 'absent' stanza, and the 'unmounted' and 'mounted' stanzas. That may be necessary for the logic, but it looks a little odd reading through the code.

system/mount.py
```diff
@@ -306,16 +321,59 @@ def umount(module, **kwargs):
...
+    rc, out, err = module.run_command(cmd)
+    allmounts = out.split('\n')
+
+    for mounts in allmounts[:-1]:
+        arguments = mounts.split()
+
+        if platform.system() == 'Linux':
+            source = LINUX_BIND_MOUNT_RE.search(arguments[1]).group(1)
+            if arguments[0] == dest and (source == src or src is None):
+                return True
```
alikins, line 359: (non blocking comment about unlikely and possibly impossible scenario)
Would it ever be valid for 'mount' or 'findmnt' output to include a bind mount entry and then undo that later in a different line? 

Was comparing this to the findmnt code in facts.py, and since that code needs to enumerate all the bind mounts, it always processes the whole file first, then checks the results, which could catch the undo case mentioned.

 **IF** contradictory lines in findmnt/mnt output is possible, that could cause a problem. 

That would be pretty weird if it's a thing, so I wouldn't worry about it now.

jtyr:
@alikins Please could you test the `mount` module from the PR #4439?

system/mount.py
alikins: I don't see an obvious way to make mount/findmnt return multiple contradictory entries, so this shouldn't be a problem.

bcoca:
lgtm

system/mount.py
abadger: That can be a thing in fstab but I don't think it's possible in mount and findmnt.  fstab is just a text file whereas mount and findmnt are reflecting state given by the kernel.  (Unless there's some implementations that use a little indirection with mtab and mtab is itself a real file rather than a pointer into kernel space on that implementation).  I know the latter is individually true but I don't know about the combination of the two.

gregdek:
Thanks @abadger for this PR. Unfortunately, it is not mergeable in its current state due to merge conflicts. Please rebase your PR. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

Issue #3877, opened 2016-06-06
ec2 module hangs if spot request fails


kuza55:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

ec2 module
##### ANSIBLE VERSION

ansible 2.1.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
##### CONFIGURATION

ANSIBLE_HOSTS=/etc/ansible/ec2.py
##### OS / ENVIRONMENT

Ubuntu on Windows 10
##### SUMMARY

When an ec2 spot bid fails, the ansible console hangs (making it seem like it's just taking a while to providion).
##### STEPS TO REPRODUCE

This task failed due to a bad volume name (/dev/sda):

``` yml
    - name: Provision a set of instances
      ec2:
         spot_price: 0.65
         spot_wait_timeout: 600
         key_name: ...
         region: us-east-1
         group_id: ...
         instance_type: g2.2xlarge  
         image: ami-d05e75b8
         wait: true
         exact_count: 1
         count_tag:
            Name: InstanceTag
         instance_tags:
            Name: InstanceTag
         vpc_subnet_id: ...
         assign_public_ip: yes
         zone: us-east-1d
         volumes:
           - device_name: /dev/sda
             volume_type: gp2
             volume_size: 20
```
##### EXPECTED RESULTS

I expected ansible to show me an error when the request failed
##### ACTUAL RESULTS

Ansible hangs

alikins:
Is https://github.com/ansible/ansible-modules-core/pull/2210/ related?

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ryansb:
I'm able to repro the issue, working on a fix for 2.2

ryansb:
Alright, got a fix out. Now if you use your sample play (with the bad volume) the task fails with:

```
"msg": "Spot instance request sir-gpf88y7m failed with status bad-parameters and fault Client.InvalidBlockDeviceMapping:Invalid device name /dev/sda (400 response code)"
```

ansibot:
[...]

PR #2210, opened 2015-10-06
Cancel spot requests before fail. 

kai11:
Fix for https://github.com/ansible/ansible-modules-core/issues/2057

Issue #2057, opened 2015-09-13
ec2 module: spot_wait_timeout exceeded, but spot instance launched anyway


kai11:
Issue Type: Bug Report

Ansible Version:
    1.9.3 from https://launchpad.net/~ansible/+archive/ubuntu/ansible
    2.0.0 (devel f0efe1ecb0)

Ansible Configuration: default \ as installed

Environment: Ubuntu 14.04

Summary: spot_wait_timeout exceeded and ec2 task failed, but spot instance launched anyway because spot request is not canceled on failure

Steps To Reproduce:
1. Run playbook https://gist.github.com/kai11/09d9bb952d422348a006
2. Playbook will fail with message "msg: wait for spot requests timeout on ..."
3. Check spot requests in AWS console - t1.micro will be open and eventually will be converted to instance without any tags

Expected Results:
Cancel sport request on spot_wait_timeout

Actual Results:
Spot request still open

kai11:
Pull request with my fix: https://github.com/ansible/ansible-modules-core/pull/2059

kai11:
Travis CI failure is os.chmod(path, 0o644). Clearly not something I've changed.

kai11:
Any progress?

sakaal:
I am experiencing the same problem. Ansible is not setting the Spot Request expiration date and time ("Request valid until").

kai11:
My fix won't set valid until; I just cancel spot request when timeout failed.
Will check "valid until" in next 2 days.

kai11:
@sakaal I played with "valid until" a bit. Biggest problem - it's up to Amazon to decide when to close spot requests. E.g. if timeout is 5 seconds and spot request set to time()+5 sec, Amazon won't guarantee it will be closed in exactly the same second when Ansible timeout expires. My approach solves problem in this issue much better.
However, it can be added as separate feature - e.g. for async workflows - if you need to launch spot cluster and init it using userdata bootstrap - instead of waiting until it come up and continue playbook.

ansibot:
@ansible, ping. This issue is still waiting on your response.
[click here for bot help](https://github.com/jctanner/ansibullbot/blob/ISSUE_TRIAGE_CLASS/ISSUE_HELP.md)

kuza55:
[...]

PR #4990, opened 2016-09-23
Check status of finished spot instance requests

ryansb:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

ec2
##### SUMMARY

Per #3877, the code to wait for spot instance requests to finish would
hang for the full wait time if any spot request failed for any reason.
This commit introduces status checks for spot requests, so if the
request fails, finishes, or is cancelled the task will fail/succeed
accordingly.

One edge case introduced here is tha if a user terminates the instance
associated with the request manually it won't fail the play, under the
presumption that the user _wants_ the instance terminated.


gregdek:
Thanks @ryansb for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

alikins:
This looks similar to https://github.com/ansible/ansible-modules-core/pull/2210/files (a fix for https://github.com/ansible/ansible-modules-core/issues/2057). Would 2057 also be address by this change?

ryansb:
No, it would not. The two PR's address different issues, and IMO #2210 is too fragile a fix. I'd rather use the AWS-provided "valid-until" spot request parameter than trying to handle it clientside.

PR #4994, opened 2016-09-24
Add the Kyup container provider API to the cloud modules

hackman:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

kyup
##### ANSIBLE VERSION

```
ansible 2.2.0
  config file = /root/api/ansible.cfg
```
##### SUMMARY

Introducing the Kyup container provider API to Ansible.
With this module, users of Ansible will be able to provision and manage containers inside the Kyup architecture.
The module has been developed and tested on Debian Jessie.

### DIFF
--- a/cloud/kyup/kyup.py
+++ b/cloud/kyup/kyup.py
@@ -222,6 +222,13 @@ def create_container(module):
         if opt[i] is None:
             module.fail_json(changed=False, msg = '%s parameter is required for creating a container' % i)
 
+    # check if the container already exists
+    req = '{"action":"cloudList","authorization_key":"%s","data":{}}'
+    ret = api_request(module, req % module.params.get('api_key'))
+    for i in ret['data']['list']:
+        if i['name'] == module.params.get('name'):
+            module.exit_json(changed=False, msg = 'Container %s already exists' % module.params.get('name'))
+
     storage_type = module.params['storage_type']
     if storage_type is None or storage_type == 'local':
         storage_type = 0
--- a/cloud/kyup/kyup.py
+++ b/cloud/kyup/kyup.py
@@ -234,7 +234,7 @@ def add_ssh_keys(module, container_id):
         api_request(module, req % ('sshInstallKey', module.params.get('api_key'), key_req % (int(key_list[key]), int(container_id) )))
 
 def create_container(module):
-    enc_key = module.params['enc_key'] or os.environ['KYUP_ENC_KEY']
+    enc_key = module.params['enc_key'] or os.environ.get('KYUP_ENC_KEY', None)
 
     if enc_key is None:
         module.fail_json(changed=False, msg = 'ENC_KEY is required for creation of containers')
--- a/cloud/kyup/kyup.py
+++ b/cloud/kyup/kyup.py
@@ -45,11 +45,6 @@ options:
       - Comma separated list of ssh key names, that have access to this container.
     required: false
     default: null
-  container_id:
-    description:
-	  - ID of the container you want to operate on.
-    required: false
-    default: null
   password:
     description:
       - Password for the root user of the container.
@@ -162,14 +157,24 @@ def api_request(module, data):
             return ret
         else:
             if 'data' in ret and 'error_code' in ret['data']:
-                module.fail_json(changed=False, msg = 'Failed to execute request. Error code: %d Error msg: %s req: %s' % (ret['data']['error_code'], ret['data']['error'], data))
-#                if ret['data']['error_code'] == 101:
-#                else:
-#                    module.fail_json(changed=False, msg = 'Failed to execute request. Error code: %d Error msg: %s' % (ret['data']['error_code'], ret['data']['error']))
+                if ret['data']['error_code'] == 101:
+                    module.fail_json(changed=False, msg = 'Failed to execute request. Error code: %d Error msg: %s req: %s' % (ret['data']['error_code'], ret['data']['error'], data))
+                else:
+                    module.fail_json(changed=False, msg = 'Failed to execute request. Error code: %d Error msg: %s' % (ret['data']['error_code'], ret['data']['error']))
     else:
         module.fail_json(changed=False, msg = 'Req: ' + data + ' Resp: ' + ret)
 
-def kyup_action(module, container_id, action):
+def kyup_action(module, action, container_id = 0):
+    req = '{"action":"cloudList","authorization_key":"%s","data":{}}'
+    ret = api_request(module, req % module.params.get('api_key'))
+    for container in ret['data']['list']:
+        if container['name'] == module.params.get('name'):
+          container_id = container['id']
+          break
+
+    if container_id == 0:
+        module.exit_json(changed=False, msg = 'Could not find container id for container name %s' % module.params.get('name'))
+
     req = '{"action":"%s","authorization_key":"%s","data":{"container_id":%d}}'
     ret = api_request(module, req % (action, module.params.get('api_key'), int(container_id)) )
     if action == 'cloudDetails':
@@ -225,9 +230,14 @@ def create_container(module):
     # check if the container already exists
     req = '{"action":"cloudList","authorization_key":"%s","data":{}}'
     ret = api_request(module, req % module.params.get('api_key'))
-    for i in ret['data']['list']:
-        if i['name'] == module.params.get('name'):
-            module.exit_json(changed=False, msg = 'Container %s already exists' % module.params.get('name'))
+    for container in ret['data']['list']:
+        if container['name'] == module.params.get('name'):
+            container['ip'] = container['ip'][0:-3]
+            module.exit_json(
+                changed=False,
+                ansible_facts = { 'container_name': container['name'], 'container_ip': container['ip'] },
+                msg = 'Container %s already exists' % module.params.get('name')
+            )
 
     storage_type = module.params['storage_type']
     if storage_type is None or storage_type == 'local':
@@ -253,7 +263,7 @@ def create_container(module):
     if 'task_id' in ret['data']:
         if ret['data']['task_id'] > 0:
             container_id = get_task_status(module, ret['data']['task_id'])
-            container = kyup_action(module, container_id, 'cloudDetails')
+            container = kyup_action(module, 'cloudDetails', container_id)
             add_ssh_keys(module, container_id)
             # remove the mask portion of the string
             container['ip'] = container['ip'][0:-3]
@@ -279,18 +289,14 @@ def core(module):
     elif action == 'create':
         create_container(module)
     else:
-        container_id = module.params['container_id']
-        if container_id is None:
-            module.fail_json(changed=False, msg = 'container_id parameter is required for this action')
-
         if action == 'destroy':
-            kyup_action(module, container_id, 'cloudDestroy')
+            kyup_action(module, 'cloudDestroy')
         elif action == 'start':
-            kyup_action(module, container_id, 'cloudStart')
+            kyup_action(module, 'cloudStart')
         elif action == 'stop':
-            kyup_action(module, container_id, 'cloudStop')
+            kyup_action(module, 'cloudStop')
         elif action == 'restart':
-            kyup_action(module, container_id, 'cloudReboot')
+            kyup_action(module, 'cloudReboot')
     
 def main():
     module = AnsibleModule(
@@ -298,7 +304,6 @@ def main():
             api_key = dict(aliases=['API_KEY'], no_log=True),
             enc_key = dict(aliases=['ENC_KEY'], no_log=True),
             action = dict(type='str', aliases=['state', 'command', 'cmd']),
-            container_id = dict(type='int'),
             dc_id = dict(type='int', aliases=['datacenter', 'datacenter_id']),
             name = dict(type='str'),
             image = dict(type='str', aliases=['template']),
@@ -309,17 +314,6 @@ def main():
             bw = dict(type='int', default=2),
             ssh_keys = dict(default=''),
             storage_type = dict(default='local', choices=['local', 'distributed']),
-        ),
-        mutually_exclusive = (
-            ['container_id', 'name'],
-            ['container_id', 'dc_id'],
-            ['container_id', 'image'],
-            ['container_id', 'password'],
-            ['container_id', 'mem'],
-            ['container_id', 'hdd'],
-            ['container_id', 'cpu_cores'],
-            ['container_id', 'bw'],
-            ['container_id', 'storage_type'],
         ),
     )
     core(module)
--- a/cloud/kyup/kyup.py
+++ b/cloud/kyup/kyup.py
@@ -118,6 +118,17 @@ EXAMPLES = '''
 
 '''
 
+RETURN = '''
+kyup_container_name:
+    returned: success
+    type: string
+    sample: "container_test"
+kyup_container_ip:
+    returned: success
+    type: string
+    sample: "10.12.131.5"
+'''
+
 import os
 import time
 import base64
--- a/cloud/kyup/kyup.py
+++ b/cloud/kyup/kyup.py
@@ -1,0 +1,294 @@
+#!/usr/bin/python -t
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: kyup
+short_description: create, terminate, start or stop a container in kyup.com, return container_id
+description:
+    - Creates or terminates Kyup container. When created optionally waits for it to be 'running'. This module has a dependency on python-json and python-requests.
+version_added: "0.1"
+options:
+  api_key:
+    description:
+	  - Kyup.com API key. The key can also be supplied by setting the Environment variable KYUP_API_KEY.
+	required: true
+	default: null
+  enc_key:
+    description:
+      - Kyup.com encryption key. The key can also be supplied by setting the Environment variable KYUP_ENC_KEY.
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @hackman for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+    required: false
+  password:
+    description:
+      - Password for the root user of the container. Required when I(action) = 'create'.
+    required: false
+  image:
+    description:
+      - Template image to use when creating the container. Required when I(action) = 'create'.
+    required: false
+  dc_id:
```
cnicolov, line 54: You can add `aliases:` for better documentation representation.

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+  - name: restart the container
+    kyup: >
+      api_key=XXXX
+      action=restart
+      name=avalon
+
+'''
+
+import os
+import requests
```
cnicolov: If executed in a python environment without requests installed the module will fail.
For example when this task needs to be delegated to a host with internet access(NAT) in a controlled environment.

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+    data = pad(data)
+    m = md5.new(key)
+    iv = m.hexdigest()[0:blocksize]
+    cipher = AES.new( key, AES.MODE_CBC, iv )
+    return base64.b64encode( cipher.encrypt( data ) )
+## End of encryption code
+
+def api_request(module, data):
+    ret = {}
+    count = retries;
```
cnicolov: semicolon token is required only when you're calling more than one statement on the same line

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+    while count != 0:
```
cnicolov: `while count:` will do the trick here

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+        # get the response
+        resp = requests.post(url, { 'request': data })
+        # parse the json
+        ret = json.loads(resp.text)
+        if ret['status']:
+            return ret
+        count -= 1
+        time.sleep(5)
+
+    if 'status' in ret:
```
cnicolov: You can do 

``` python
status = ret.get('status', False)
```

and use

``` python
if not status:
```

and get rid of the `else` statements?

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+    ret = api_request(module, req % (action, module.params.get('api_key'), int(container_id)) )
+    if action == 'cloudDetails':
+        return ret['data']['container']
+    else:
+        module.exit_json(changed=True, msg = '%s successfull' % action)
+
+def get_task_status(module, task_id):
+    req = '{"action":"getTask","authorization_key":"%s","data":{"task_id":%d}}'
+    count = retries;
+    while count != 0:
```
cnicolov: `while count` will do the trick

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+        count -= 1
+    module.fail_json(changed=False, msg = ret)
+
+def add_ssh_keys(module, container_id):
+    key_list = {}
+    ssh_keys = module.params['ssh_keys']
+    if ssh_keys is None or ssh_keys == '':
+        return
+    # cleanup any spaces put by the user
+    ssh_keys.replace(' ', '')
```
cnicolov, line 234: IMO list is better fit for the `ssh_keys` argument.

The syntax will become

``` yaml
kyup:
  ssh_keys:
    - foo
    - bar

kyup: >
  ssh_keys=["foo", "bar"]
```

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+            key_list[key['title']] = key['key_id']
+    # add the keys to the container
+    key_req = '"key_id":%d,"container_id":%d'
+    for key in key_list:
+        if key_list[key] == 0:
+            module.fail_json(changed=True, msg = 'Unable to find key ID for key %s' % key)
+        api_request(module, req % ('sshInstallKey', module.params.get('api_key'), key_req % (int(key_list[key]), int(container_id) )))
+
+def create_container(module):
+    enc_key = module.params['enc_key'] or os.environ['KYUP_ENC_KEY']
```
cnicolov: If `module.params['enc_key']` is false and `KYUP_ENC_KEY` env var isn't set this will raise an KeyError exception.

`os.environ.get('KYUP_ENC_KEY', None)` will do the trick

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # check if the container already exists
+    req = '{"action":"cloudList","authorization_key":"%s","data":{}}'
+    ret = api_request(module, req % module.params.get('api_key'))
+    for container in ret['data']['list']:
+        if container['name'] == module.params.get('name'):
+            container['ip'] = container['ip'][0:-3]
+            module.exit_json(
+                changed=False,
+                ansible_facts = { 'container_name': container['name'], 'container_ip': container['ip'] },
```
cnicolov: You can prefix the facts with the module name to avoid overrides

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+        module.fail_json(changed=False, msg = 'you can not continue without api_key')
+    module.params['api_key'] = api_key
+
+    action = module.params['action']
+    if action is None:
+        module.fail_json(changed=False, msg = 'action parameter is required by this module')
+    elif action == 'create':
+        create_container(module)
+    else:
+        if action == 'destroy':
```
cnicolov, line 328: You can create a class with static method for each action and call them with getattr, this might simplify the extensibility for the `action` argument. for example

``` python
class Actions:
    @staticmethod
    def create(module):
        return create_container(module)

action = getattr(Actions, action)
action(module)
```

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,338 @@
...
+        elif action == 'restart':
+            kyup_action(module, 'cloudReboot')
+    
+def main():
+    module = AnsibleModule(
+        argument_spec = dict(
+            api_key = dict(aliases=['API_KEY'], no_log=True),
+            enc_key = dict(aliases=['ENC_KEY'], no_log=True),
+            action = dict(type='str', aliases=['state', 'command', 'cmd']),
+            dc_id = dict(type='int', aliases=['datacenter', 'datacenter_id'], default=1),
```
cnicolov: I think `dc` is also an suitable alias for this argument

cnicolov (commented):
Also you can look at [PEP8](https://www.python.org/dev/peps/pep-0008/) to make the code more readable.

This is only an opinion, so feel free to ignore my comments

cloud/kyup/kyup.py
hackman: It took me some time to replace requests with fetch_url from the urls.py, but I did it :)

cloud/kyup/kyup.py
hackman: I found somewhere in the documentation that aliases should be removed from DOCUMENTATION and that is why I removed them

cloud/kyup/kyup.py
hackman: yup... a habit from other languages. Fixed

cloud/kyup/kyup.py
hackman: I do prefer to actually write it this way... but if you require it like that. I'll change it.

cloud/kyup/kyup.py
hackman: I do prefer to actually write it this way... but if you require it like that. I'll change it.

cloud/kyup/kyup.py
hackman: I'm sorry but I do not agree. List is harder to be written in the yml file. And I prefer the comma separated list

cloud/kyup/kyup.py
hackman: Yup, I saw that one, also :)

cloud/kyup/kyup.py
hackman: for now, if this is not required I would not do that.

cloud/kyup/kyup.py
hackman: agreed

cloud/kyup/kyup.py
hackman: done

cloud/kyup/kyup.py
hackman: I did what you requested... however I don't like it this way :(

cloud/kyup/kyup.py
cnicolov: I think this is valid only for the empty `aliases: []` and `choices: []` statements but I'm not sure

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+    if http_status != 200:
+        module.fail_json(changed=False, msg = 'Request failed with status %d' % int(status))
+
+    status = ret.get('status', False)
+    # Exit if we did not found status
+    if not status:
+        module.fail_json(changed=False, msg = 'Req: ' + data + ' Resp: ' + ret)
+
+    # We have found status, check its value here
+    if status:
```
cnicolov, line 191: You can omit the `if/else` statement and just return here.

The else statement won't get executed, the module will exit at line 188 if the status is `False`

cloud/kyup/kyup.py
cnicolov: I was thinking for something like this: 

``` yaml
# group_vars/prod_web/keys.yml
devops_organization_pubkeys:
  - name: peter
    pubkey: ssh-blabla really long string of nonsense peter@devops.organization
  - name: bob
    pubkey: ssh-blalbla really long string of nonsense bob@devops.organization
```

If you decide to extend the module with a feature which allows the users to create new keys:

``` yaml
- name: "Add ssh pubkeys"
  kyup:
    action: ssh_key
    name: "{{ item.name }}"
    pubkey: "{{ item.pubkey }}"
  with_items:
    - "{{ devops_organization_pubkeys }}"
```

Now let's create new container:

``` yaml
- name: "Create new container"
  kyup:
    action: create
    ssh_key: "{{ devops_organization_keys|map(attribute='name')|list }}"
```

This also can be done with comma separated string:

``` yaml
"{{ devops_organization_keys|map(attribute='name')|join(",") }}"
```

Now when the user needs a new pubkey he/she can add it as a group var.

Anyway..

IMO if something can get multiple values as input it should be a collection. It is cleaner and easier to read/follow. And IMO it's not that important how hard it was to write the thing once but how you'll read it for example 500 times.

Consider the following scenario:
 There is a typo in the code - a comma in the middle of the string `peter`:

``` yaml
kyup:
  ssh_keys: "bob,pe,ter,alice"
```

The error might look like:

```
Error! SSH key pe does not exist
```

Example with a list:

``` yaml
kyup:
  ssh_keys:
    - "bob"
    - "pe,ter"
    - "alice"
```

```
Error! SSH key pe,ter does not exist
```

I think the second case is cleaner.

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+                ansible_facts = { 'kyup_container_name': container['name'], 'kyup_container_ip': container['ip'] },
+                msg = 'Container: Name: %s IP: %s' % (container['name'], container['ip'])
+            )
+        else:
+            module.fail_json(changed=True, msg = 'Invalid task_id')
+    else:
+        module.fail_json(changed=True, msg = 'Unable to get task_id')
+
+def core(module):
+    api_key = module.params['api_key']
```
cnicolov, line 314: You can do:

``` python
api_key = os.environ.get('KYUP_API_KEY', None) or module.params['api_key']

if not api_key:
     module.fail_json(changed=False, msg='you can not continue without api_key')
```

and omit the two `if` statements below

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+            module.fail_json(changed=True, msg = 'Unable to find key ID for key %s' % key)
+        api_request(module, req % ('sshInstallKey', module.params.get('api_key'), key_req % (int(key_list[key]), int(container_id) )))
+
+def create_container(module):
+    enc_key = module.params['enc_key'] or os.environ.get('KYUP_ENC_KEY', None)
+
+    if enc_key is None:
+        module.fail_json(changed=False, msg = 'ENC_KEY is required for creation of containers')
+
+    opt = { 'name' : None, 'password' : None, 'image' : None, 'dc_id' : None }
```
cnicolov, line 257: If more than one of the parameters is None, you can detect it in one shot for better user experience:

``` python
mandatory_parameters = ["name', 'password', 'image', 'dc_id']
missing = [p for p in mandatory_parameters if module.params.get(p, None) is None]
if len(missing):
    module.fail_json(
        changed=False,
        msg='One or more parameters are missing: %s' % missing.join(","))
```

gregdek:
@hackman A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

cloud/kyup/kyup.py
hackman: I do not agree. The status variable can be any of the following values:
0, 1, 101, 102 and so on.
The first check is if the status variable has assigned value.
The second check is if the value is true, if not do something else.

cloud/kyup/kyup.py
hackman: I want to:
1. get the api_key form the playbook
2. if env variable is existing overwrite the api_key
3. check if the api_key is not None
As far as my tests go, the code does exactly what I need.

hackman:
All issues have been resolved.
ready_for_review

cloud/kyup/kyup.py
hackman: commit 63597f8f78c8db2fd8b9007654e9d171264a05f3
removes aliases and choices as http://docs.ansible.com/ansible/developing_modules.html#module-checklist requires.
The checklist is not specific if this rule is intended only for empty lists. But I prefer to be safe. So I removed them.

gregdek:
Thanks @hackman for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

cloud/kyup/kyup.py
cnicolov:
```
>>> var.update({"status": 0})
>>> if var.get("status", False):
...     print "wont print"
...
>>> var.get("status")
0
>>> var.get("status") == False
True
```

If you don't like this behavior you must use `if status != 0:` to make difference if `status` is `integer` or even better, use the get method with `None` as fallback.

```
>>> var.get('doesnotexist', None) is None
True
```

hackman:
Is there any news for this?

hackman:
Ping

cloud/kyup/kyup.py
ryansb: That sort of thing (using getattr to avoid if/else checks) isn't something we usually ask for as part of modules, since it's primarily a question of style.

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+  - name: ensure kyup container
+    kyup: >
+      api_key=XXXX
+      enc_key=YYYY
+      action=create
+      image="Centos Plain"
+      dc_id=1
+      name=avalon
+      password=e4a17265b28f6a9e8657
+      ssh_keys=hackman,piele
```
ryansb, line 100: As mentioned in a later comment, a `list` type is preferred for this parameter since it is actually a list of things.

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+    description:
+      - Storage size in GB. Minimum 20 GB.
+    required: false
+    default: 20
+  cpu_cores:
+    description:
+      - Number of CPU cores to be used. Minimum 2 CPU cores.
+    required: false
+    default: 2
+  bw:
```
ryansb, line 74: Super short names like this aren't obvious to everyone, can you name this something like "bandwidth" and use an alias to `bw` for power users?

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+  image:
+    description:
+      - Template image to use when creating the container. Required when I(action) = 'create'.
+    required: false
+  dc_id:
+    description:
+      - DataCenter location ID. Required when I(action) = 'create'.
+    required: false
+    default: 1
+  mem:
```
ryansb, line 59: I'd prefer using `memory` for this parameter to be more consistent with other modules. Also, is a decimal value ok for this field? For example, does kyup allow 1.5GB machines?

cloud/kyup/kyup.py
ryansb: That's only for empty ones in docs. It's not very nice to read (as a user) a bunch of empty `aliases:` lists when looking for something. Aliases are still allowed.

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+## Code used for encryption of the passwords
+blocksize = AES.block_size
+
+def pad( data ):
+    padding = blocksize - ( len(data) % blocksize )
+    return data + (chr(padding) * padding)
+
+def encrypt( key, data ):
```
ryansb, line 157: Can you please remove the whitespace around these params to fit with PEP8 code style?

cloud/kyup/kyup.py
```diff
@@ -0,0 +1,357 @@
...
+            )
+
+    storage_type = module.params['storage_type']
+    if storage_type is None or storage_type == 'local':
+        storage_type = 0
+    elif storage_type == 'distributed':
+        storage_type = 1
+
+    container_id = 0
+    req = '{"action":"cloudCreate","authorization_key":"%s","data":{"name":"%s","password":"%s","image_name":"%s","datacenter_id":%d,"storage_type":%d,"resources":{"mem":%d,"hdd":%d,"cpu":%d,"bw":%d}}}'
```
ryansb, line 282: Please use the JSON serializer instead. Using string formatting to inject things into JSON could have unfortunate consequences.

ryansb (changes requested):

PR #5069, opened 2016-09-28
Fix docker image always changed even if image has not been updated.

cuckoohello:
While using force mode in docker module,  always return Changed=True even if the image has not been updated.

This pull request fixes this issue, so we can always use force in docker image and recreate docker container only if the upstream image is updated.

### DIFF
--- a/cloud/docker/docker_image.py
+++ b/cloud/docker/docker_image.py
@@ -335,6 +335,8 @@ class ImageManager(DockerBaseClass):
                 self.results['changed'] = True
                 if not self.check_mode:
                     self.results['image'] = self.client.pull_image(self.name, tag=self.tag)
+            if image and image['Id'] == self.results['image']['Id']:
+                self.results['changed'] = False
 
         if self.archive_path:
             self.archive_image(self.name, self.tag)


gregdek:
Thanks @cuckoohello for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

chouseknecht:
@cuckoohello 

Can you provide an example playbook that produces the bug. I'd like to reproduce the error and test this PR before merging it. An example will make it easier.

Thank you for doing the work on this and bringing it to our attention. Nice job!

cuckoohello:
```
---

- name: Test docker image always changed even if image has not been updated bug while using force mode.
  hosts: servers
  tasks:
    - name: Download image alpine first time
      docker_image:
        name: alpine
        force: yes
        state: present

    - name: Download image alpine second time with same config
      docker_image:
        name: alpine
        force: yes
        state: present
```

Using the playbook above, they are two tasks using the same config except the name.

Without my patch, the outpout looks like below:

```
PLAY [Test docker image always changed even if image has not been updated bug while using force mode.] ***

ok: [btsync]

TASK [Download image alpine first time] ****************************************
changed: [btsync]

TASK [Download image alpine second time with same config] **********************
changed: [btsync]

btsync                     : ok=3    changed=1    unreachable=0    failed=0 
```

The task "Download image alpine second time with same config" should return ok because the upstream docker image alpine is not updated.

With the patch, it will looks like：

```
PLAY [Test docker image always changed even if image has not been updated bug while using force mode.] ***

ok: [btsync]

TASK [Download image alpine first time] ****************************************
changed: [btsync]

TASK [Download image alpine second time with same config] **********************
ok: [btsync]

btsync                     : ok=3    changed=1    unreachable=0    failed=0   
```

Only the first pull should result in changed(alpine image is not existed in the server before test),  task "Download image alpine second time with same config" will detect the image id is as same as previous, so it should return ok.

cuckoohello:
@chouseknecht 
Did you have time to test this pull request?

PR #5080, opened 2016-09-28
fixes junos_template ignoring the action directive

privateip:
This updates the junos_template to properly process the action
directive when loading the configuration.


dgarros:
Hi 

I'm still seeing this issue in `2.2.0.0`.
The action parameter is not properly applied when using `overwrite`, here is my task

``` yaml
    - name: Pushing config ... please wait
      junos_template:
        src: "{{ junos_conf }}"
        action: overwrite
        comment: "Applied by Ansible"
        provider: "{{ netconf }}"
```

Should I open a new issue or should we reopen this one ?

Also, I noticed that there is no option `overwrite` in **junos_config** in 2.2.0.0, 

cc / @stacywsmith, @vnitinv

Issue #5118, opened 2016-10-01
win_file: Failed to delete folder with broken symlink


jborean93:
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

ansible-modules-core/windows/win_file.ps1
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 7e0074263d) last updated 2016/10/02 10:15:29 (GMT +1100)
  lib/ansible/modules/core: (detached HEAD 18f710fe32) last updated 2016/10/02 10:15:41 (GMT +1100)
  lib/ansible/modules/extras: (detached HEAD a58e1d59c0) last updated 2016/10/02 10:15:53 (GMT +1100)
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/usr/share/ansible']
```
##### OS / ENVIRONMENT

Running: Bash for Ubuntu on Windows
Managing: Windows Server 2012 R2
##### SUMMARY

When setting win_file state: absent for a directory that contains a symlink that is no longer pointing to a valid directory it fails with DirectoryNotFoundException
##### STEPS TO REPRODUCE

Inventory.ini

```
[windows]
host

[windows:vars]
ansible_user=.\User
ansible_password=Password
ansible_connection=winrm
ansible_port=5986
ansible_winrm_transport=ntlm
ansible_winrm_server_cert_validation=ignore
```

Playbook.yml

```
- name: test windows
  hosts: windows

  tasks:
    - name: create test folder
      win_file:
        path: C:\temp\test
        state: directory

    - name: create symlink
      raw: CMD.EXE /C mklink /d C:\temp\link C:\temp\test

    - name: delete folder that link is pointing to
      win_file:
        path: C:\temp\test
        state: absent

    - name: delete parent folder containing broken link
      win_file:
        path: C:\temp
        state: absent
```
##### EXPECTED RESULTS

Expected C:\temp to be deleted correctly without exception
##### ACTUAL RESULTS

```
An exception occurred during task execution. The full traceback is:
At C:\Users\User\AppData\Local\Temp\ansible-tmp-1475364906.0-88571632788975\win_file.ps1:317 char:9
+         Remove-Item -Recurse -Force $fileinfo
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
fatal: [192.168.1.13]: FAILED! => {
    "changed": false,
    "error_record": {
        "CategoryInfo": {
            "Activity": "Remove-Item",
[...]

PR #5120, opened 2016-10-02
Updated win_file with logic on handling sym link deletions

jborean93:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

win_file
##### ANSIBLE VERSION

```
ansible 2.2.0 (devel 7e0074263d) last updated 2016/10/02 09:15:29 (GMT +1000)
  lib/ansible/modules/core: (detached HEAD 18f710fe32) last updated 2016/10/02 09:15:41 (GMT +1000)
  lib/ansible/modules/extras: (detached HEAD a58e1d59c0) last updated 2016/10/02 09:15:53 (GMT +1000)
  configured module search path = ['/usr/share/ansible']
```
##### SUMMARY

This pull request fixes the issue described in https://github.com/ansible/ansible-modules-core/issues/5118. This issue currently occurs when you use win_file to delete a folder or symbolic link directory that is pointing to a folder/file that no longer exists.


gregdek:
Thanks @jborean93. To the current maintainers, @jhawkesworth please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jborean93:
@jhawkesworth do you believe you would get time to look at this pull request and potentially merge it in before 2.2 is finally released?

jhawkesworth:
My apologies @jborean93 I did review this a couple of weeks ago but clearly I didn't manage to post my comments, probably because I did it on the train where net connection is not reliable.

I'm sorry this missed 2.2.  At least ansible is flexible enough for you to put your version in your /etc/ansible/library (or wherever you have 'library' configured to point in your ansible.cfg).  Your version will be preferred, so you can at least use your change until there is a fix merge.

Your code looks like it will fix the issue.  

If you could add an integration test that would be great, or at least share a reproducer playbook.  The existing integration tests for win_file are here (in the main project):  https://github.com/ansible/ansible/tree/devel/test/integration/targets/win_file/tasks/main.yml

If  you need help running the integration tests please ask on #ansible-devel irc channel or post in the google group of the same name and I will get back to you as soon as I can.

Personally I'd rather try and find a .net way of removing (possibly Directory.Delete rather than using cmd.exe /c rmdir ), but what you have now would work.

Only other thing is stylistic: I like it when a module lets you create and remove whatever entities it is dealing with as it plays nicely with the state present/absent pattern.  However that would mean adding support for creating links which looks like it would need to use .net integration.

Since this is a core module, I think I'd really prefer it it there was a test to cover this change.  Hopefully this wouldn't be too much to add as I imagine you already have already encountered a use case.

needs_revision

jhawkesworth:
Looking back at win_file again, I think it needs an overhaul as I've learned a bit more about powershell and dotnet since I first tackled this.  I'd really like to add some proper exception handling and tackle adding support for creating and removing links (soft, hard and I guess Junctions too).  I've put it on my todo list although I can't make a promise about when I will get to it.

jhawkesworth:
Just to be clear, I think this could be merged as is, the only change I want is to add a test which, at the moment means a separate PR against ansible/ansible project. This will all get a lot simpler soon once we go back to a single repo.  

Jon

jborean93:
@jhawkesworth thanks for the info I was planning on adding some tests for this once it was merged (not sure if that is the common practice). For symlinks I can interact with the Win32 API to do this and not through mklink but it doesn't look like there is a .net equivalent. I wasn't sure what the preference of doing actions in powershell was but my assumption was powershell < .net < wmi < cmd and am not sure where calling dll's sits in there. I can also try and check the version of powershell and use the new PS cmdlets with powershell v5 that added support for symlinks and only use mklink or the Win32 API if it is an earlier verison.

I was planning on creating a win_link module to create/delete symlinks/junctions as I have seen some previous conversations between yourself and others that the actions should be in separate modules. I am happy to make the changes in win_file in a separate PR for win_file if you now believe otherwise.

jhawkesworth:
Thanks for this.  Generally for core stuff I'd say its best to prepare tests and code PRs at the same time.  Extras modules are less exacting in requirements but tests always help PRs get to merged as it reduces time needed by reviewers to get confident with the changes.

I would put calling dlls below calling .net methods and probably above wmi for the most stable dlls ( e.g. mscorlib, kernel ).  

These days I feel wrapping native .exe programs is unlikely to be appropriate for core modules, but if it proves necessary I learned the hard way that its essential to check return codes as a bare minimum.

Where possible I like to make win modules have equivalent functionality to the linux ones so I wouldn't object to having actual links in win_file, as opposed to shortcuts (which I feel are windows-specific), unless it made for a huge bloated module that was slow to execute.  

I'm going to tag @nitzmahone in on this conversation for his input as 'keeper of windows things for ansible' here, before sending you off in a particular direction though.  Matt any thoughts on the above?

jborean93:
While waiting for @nitzmahone I've added some tests to test out this scenario here https://github.com/ansible/ansible/pull/18328. Once Matt has given some direction I will look at integrating the creation on links with this module in another pull request

windows/win_file.ps1
```diff
@@ -50,6 +50,33 @@ $result = New-Object psobject @{
     changed = $FALSE
 }
 
+# Used to delete directories and files with logic on handling symbolic links
+Function Delete($file)
+{
+    $filepath = $file.FullName
+    If ( $file.Attributes.ToString().contains("ReparsePoint") )
```
nitzmahone: Since this is an enum, it'd be better to actually test the enum value instead of using magic strings, eg: `If ($file.Attributes -band [System.IO.FileAttributes]::ReparsePoint)`

windows/win_file.ps1
```diff
@@ -50,6 +50,33 @@ $result = New-Object psobject @{
...
+    {
+        cmd.exe /c rmdir $filepath
```
nitzmahone: I don't love this either, but the P/Invoking the Win32 side would be a bit of pain (esp junction points via DeviceIoControl), fsutil has its own warts for this, and the PS FileSystem provider doesn't get symlink support until 5.0... So this works for now. Especially given the recursion, we should Try/Catch around this and include the filepath with the error message from cmd...

nitzmahone (changes requested):
Couple little things, but otherwise LGTM

jborean93:
@nitzmahone I have added some more checks and exception handling when deleting any items. I have also changed the ReparsePoint check to use the bitwise operator on the enum value. As well as this I have modified the symlink deletion to use the Powershell cmdlet if running on v5 or higher or use the Win32 API if on an older version.

Let me know if you had any more suggestions.

gregdek:
Thanks @jborean93 for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

jborean93:
ready_for_review

gregdek:
Thanks @jborean93. To the current maintainers, @jhawkesworth please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

windows/win_file.ps1
```diff
@@ -50,6 +50,52 @@ $result = New-Object psobject @{
...
+"@
+
+# Used to delete directories and files with logic on handling symbolic links
+Function Remove-File($file) {
+    Try {
+        If ($file.Attributes -band [System.IO.FileAttributes]::ReparsePoint) {
+            If ($PSVersionTable.PSVersion.Major -ge 5) {
+                Remove-Item -Path $file.FullName -Force
+            } Else {
+                Add-Type -TypeDefinition $symlink_util
```
nitzmahone: Minor perf nit- calling this in a recursive function gets kind of expensive (even though the generated type is cached, it's still parsing the code on each call). Ideally, the `Add-Type` should occur outside the loop.

windows/win_file.ps1
```diff
@@ -50,6 +50,52 @@ $result = New-Object psobject @{
...
+using System;
+using System.Runtime.InteropServices;
+
+namespace Ansible.Command
+{
+    public class DeleteSymLink
+    {
+        [DllImport("kernel32.dll", CharSet=CharSet.Unicode, SetLastError=true)]
+        public static extern bool RemoveDirectory(string lpPathName);
+    }
```
nitzmahone, line 72: Should probably have a managed wrapper around this that does the return-val check / `Marshal.GetLastWin32Error()` and passes it into/throws a `Win32Exception`.

windows/win_file.ps1
```diff
@@ -50,6 +50,52 @@ $result = New-Object psobject @{
...
+Function Remove-File($file) {
+    Try {
+        If ($file.Attributes -band [System.IO.FileAttributes]::ReparsePoint) {
+            If ($PSVersionTable.PSVersion.Major -ge 5) {
+                Remove-Item -Path $file.FullName -Force
+            } Else {
+                Add-Type -TypeDefinition $symlink_util
+                $result = [Ansible.Command.DeleteSymLink]::RemoveDirectory($file.FullName)
+                If (-not $result) {
+                    throw [System.SystemException] "Unknown Exception"
```
nitzmahone: If you do this on the C# side as suggested above, no need to check anything here as the managed exception will do the right thing (and automatically translate into a more useful error).

nitzmahone (changes requested):
Almost there... :)

jborean93:
@nitzmahone thanks for the info I have incorporated the changes you suggested into the latest commit.

ready_for_review

jborean93:
Changes moved to https://github.com/ansible/ansible/pull/19146 after repo merge

PR #5168, opened 2016-10-05
fixes user module not adding groups

### LINKED ISSUE #5163: User module does not add groups
##### ISSUE TYPE
- Feature Idea
- ~~Bug Report~~
##### COMPONENT NAME

user module
##### ANSIBLE VERSION

```
ansible 2.1.2.0
```
##### OS / ENVIRONMENT

Linux
##### SUMMARY

The `groups`argument for the `user` module is not parsed properly in this [line](https://github.com/diego-treitos/ansible-modules-core/blob/devel/system/user.py#L521).
The final group name has the format `'sftpjail']`. 
Error example:

```
fatal: [anstest]: FAILED! => {"changed": false, "failed": true, "msg": "Group  'sftpjail'] does not exist"
```
##### STEPS TO REPRODUCE

Just use the `user` module with the `groups` argument adding at least 2 groups (although it may fail with only 1).

```
- name: Create os user
  user:
      name: john
      password: smith
      shell: /usr/bin/rssh
      home: /path/to/jail
      group: john
      groups:
        - sshgroup
        - sftpjail
```
##### EXPECTED RESULTS

User created and added to groups
##### ACTUAL RESULTS

```
fatal: [anstest]: FAILED! => {"changed": false, "failed": true, "msg": "Group  'sftpjail'] does not exist"
```
##### COMMENTS

I think this is an important regression as it affects an important core module. It would be great to have a fix asap.

diego-treitos:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

`user` module
##### ANSIBLE VERSION

```
ansible 2.1.2.0
```
##### SUMMARY

The groups were not being properly parsed so I changed the way they are parsed from a list of strings converted to an string: stripping ' [' and ' ]' and then individually removing the string delimiters `'` or `"`.
## before

```
fatal: [anstest]: FAILED! => {"changed": false, "failed": true, "msg": "Group  'sftpjail'] does not exist"
```
## after

```
changed: [anstest]
```

Fixes #5163

### DIFF
--- a/system/user.py
+++ b/system/user.py
@@ -518,8 +518,10 @@ class User(object):
         if self.groups is None:
             return None
         info = self.user_info()
-        groups = set(filter(None, self.groups.split(',')))
-        for g in set(groups):
+        groups = set()
+        for g in filter(None, self.groups.strip('[]').split(',')):
+            groups.add( g.strip().strip( '"\'') )
+        for g in groups:
             if not self.group_exists(g):
                 self.module.fail_json(msg="Group %s does not exist" % (g))
             if info and remove_existing and self.group_info(g)[2] == info[3]:


bcoca:
The actual needed change here is to switch the arg spec from 'string' to 'list' so you don't have to do this string manipulation here.

PR #5176, opened 2016-10-06
Allow creation of Openstack volumes from an existing volume

manics:
##### ANSIBLE VERSION

```
ansible 2.1.2.0
  config file =
```
##### SUMMARY

Adds support to `os_volume` for creating volumes by cloning an existing volume.

`os_volume` supports creating Openstack volumes from scratch, or based on an existing image or snapshot, but not from an existing volume.
Depending on the storage implementation on the Openstack cloud in some cases creating a volume from an existing volume can be a very quick copy-on-write operation, compared to a time-consuming full copy when using an image or snapshot.

This PR adds a new parameter `volume_src` to the `os_volume` module, which behaves similarly to the existing `image` parameter in taking either the name or the ID of a volume.

For example:

```
- hosts: localhost
  tasks:
  - os_volume:
      state: present
      size: "500"
      display_name: "new-volume"
      volume_src: "existing-volume"
```

```

ok: [localhost]

TASK [os_volume] ***************************************************************
changed: [localhost]

localhost                  : ok=2    changed=1    unreachable=0    failed=0
```

If an invalid volume is set this returns an error:

```

ok: [localhost]

TASK [os_volume] ***************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to find volume source 'non-existent-volume'"}

```

If you have the `openstack` command line client installed then the output of `openstack volume show new-volume` should include a `source_volid`:

```
| source_volid                         | f383e3b3-3382-4573-9526-aa8b32a28aff  |
```

An unrelated issue: I've noticed that if a non-existent `image` parameter is passed to `os_volume` it silently creates a new volume. If you want I can add another commit along the lines of f1222f3.

### DIFF
--- a/cloud/openstack/os_volume.py
+++ b/cloud/openstack/os_volume.py
@@ -68,6 +68,7 @@ options:
        - Volume source to create from
      required: false
      default: None
+     version_added: "2.3"
    state:
      description:
        - Should the resource be present or absent.
--- a/cloud/openstack/os_volume.py
+++ b/cloud/openstack/os_volume.py
@@ -63,9 +63,9 @@ options:
        - Volume snapshot id to create from
      required: false
      default: None
-   volume_src:
+   volume:
      description:
-       - Volume source to create from
+       - Volume name or id to create from
      required: false
      default: None
      version_added: "2.3"
@@ -111,10 +111,10 @@ def _present_volume(module, cloud):
         image_id = cloud.get_image_id(module.params['image'])
         volume_args['imageRef'] = image_id
 
-    if module.params['volume_src']:
-        volume_id = cloud.get_volume_id(module.params['volume_src'])
+    if module.params['volume']:
+        volume_id = cloud.get_volume_id(module.params['volume'])
         if not volume_id:
-            module.fail_json(msg="Failed to find volume source '%s'" % module.params['volume_src'])
+            module.fail_json(msg="Failed to find volume '%s'" % module.params['volume'])
         volume_args['source_volid'] = volume_id
 
     volume = cloud.create_volume(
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @manics. To the current maintainers, @emonty, @shrews, @juliakreger, @j2sol, @rcarrillocruz please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

cloud/openstack/os_volume.py
```diff
@@ -63,6 +63,12 @@
        - Volume snapshot id to create from
      required: false
      default: None
+   volume_src:
+     description:
+       - Volume source to create from
```
omgjlk: Can this be a volume name, or does it have to be a UUID? You should clarify.

omgjlk: Also, we've got three nouns for potential input. `image`, `snapshot_id`, and `volume_src`. That's three different nouns with differing identifiers (none, id, src). It may be too late to fix `image`, but it would be good if the other two matched.

omgjlk (commented):
This looks good, I only have minor nits on the documentation and naming of arguments.

manics:
@j2sol Thanks for your review, I changed the parameter name to `volume` and updated the description as suggested.

It looks like `snapshot_id` is correctly named since the current implementation only works with IDs, though I think it could be changed to also accept a snapshot name by calling `cloud.get_volume_snapshot` first.

omgjlk (approved):

omgjlk:
shipit

gregdek:
Thanks again to @manics for this PR, and thanks @emonty, @shrews, @juliakreger, @j2sol, @rcarrillocruz for reviewing. Marking for inclusion.

PR #5257, opened 2016-10-13
Run validate-modules from ansible/ansible@devel

gundalow:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

validate_modules
##### Summary

Now that `validate-modules` is in `devel` thanks to https://github.com/ansible/ansible/pull/18001 we should use that version, rather than running from sivel/ansible-validate-modules. 

Until the three repos are merged this means we need access to the ansible/ansible source, so check that out. This will be removed post-merge.

To prove this worked I removed the `ANSIBLE_DEPRECATION_WARNINGS=false` line, which gave

```
2016-10-13 18:44:50 + python3.5 -m compileall -fq . -x '()'
2016-10-13 18:44:55 + /tmp/ansible/test/sanity/validate-modules/validate-modules --exclude '/utilities/|/shippable(/|$)' .
2016-10-13 18:44:56 [DEPRECATION WARNING]: ec2_ami_search is kept for backwards compatibility but 
2016-10-13 18:44:56 usage is discouraged. The module documentation details page may explain more 
2016-10-13 18:44:56 about this rationale..
```

NOTE: We now depend on `mock` and `voluptuous==0.8.8`, these will need adding into `ansible-test` requirements


test/utils/shippable/sanity.sh
```diff
@@ -13,18 +13,32 @@ if [ "${install_deps}" != "" ]; then
     apt-get update -qq
     apt-get install shellcheck
 
+    # FIXME REPOMERGE, remove the following
     pip install git+https://github.com/ansible/ansible.git@devel#egg=ansible
```
mattclay: I'm not sure we want to install ansible this way, since we basically end up cloning the repo twice. Once here and then again to get access to the validation script.

We should clone the repo once and use hacking/env-setup to make ansible available. Once the repos are merged, we'll be able to drop the clone (we'll already have the source) and just use hacking/env-setup.

gregdek:

test/utils/shippable/sanity.sh
sivel: In a way it is required. We need a copy of Ansible before and after. This is how we detect new modules. By effectively checking if the module in the PR is in the installed version of Ansible.

There may be other ways to work around this, but that would have to be solved.

sivel: Oh, I may have misread, in which case ignore me.

gundalow:
Dependencies added in https://github.com/ansible/ansible/pull/18028

test/utils/shippable/sanity.sh
gundalow: Yup, good point. I've got this working now. Will require work after the repos have been merged. We can perhap's use @mattclay's gitdiff tool to spot for new modules and check if they don't contain the correct top level `version_added:` line

gundalow:
ready_for_review

gregdek:
Thanks @gundalow for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

test/utils/shippable/sanity.sh
```diff
@@ -1,30 +1,44 @@
...
 if [ "${install_deps}" != "" ]; then
     add-apt-repository ppa:fkrull/deadsnakes && apt-get update -qq && apt-get install python2.4 -qq
 
     apt-add-repository 'deb http://archive.ubuntu.com/ubuntu trusty-backports universe'
     apt-get update -qq
     apt-get install shellcheck
 
-    pip install git+https://github.com/ansible/ansible.git@devel#egg=ansible
-    pip install git+https://github.com/sivel/ansible-testing.git#egg=ansible_testing
+    # FIXME REPOMERGE: No need to checkout ansible
```
mattclay: This looks like a leftover comment from a previous commit.

test/utils/shippable/sanity.sh
```diff
@@ -1,30 +1,44 @@
...
+    # Install dependencies for ansible and validate_modules
+    pip install -r "${build_dir}/test/utils/shippable/sanity-requirements.txt" --upgrade
+    pip list
+
 fi
 
+validate_modules="${build_dir}/test/sanity/validate-modules/validate-modules"
+
+
```
mattclay: Extra blank line.

test/utils/shippable/sanity.sh
```diff
@@ -1,30 +1,44 @@
 #!/bin/bash -eux
 
 source_root=$(python -c "from os import path; print(path.abspath(path.join(path.dirname('$0'), '../../..')))")
+ansible_repo_url="https://github.com/ansible/ansible.git"
```
mattclay: Move this inside the repomerge block below since it is only used there. Also, since it's only used once it can be inline in the clone command.

test/utils/shippable/sanity.sh
```diff
@@ -1,30 +1,44 @@
...
 
 install_deps="${INSTALL_DEPS:-}"
 
 cd "${source_root}"
 
+# FIXME REPOMERGE: No need to checkout ansible
+build_dir="/tmp/ansible"
+mkdir -p "${build_dir}"
```
mattclay: Git will create the directory when cloning.

test/utils/shippable/sanity.sh
```diff
@@ -6,25 +6,35 @@ install_deps="${INSTALL_DEPS:-}"
 
 cd "${source_root}"
 
+# FIXME REPOMERGE: No need to checkout ansible
+build_dir="/tmp/ansible"
+git clone "https://github.com/ansible/ansible.git" "${build_dir}" --recursive
```
mattclay, line 13: I'd install a trap after the clone to remove the temporary directory on exit, in case the script is run locally.

``` shell
trap 'rm -rf "${build_dir}"' EXIT
```

The rest looks good.

gundalow:
Added `trap`, from the logs

``` bash
2016-10-18 13:24:17 + build_dir=/tmp/tmp.AUtrprdtPt
2016-10-18 13:24:17 + trap 'rm -rf /tmp/tmp.AUtrprdtPt' EXIT
....
2016-10-18 13:26:06 + shellcheck test/utils/shippable/ci.sh test/utils/shippable/docs.sh test/utils/shippable/integration.sh test/utils/shippable/sanity.sh
2016-10-18 13:26:06 + rm -rf /tmp/tmp.AUtrprdtPt
```

test/utils/shippable/sanity.sh
```diff
@@ -7,7 +7,10 @@ install_deps="${INSTALL_DEPS:-}"
 cd "${source_root}"
 
 # FIXME REPOMERGE: No need to checkout ansible
-build_dir="/tmp/ansible"
+build_dir=$(mktemp -d)
+# shellcheck disable=SC2064
+trap "rm -rf $build_dir" EXIT
```
mattclay: You can put the trap in single quotes and avoid the shellcheck disable comment. There's no harm in evaluating the build_dir variable when the trap executes:

``` shell
trap 'rm -rf "{$build_dir}"' EXIT
```

mattclay: Sorry about the typo here. I meant to write this:

``` shell
trap 'rm -rf "${build_dir}"' EXIT
```

PR #5258, opened 2016-10-13
Add audit argument to put the module in noop

phil-dileo:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

eos_config
##### ANSIBLE VERSION

```
ansible 2.2.0.0 (detached HEAD 44faad0593) last updated 2016/10/13 17:08:12 (GMT +000)
  lib/ansible/modules/core: (detached HEAD d66983b43e) last updated 2016/10/13 17:08:40 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 35132b892f) last updated 2016/10/13 17:08:40 (GMT +000)
  config file =
```
##### SUMMARY

This argument allows a user to run the playbook in normal mode, and report back
what config changes _would_ have been made without actually making them.  This is
useful since some implementations require playbooks that query external APIs to gather
information that influences the configuration template.  This way, a user does not have to run
in check mode.

Sample playbook:

```

---
- hosts: eos
  gather_facts: no
  connection: local

  tasks:
    - name: "Tests audit feature"
      eos_config:
        "provider": "{{ eapi }}"
        "lines":
          - "hostname newhostname"
        "audit": True
```

### DIFF
--- a/network/eos/eos_config.py
+++ b/network/eos/eos_config.py
@@ -111,6 +111,20 @@ options:
     required: false
     default: false
     choices: ['yes', 'no']
+  audit:
+    description:
+      - This argument will cause the module to run as usual and perform
+        all necessary analysis between the designed configuration and
+        the current running-config, however, it will prohibit the module
+        from sending any commands to the switch. It will still output
+        which commands would have been run, and will return changed=True
+        since a change would have occurred. This argument allows you to run
+        the playbook in normal mode (not --check) and still gather useful
+        audit information.
+    required: false
+    default: false
+    choices: ['yes', 'no']
+    version_added: "2.2"
   backup:
     description:
       - This argument will cause the module to create a full backup of
--- a/network/eos/eos_config.py
+++ b/network/eos/eos_config.py
@@ -281,10 +281,13 @@ def run(module, result):
             result['updates'] = commands
 
         module.log('commands: %s' % commands)
-        load_config(module, commands, result)
+        if not module.params['audit']:
+            load_config(module, commands, result)
+        else:
+            result['changed'] = True
 
     if module.params['save']:
-        if not module.check_mode:
+        if not module.check_mode and not module.params['audit']:
             module.config.save_config()
         result['changed'] = True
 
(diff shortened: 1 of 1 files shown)


bcoca:
why not just use the built in 'check mode'?

phil-dileo:
Because the uri module doesn't support check mode you can't gather data from an API before running eos_config.  EG

```
---
- hosts: eos
  gather_facts: no
  connection: local

  tasks:
    - name: "Tests audit feature"
      uri:
        url: https://api.github.com/users/phil-dileo
        method: GET
        return_content: yes


    - name: "Tests audit feature"
      eos_config:
        "provider": "{{ eapi }}"
        "lines":
          - "hostname bobbyBoy"
        "audit": True
```

In check mode yields:

```
root@b59be8baeaf7:/test# ansible-playbook -i hosts test-audit.yaml -v --check
No config file found; using defaults

PLAY [eos] *********************************************************************

TASK [Tests audit feature] *****************************************************
skipping: [dc1-spine1] => {"changed": false, "msg": "remote module (uri) does not support check mode", "skipped": true}

TASK [Tests audit feature] *****************************************************
ok: [dc1-spine1] => {"changed": false, "warnings": []}

dc1-spine1                 : ok=1    changed=0    unreachable=0    failed=0
```

The motivation behind this enhancement is to avoid touching the uri module.

gregdek:
Thanks @phil-dileo for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

cidrblock:
I'm in favor of all network modules having this capability.  As enterprises transition to infrastructure as code, it's unrealistic to believe our ITSM process will follow as fast.  The ability to preview changes that would be made across all network modules is going to be key for enterprise adoption.

The alternative, if "check mode"  is the desired way to handle this, we'll need the ability to enable specific modules as needed.  For instance, we keep our snmp location strings in a NoSQL DB. But check mode disables URI.

bcoca:
yes, check_mode is something each module needs to implement, not unlike this patch, but it already exists and has a convention.

 If a module does NOT support check_mode it is skipped to avoid destructive actions (overridable if user is sure the action is not destructive) which makes it 'safe by default'.

phil-dileo:
Cool, just looked up http://docs.ansible.com/ansible/playbooks_checkmode.html. I didn't realize you could force it to run within the task.  Thanks for the tip.

PR #5261, opened 2016-10-13
amazon returns connection_draining_timeout as an int

sbussetti:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

cloud/amazon/ec2_elb_lb.py
##### ANSIBLE VERSION

```
ansible 2.3.0
```
##### SUMMARY

All this does is cast `connection_draining_timeout` to a string on both sides of the comparison (local vs the returned data from amazon), as amazon returns this value as a string but locally it's an integer.

Before the change, an elb command would always be marked as having changed, afterwards if everything remains the same, it is correctly noted as no change.

### DIFF
--- a/cloud/amazon/ec2_elb_lb.py
+++ b/cloud/amazon/ec2_elb_lb.py
@@ -962,7 +962,7 @@ class ElbManager(object):
         attributes = self.elb.get_attributes()
         if self.connection_draining_timeout is not None:
             if not attributes.connection_draining.enabled or \
-                    attributes.connection_draining.timeout != self.connection_draining_timeout:
+                    str(attributes.connection_draining.timeout) != str(self.connection_draining_timeout):
                 self.changed = True
             attributes.connection_draining.enabled = True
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @sbussetti. To the current maintainers, @jsdalton please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

jonhadfield:
+1 Fixes idempotency issue.
Not sure the second casting 'str(self.connection_draining_timeout)' is required, but don't see it doing any harm.
Also, it does add another future python 2 > 3 task, but assume the code is riddled with them.

sbussetti:
Second casting is admittedly just for readability. And str() is all over sure, but 2to3 should get those.

sbussetti:
Second casting is admittedly just for readability. And str() is all over
sure but 2to3 should get those.

~ Steve

PR #5265, opened 2016-10-13
Fix unarchive on python3

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

unarchive
##### SUMMARY

Since handler.files_in_archive is a list of files coming from
various executables output, that's a bytes list, and dest is a
str. So we need to convert to get the path.

I am however not 100% that's the right place and if that work fine with utf8 path.

### DIFF
--- a/files/unarchive.py
+++ b/files/unarchive.py
@@ -251,7 +251,7 @@ class ZipArchive(object):
             try:
                 for member in archive.namelist():
                     if member not in self.excludes:
-                        self._files_in_archive.append(member)
+                        self._files_in_archive.append(to_native(member))
             except:
                 archive.close()
                 raise UnarchiveError('Unable to list files in the archive')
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @mscherer. To the current maintainers, @dagwieers, @pileofrogs please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

mscherer:
Once this is done, we also need to remove 'test_unarchive' from the python3 blacklist ( https://github.com/ansible/ansible/blob/devel/test/utils/shippable/python3-test-tag-blacklist.txt )

abadger:
<nod>  It will work for utf8.  It won't work for random other encodings.   It's probably best to convert inside of each handler (convert at the borders-strategy).

abadger:
To be clear, I don't think we need to worry about other encodings right now, just utf8.  But applying the conversion inside the handlers would be good o do correctly now.

needs_revision

gregdek:

mscherer:
ready_for_review

abadger:

PR #5275, opened 2016-10-15
add new module do_sshkeys

pmarques:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME
- cloud/digital_ocean/do_sshkeys
- cloud/digital_ocean/do_sshkeys_facts
##### SUMMARY

Add module to manage (create, delete and rename)  Digital Ocean SSH Keys

### DIFF
--- a/cloud/digital_ocean/_digital_ocean_sshkey.py
+++ b/cloud/digital_ocean/_digital_ocean_sshkey.py
@@ -27,6 +27,7 @@ description:
      - Create/delete an SSH key.
 version_added: "1.6"
 author: "Michael Gregson (@mgregson)"
+deprecated: Deprecated in 2.3. Use do_sshkeys instead
 options:
   state:
     description:
--- a/cloud/digital_ocean/do_sshkeys.py
+++ b/cloud/digital_ocean/do_sshkeys.py
@@ -1,0 +1,262 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# (c) 2015, Patrick F. Marques <patrickfmarques@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+DOCUMENTATION = '''
+---
+module: do_sshkeys
+short_description: Manage DigitalOcean SSH keys
+description:
+     - Create/delete DigitalOcean SSH keys.
+version_added: "2.3"
+author: "Patrick Marques (@pmarques)"
+options:
+  state:
+    description:
+     - Indicate desired state of the target.
+    default: present
+    choices: ['present', 'absent']
+  fingerprint:
+    description:
+     - This is a unique identified for the SSH key used to delete a key
+    required: false
+    default: None
+  name:
+    description:
+     - The name for the SSH key
+    required: false
+    default: None
+  ssh_pub_key:
+    description:
+     - The Public SSH key to add.
+    required: false
+    default: None
+  oauth_token:
+    description:
+     - DigitalOcean OAuth token.
+    required: true
+
+notes:
+  - Version 2 of DigitalOcean API is used.
+requirements:
+  - "python >= 2.6"
+'''
+
+
+EXAMPLES = '''
+- name: "Create ssh key"
+  do_sshkeys:
+    name: "My SSH Public Key"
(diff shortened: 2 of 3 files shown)


harneksidhu:
I added additional error checking that may be useful for your modules.

HTTP Code>=500
https://github.com/harneksidhu/ansible-modules-core/blob/35119313a3a667d1067d6cce934c815f851d3db8/cloud/digital_ocean/do_droplet.py#L314

Since that logic is abstracted in the `REST` class, it could be a good candidate to move out to the `do_utils.py` file we will make in the near future.

HTTP Code >=400

I'm not sure if this is required, but I added further checks for any 4XX error that might be returned by the API that isn't documented. 

https://github.com/harneksidhu/ansible-modules-core/blob/35119313a3a667d1067d6cce934c815f851d3db8/cloud/digital_ocean/do_droplet.py#L359

gregdek:
Thanks @pmarques for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

pmarques:
@harneksidhu I'm also expect to move that "shared code" to the utils.
Toward the 5xx status code, I'm not a python expert and didn't check other modules but I would prefer to "exit" using **module.fail_json** instead of raising the exception.
If we raise the exception, I would like to see the stack trace printed alongside the message.

harneksidhu:
@pmarques 

The exception would be caught here

https://github.com/harneksidhu/ansible-modules-core/blob/4219_do_rework/cloud/digital_ocean/do_droplet.py#L545

which would then redirect to module.fail_json. It helps aggregate similar exceptions into one place so that module.fail_json would not be repeated in multiple lines.

pmarques:
Yes, I know that :)
After thinking a little bit more about it, that exception can be useful, for instance: I can ignore the 5xx error while wait/pool action status.

harneksidhu:
@pmarques :) We can also have it on the `do_utils.py` file and abstract 500 error handling from all modules through abstraction.

harneksidhu:
As part of this pull request, can you deprecate `digital_ocean_sshkey`? http://docs.ansible.com/ansible/developing_modules.html#deprecating-and-making-module-aliases

harneksidhu:
Edge case Bug:

If I run the following two times:

```
- name: Store public ssh key in digital ocean
    do_sshkeys:
      state: present
      ssh_pub_key: "{{ public_key.stdout }}"
      oauth_token: "{{ digital_ocean_api_token }}"
    register: key1
```

The output of running the playbook the second is:

```
TASK [Store public ssh key in digital ocean] ***********************************
changed: [localhost]
```

Expected output is:

```
TASK [Store public ssh key in digital ocean] ***********************************
ok: [localhost]
```

There is a loss of idempotency in this edge case.

pmarques:
@harneksidhu thanks reporting that, fixed :)

I was checking your [comment](https://github.com/ansible/ansible-modules-core/pull/5275#issuecomment-254019276) about the 5xx response and I don't think that return a message _Unable to reach api.digitalocean.com_ is the best message for the user when the error is a "Server side error".
I would suggest to return something like `"<status code> : <message>"` for instance (`"500: Internal server error"`) but I don't know if the API returns a JSON message. This should be more useful in case of bug reporting not only for ansible but also for Digital Ocean.

As you can infer from my last comments, I don't like to mask too much the errors from API, but is just my 2 cents :)

harneksidhu:
@pmarques Good suggestion! I'll make the change on my PR to `"500: Internal server error"`

harneksidhu:
@pmarques What is the status on the error handling for 5XX and 4XX errors on this module?

pmarques:
@harneksidhu please review the last commit, instead of hardcode the `"500 Internal Server Error"` I have used `response.info['msg']` let me know what do you think.

harneksidhu:
@pmarques I could not find on digital ocean api documentation if 500 errors return a message. https://developers.digitalocean.com/documentation/v2/#statuses

Have you been able to confirm that a message is returned?

pmarques:
@harneksidhu that 'msg' attribute comes from [the function fetch_url urls.py](https://github.com/ansible/ansible/blob/v2.2.0.0-1/lib/ansible/module_utils/urls.py#L910). You can see some lives bellow in [exception handling](https://github.com/ansible/ansible/blob/v2.2.0.0-1/lib/ansible/module_utils/urls.py#L965) that a message is added to info.
At least using urllib2 it will be a message [like this one](https://github.com/python-git/python/blob/master/Lib/urllib2.py#L165).
I've tested it manually

andrewsomething:
The DigitalOcean API will attempt to return an error message for 5XX errors in the same format as other error messages. E.g. a 500 error will return:

```
{
  "id": "api_error",
  "message": "Looks like something went wrong. If this continues, please open a support ticket so we can help you out."
}
```

harneksidhu:
@pmarques I will add `response.info['msg']` to my PR for 500 errors. Thank you and @andrewsomething for clarifying this. I commented on one minor fix in your commit and I should be good with this afterwards!

pmarques:
@andrewsomething thanks for the details. If the message gives some error ID to help support I'll update the module to print that _id_ and _message_ to the user. If the message is similar to that one I think we can do that when we merge info a common library.

The message that I'm using right know is just a HTTP generic message and I don't like to trick the user with generic messages, even being a edge case :)

Please let me know your thoughts.

pmarques:
Rebased, migration to [ansible repo](https://github.com/ansible/ansible) was not working.

PR #5281, opened 2016-10-16
Make pip module use pip3 on python 3

mscherer:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

pip
##### SUMMARY

This is linked to https://github.com/ansible/ansible/pull/18063

On python 3, if we want to be sure to run pip for python 3, we have to call pip3.

I couldn't test it locally for now, and the CI might not be run for that (since python 3 test is blacklisted, and this requires a PR on another git repository)

### DIFF
--- a/packaging/language/pip.py
+++ b/packaging/language/pip.py
@@ -114,6 +114,8 @@ options:
         example C(pip-3.3), if there are both Python 2.7 and 3.3 installations
         in the system and you want to run pip for the Python 3.3 installation.
         It cannot be specified together with the 'virtualenv' parameter (added in 2.1).
+        By default, it will take the appropriate version for the python interpreter
+        use by ansible, e.g. pip3 on python 3, and pip2 or pip on python 2.
     version_added: "1.3"
     required: false
     default: null
@@ -132,6 +134,8 @@ notes:
    - Please note that virtualenv (U(http://www.virtualenv.org/)) must be
      installed on the remote host if the virtualenv parameter is specified and
      the virtualenv needs to be created.
+   - By default, this module will use the appropriate version of pip for the
+     interpreter used by ansible (e.g. pip3 when using python 3, pip2 otherwise)
 requirements: [ "virtualenv", "pip" ]
 author: "Matt Wright (@mattupstate)"
 '''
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @mscherer. To the current maintainers, @mattupstate please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

abadger:
Questions:
- Do we want to do this or do we want to make the user use executable=pip3 in this instance?  It's technically a change in behaviour but since 2.2.0 is the first release to include python3 support, I think it's a change we can make if we want to.  The two choices I see are:
  - (What this PR does) Consistently use the python version which implements the module.  This is nice because people frequently use the pip module to install python modules that a later ansible module will need.
  - Consistently use the python version that the executable uses.  This is nice because people who use it to setup a system can depend on it always selecting the same version.
  - In either of the above cases, the default value can be changed by being explicit about what pip executable to use.
- Does this work in a virtualenv? ( I think if upstream is creating the pip3 name then it will but if it is in the distro packaging then it won't).  Maybe we need to push pip3 to the front of the list but allow for fallbacks?

mscherer:
So maybe we could place pip3 in the list, either first, or second, so we would cover most of the use case.

As discussed on irc, the biggest problem is when there is more than 1 version of python installed.
We have one one hand "entreprise distro" with a newer python. I imagine someone doing that using python2 for the time being, and so wanting pip3. But those people would specifies pip3 explicitly already (since that's how it work now).

On the other hand, we have people who want to deploy something on a python3 distribution (like, recent Ubuntu/Fedora), but did install python2. For example, for ansible (because this is still beta, and they didn't heard the good news), or because they need it for some reasons.

And that's the tricky case, because if they start to use python3 for ansible, this might switch the deps being used silently.

abadger:
jimi-c joined the discussion later and we decided that this approach (install for the python version that the ansible module is being run under by default) is sound.  Just some implementation changes to make:
- Change to using ansible.module_utils.six.PY3 for the python version check
- Update documentation (executable parameter and a note entry) to say that by default the module installs for the version of python that the ansible module is run under.  Use executable to specify a different target.
- Add pip2 to the front of https://github.com/ansible/ansible-modules-core/pull/5281/files#diff-1c140292edf64101e06d34f001c1d4e8R268 so that we first try a version of pip that we know is for python2 in that case.

abadger:
Merged to devel.  Cherry-picked to stable-2.2

Issue #4780, opened 2016-09-10
"authorized_key" re-orders the keys and comments in the pre-existing authorized_keys file


gerhard-tinned:
##### ISSUE TYPE

- Bug Report
##### COMPONENT NAME

system/authorized_key
##### ANSIBLE VERSION

```
ansible 2.1.1.0
```
##### OS / ENVIRONMENT

Managing from: LinuxMint
Managing: CentOS 6
##### SUMMARY

authorized_key is re-ordering the lines in a pre-existing authorized_keys file. This causes comments and the related lines to be re-ordered.
##### STEPS TO REPRODUCE

The managed host (B) should have a authorized_keys file containing multiple keys and comments. Example could look like this:

```
# The following is allice's key
ssh-rsa <key content-removed> ssh-key-comment-allice
# The following is bob's key
ssh-rsa <key content-removed> ssh-key-comment-bob
# The following is eve's key
ssh-rsa <key content-removed> ssh-key-comment-eve

```

After adding a key from another host (A) to the host (B) using the "authorized_key" task, below the pre-existing authorized_keys file as shown above is completely reordered.

```

---
- name: Generate ssh key on host (A)
  user: name=root generate_ssh_key=yes ssh_key_bits=4096 ssh_key_file=.ssh/id_rsa ssh_key_type=rsa

- name: Download generated ssh key from host (A)
  fetch: src=/root/.ssh/id_rsa.pub dest=/tmp/ansible/{{ ansible_hostname }}_id_rsa.pub 
         flat=yes validate_checksum=yes

- name: Upload/install the ssh-key from host (A) to host (B)
  authorized_key: user=root state=present 
      key="{{ lookup('file', '/tmp/ansible/{{ ansible_hostname }}_id_rsa.pub') }}"
  delegate_to: "{{ host_B }}"
```

##### EXPECTED RESULTS

```
# The following is allice's key
ssh-rsa <key content-removed> ssh-key-comment-allice
# The following is bob's key
ssh-rsa <key content-removed> ssh-key-comment-bob
# The following is eve's key
ssh-rsa <key content-removed> ssh-key-comment-eve
ssh-rsa <key content-removed> ssh-key-comment-joe
```

As shown in the example, it would be expected to see the added key (in the example "joe") at the end of the file keeping the pre-existing lines unchanged.
##### ACTUAL RESULTS

The result of the reordering might look like this containing the added key for "joe". The exact reason and ordering is not known at this point as of such the following example is just an example.

[...]

PR #3573, opened 2016-04-29
Allow authorized_keys module to accept multiple permitopen

linuxdynasty:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

authorized_key
##### ANSIBLE VERSION

```
ansible 2.0.2.0
```
##### SUMMARY

Allow for ssh options that can be included multiple times.
[permitopen](http://www.freebsd.org/cgi/man.cgi?sshd%288%29) can be passed multiple times.
Example
'command="/bin/false",no-pty,no-user-rc,no-X11-forwarding,no-agent-forwarding,permitopen="production-foo.com:5439",permitopen="development-foo.com:5439"'

This will allow the authorized_module to accept options that can be
passed multiple times into ssh options. For instance permitopen.

---

This change is [Reviewable](https://reviewable.io/reviews/ansible/ansible-modules-core/3573)

Fixes #1715

PR #5329, opened 2016-10-20
Fix authorized_key module to preserve the order of options

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

system/authorized_key.py
##### ANSIBLE VERSION

```
devel 2.2
```
##### SUMMARY

The last fix ( https://github.com/ansible/ansible-modules-core/pull/3573 ) allowing multiple definitions of the same option key (for
permitopen support) introduced a set() which removed the guaranteed
ordering of the options. 

This change restores ordering.  The change is larger than simply
removing the set because we do need to handle the non-dict semantics
around keys not being unique in the data structure.  The new code make
use of **setitem**() and items() to do its work.  Trying to use
getitem() or keys() should be looked upon with suspicion as neither of
those follow dictionary semantics and it is quite possible the coder
doesn't realize this.  The next time we need to touch or enhance the
keydict code it should probably be rewritten to not pretend to extend
the dictionary interface.

### DIFF
--- a/system/authorized_key.py
+++ b/system/authorized_key.py
@@ -143,7 +143,6 @@ EXAMPLES = '''
 #
 # see example in examples/playbooks
 
-import sys
 import os
 import pwd
 import os.path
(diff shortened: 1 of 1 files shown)


abadger:
This error becomes apparent on python3 because of the randomhashseed there making set order non-stable between runs.  It's probably also possible to provoke this in python2 if we use different initial data than in the current integration test.

abadger:
I'll merge this into devel tomorrow.  Want to talk to other people about whether it should be targetted at 2.2.0 or 2.2.1 before cherry-picking it to other branches.

abadger:
Once this is merged we should be able to merge this: https://github.com/ansible/ansible/pull/18053  to enable the tests.

gregdek:
Thanks @abadger for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

alikins:
Add a 'Fixes #4780' to commit messages

Assuming https://github.com/ansible/ansible-modules-core/issues/4780 is the correct issue.

abadger:
Took a look at #4780... It won't be fixed by this as this one only applies to options, not to the keys themselves.

abadger:
Merged to  devel and stable-2.2

PR #5348, opened 2016-10-21
Add a redhat_repos module to handle repos with subscription-manager

karmab:
##### ISSUE TYPE
- New Module Pull Request
##### COMPONENT NAME

redhat_repos
##### ANSIBLE VERSION

latest
##### SUMMARY

Add a redhat_repos module to handle repos with subscription-manager

```

```

### DIFF
--- a/packaging/os/redhat_repos.py
+++ b/packaging/os/redhat_repos.py
@@ -1,0 +1,107 @@
+#!/usr/bin/python
+
+# Karim Boumedhel (karim@redhat.com)
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from ansible.module_utils.basic import AnsibleModule
+import os
+
+DOCUMENTATION = '''
+module: redhat_repos
+short_description: Handles repositories for rhel machines
+description:
+    - Handles repositories for rhel machines
+version_added: "2.3"
+author: "Karim Boumedhel, @karmab"
+notes:
+    - This module doesn't handle subscriptions of the machine, only its repositories
+requirements:
+    - subscription manager and a rhel machine
+options:
+    repos:
+        description:
+            - a list of repositories to either add or remove
+        required: true
+        default: null
+    only:
+        description:
+            - whether the indicated repos should be the only one left to the system
+        required: false
+        default: no
+    state:
+        description:
+            - whether the repositories should be made present or absent
+        required: false
+        default: present
+'''
+
+EXAMPLES = '''
+- name: Assign Openstack Liberty Repositories
+  redhat_repos:
+   repos:
+     - rhel-7-server-rpms
+     - rhel-7-server-rh-common-rpms
+     - rhel-7-server-openstack-8-rpms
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @karmab for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

karmab:
ready_for_review

gregdek:
Thanks @karmab for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

karmab:
anything more i can do to speed this up ?

samdoran:
@karmab It looks like this module always returns a change. Is that the intent?

Also, I'm wondering if it would be better to enumerate the list of current repos with `subscription-manager repos --list-enabled` and then only add the ones that need to be added.

samdoran:
Probably also worth looking at the `yum` module to see how it handles a list as input vs individual items via a loop efficiently. Currently, passing in a list of items via a loop results in multiple calls to `subscription-manager` and does not combine them together into a single operation. It takes ~11s per repository.

```
TASK [redhat-subscription : Configure repository subscriptions] ****************
Friday 11 November 2016  13:26:45 -0500 (0:00:02.468)       0:00:03.791 *******
changed: [rhel7] => (item={u'state': u'present', u'name': u'rhel-7-server-extras-rpms'})
changed: [rhel7] => (item={u'name': u'rhel-7-server-rh-common-rpms'})
changed: [rhel7] => (item={u'name': u'rhel-7-server-openstack-8-rpms'})

rhel7                      : ok=5    changed=1    unreachable=0    failed=0

Friday 11 November 2016  13:27:19 -0500 (0:00:34.679)       0:00:38.471 *******
===============================================================================
redhat-subscription : Configure repository subscriptions --------------- 34.68s
redhat-subscription : Manage Red Hat subscription ----------------------- 2.47s
redhat-subscription : Configure Red Hat subscirption manager ------------ 0.67s
setup ------------------------------------------------------------------- 0.59s
redhat-subscription : Check that this is Red Hat ------------------------ 0.03s
```

karmab:
@samdoran actually no, there s a single call to subscription-manager
also speed is the reason why i dont check the current enabled repositories, hence changing every time the module is called.

Fale:
I think that the fact of returning "ok" or "changed" in the proper way is more important than speed. my2cc

evgeni:
Calling `subscription_manager` is quite expensive, as it will try to talk to whatever provides your subscriptions. But it also has an Python API (https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/repolib.py#L707) which could be used here:

```python
>>> from subscription_manager.repolib import RepoFile
>>> x = RepoFile()
>>> x.read()
>>> x.get('rhel-7-server-satellite-6.2-rpms', 'enabled')
'0'
>>> x.get('rhel-7-server-satellite-6.1-rpms', 'enabled')
'1'
```

This would have the disadvantage that we do not update the list of *available* repos from our upstream distributor, but instead we're *a lot* quicker.

Thoughts?

samdoran:
I think we need to take different approaches based on the combination of parameters passed to the module.

1) If `only: yes`, then list existing repos, diff with the list of declared repos, and create _one_ call to `subscription-manager` to enable and one to disable (or possible disable `*` how it is currently — not sure which is more expensive)

2) If a list of repos is passed without `only: yes`, either via a loop or directly as a list to the `repos` parameter, then construct a single call to `subscription-manager` (how it is currently).

Also need to account for passing multiple repos via a loop with differing state values as well as passing values that are regular expressions. Checking for a regexp is hopefully as straightforward as checking for a single repo via the API.

```yaml
- name: Add Red Hat repositories
  redhat_repository:
    repos: "{{ item.repo }}"
    state: "{{ item.state }}"
  with_items:
    - repo: rhel-7-server-extras-rpms
      state: present
    - repo: rhel-7-server-rpms
      state: absent
```

A loop like this currently results in one call to `subscription-manager` per item, making the execution time increase lineraly with the number of items.

PR #5350, opened 2016-10-21
Remove the yaml dep from the git module

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

source_control/git.py
##### ANSIBLE VERSION

```
devel 2.2
```
##### SUMMARY

Because this code was copied from the controller-side, it used the yaml library to parse key: value pairs.  We don't want to introduce that dep in module code.  The file format is simple enough for us to parse manually: https://git-scm.com/docs/gitrepository-layout


abadger:

PR #5353, opened 2016-10-21
Only change to short IDs for delete

### LINKED ISSUE #5237: [security] apt_key module does not verify key fingerprints
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

apt_key module
##### ANSIBLE VERSION

devel
##### CONFIGURATION

not relevant
##### OS / ENVIRONMENT

not relevant
##### SUMMARY

apt_key module does not verify key fingerprints and imports keys based on 16 digits long id.
This is a serious problem because it is easy to generate a OpenPGP key with a desired 16 digit key IDs.
Also, hkp is unauthenticated so it it becomes trivial to offer a wrong key to a victim.

The problem is [this workaround](https://github.com/ansible/ansible-modules-core/blob/683e5e4d1ad1e779d27d58af4ae7fdcda9881ab2/packaging/os/apt_key.py#L240-L250).
##### STEPS TO REPRODUCE

``` yaml

- name: Try to import key without verifying that we import the correct key
  hosts: 'dev.example.com'
  become: True

  tasks:

    - apt_key:
        id: '00000000000000000000000047AE7F72479BC94B'
        keyserver: 'hkp://pool.sks-keyservers.net'

```
##### EXPECTED RESULTS

That key does not exist. No key is added.
##### ACTUAL RESULTS

The key id 479BC94B is imported.
##### EXAMPLE FIX

Enforce correct key fingerprint with a trick such as:

``` shell

mkdir /tmp/gpg && chmod 700 /tmp/gpg
gpg --homedir /tmp/gpg --recv-keys --keyserver hkp://pool.sks-keyservers.net --recv 479BC94B
gpg --homedir /tmp/gpg --export DDA2C105C4B73A6649AD2BBD47AE7F72479BC94B | apt-key add -

```

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

apt_key.py
##### ANSIBLE VERSION

```
devel stable-2.2
```
##### SUMMARY

If the user specifies long IDs, use them for all commands except for
deleting a key.  Need to use short IDs there because of an upstream
apt_key bug.  Fixed in apt_key 1.10 (fix is present in Ubuntu 16.04 but
not Ubuntu 14.0 or some Debians).

Fixes #5237

Also:
- Some style cleanups
- Return a nice user error message if the key was not found on the
  keyserver
- Make file and keyring parameters type='path' so envars and tilde are
  expanded


gregdek:
Thanks @abadger. To the current maintainers, @jvantuyl please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

abadger:
Merged to devel and stbale-2.2.  This should be in the 2.2.0 release.

packaging/os/apt_key.py
```diff
@@ -237,64 +284,70 @@ def main():
...
-
+    fingerprint = short_key_id = key_id
+    short_format = False
     if key_id:
         try:
-            _ = int(key_id, 16)
-            if key_id.startswith('0x'):
-                key_id = key_id[2:]
-            key_id = key_id.upper()[-8:]
+            key_id, fingerprint, short_key_id = parse_key_id(key_id)
```
ypid, line 291: @abadger You mixed up the return order. It should be: `short_key_id, fingerprint, key_id = parse_key_id(key_id)`

ypid:
@abadger Thanks but even after the return fix it still showed ok for this task when the key is already present (the short key id only obvious):

``` YAML
- apt_key:
    # id: 'DDA2C105C4B73A6649AD2BBD47AE7F72479BC94B'
    id: '00000000000000000000000047AE7F72479BC94B'
  keyserver: 'hkp://pool.sks-keyservers.net'
```

Can someone else also review this? cc @jvantuyl

abadger:
Fixed the problem ypid identified here: https://github.com/ansible/ansible-modules-core/pull/5357

If we decide to backport this to stable-2.1, we'll need to cherry-pick that commit as well.

PR #5357, opened 2016-10-22
Order of return values was reversed

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

apt_key.py
##### ANSIBLE VERSION

```
devel 2.2
```
##### SUMMARY

My previous fix to apt_key reversed the order of return values and therefore compared the wrong length of ids.  This PR fixes that.


ypid (approved):
LGTM thanks! Fixes: #5237

abadger:
Thanks for testing @ypid!  Sorry that I broke it immediately after fixing it before.  

If you want to work on tests for apt_key, we have both integration and unit tests for modules.  The integration tests are just ansible roles and live here: https://github.com/ansible/ansible/tree/devel/test/integration/targets  You can look at how the apt tests are set up for ideas on how to setup the apt_key tests.

unittests live here: https://github.com/ansible/ansible/tree/devel/test/units/modules/core  I've mostly used those for very algorithmic code.  (The function that splits up key_id into 8, 16, and everything segments would be something easily tested via the unittests).

abadger:
Merged to devel and stable-2.2 branches.  This should be in the 2.2.0 release.

ypid:
@abadger Thanks for your work. I will look into unit testing when I come around to submit a new Ansible module but I can not promise it. For now, i pass, sorry.

PR #5358, opened 2016-10-22
Stat attr

bcoca:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

stat
##### ANSIBLE VERSION

```
2.3
```
##### SUMMARY

Depends on https://github.com/ansible/ansible/issues/18213

### DIFF
--- a/files/stat.py
+++ b/files/stat.py
@@ -53,16 +53,24 @@ options:
     default: sha1
     aliases: [ 'checksum_algo', 'checksum' ]
     version_added: "2.0"
-  mime:
+  get_mime:
     description:
       - Use file magic and return data about the nature of the file. this uses
         the 'file' utility found on most Linux/Unix systems.
       - This will add both `mime_type` and 'charset' fields to the return, if possible.
+      - In 2.3 this option changed from 'mime' to 'get_mime' and the default changed to 'Yes'
     required: false
     choices: [ Yes, No ]
-    default: No
+    default: Yes
     version_added: "2.1"
-    aliases: [ 'mime_type', 'mime-type' ]
+    aliases: [ 'mime', 'mime_type', 'mime-type' ]
+  get_attributes:
+    description:
+      - Get file attributes using lsattr tool if present.
+    required: false
+    default: True
+    version_added: "2.3"
+    aliases: [ 'attributes', 'attr' ]
 author: "Bruce Pennypacker (@bpennypacker)"
 '''
 
@@ -107,7 +115,7 @@ EXAMPLES = '''
 
 RETURN = '''
 stat:
-    description: dictionary containing all the stat data
+    description: dictionary containing all the stat data, some platforms might add additional fields
     returned: success
     type: dictionary
     contains:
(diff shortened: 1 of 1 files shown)


gregdek:

files/stat.py
```diff
@@ -469,15 +518,30 @@ def main():
...
+        filecmd = [module.get_bin_path('file'), '-i', path]
+        if filecmd:
+            try:
+                rc, out, err = module.run_command(filecmd)
+                if rc == 0:
+                    mimetype, charset = out.split(':')[1].split(';')
+                    output['mimetype'] = mimetype.strip()
+                    output['charset'] = charset.split('=')[1].strip()
+            except:
+                pass
```
abadger, line 500: It feels like we should change the default value of mime to True if we do this.  Right now the UI is:
- User explicitly requests that we get mime information for this file.  If we can't do that, because file is unavailable then fail. (Bug: if we can't do that because file fails then we should also fail).

If we make your change but don't change the default it reads as:
- User explicitly asked for mime information about this file.  Maybe we'll give that to them or maybe we won't.

If we make your change and change the default it then reads:
- Normally we get the mime information iif we can.  If the user explicitly says not to then we don't bother.

I like the last one the best.  The first one is my second choice (and we should fix non-zero exit to also fail_json if we do that)

files/stat.py
```diff
@@ -469,15 +518,30 @@ def main():
...
+
+    # try to get attr data
+    if get_attr:
+        output['version'] = None
+        output['attributes'] = []
+        filecmd = [module.get_bin_path('lsattr', False), '-vd', path]
```
abadger: Just a style nit (not a blocker) maybe name this attrcmd.

files/stat.py
```diff
@@ -331,6 +347,37 @@
...
+        'j': 'journalled',
+        'N': 'inline',
+        's': 'zero',
+        'S': 'synchronous',
+        't': 'notail',
+        'T': 'blockroot',
+        'u': 'undelete',
+        'X': 'compressedraw',
+        'Z': 'compresseddirty',
+    }
```
abadger: style nit, not a blocker.  If we move this list to a global then it keeps code and data separate  Can make the code's flow more readable.  Also means the data is only instantiated once but that's not a concern here.

bcoca: i was thinking of moving it into 'basic.py' for future usage (basic/file.py eventually)

files/stat.py
bcoca: changed

abadger:
Looks good!  +1 to merge

gregdek:

files/stat.py
```diff
@@ -469,15 +519,30 @@ def main():
...
-        try:
-            rc, out, err = module.run_command(filecmd)
-            if rc == 0:
-                mimetype, charset = out.split(':')[1].split(';')
-                output['mimetype'] = mimetype.strip()
-                output['charset'] = charset.split('=')[1].strip()
-        except:
-            pass
+        mimecmd = [module.get_bin_path('file'), '-i', path]
+        if mimecmd:
```
mattclay, line 491: This conditional will always be true. The code should be something like:

``` python
mimecmd = module.get_bin_path('file')
if mimecmd:
    mimecmd = [mimecmd, '-i', path]
```

If the command is not found, the module will fail.

files/stat.py
```diff
@@ -469,15 +519,30 @@ def main():
...
+                    output['charset'] = charset.split('=')[1].strip()
+            except:
+                pass
+
+    # try to get attr data
+    if get_attr:
+        output['version'] = None
+        output['attributes'] = []
+        attrcmd = [module.get_bin_path('lsattr', False), '-vd', path]
+        if attrcmd:
```
mattclay: This conditional will always be true. The code should be something like:

``` python
attrcmd = module.get_bin_path('lsattr')
if attrcmd:
    attrcmd = [attrcmd, '-vd', path]
```

If the command is not found, the module will fail.

gregdek:

gregdek:
@bcoca A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

PR #5359, opened 2016-10-23
change order when getting the pip executable (pip or pip2) in the cas…

### LINKED ISSUE #5347: pip uses pip2 and ignores virtualenv
##### ISSUE TYPE
- Bug Report
##### COMPONENT NAME

pip
##### ANSIBLE VERSION

```
ansible 2.3.0 (devel 2804e64ed5) last updated 2016/10/21 09:59:53 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 6f15dfd464) last updated 2016/10/18 16:01:36 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 3f77bb6857) last updated 2016/10/18 16:01:38 (GMT +200)
  config file = 
```
##### CONFIGURATION

ø
##### OS / ENVIRONMENT

Debian 8.6 jessie on both machines.
##### SUMMARY

Using the ansible `pip` module with a virtualenv, ansible uses `/usr/bin/pip2` instead of the shiny pip3 contained in the virtualenv.
##### STEPS TO REPRODUCE

```
  - name: pip install requirements
    become: yes
    become_user: test
    pip:
      chdir: /home/test/
      requirements: requirements.txt
      virtualenv: /home/test/
      virtualenv_command: pyvenv
      virtualenv_python: python3
```

```
fatal: [dev]: FAILED! => {"changed": false, "cmd": "/usr/bin/pip2 install -r requirements.txt",
```
##### EXPECTED RESULTS

I expect the `ansible pip module` to use `python3 -m pip`, or at least a `pip3` found in the virtualenv (there is one).
##### ACTUAL RESULTS

Ansible uses `/usr/bin/pip2`.

zserg:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

pip
##### ANSIBLE VERSION

2.3.0 0.0.devel
##### SUMMARY

Thist PR fixes #5347 in pip module.
Problem: _get_pip function firstly select pip2 executable in system directory instead pip in virtualenv.
This PR changes _get_pip function in the following way: if the virtualenv is used we search virtualenv directory firstly for all pip variants (pip, pip2, pip3 etc). If pip is not found in virtualenv we search then in the system path.


gregdek:
Thanks @zserg for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

zserg:
Shippable failed because of changes in ./files/ini_file.py. It's not from my PR. 
CI on shippable was broken today from run 1987 https://app.shippable.com/projects/573f79d02a8192902e20e34e/status/history.
ready_for_review

tntC4stl3:
I have the similar solution in my local environment, but I'm not sure whether it will effect other part or not.

The only different is, I think `pip` should be add to the `candidate_pip_basenames` as well when it is `PY3`.
Because when the default python interpreter is `Python3` and we want to create a `python2` virtualenv, we will still get the same problem.

zserg:
@tntC4stl3 Existing `_get_pip` search every item from the `candidate_pip_basenames` in the virutalenv path and also in the system path. I think, at first we have to  search all of the `candidate_pip_basenames` in the virtualenv path and then if no pip variant is found to repeat a search in the system path.

tntC4stl3:
@zserg What I mean is, for your solution, if the default Python interpreter in system (or the environment that Ansible run) is Python3, which means the `PY3` is `True`, the `candidate_pip_basenames` will equal to `['pip3']`. Then the program will only search `pip3` in virtualenv path and all system paths.

But what if I want to create a virtualenv of Python2 under the same system? The binary I need will be `pip` or `pip2`, not `pip3`, so I believe the `_get_pip` function can not work correctly in that situation....

JulienPalard:
Question may be why the executable parameter "cannot be specified together with the 'virtualenv' parameter"? Looks like we need it. Also, again, I'd like to be able to use `python3 -m pip` instead of `pip3`, which won't work as an `executable` argument, but this may be a completly unrelated issue.

zserg:
I think we have to use the executable parameter to select pip command in the system environment where we may have got the multiple versions of python installed. In virtualenv we have the one version of python and pip as well. As I see in [virtualenv doc](https://virtualenv.pypa.io/en/stable/userguide/), the executable should be only `pip` independently of python version we use in  virtualenv.

zserg:
ready_to_review

zserg:
ready_for_review

packaging/language/pip.py
```diff
@@ -273,7 +273,7 @@ def _get_pip(module, env=None, executable=None):
     # interpreter, so people can use pip to install modules dependencies
     candidate_pip_basenames = ['pip2', 'pip', 'python-pip', 'pip-python']
     if PY3:
-        candidate_pip_basenames = ['pip3']
+        candidate_pip_basenames = ['pip3', 'pip']
```
abadger, line 276: I'm not sure about this.  Linux distros are definitely making /usr/bin/pip the python2 version.  Adding pip here will mean that a site that has python2-pip installed from system packages but python3-pip is not installed will end up using the python2 version even though the python3 version is expected.  We should return an error to the user in that case instead.

packaging/language/pip.py
```diff
@@ -288,11 +288,17 @@ def _get_pip(module, env=None, executable=None):
...
+            for basename in candidate_pip_basenames:
+                pip = module.get_bin_path(basename, False, opt_dirs)
+                if pip is not None and pip == os.path.join(env,'bin',basename):
+                    break
+
+        if pip is None:
+          for basename in candidate_pip_basenames:
+              pip = module.get_bin_path(basename, False, opt_dirs)
+              if pip is not None:
+                  break
```
abadger, line 301: This looks like it might not go far enough.  I think if we're operating on a virtualenv we should check only the bin paths in the virtualenv.  If pip does not exist there we should return an error.

abadger:
@zserg I just submitted this alternative PR which implements my suggestions: https://github.com/ansible/ansible-modules-core/pull/5453  Do you think that will work or do you think that checking for a system pip when operating on a venv has value I'm not seeing?

zserg:
@abadger I think you are right. If we work in the virtualenv we have to use pip from the virtualenv. Your PR #5453 looks good.

PR #5360, opened 2016-10-23
Make service work when the service is not present in rc.conf

mscherer:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

service
##### SUMMARY

After installing a package from the ports collection on a
fresh FreeBSD 11.0, Ansible was unable to enable it, failing with
"unable to get current rcvar value". Debugging showed that sysrc
didn't see the variable from /usr/local/etc/rc.d/myservice, but
adding the value was working.

So we will just fallback to the default value if we can't find it.

### DIFF
--- a/system/service.py
+++ b/system/service.py
@@ -988,7 +988,7 @@ class FreeBsdService(Service):
         # and hope for the best.
         for rcvar in rcvars:
             if '=' in rcvar:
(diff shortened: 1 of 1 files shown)


gregdek:

kadecole:
I have verified that I also see this behavior on FreeBSD 11. This fix would be great. Thanks.

PR #5367, opened 2016-10-24
Fix uri for change in case in response

abadger:
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME

uri.py
##### ANSIBLE VERSION

```
devel 2.2
```
##### SUMMARY

In python3, response fields are title cased whereas in python2 they were
not.  We return these fields to the module's caller so we need to
normalize all of them to be lower case.

This reverts the lowercase check from 454f741ef5b56cccd123e12d7b2e6fe31d47c755
as that one was only targetted as a single field.


abadger:

PR #5381, opened 2016-10-25
Example of async reading taken from

caphrim007:
##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

windows/win_shell.ps1
##### ANSIBLE VERSION

```
ansible 2.3.0
  config file = /Users/trupp/src/trupp/azure-stack-bvt/ansible.cfg
```
##### SUMMARY

http://stackoverflow.com/questions/24370814/how-to-capture-process-output-asynchronously-in-powershell/24371479#24371479

That consistently works for my tests when using win_shell.

To illustrate my point.


caphrim007:
I forgot to add the stdout and stderr to the $result variable, but you get the idea.

PR #5389, opened 2016-10-26
uri: Add a second list of status codes, for indicating "changed" result

towolf:
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME

uri
##### ANSIBLE VERSION

```
ansible 2.2.0.0 (stable-2.2 deb1e3ebc7) last updated 2016/10/26 12:28:27 (GMT +200)
  lib/ansible/modules/core: (uri_add_status_codes_for_changed 8716e63075) last updated 2016/10/26 16:17:10 (GMT +200)
  lib/ansible/modules/extras: (fix_type_in_debconf 878c419f83) last updated 2016/09/12 09:44:07 (GMT +200)

```
##### SUMMARY

During URI requests we handle `status_code` to see if the Task failed or
not. But we _also_ need to distinguish `changed` or not. 

Currently the module always returns ok-green, but a REST API may be so
rich to return the HTTP status depending whether something was effected.

For example Graylog returns 400 when a user was already present and 201
when it was created. Similar for other endpoints.

I propose adding the `changed_code` option to the module to augment
status_code.

### DIFF
--- a/network/basics/uri.py
+++ b/network/basics/uri.py
@@ -114,6 +114,14 @@ options:
         request. Can also be comma separated list of status codes.
     required: false
     default: 200
+  changed_code:
+    description:
+      - A valid, numeric, HTTP status code that signifies a "changed" status
+        effected by the request. Can also be comma separated list of status
+        codes.
+    required: false
+    default: 200
+    version_added: "2.3"
   timeout:
     description:
       - The socket level timeout in seconds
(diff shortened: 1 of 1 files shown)


bcoca:
changed_when should provide this functionality

bcoca:
I would argue that we need to deprecate status_code in favor of `failed_when` as modules should not build these things internally when we have a general way to handle these conditions.

towolf:
`changed_when`, interesting, this didn't occur to me at all.

So should I withdraw this PR then?

bcoca:
Possibly, I'm not going to close it w/o other committers weighing in, in case there is an argument I'm missing.

sivel:
I'm in agreement that the proper solution is to leverage `changed_when`.

Possibly we could deprecate `status_code`, but with the number of people likely using it, might just be easiest to leave it around, and maybe add a note somewhere that the more flexible option is to utilize `failed_when` in that instance.

towolf:
Alright then, closing.

PR #5395, opened 2016-10-26
gce: Allow registering compute instance under service account parameter value

jbaublitz:
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

gce
##### ANSIBLE VERSION

```
ansible 2.2.0
```
##### SUMMARY

Currently, ansible only allows provisioning of Google Compute Instances under the default service account for the Google project in which it is provisioned. This change will take the unused `service_account_email` in the code or the `GCE_EMAIL` environment variable supported by libcloud (if defined) and register the compute node under that service account. I think there is room for discussion about if there should be a separate parameter for this given a few use cases I can think of (provisioning the node for use by another service account once provisioning is done for the purpose of more restrictive policy control). I can see one additional strength in this approach which is that currently libcloud handles all of the methods of authentication (passing in information via ansible parameters, environment variables, and libcloud secret file). If we have a separate parameter, we could cover the use case I mentioned and also make sure that there is no confusion if someone is using the libcloud secrets file. The secrets file would not currently trigger the ansible module to register the new instance under the desired service account which really could be considered a bug. While we could replicate the code to parse the libcloud secrets file, I don't particularly like that idea either for the sake of code duplication, abstracting away libcloud implementation details, etc. I have decided to take the approach I have at the moment mainly because I think this will make the API more straightforward in most cases and (in my experience at least) will cover the majority of use cases. That being said, I think the changes to the available parameters might be more flexible and am happy to implement that instead if that is decided to be the best option.

The output of the module does not change, only the state of the machine that has been provisioned.


gregdek:
Thanks @jbaublitz. To the current maintainers, @supertom please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

supertom:
Hey @jbaublitz, thanks for taking a look at this.  In fact, this just came up in #5376.  The use-case you bring up (one account to provision, one account for use) is the exact case discussed in that issue.

I'm reluctant to just use service_account_email as it changes existing behavior (using default account to using specified account), even though that behavior isn't great.

I want to talk it over with @ryansb and gut-check my assumption in the behavior change (as now you're the second person to bring it up) and then discuss a design with everyone.  And of course, happy to have your help!

jbaublitz:
Hi @supertom, I've been meaning to discuss this with Ansible/Google people for a bit but haven't had the opportunity until now. This has mainly come out of wanting `gcloud` and `gsutil` to work out of the box on a fresh Google Cloud instance deployment from ansible. Ping me when you've sorted out the implementation overview that you're looking for with @ryansb and I will update the pull request. Thanks!

ryansb:
This seems reasonable, since (as I read it) the change will work with existing plays just fine, but users that need the more specific control w/ service accounts. The golden rule is, of course, don't break playbooks.

In this case, I see your PR will omit the "default" parameter if there's no GCE email defined. Will that change the instances that are provisioned, or is that param filled in by the SDK?

jbaublitz:
Hmmm so I just checked to make sure and it appears the `email` parameter is required. I missed this the first time around. I'm guessing that the `default` value would not be required in terms of control flow because without an email defined either in an environment variable or parameter passed in directly, the module would fail authorization/authentication with the Google Cloud API. If we decide to add a new optional parameter for service account to register the instance under, then I think the `default` value would need to come back for the `email` parameter. Either way I am open to feedback. Nonetheless, I think it does work as long as all of the required parameters are passed in.

PR #5407, opened 2016-10-27
adding id: back to documentation fixes #2280

### LINKED ISSUE #2280: Documentation Bug: AWS EC2 idempotency
**Summary:**
Idempotency is not documented in the EC2 module even though it's supported.

**Description:**
The AWS EC2 cloud module supports idempotency, by passing 'id' to the module.
This used to be documented, but in commit c6b0d469, the following lines were removed. 

``` diff
-  id:
-    description:
-      - identifier for this instance or set of instances, so that the module will be idempotent with respect to EC2 instances. This identifier is valid for at least 24 hours after the termination of the instance, and should not be reused for another call later on. For details, see the desc
-    required: false
-    default: null
-    aliases: []
```

**Expected Results:**
Idempotency should be documented.

**Actual Results:**
Ablility to use the ec2 module idempotently is not documented.

chrrrles:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

ec2 module
##### ANSIBLE VERSION

```
ansible 2.3.0
  config file =
```
##### SUMMARY

Ansible docs do not explain the idempotent use of the ec2 module.  Adding back the description of the "id" field for the convenience of future users.

### DIFF
--- a/cloud/amazon/ec2.py
+++ b/cloud/amazon/ec2.py
@@ -29,6 +29,13 @@ options:
     required: false
     default: null
     aliases: ['keypair']
+  id:        
+    version_added: "1.1"
+    description:        
+      - identifier for this instance or set of instances, so that the module will be idempotent with respect to EC2 instances. This identifier is valid for at least 24 hours after the termination of the instance, and should not be reused for another call later on. For details, see the description of client token at U(http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Run_Instance_Idempotency.html).        
+    required: false        
+    default: null        
+    aliases: []        
   group:
     description:
       - security group (or list of groups) to use with the instance


gregdek:
Thanks @chrrrles for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

chrrrles:
Adding "version_added: 1.1"

chrrrles:
This might need discussion -- the PR won't pass CI because it believes the version_added should be 2.3, not 1.1.  What to do?

chrrrles:
@docschick - for the sake of passing CI I am wondering if I should bump `version_added` to 2.3, even though this feature was added in 1.1.

docschick:
I'm not in the loop enough to know the answer needed. @dharmabumstead -- what are your thoughts?

mattclay:
Since the only CI failure is the version check, which happens to be wrong in this case, this is OK to merge. Once the submodule refs are updated in ansible/ansible to point to the updated module with the "new" option documented, the problem will resolve itself.

dharmabumstead:
Thanks @chrrrles!

PR #5431, opened 2016-10-29
Improvement of Documentations with Native Yaml

rvben:
##### ISSUE TYPE
- Docs Pull Request
##### COMPONENT NAME

shell, add_host, group_by, git, hg, subversion, authorized_key, cron, group, hostname, seboolean, selinux, service, sysctl, systemd, user
##### ANSIBLE VERSION

```
n/a
```
##### SUMMARY

According to https://www.ansible.com/blog/ansible-best-practices-essentials, usage of Native Yaml is a best practice and should also be adopted in the documentation. Part three, on this topic, ended with "Documentation pull requests accepted.", so I went throught the different modules to see where Native Yaml was not yet used, and edited accordingly.

### DIFF
--- a/commands/shell.py
+++ b/commands/shell.py
@@ -68,11 +68,11 @@ notes:
       playbooks will follow the trend of using M(command) unless M(shell) is
       explicitly required. When running ad-hoc commands, use your best
       judgement.
-   -  To sanitize any variables passed to the shell module, you should use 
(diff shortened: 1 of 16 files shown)


bcoca:
# Question About This One

Hi!

Unfortunately, we're not sure if we want this feature in the program, and I don't want this to seem confrontational. We don't take style only changes.

PR #5436, opened 2016-10-31
fix the behavior of copy module using "force" args

zuiurs:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

copy

##### ANSIBLE VERSION

```
ansible 2.3.0
```

##### SUMMARY

This fixes the behavior that the dest is directory,
when we set the "force: no" argument.
To be join the dest and the src's basename,
before checking the "force" argument.


abadger:
This looks right to me.  I'm going to ask for a second reviewer to look since this is the copy module (pretty essential, don't want to break it).  If they think it's good as well we can merge.

PR #5471, opened 2016-11-03
added in extra info to win_stat that is in win_find. Added functional…

jborean93:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
win_stat

##### ANSIBLE VERSION
```
ansible 2.2.0.0
  config file =
```

##### SUMMARY
The following changes were made
- started to use newer powershell shared functions like Parse-Args and Get-AnsibleParam
- added support for choosing checksum algorithms similar to the stat module
- set the md5 return value to be md5 instead of sha1
- added support for getting info on hidden files
- synchronised the return value with the upcoming win_find module so we can find share values, if the file is a link/dir/readonly as well as many others
- updated the documentation to add in all this information


jborean93:
Tests for these changes are in https://github.com/ansible/ansible/pull/18339

jhawkesworth:
Hey thanks for this.

If I recall a little while back ansible switched from md5 to sha1 for checksums because md5 is not considered good enough for systems that have to reach FIPS standards.  

win_stat is used behind the scenes by other modules (win_template for example) so the ansible controller and the stat modules need to use a common file hashing algorithm otherwise the controller won't be able to tell if a file has changed in action plugins and then do correct processing.

Rather than having a private Get-Hash method in win_stat, I think it would be better to modify the Get-FileChecksum in the common powershell.ps1 code if necessary to provide non SHA1 hashes.

Personally I would favour removing md5 from win_stat and just having it return SHA1 checksum, and allowing a custom checksum to be returned based on the options available in .net (which could be md5 if so desired).

jborean93:
Thanks, I will look at modifying Get-FileChecksum to be able to return different hashes (sha1 by default). Is there a process of deprecating a return field or argument i.e removing the md5 return field and get_md5 argument? Or are you suggesting I keep the status quo as it is and allow the changing of the checksum field based on the changes I have made?

jborean93:
@jhawkesworth I've raised a change to modify the Get-FileChecksum method to allow different hash types here https://github.com/ansible/ansible/pull/18374. Once that is through I will modify this pull request based on your suggestions on what to do with the existing md5 field and argument.

Thanks for your help with my various pull requests I am very eager to learn the best ways of doing things and to contribute back to make Ansible and Windows easier to use for everyone.

gregdek:
Thanks @jborean93 for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

gregdek:
@jborean93 A friendly reminder: this pull request has been marked as needing your action. If you still believe that this PR applies, and you intend to address the issues with this PR, just let us know in the PR itself and we will keep it open pending your changes. When you do address the issues, please respond with ready_for_review in your comment, so that we can notify the maintainer.

jborean93:
Changes moved to https://github.com/ansible/ansible/pull/19148 after repo merge

PR #5482, opened 2016-11-03
docker_container - make json-file default logging driver PR #5269

Abdul-Anshad-A:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
docker_container

##### ANSIBLE VERSION
ansible 2.3.0

##### SUMMARY
According to docker documentation(https://docs.docker.com/engine/admin/logging/overview/), If the --log-driver option is not set, docker uses the default (json-file) logging driver.

Ansible should also handle this in a similar way as reported on  #PR #5269 

### Playbook

docker_container:
  name: test_container
  image: bash
  log_options:
    max-size: "512m"
    max-file: "5"

### Before - docker inspect

        "LogConfig": {
            "Type": "json-file",
            "Config": {}
        },

### After - docker inspect

        "LogConfig": {
            "Type": "json-file",
            "Config": {
                "max-file": "5",
                "max-size": "512m"
            }
        },

### DIFF
--- a/cloud/docker/docker_container.py
+++ b/cloud/docker/docker_container.py
@@ -1089,8 +1089,6 @@ class TaskParameters(DockerBaseClass):
         '''
         Create a LogConfig object
         '''
-        if self.log_driver is None:
-            return None
 
         options = dict(
             Type=self.log_driver,
@@ -1949,7 +1947,7 @@ def main():
         kill_signal=dict(type='str'),
         labels=dict(type='dict'),
         links=dict(type='list'),
-        log_driver=dict(type='str', choices=['json-file', 'syslog', 'journald', 'gelf', 'fluentd', 'awslogs', 'splunk'], default=None),
+        log_driver=dict(type='str', choices=['json-file', 'syslog', 'journald', 'gelf', 'fluentd', 'awslogs', 'splunk'], default='json-file'),
         log_options=dict(type='dict', aliases=['log_opt']),
         mac_address=dict(type='str'),
         memory=dict(type='str', default='0'),


gregdek:
Thanks @Abdul-Anshad-A for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

Kukunin:
I against this PR because default `log_driver` for docker daemon may be different that `json-file`. 
In this case, all new container manually created  without explicit `log_driver` will use global one (other than `json-file`). 
But all containers created with Ansible without explicit `log_driver` will fall down to `json-file` in all cases, even when global `log_driver` is different, which is wrong behaviour and hard to workaround.

Abdul-Anshad-A:
@Kukunin Thanks for you views.

But from official docker documentation ( https://docs.docker.com/engine/admin/logging/overview/ ),

`The container can have a different logging driver than the Docker daemon. Use the --log-driver=VALUE with the docker run command to configure the container’s logging driver. If the --log-driver option is not set, docker uses the default (json-file) logging driver.`

I still believe that its a legitimate PR.

Kukunin:
@Abdul-Anshad-A I've double checked, and posted issue to docker about wrong documentation. 

There are two levels of `log_driver` options: daemon-level and container-level. If container-level isn't explicitly set, it falls back to daemon-level, which isn't always `json-file`. 

Ansible breaks this behaviour.

I'm copying proof from that issue:
<details>
I have docker daemon with syslog log_driver:

    $ ps aux | grep dockerd
    root     12509  0.0  1.9 776004 39412 ?        Ssl  Nov25   0:21 /usr/bin/dockerd --log-driver=syslog --log-opt tag={{.Name}} --raw-logs

And I run new container without explicit `log_driver` option

    docker run --name test debian:jessie true

It shows `syslog` log_driver for that container:

    $ docker inspect test | grep -A 2 'LogConfig'
                "LogConfig": {
                    "Type": "syslog",
                    "Config": {
</details>

Kukunin:
Actually, behaviour provided by this PR is already used in Ansible 2.1 version's branch, and was changed by https://github.com/ansible/ansible-modules-core/commit/de0122fdaf5975cf2df87fc36b059c74032396e6 commit. 

What was reason for that commit?

Kukunin:
Actually, behaviour provided by this PR is already used in Ansible 2.1 version's branch, and was changed by https://github.com/ansible/ansible-modules-core/pull/4774

Abdul-Anshad-A:
@Kukunin Good work !

The confusion was created by incorrect documentation.

Updated doc ( https://deploy-preview-722--docker-docs.netlify.com/engine/admin/logging/overview/ ) makes much sense now.

I agree that this PR is not valid anymore. Closing it !

PR #5487, opened 2016-11-03
apt: fix changed when cache updated but not pkg

### LINKED ISSUE #5484: apt update_cache creates "changed" since ansible 2.2
##### ISSUE TYPE
 - Bug Report

##### COMPONENT NAME
apt

##### ANSIBLE VERSION
ansible 2.2.0.0 (stable-2.2 b229898f80)

##### CONFIGURATION
none

##### OS / ENVIRONMENT
Debian 8 64bit

##### SUMMARY
In ansible 2.1.3.0 using the apt module with update_cache: yes created an "ok" when deb package was already installed.

In ansible 2.2.0.0 (stable-2.2 b229898f80) the same task is creating a "changed" because of update_cache.

##### STEPS TO REPRODUCE

```
- name: testing
  hosts: all
  become: True
  tasks:
    - name: testing apt module
      apt:
        name: nano
        state: present
        update_cache: yes
```

##### EXPECTED RESULTS
Expected an "ok" if package nano is already installed,

Output in ansible 2.1.3.0
```
ok: [ansible-debian] => {"cache_update_time": 1478206604, "cache_updated": true, "changed": false, "invocation": {"module_args": {"allow_unauthenticated": false, "autoremove": false, "cache_valid_time": null, "deb": null, "default_release": null, "dpkg_options": "force-confdef,force-confold", "force": false, "install_recommends": null, "name": "nano", "only_upgrade": false, "package": ["nano"], "purge": false, "state": "present", "update_cache": true, "upgrade": null}, "module_name": "apt"}}
```

##### ACTUAL RESULTS
Return "changed" even package nano is already been installed on the system.

Output in ansible 2.2.0.0
```
changed: [ansible-debian] => {
    "cache_update_time": 1478206867,
    "cache_updated": true,
    "changed": true,
    "invocation": {
        "module_args": {
            "allow_unauthenticated": false,
            "autoremove": false,
            "cache_valid_time": 0,
            "deb": null,
            "default_release": null,
            "dpkg_options": "force-confdef,force-confold",
            "force": false,
            "install_recommends": null,
            "name": "nano",
            "only_upgrade": false,
            "package": [
                "nano"
            ],
            "purge": false,
            "state": "present",
            "update_cache": true,
            "upgrade": null
        },
        "module_name": "apt"
    }
}

```

resmo:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
apt

##### ANSIBLE VERSION

```
devel, 2.2
```

##### SUMMARY
In case `update_cache=yes` with `pkg` e.g. `apt: pkg=nginx update_cache=yes` don't report changed if `pkg` was not updated but cache was.

Fixes #5484 

/cc @cloudnull


bcoca:
so, I would argue that an updated cache IS changed.

resmo:
@bcoca I see your point and agree if `pkg=None`. 

if a `pkg` is given with `update_cache` then _changed_ it is misleading because the user can not find out if the pkg was updated or only cache. Updating cache does not really change a system. 

On the other hand the user can pretty easy find out if the cache was updated even if `changed=False`, the module returns `cache_updated: true`

Ansible 1.9 does not return changed for updating cache at all, even without `pkg`

And last, if a user really want it to be changed on cache update it is also pretty easy doable

```
apt: pkg=hello update_cache=yes
register: result
changed_when: "result|changed or result.cache_updated|bool == True"
```

Therefore I still vote for my PR to fix #5484 as it is a good compromise and the best user experience IMHO.

bcoca:
I would say that 'changed' applies in all cases when there is actual change, and updating cache is a change. 

What we might want to make sure is that the nature of the changes are reflected in both `msg` and `diff` returned from the module.

A similar case would be:

`apt: pkg=hello,goodbye`

when only 1 package is updated. Using the return a user could control the `changed_when` to suit their needs.

towolf:
Agree with @resmo. In the case of `apt: pkg=hello update_cache=yes`, changed should only be True when the package is upgraded.

mscherer:
So since resmo and bcoca asked for opinions:

```
18:34:14|  misc> mhh, that's a complex topic  
18:34:46|  misc> but I would be on resmo side
18:35:15|  misc> if I listen to a change for pkg, I am interested to see when the package got changed, not much for cache that do change every 6h
18:35:32|  misc> even if I see your point too, and technically, you are right
```

I would also like to see how do others packages manager act, but as said on irc, that's likely a specifity of the apt module, so we maybe can't abstract that.

abadger:
like @resmo, I would also expect that update_cache=True name=hello state=X  should return changed=False unless the _package_  was changed.  updating the cache is incidental to installing the package.

The apt module's ui is "special", though, in that it allows "update_cache=True" without specifying a name.  In that instance, the primary purpose of the command is to update the cache, not to modify the state of packages on the system.  So for that command I'd expect changed to depend on whether the cache was updated or not.

I think  update_cache functioning both as a modifier and a toplevel command is the problem here; it should either only be specified along with a packaging option (in which case updating the cache alone would not cause a change) or it should only be specified if a packaging option is not specified (in which case updating the cache would be entirely in control).  Either of those changes would break playbooks at this point, though, so they probably aren't feasible.

If I had to choose between update_cache always being reflected in changed or never being reflected in changed, I would say it should never be reflected.  As a whole, the apt module is primarily about installing, removing, and updating packages.  update_cache is an ancillary function to that.  Good UI makes the common path easy when there's a conflict between the packaging and cache functions we should be making changed reflect what happened to the packages on the system rather than what happened to the cache.

resmo:
merged based on comments.

dmsimard:
So what about this ?

```
tasks:
    - name: We don't care about this status
       apt:
           update_cache: yes
       changed_when: false

    - name: Install or update the package
       apt:
           package: foo
           state: latest
```

techraf:
- There is a boolean return value `cache_updated`

     If someone wants to verify if the cache update happened (regardless of the packages updated) - they can use it in `changed_when:` parameter.

- There is no return value to tell the user if the package was updated

    So once the cache was updated, there are no means to check for the actual package(s) - at least without splitting into multiple tasks.

The intention for having `cache_updated`-value seems clear: by default the task should **not** report "changed" in a situation when only cache is updated. And this PR fixes the bug.

PR #5507, opened 2016-11-06
systemd: Add boolean option to enable --no-block

towolf:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME

systemd

##### SUMMARY

In our environment we have custom services that need to be stopped and
restarted very gracefully to not interrupt active sessions.

A stop job, depending on the state, can take up to 20 minutes until the
process exits. It simply reacts to SIGTERM with a graceful shutdown.

By default, systemctl blocks until the job has completed, which leads to
Ansible hanging on this task for up to 20 minutes.

Thankfully systemctl supports the `--no-block` flag which lets the job
continue in the background.

This PR adds support for that flag as the `no_block` boolean option.

From the man page:

     --no-block
       Do not synchronously wait for the requested operation to
       finish. If this is not specified, the job will be
       verified, enqueued and systemctl will wait until the
       unit's start-up is completed. By passing this argument,
       it is only verified and enqueued. This option may not be
       combined with --wait.

### DIFF
--- a/system/systemd.py
+++ b/system/systemd.py
@@ -64,6 +64,14 @@ options:
         description:
             - run systemctl talking to the service manager of the calling user, rather than the service manager
               of the system.
+    no_block:
+        required: false
+        default: no
+        choices: [ "yes", "no" ]
+        description:
+            - Do not synchronously wait for the requested operation to finish.
+              Enqueued job will continue without Ansible blocking on its completion.
+        version_added: "2.3"
 notes:
     - One option other than name is required.
 requirements:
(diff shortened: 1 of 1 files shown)


mscherer:
Seems ok to me

bcoca:
I would reverse this setting `block=yes|no` also alias to `wait` which also applies to other init systems.

towolf:
Hm, well, this is complicated by the fact that systemd offers both with different functions:

```
   --no-block
       Do not synchronously wait for the requested operation to
       finish. If this is not specified, the job will be
       verified, enqueued and systemctl will wait until the
       unit's start-up is completed. By passing this argument,
       it is only verified and enqueued. This option may not be
       combined with --wait.

   --wait
       Synchronously wait for started units to terminate again.
       This option may not be combined with --no-block. Note
       that this will wait forever if any given unit never
       terminates (by itself or by getting stopped explicitly);
       particularly services which use "RemainAfterExit=yes".
```

`--no-block` disables waiting while the service is in the process of starting until it is _up_. And `--wait` enables waiting for the whole runtime of the service until it is _down again_.

Uniquely, systemd distinguishes between states `activating` and `active`, and `stopping` and `stopped`.

Since this is not a generic module but specifically for systemd I simply wanted to mirror the option of the underlying tool.

towolf:
If we added both `block:` and `wait:` it would be very unclear what does what and necessitate lengthy explanations. OTOH `--no-block` means absolutely no blocking/waiting.

Mapping both to the same function (because `--wait` in the context of Ansible doesn’t seem to make a lot of sense) would collide with the semantics of the underlying tool.

gregdek:
Thanks @towolf for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

PR #5547, opened 2016-11-10
add docker_swarm_service module

### LINKED ISSUE #4607: Support of new docker 1.12 features (swarm, service, stack)
##### ISSUE TYPE
- Feature Idea
##### COMPONENT NAME
- docker_service
- docker_stack
- docker_swarm
##### SUMMARY

Docker released new features with version 1.12. It would be great to have docker modules that supports them. Those features are:
###### # Docker service
- [service create](https://docs.docker.com/engine/reference/commandline/service_create/)
- [service inspect](https://docs.docker.com/engine/reference/commandline/service_inspect/)
- [service ls](https://docs.docker.com/engine/reference/commandline/service_ls/)
- [service rm](https://docs.docker.com/engine/reference/commandline/service_rm/)
- [service scale](https://docs.docker.com/engine/reference/commandline/service_scale/)
- [service ps](https://docs.docker.com/engine/reference/commandline/service_ps/)
- [service update](https://docs.docker.com/engine/reference/commandline/service_update/)
###### # Docker stack
- [stack deploy](https://docs.docker.com/engine/reference/commandline/stack_deploy/)
- [stack config](https://docs.docker.com/engine/reference/commandline/stack_config/)
- [stack rm](https://docs.docker.com/engine/reference/commandline/stack_rm/)
- [stack services](https://docs.docker.com/engine/reference/commandline/stack_services/)
- [stack tasks](https://docs.docker.com/engine/reference/commandline/stack_tasks/)
###### # Docker swarm
- [swarm init](https://docs.docker.com/engine/reference/commandline/swarm_init/)
- [swarm join](https://docs.docker.com/engine/reference/commandline/swarm_join/)
- [swarm leave](https://docs.docker.com/engine/reference/commandline/swarm_leave/)
- [swarm update](https://docs.docker.com/engine/reference/commandline/swarm_update/)
- [swarm join-token](https://docs.docker.com/engine/reference/commandline/swarm_join_token/)
- [node rm](https://docs.docker.com/engine/reference/commandline/node_rm/)

dariko:
##### ISSUE TYPE
 - New Module Pull Request

##### COMPONENT NAME
swarm_service

##### ANSIBLE VERSION
```
ansible 2.3.0 (devel 3072d75bcc)
```

##### SUMMARY
This PR introduces a new module: swarm_service
this module accepts parameters mapping those of the ``docker service (create|update)`` command.
Service updates will use swarm facilities, requiring service rebuild (remove -> create) only if strictly necessary (service network updating, as per https://github.com/docker/docker/issues/25876). Other modification will be run 'live'

This module requires a docker-py version more recent than commit https://github.com/docker/docker-py/commit/3ac73a285b2f370f6aa300d8a55c5af55660d0f4, which corrects some parameters.

Partially fixes #4607

### DIFF
--- a/cloud/docker/docker_swarm_service.py
+++ b/cloud/docker/docker_swarm_service.py
@@ -1,0 +1,806 @@
+#!/usr/bin/python
+
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+
+DOCUMENTATION = '''
+---
+module: docker_swarm_service
+author: "Dario Zanzico (@dariko)"
+short_description: docker swarm service
+description: |
+  Manage docker services. Allows live altering of already defined services
+  (see examples)
+version_added: "2.3"
+options:
+  name:
+    required: true
+    description:
+    - Service name
+  image:
+    required: true
+    description:
+    - Service image path and tag.
+      Maps docker service IMAGE parameter.
+  state:
+    required: true
+    description:
+    - Service state.
+    choices:
+    - present
+    - absent
+  args:
+    required: false
+    default: []
+    description:
+    - List comprised of the command and the arguments to be run inside
+      the container
+  constraints:
+    required: false
+    default: []
+    description:
+    - List of the service constraints.
+      Maps docker service --constraint option.
+  labels:
+    required: false
+    description:
+    - List of the service labels.
+      Maps docker service --label option.
+  container_labels:
+    required: false
+    description:
+    - List of the service containers labels.
+      Maps docker service --container-label option.
+    default: []
+  endpoint_mode:
+    required: false
+    description:
+    - Service endpoint mode.
+      Maps docker service --endpoint-mode option.
+    default: vip
+    choices:
+    - vip
+    - dnsrr
+  env:
+    required: false
+    default: []
+    description:
+    - List of the service environment variables.
+      Maps docker service --env option.
+  limit_cpu:
+    required: false
+    default: 0.000
+    description:
+    - Service CPU limit. 0 equals no limit.
+      Maps docker service --limit-cpu option.
+  reserve_cpu:
+    required: false
+    default: 0.000
+    description:
+    - Service CPU reservation. 0 equals no reservation.
+      Maps docker service --reserve-cpu option.
+  limit_memory:
+    required: false
+    default: 0
+    description:
+    - Service memory limit in MB. 0 equals no limit.
+      Maps docker service --limit-memory option.
+  reserve_memory:
+    required: false
+    default: 0
+    description:
+    - Service memory reservation in MB. 0 equals no reservation.
+      Maps docker service --reserve-memory option.
+  mode:
+    required: false
+    default: replicated
+    description:
+    - Service replication mode.
+      Maps docker service --mode option.
+  mounts:
+    required: false
+    description:
+    - List of dictionaries describing the service mounts.
+      Every item must be a dictionary exposing the keys
+      source
+      target
+      type (defaults to 'bind')
+      readonly (defaults to false)
+      Maps docker service --mount option.
+    default: []
+  networks:
+    required: false
+    default: []
+    description:
+    - List of the service networks names.
+      Maps docker service --network option.
+  publish:
+    default: []
+    required: false
+    description:
+    - List of dictionaries describing the service published ports.
+      Every item must be a dictionary exposing the keys
+      published_port
+      target_port
+      protocol (defaults to 'tcp')
+  replicas:
+    required: false
+    default: -1
+    description:
+    - Number of containers instantiated in the service. Valid only if
+      ``mode=='replicated'``.
+      If set to -1, and service is not present, service replicas will be
+      set to 1.
+      If set to -1, and service is present, service replicas will be
+      unchanged.
+      Maps docker service --replicas option.
+  restart_policy:
+    required: false
+    default: none
+    description:
+    - Restart condition of the service.
+      Maps docker service --restart-condition option.
+    choices:
+    - none
+    - on-failure
+    - any
+  restart_policy_attempts:
+    required: false
+    default: 0
+    description:
+    - Maximum number of service restarts.
+      Maps docker service --restart-max-attempts option.
+  restart_policy_delay:
+    required: false
+    default: 0
+    description:
+    - Delay between restarts.
+      Maps docker service --restart-delay option.
+  restart_policy_window:
+    required: false
+    default: 0
+    description:
+    - Restart policy evaluation window.
+      Maps docker service --restart-window option.
+  user:
+    required: false
+    default: root
+    description: username or UID
+requirements:
+  - "docker-py > https://github.com/docker/docker-py/commit/3ac73a285b2f370f6aa300d8a55c5af55660d0f4"
+'''
+
+RETURN = '''
+ansible_swarm_service:
+  description:
+  - Dictionary of variables representing the current state of the service.
+    Matches the module parameters format.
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @dariko for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

dariko:
ready_for_review

mscherer:
I ponder if it shouldn't be named docker_swarm_service

gregdek:
Thanks @dariko for this new module. When this module receives 'shipit' comments from two community members and any 'needs_revision' comments have been resolved, we will mark for inclusion.

dariko:
You're right, i renamed it

cloud/docker/docker_swarm_service.py
```diff
@@ -0,0 +1,594 @@
...
+    restart_policy: any
+    restart_policy_attempts: 5
+    restart_policy_window: 30
+- name: |
+    give mysql_service access to a network named backend.
+    this operation will destroy and rebuild the service as per
+    https://github.com/docker/docker/issues/25876
+  docker_swarm_service:
+    name: mydb
+    image: mysql:5.7
```
mscherer: I think the value need to be quoted (in case someone do cut and paste)

mscherer:
So looking at the whole code, I think part of it could be refactored to use the docker_common module code. See  https://github.com/ansible/ansible-modules-core/blob/devel/cloud/docker/docker_service.py

The code would take care of the path parameter and likely other, just see https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/docker_common.py

dariko:
as per @mscherer comment I quoted the image parameters and made the examples to be copypaste-able (renamed ports parameter to publish, added required env for mysql:5.7 container)

cloud/docker/docker_swarm_service.py
```diff
@@ -0,0 +1,615 @@
...
+    ds.limit_memory             = params['limit_memory']
+    ds.reserve_cpu              = params['reserve_cpu']
+    ds.reserve_memory           = params['reserve_memory']
+    ds.publish                  = params['publish']
+    return ds
+
+  def compare_services(self, s1, s2):
+    changes = []
+    need_recreate = False
+    if s1.endpoint_mode!=s2.endpoint_mode:
```
derfred: This function does not detect changes in the image. I added this:

```
if s1.image!=s2.image:
      changes.append('image')
```

cloud/docker/docker_swarm_service.py
```diff
@@ -0,0 +1,615 @@
...
+        endpoint_spec['Ports'] = []
+        for port in self.publish:
+          endpoint_spec['Ports'].append({
+            'Protocol': port['protocol'],
+            'PublishedPort': port['published_port'],
+            'TargetPort': port['target_port']
+          })
+        return task_template, networks, endpoint_spec, mode, self.labels
+
+  def __init__(self, socket_path=None):
```
derfred: The other docker modules allow setting the `docker_url` or `docker_host`. This is useful if you want to access a docker engine on a different machine via the network.

cloud/docker/docker_swarm_service.py
dariko: yep, thanks

cloud/docker/docker_swarm_service.py
dariko: I'm nearly done reingeneering the module for using the classes defined in docker_common.py (as suggested by @mscherer ), thus making those shared parameters available in this module.

dariko:
Rewrote the module, now uses `AnsibleDockerClient` to manage the docker connection.

cloud/docker/docker_swarm_service.py
```diff
@@ -0,0 +1,795 @@
...
+      restart_policy=types.RestartPolicy(
+        condition     = self.restart_policy,
+        delay         = self.restart_policy_delay,
+        max_attempts  = self.restart_policy_attempts,
+        window        = self.restart_policy_window
+      )
+      resources={
+        'Limits': {
+          'NanoCPUs': int(self.limit_cpu * 1000000000),
+          'MemoryBytes': self.limit_memory
```
derfred: to be consistent with the definition in the docs this needs to be multiplied by ```1024*1024```

cloud/docker/docker_swarm_service.py
```diff
@@ -0,0 +1,795 @@
...
+        },
+        'Reservations': {
+          'NanoCPUs': int(self.reserve_cpu * 1000000000),
+          'MemoryBytes': self.reserve_memory
```
derfred: same here wrt to the definition of the field being in megabytes.

derfred:
I would like to see the ```replicas``` param be handled differently from the others. In my setup the scaling is separate from the deployment. Currently with the default value being 1, each deploy scales the service back to one replica.

For my purposes the following logic would work:
1. If replicas is set, change the number of replicas.
2. If replicas is not set, create a service with one replica if the service is new but do not change the number of replicas for existing services.

cloud/docker/docker_swarm_service.py
dariko: thank you

dariko:
@derfred The approach you propose would break idempotence, maybe we could have an additional boolean parameter `ignore_replicas`, defaulting to false, which acts as you suggested (with a warning in the docs about the idempotence breaking problem) ?

| ignore_replicas  | replicas  | current_replicas |behaviour |
|---|---|---|---|
| false | X\|default(1)  | * | scale the service to X |
| true  | X\|default(1) | `new service` | scale the service to X\|default(1) |
| true | X\|default(1) | Y | scale the service to Y |

cloud/docker/docker_swarm_service.py
dariko: yep

derfred:
I don't think the behaviour in the original version or in the default of your proposal is intuitive. If I manually scale a service and then run this module without specifying a replica count, I would not expect it change my system settings.

As to my proposal I don't think this breaks the idempotency of the operation. As per my understanding in order for it to be idempotent multiple applications of the operation should yield the same result.

Looking at the logic:

situation | first run | second run
-|--------|-----
new service & replicas=X | actual replicas: X | actual replicas: X
new service & replicas not set | actual replicas: 1 | actual replicas: 1
existing service with Y replicas & replicas = X | actual replicas: X | actual replicas: X
existing service with Y replicas & replicas not set | actual replicas: Y | actual replicas: Y

What do you think?

dariko:
You're right, I'm gonna change che module logic.

cloud/docker/docker_swarm_service.py
```diff
@@ -0,0 +1,795 @@
...
+      service_m['readonly'] = bool(param_m.get('readonly', False))
+      service_m['type']     = param_m.get('type', 'bind')
+      service_m['source']   = param_m['source']
+      service_m['target']   = param_m['target']
+      s.mounts.append(service_m)
+    return s
+
+  def compare(self,os):
+    differences = []
+    can_edit_live = True
```
derfred: The ```needs_rebuild``` logic does not work. If the service exists it is always rebuilt, because this value is always ```True```. Changing this line to ```can_edit_live = False``` fixes it.

dariko: yes, I also renamed the variable for consistency.

dariko:
@derfred The last commit adds a special value of `-1` (default) to the `replicas` parameter. I tested it with these tasks:
```
- name: remove service
  docker_swarm_service:
    name: myservice
    state: absent
- name: create service without replicas parameter
  docker_swarm_service:
    name: myservice
    image: "nginx"
  register: out1
- name: test replica count
  fail:
    msg: wrong replica count
  when: "{{out1['ansible_docker_service']['replicas']!=1}}"
- name: manually scale service
  command: docker service scale myservice=4
- name: keep service as is
  docker_swarm_service:
    name: myservice
    image: "nginx"
  register: out2
- name: test replicas unchanged
  fail:
    msg: "replicas!=4"
  when: "{{out2['ansible_docker_service']['replicas']!=4}}"
```
Does it work as you suggested?

dariko:
i squashed the commits and rebased to  ansible@567a9675b233d062535f0fefd13eb93322b4dd30 but I don't understand what the shippable error is about

derfred:
@dariko I will add it to my deploy today and test it

dariko:
reopened as ansible/ansible#18962

derfred:
@dariko I've added this to our deploy and tested it in various situations. Works great, thank you!

wilkko:
What about `--with-registry-auth`? Does the module work with private registries?

dariko:
No: I'm gonna take a look at what the command line client does with that option.
I'm closing this PR, it's been migrated to ansible/ansible/pull/19229

PR #5599, opened 2016-11-14
Revert "stat: doc: add version for new returns"

bcoca:
Reverts ansible/ansible-modules-core#5594

using version_added field instead of adding to description


resmo:
got it ;)

PR #5601, opened 2016-11-14
Do not require password when deleting os_user

omgjlk:
- Bugfix Pull Request

os_user

```
devel
```

```

```

I broke backwards compat with the addition to define when a password
should be updated. It was requiring that a password value be passed when
deleting a user, which seems silly.

This moves the argument logic out of the argument spec and into when it
would be needed, when state is present.

### DIFF
--- a/cloud/openstack/os_user.py
+++ b/cloud/openstack/os_user.py
@@ -191,10 +191,6 @@ def main():
     module_kwargs = openstack_module_kwargs()
     module = AnsibleModule(
         argument_spec,
-        required_if=[
-            ('update_password', 'always', ['password']),
-            ('update_password', 'on_create', ['password']),
-        ],
         **module_kwargs)
 
     if not HAS_SHADE:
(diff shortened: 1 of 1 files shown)


cloud/openstack/os_user.py
```diff
@@ -219,6 +215,10 @@ def main():
             domain_id = _get_domain_id(opcloud, domain)
 
         if state == 'present':
+            if update_password in ('always', 'on_create'):
+                if not password:
+                    msg = ("update_password is %s but a password value is "
```
Thingee, line 220: Where does msg actually get used?

omgjlk: Lol, good catch, I was a bit too hurried this morning. I intended to use it in a `fail_json` call, which I have just done with a newer change set.

gregdek:
Thanks @j2sol for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

omgjlk:
This build failure is due to a module outside of this PR.

Thingee:
Tested, looks good to me!

PR #5609, opened 2016-11-15
Update system/group.py module on Darwin so system=yes works.

rqelibari:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME
system/group.py module

##### ANSIBLE VERSION
```
ansible 2.3.0 (devel 11465134fa) last updated 2016/11/14 21:27:46 (GMT +200)
  lib/ansible/modules/core: (detached HEAD 8d1fe3b444) last updated 2016/11/14 21:34:36 (GMT +200)
  lib/ansible/modules/extras: (detached HEAD 88e58df86b) last updated 2016/11/14 21:34:49 (GMT +200)
  config file = 
```

##### SUMMARY
Extend group module for Darwin operating systems such that system options now generates a system
gid (< 500) if possible.

Add `get_lowest_available_system_gid` method to `DarwinGroup` and use it, when user has not
provided a `gid` and has set `system=yes` option. Only call `get_lowest_available_system_gid` in `group_add` method. Like that it stays idempotent.

Commit message: Add ability to add system groups with next free system gid (< 500) on macOS.

### DIFF
--- a/system/group.py
+++ b/system/group.py
@@ -269,6 +269,11 @@ class DarwinGroup(Group):
         cmd += [ '-o', 'create' ]
         if self.gid is not None:
             cmd += [ '-i', self.gid ]
+        elif 'system' in kwargs and kwargs['system'] == True:
+            gid = self.get_lowest_available_system_gid()
+            if gid != False:
+                self.gid = str(gid)
+                cmd += [ '-i', self.gid ]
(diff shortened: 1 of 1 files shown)


rqelibari:
What exactly did here go wrong?

PR #5613, opened 2016-11-15
AWS Elasticache module check mode, replication group, and security group names

calvinli723:
##### ISSUE TYPE
 - Feature Pull Request

##### COMPONENT NAME

AWS elasticache module

##### ANSIBLE VERSION

```
ansible 1.9.1
```

##### SUMMARY

The module previously had not supported check mode, so this pull request was partly made to address that by only making API calls when check mode is not activated. Also, there was no support for managing replication groups for elasticache clusters, so this pull request also attempts to address that. Also, it may be useful to pass in security group names instead of ids, since you might not want to hardcode security group ids in your configuration management, so the module can query for the ids using these names.


gregdek:
Thanks @calvinli723. To the current maintainers, @jsdalton, @alachaum please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

PR #5624, opened 2016-11-15
Examples syntax batch7

samdoran:
##### ISSUE TYPE
 - Docs Pull Request

##### COMPONENT NAME
- nxos_feature
- nxos_hsrp
- nxos_igmp
- nxos_interface
- nxos_interface_ospf
- nxos_ip_interface
- nxos_ping
- nxos_switchport
- nxos_vlan
- nxos_vrf
- nxos_vrf_interface
- nxos_vrrp
- meta
- set_fact
- win_copy
- win_file
- win_get_url
- win_lineinfile
- win_msi
- win_stat

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = /Users/sdoran/.ansible.cfg
```

##### SUMMARY

Last batch!

Per Ansible Core team meeitng on [2016-11-03](https://meetbot.fedoraproject.org/ansible-meeting/2016-11-03/public_core_meeting.2016-11-03-15.02.html), change syntax in EXAMPLES in Core modules to use multi-line YAML.


windows/win_copy.py
```diff
@@ -48,16 +48,22 @@
 
 EXAMPLES = '''
 # Copy a single file
-- win_copy: src=/srv/myfiles/foo.conf dest=c:\\TEMP\\foo.conf
+- win_copy:
+    src: /srv/myfiles/foo.conf
+    dest: c:\TEMP\foo.conf
```
samdoran, line 53: Quotes aren't needed for paths in the Windows modules.

windows/win_file.py
```diff
@@ -56,17 +56,26 @@
...
 EXAMPLES = '''
 # create a file
-- win_file: path=C:\\temp\\foo.conf 
+- win_file:
+    path: C:\temp\foo.conf
 
 # touch a file (creates if not present, updates modification time if present)
-- win_file: path=C:\\temp\\foo.conf state=touch 
+- win_file:
+    path: C:\temp\foo.conf
```
samdoran, line 65: Double backslashes aren't needed for unquoted YAML, but _are_ needed for `k=v` pairs and when single quoting the path in YAML. This may be a case where we should explicitly put in a `k=v` example showing that.

gregdek:
Thanks @samdoran for this PR. This PR requires revisions, either because it fails to build or by reviewer request. Please make the suggested revisions. When you are done, please comment with text 'ready_for_review' and we will put this PR back into review.

windows/win_copy.py
```diff
@@ -48,16 +48,22 @@
...
 
 # Copy the contents of files/temp_files dir into c:\temp\.  Includes any sub dirs under files/temp_files
-# Note the use of unix style path in the dest.  
+# Note the use of unix style path in the dest.
 # This is necessary because \ is yaml escape sequence
-- win_copy: src=files/temp_files/ dest=c:/temp/
+- win_copy:
+    src: files/temp_files/
+    dest: c:/temp/
```
gundalow: `/` or `\` for `dest:` on Windows machines, we have both listed in this module

The note on line 56 isn't clear

windows/win_file.py
gundalow: I think a note here (and in all the windows modules) would be good.

We could maybe setup a docs_fragment note explaining this that we include in all of `windows/` modules. What do you think @nitzmahone  ?

gundalow: Using the Windows modules can you write directly to network shares `//server/foo/bar`, if so is anything needed to escape that correctly?

network/nxos/nxos_igmp.py
```diff
@@ -61,21 +61,28 @@
...
+    enforce_rtr_alert: true
+    host: "{{ inventory_hostname }}"
+
+- name: Restart the igmp process
+  nxos_igmp:
+    restart: true
+    host: "{{ inventory_hostname }}"
+
+- name: Configure a simple ASN
+  nxos_bgp:
```
gundalow: `nxos_bgp` in the `nxos_igmp` module, this doesn't seem to follow on

gundalow (changes requested):
needs_revision

windows/win_lineinfile.py
```diff
@@ -100,20 +100,43 @@
 """
 
 EXAMPLES = """
```
gundalow:
```
python3.5 -m compileall -fq windows/win_lineinfile.py"
2016-11-15 20:47:21 *** Error compiling './windows/win_lineinfile.py'...
2016-11-15 20:47:21   File "./windows/win_lineinfile.py", line 142
2016-11-15 20:47:21     """
2016-11-15 20:47:21      ^
2016-11-15 20:47:21 SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 663-664: truncated \uXXXX escape
```

network/nxos/nxos_igmp.py
samdoran: I removed this example. Seems to be a copy/paste error from the `nxos_bgp` module.

windows/win_lineinfile.py
samdoran: I marked examples section as containing regexp. Hopefully that will fix the error.

windows/win_copy.py
nitzmahone: Yeah, with an unquoted arg, backslashes are fine- no need for the comment, though this module is also fine with forward-slashes.

windows/win_file.py
```diff
@@ -56,17 +56,26 @@
 
 EXAMPLES = '''
 # create a file
-- win_file: path=C:\\temp\\foo.conf 
+- win_file:
+    path: C:\temp\foo.conf
```
nitzmahone, line 60: This example is actually broken- path without state is just an error.

windows/win_file.py
nitzmahone: I generally hate doc fragments, but building one that just links to a page describing all the different ways to deal with Windows paths (only for Windows modules that have path args) might not be a bad thing. I'd rather not put all that content inline though. I'm doing a bunch of Windows doc work, so I can add that page to my list of things to build.

windows/win_lineinfile.py
samdoran: Test 16 seems to be passing with this change.

windows/win_get_url.py
```diff
@@ -96,18 +96,18 @@
...
 
-- name: Download earthrise.jpg to 'C:\\Users\\RandomUser\\earthrise.jpg' through a proxy server.
+- name: Download earthrise.jpg to 'C:\Users\RandomUser\earthrise.jpg' through a proxy server.
   win_get_url:
     url: 'http://www.example.com/earthrise.jpg'
-    dest: 'C:\\Users\\RandomUser\\earthrise.jpg'
+    dest: 'C:\Users\RandomUser\earthrise.jpg'
     proxy_url: 'http://10.0.0.1:8080'
     proxy_username: 'username'
     proxy_password: 'password'
```
nitzmahone: I'd just nuke the quotes on all these- we'll handle that in the "how to do Windows paths" doc fragment.

windows/win_lineinfile.py
```diff
@@ -99,21 +99,44 @@
...
+    encoding: "utf-16"
+    line: "This is a utf-16 encoded file"
 
 # Add a line to a file and ensure the resulting file uses unix line separators
-- win_lineinfile: dest=C:\\temp\\testfile.txt line="Line added to file" newline="unix"
-
+- win_lineinfile:
+    dest: C:\temp\testfile.txt
+    line: "Line added to file"
+    newline: "unix"
```
nitzmahone: We can ditch all these quotes with the k=v conversion (well, except maybe the ones with embedded =, not sure about that).

nitzmahone (changes requested):
A few little things to change. Let's just go consistent with unquoted and backslashed for now- we can cover the other stuff with a doc fragment once I write it.

windows/win_file.py
samdoran: Does that mean `state` is required? It's currently listed as not required.

gundalow: If state is required then it looks like the module docs are wrong _and_ `win_file.ps1` needs updating to enforce that.

windows/win_copy.py
```diff
@@ -47,25 +47,22 @@
...
+    src: files/temp_files/
+    dest: c:\temp
 '''
 RETURN = '''
 dest:
     description: destination file/path
     returned: changed
     type: string
-    sample: "c:/temp/"
+    sample: "c:\temp"
```
nitzmahone: missed one, no quotes necessary here

samdoran: That upset one of the tests since it's in the results section.

nitzmahone: Needs single quotes if you're going to leave it quoted then, otherwise the path is c:[tab]emp

samdoran: Ok. I put it back to how it was originally, which is why I used double quotes.

nitzmahone:
@gundalow / @samdoran Windows stuff looks good

samdoran:
Thanks @nitzmahone!

gundalow (approved):

gundalow:
Cool, merged

Issue #5603, opened 2016-11-14
mount module is broken on CentOS 5 in Ansible 2.2


mkoshelev:
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

mount

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```

##### OS / ENVIRONMENT

controller: Fedora 23
client host: CentOS 5.11

##### SUMMARY

'mount' module in Ansible 2.2 fails on CentOS 5 systems due to lack of /proc/self/mountinfo file. Support for mountinfo was added in kernel 2.6.26, but CentOS 5 is on kernel version 2.6.18.
The issue seems to be introduced by https://github.com/ansible/ansible-modules-core/pull/5055

##### STEPS TO REPRODUCE

Attempt to add any mount on CentOS 5 system via 'mount' module.

```
- name: mount module in Ansible 2.2 is broken on CentOS 5
  hosts: test-centos5
  remote_user: root

  tasks:
  - name: add mount to fstab
    mount: src="nfs.example.com:/SHARE" name="/mnt" fstype="nfs" opts="noauto" state=present
```

##### EXPECTED RESULTS

Mount added to /etc/fstab

##### ACTUAL RESULTS

Playbook fails due to missing /proc/self/mountinfo file.

```
PLAY [mount module in Ansible 2.2 is broken on CentOS 5] ***********************

TASK [setup] *******************************************************************
ok: [test-centos5]

TASK [add mount to fstab] ******************************************************
fatal: [test-centos5]: FAILED! => {"changed": false, "failed": true, "msg": "Cannot open file /proc/self/mountinfo"}

PLAY RECAP *********************************************************************
test-centos5               : ok=1    changed=0    unreachable=0    failed=1   
```

ansibot:
@ansible ping, this issue is waiting for your response.
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

fdutheil:
Same issue here on a RHEL 5.11 host with ansible 2.2.0.0 on Ubuntu 14.04.5. Ansible 2.1.2.0 is not affected by this issue.

jctanner:
@jtyr can you please take a look?

jtyr:
[...]

PR #5689, opened 2016-11-21
Fall back if mountinfo reading failed (fixing #5603)

jtyr:
##### ISSUE TYPE

Bugfix Pull Request

##### COMPONENT NAME

`mount`

##### ANSIBLE VERSION

```
ansible 2.2.0.0
```

##### SUMMARY

This PR should fix the case where the kernel doesn't support the `mountinfo` file yet as reported in the issue #5603.

### DIFF
--- a/system/mount.py
+++ b/system/mount.py
@@ -398,7 +398,7 @@ def is_bind_mounted(module, linux_mounts, dest, src=None, fstype=None):
 
     is_mounted = False
 
-    if get_platform() == 'Linux':
+    if get_platform() == 'Linux' and linux_mounts is not None:
         if src is None:
             # That's for unmounted/absent
             if dest in linux_mounts:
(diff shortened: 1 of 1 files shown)


gregdek:
Thanks @jtyr for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

bcoca:
actually, there are several reasons mountinfo can be unavailable, it would be a better fix to just try and open and issue a warning (or use fallback method) if unavailable.

jtyr:
@bcoca This PR allows to fall back to the behavior before we have implemented the mountinfo functionality.

bcoca:
not always, only for specific kernel versions, which is not a good test. File access in the end is the only definitive one.

bcoca:
I'm thinking this should use the code we have in fact gathering which does this correctly and works in a multitude of scenarios.

jtyr:
@bcoca You are right. It will be better to fall back to the original behavior whenever the `mountinfo` is not readable. Please check the latest update.

abadger:
Thank @jtyr!  Merged to devel and stable-2.2.

PR #5703, opened 2016-11-22
Support script interpreters in async_wrapper.

mattclay:
##### ISSUE TYPE

Feature Pull Request

##### COMPONENT NAME

async_wrapper

##### ANSIBLE VERSION

```
ansible 2.3.0 (ansible-test d6bf3e9e05) last updated 2016/11/22 13:39:45 (GMT -700)
  lib/ansible/modules/core: (detached HEAD 5c986306be) last updated 2016/11/21 15:50:52 (GMT -700)
  lib/ansible/modules/extras: (detached HEAD 9511de1e3d) last updated 2016/11/18 13:10:36 (GMT -700)
  config file = 
```

##### SUMMARY

Support script interpreters in async_wrapper.

By calling the interpreter (extracted from the shebang) directly, a script can be used for the interpreter on non-Linux platforms, which typically require the interpreter to be a binary.


utilities/logic/async_wrapper.py
```diff
@@ -115,6 +115,18 @@ def _filter_non_json_lines(data):
...
     return ('\n'.join(lines), warnings)
 
+
+def _get_interpreter(module_path):
+    module_fd = open(module_path, 'rb')
+    try:
+        head = module_fd.read(1024)
+        if head[0:2] != '#!':
+            return None
+        return head[2:head.index('\n')].split(' ', 1)
```
nitzmahone: should probably add a `strip()` to make this handle leading spaces, since they are technically legal (I'm guessing that'd be handled transparently by the popen, but maybe not)

nitzmahone (changes requested):

nitzmahone (approved):

PR #5705, opened 2016-11-23
fix apt_key imports

jamespharaoh:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
apt_key module

##### ANSIBLE VERSION
```
ansible 2.1.3.0 (stable-2.1 ff0b525608) last updated 2016/11/23 10:09:48 (GMT +000)
  lib/ansible/modules/core: (detached HEAD 36121c5404) last updated 2016/11/23 10:10:31 (GMT +000)
  lib/ansible/modules/extras: (detached HEAD 91a142d625) last updated 2016/11/23 10:10:31 (GMT +000)
  config file = 
```

##### SUMMARY
The apt_key.py modules fails with the following error:

```
fatal: [wistla-berk-dev]: FAILED! => {"failed": true, "msg": "error importing module in /home/ubuntu/wistla-devops/third-party/ansible-modules-core/packaging/os/apt_key.py, expecting format like 'from ansible.module_utils.<lib name> import *'"}
```

It looks like ansible is manually processing these imports, and this module doesn't use the correct `import *` format. This change fixes that.

```
ok: [wistla-berk-dev] => (item={u'file': u'security/apt-keys/postgresql-apt-key'})
```

### DIFF
--- a/packaging/os/apt_key.py
+++ b/packaging/os/apt_key.py
@@ -108,8 +108,8 @@ EXAMPLES = '''
 # FIXME: standardize into module_common
 from traceback import format_exc
 
-from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.basic import *
+from ansible.module_utils.urls import *
 
 
 apt_key_bin = None


gregdek:
Thanks @jamespharaoh. All backport requests must be reviewed by the core team, and this can take time. We appreciate your patience. 

gundalow:
http://docs.ansible.com/ansible/dev_guide/developing_modules.html#common-module-boilerplate

>Prior to Ansible-2.1.0, importing only what you used from ansible.module_utils.basic did not work. You needed to use a wildcard import like this:
> `from ansible.module_utils.basic import *`

gundalow:
Note: This PR is against `stable-2.1`

jamespharaoh:
@gundalow And yet this bug exists in 2.1, perhaps the quoted documentation should refer to version 2.2?

abadger:
The 2.1.3 code is here: https://github.com/ansible/ansible/blob/v2.1.3.0-1/lib/ansible/executor/module_common.py#L328

It uses the AnsiBallZ method of constructing modules so the imports will properly work.  The error you're getting means that somewhere on your system there's an old version of the ansible code and that for some reason, that's what's being used when you're invoking ansible.

(The error message isn't present in the 2.1.3 code that I linked to above... You can see where it was in the old 2.0.x code like here: https://github.com/ansible/ansible/blob/v2.0.2.0-1/lib/ansible/executor/module_common.py#L100  )

jamespharaoh:
ok i think git subtree has done something nasty to cause this, i think i've got it working now, thanks.

PR #5715, opened 2016-11-23
Force BSDs to use umount/mount instead of trying to use remount.

### LINKED ISSUE #5591: mount module can't update flags on a mounted FS on OpenBSD
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

mount module

##### ANSIBLE VERSION

```
ansible 2.2.0.0
```

##### OS / ENVIRONMENT

OpenBSD

##### SUMMARY

https://github.com/ansible/ansible-modules-core/blob/devel/system/mount.py#L319 uses `-o remount`, which doesn't exist on OpenBSD. if you try to update the flags for an existing mountpoint, it will update fstab correctly, but then will try to remount/update the flags on the living FS, and mount module will fail.
##### STEPS TO REPRODUCE

```
#grep local /etc/fstab                                                                                       
/dev/wd0g /usr/local ffs rw,wxallowed 0 0

#ansible master -m mount -a 'name=/usr/local opts=rw,wxallowed,nosuid state=mounted fstype=ffs src=/dev/wd0g'
master | FAILED! => {
    "changed": false, 
    "failed": true, 
    "msg": "Error mounting /usr/local: mount_ffs: -o remount: option not supported\n"
}
#grep local /etc/fstab                                                                                       
/dev/wd0g /usr/local ffs rw,wxallowed,nosuid 0 0
#mount |grep /usr/local
/dev/wd0g on /usr/local type ffs (local, wxallowed)
```

##### EXPECTED RESULTS

the living FS should have seen its opts updated.
##### ACTUAL RESULTS

mount module failed...  a possible fix would be to use `mount -u` which is more or less the equivalent of `mount -o remount`

```
#mount -u -o nosuid /usr/local
#mount |grep /usr/local       
/dev/wd0g on /usr/local type ffs (local, nosuid, wxallowed)

```

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME
mount.py

##### ANSIBLE VERSION
```
devel and 2.2.0
```

##### SUMMARY

Fixes #5591


system/mount.py
```diff
@@ -367,9 +367,18 @@ def remount(module, mount_bin, args):
...
     cmd += [ args['name'], ]
     try:
-        rc, out, err = module.run_command(cmd)
+        if get_platform().lower().endswith('bsd'):
+            # Note: Forcing BSDs to do umount/mount due to BSD remount not
+            # working as expected (suspect bug in the BSD mount command)
+            # Interested contributor could rework this to use mount options on
+            # the CLI instead of relying on fstab
+            # https://github.com/ansible/ansible-modules-core/issues/5591
+            rc = 1
```
bcoca, line 377: need to initialize out/err to '' as they are used below

abadger:
@bcoca thanks.  Fixed.

abadger:

PR #5739, opened 2016-11-26
Support for environment variable in setup.ps1

Gianlu:
##### ISSUE TYPE

 - Feature Pull Request
 
##### COMPONENT NAME

Windows setup module (setup.ps1)

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  configured module search path = ['/usr/share/ansible/modules/']
```

##### SUMMARY

Hi,
I introduced a little mod to setup.ps1 in order to check if it is present a specific environment variable in Windows nodes. If no value is passed to _fact_path_ parameter, the script tries to get extra facts directory   from "ANSIBLE_FACTS" environment variable. This behaviour is useful in playbooks, when setup module is invoked without passing any parameter

```
ansible -m setup myserver --args "fact_path=C:\facts"
```
is the same than
```
ansible -m setup myserver 
```
if $env:ANSIBLE_FACTS is "C:\facts".


gregdek:
Thanks @Gianlu for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

nitzmahone:
After some internal discussion, we're not big fans of a hidden abstraction point for custom fact location. https://github.com/ansible/ansible-modules-core/pull/5789 enables the explicit usage of an envvar for *nix setup, and I'll figure out a way to enable it on the Windows side as well (probably allowing environment string expansion with a `path` type var as part of the declarative argspec work for 2.3). I realize this doesn't cover the "implicit play setup" case, but there's been talk of adding a "setup_args" play key- this would fit nicely with that. Meantime, `gather_facts: no` and an explicit setup call will get it done. If you want to argue for any of these features, feel free to come to an [IRC meeting](https://github.com/ansible/community/blob/master/MEETINGS.md) to discuss further. Thanks!

PR #5761, opened 2016-11-29
junos_config: Fix the zeroize function

ikelos:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

junos_config

##### ANSIBLE VERSION

```
ansible 2.2.0.0
  config file = 
```

##### SUMMARY

Currently this function directs to the standard NetworkModule,
whose run_commands function takes no arguments (other than self) and
so throws an exception that it's given two arguments when it only takes one.

This directs the call to the connection's cli method to run the command
directly on the device.

### DIFF
--- a/network/junos/junos_config.py
+++ b/network/junos/junos_config.py
@@ -274,7 +274,7 @@ def rollback_config(module, result):
 
 def zeroize_config(module, result):
     if not module.check_mode:
-        module.cli.run_commands('request system zeroize')
+        module.connection.cli('request system zeroize')
     result['changed'] = True
 
 def confirm_config(module, result):


gundalow:
Hi,
Thanks for the patch. Could you please provide a snippit of a playbook that current fails so I can add it to our test suite.

ikelos:
Sure, you'll need to fill in the provider bit yourself though...
```
 - name: Zeroize the config
   junos_config:
     zeroize: yes
     provider: "{{ provider }}"
   delegate_to: 127.0.0.1
```

gundalow:
That's great, thanks

ikelos:
Just to note, this might be a little drastic to keep as a test-case, since it will completely reset the device to factory default, removing all existing access and requiring that the device be setup again (including credentials for access by systems such as ansible).

gundalow:
That is very true, good message to add in case anyone else looks at this.

Currently we have some test VMs setup for this purpose

gregdek:
Thanks @ikelos. To the current maintainers, @privateip, @gundalow, @qalthos, @ganeshnalawade please review according to guidelines (http://docs.ansible.com/ansible/developing_modules.html#module-checklist) and comment with text 'shipit', 'needs_revision' or 'close_me' as appropriate.

PR #5766, opened 2016-11-29
[git] Fix fetching branch of shallow clone

veger:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
git

##### ANSIBLE VERSION
```
ansible 2.2.0.0
```

##### SUMMARY
After shallow cloning a repository (non-master branch), updates to this branch are not fetched by the git module. Very similarly as #3817 except I am not switching from a tag to a branch, but merely updating the branch, but I suspect that the same code is activated.

As @robinro suggested in  https://github.com/ansible/ansible-modules-core/issues/3817#issuecomment-224223316, I started with removing the `currenthead != version` check, but fetching a remote ref to the active local ref with
```python
refspecs.append('+refs/heads/%s:refs/heads/%s' % (version, version))
```
gives the following error:
> Failed to download remote objects and refs:  fatal: Refusing to fetch into current branch refs/heads/mybranch of non-bare repository.

Seems logical, hence the check is there (although I do not really understand the comments next to it).

The other refspec addition did the trick for my situation, so I replaced the original with this one.

### DIFF
--- a/source_control/git.py
+++ b/source_control/git.py
@@ -646,9 +646,7 @@ def fetch(git_path, module, repo, dest, version, remote, depth, bare, refspec, g
                 # 1.8.3 is broken, 1.9.x works
                 # ensure that remote branch is available as both local and remote ref
                 refspecs.append('+refs/heads/%s:refs/heads/%s' % (version, version))
-                refspecs.append('+refs/heads/%s:refs/remotes/%s/%s' % (version, remote, version))
-            else:
-                refspecs.append(version)
+            refspecs.append('+refs/heads/%s:refs/remotes/%s/%s' % (version, remote, version))
         elif is_remote_tag(git_path, module, dest, repo, version):
             refspecs.append('+refs/tags/'+version+':refs/tags/'+version)
         if refspecs:


gregdek:
Thanks @veger for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

robinro:
@veger 
Thanks for this fix! This is pretty much a follow-up extension of #3492, which introduced all lines you touch.
I didn't test it yet, but if the current tests pass and your usecase works it should be enough to merge it.

Which git version are you using? I assumed updating a branch is covered by #3492 and also tested it locally back then, but that might be git version specific.

Can you please add an integration test for your usecase in ansible/ansible test/integration/targets/git/tasks ?
The ansible repos will be merged soon, so if you can wait ~2 weeks you can add the fix and the test in one PR, see https://groups.google.com/forum/#!topic/ansible-devel/mIxqxXRsmCI.

The whole depth/fetch code is still rather messy and to prevent future regressions we need higher test coverage.

robinro:
Can you also fix #3817? This should be in the same lines, or might already be fixed by your change.

veger:
I am using git version 2.10.2 on Arch Linux
I rather not wait for ~ 2 weeks, as I now have the issue fresh in my mind. (Lots of projects going on, so details tend to blur... ;) )

I attempted an integration test on my usecase (updating a branch clone with depth specified): https://github.com/veger/ansible/commit/f4b969aa310b48ad03fad072a4643d64e1a0efd2

It is my first ansible test (took a while to get the test running). I would appreciate it if you could take a (short) look at it, in order to check if it makes some sense.

* I could only reproduce it with a local created repo. A prepared GitHub one did not work, as the new commits should not be available yet, so they are 'unknown to the system' (or something similar to these lines...)
* I have to switch back to master after the commits on the branch in order to break the test without the update of this PR (without switching back to master the test succeeds..?!)
* Tested using Ubuntu 1604 container only... (both breaking reproduction as well as fixed module)

If this is ok, I am willing to try and reproduce the issue of #3817 in a test and see if the fix of this PR indeed fixes it as well.

Should I create a PR on ansible/ansible for this test?

robinro:
@veger The test looks good to me. Using external repos is frowned upon and your on-the-fly creation of the test repo looks fine to me and makes the test self-consistent.

The "switching to master" is probably why I did not see the bug earlier, so it's important to have the concrete test case. Once the PR is open all OSes will run (and fail, since your fix is not yet merged). Once the fix is in the test can be rerun and pass. It might happen that the fix does not work for old git versions (1.8.x), so it's a good idea to have an eye on centos6 in the tests.

Please open a PR against ansible/ansible. Someone from ansible core has to review it. To speed this up it's best to show up in the IRC channel and ask for a review, answer questions and implement suggestions.

robinro (approved):

veger:
Thanks for your quick response.

Tomorrow I'll have a go at reproducing (and hopefully seeing that this PR fixes) the issue of #3817. After that I'll open a PR in ansible/ansible.

veger:
Looked a couple of hours at reproducing #3817, but was not able to do so. Maybe this got fixed already?

(I did find an oddity though, I'll create a PR to fix that)

veger:
Merged into https://github.com/ansible/ansible/pull/18728

PR #5805, opened 2016-12-02
Do not use the fstab parameter on openbsd for mounting

### LINKED ISSUE #5591: mount module can't update flags on a mounted FS on OpenBSD

abadger:
##### ISSUE TYPE

 - Bugfix Pull Request

##### COMPONENT NAME

mount.py

##### ANSIBLE VERSION
```
devel
```

##### SUMMARY

OpenBSD's mount command doesn't allow selecting which fstab file to use.
So if we're operating on the live filesystem (mount or remount) return
an error if the user specified an fstab file.

Fixes #5591


abadger:
@landryb here's the PR I promised you for the fstab parameter on OpenBSD.  See if it works for you.  I'm tempted to push it to devel but not to 2.2... I might just hold off on pushing it to 2.2.1 but queue it up for 2.2.2 if there aren't issues.

gregdek:
Thanks @abadger for this PR. This module is maintained by the Ansible core team, so it can take a while for patches to be reviewed. Thanks for your patience.

landryb:
Well, with mount.py taken from your pr on top of 2.2.0, i get the 'Do not specify the fstab parameter for OpenBSD hosts' error message but... i didnt specify one.
```
doas ansible -c local localhost -m mount -a 'name=/home opts=rw,dev,nosuid,wxallowed state=mounted fstype=ffs src=/dev/sd0d dump=1 passno=2'
```

system/mount.py
```diff
@@ -328,7 +330,13 @@ def mount(module, args):
     if ismount(name):
         return remount(module, mount_bin, args)
 
-    cmd += _set_fstab_args(args)
+    if get_platform().lower() == 'openbsd':
+        # Use module.params['fstab'] here as args['fstab'] has been set to the
+        # default value.
+        if module.params['fstab'] is None:
```
landryb: Should be inverted ?

system/mount.py
```diff
@@ -364,7 +372,13 @@ def remount(module, mount_bin, args):
     else:
         cmd += ['-o', 'remount' ]
 
-    cmd += _set_fstab_args(args)
+    if get_platform().lower() == 'openbsd':
+        # Use module.params['fstab'] here as args['fstab'] has been set to the
+        # default value.
+        if module.params['fstab'] is None:
```
landryb: Same thing, invert the condition ?

landryb:
If i invert the logic (ie 'is not None' instead of 'is None'), i dont have the error message, but then the module still fails because it tries to umount a busy device. Still not ideal :)

abadger:
Thanks.  Logic inversion fixed.

abadger:
Merged to devel and pushed to temp-post-2.2.1-staging for inclusion in 2.2.2.

PR #5810, opened 2016-12-03
Added API throttling to RDS module

cooniur:
##### ISSUE TYPE
 - Bugfix Pull Request

##### COMPONENT NAME
AWS RDS module

##### ANSIBLE VERSION
```
$ ansible --version
ansible 2.2.0.0
  config file =
```

##### SUMMARY
```
Added a new parameter `api_throttling` to RDS module to allow specifying a time before calling AWS API to avoid throttling issue. 

Fixes my bug report in Issue #143 (see my comment there).
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,5 @@ venv
 Vagrantfile
 .vagrant
 ansible.egg-info/
+# Visual Studio Code
+.vscode
--- a/cloud/amazon/rds.py
+++ b/cloud/amazon/rds.py
@@ -184,7 +184,7 @@ options:
   aws_throttling:
     description:
       - how long before calling AWS API to check resource, in seconds
-    default: 10
+    default: 5
   apply_immediately:
     description:
       - Used only when command=modify.  If enabled, the modifications will be applied as soon as possible rather than waiting for the next preferred maintenance window.
@@ -1057,7 +1057,7 @@ def main():
             subnet            = dict(required=False),
             wait              = dict(type='bool', default=False),
             wait_timeout      = dict(type='int', default=300),
-            api_throttling    = dict(type='int', default=30),
+            api_throttling    = dict(type='int', default=5),
             snapshot          = dict(required=False),
             apply_immediately = dict(type='bool', default=False),
             new_instance_name = dict(required=False),
--- a/cloud/amazon/rds.py
+++ b/cloud/amazon/rds.py
@@ -162,7 +162,7 @@ options:
     default: null
   aws_secret_key:
     description:
-      - AWS secret key. If not set then the value of the AWS_SECRET_KEY environment variable is used. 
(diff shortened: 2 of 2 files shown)


cooniur:
How to fix the following error?
```
2016-12-03 07:04:05 ERROR: version_added for new option (aws_throttling) should be 2.3. Currently 0.0
```

cloud/amazon/rds.py
```diff
@@ -647,8 +652,9 @@ def get_data(self):
 
 def await_resource(conn, resource, status, module):
     wait_timeout = module.params.get('wait_timeout') + time.time()
+    api_throttling = module.params.get('api_throttling')
```
ryansb, line 655: Instead of manually doing retries, use [AWSRetry](https://github.com/ansible/ansible-modules-core/blob/db66833a80bcd7f0d33ccbfd6555d69c5ae88003/cloud/amazon/cloudformation.py#L317) which is now part of the EC2 module_utils. It got added early in 2.3, and should cut down on duplicated retry/backoff logic.

cloud/amazon/rds.py
```diff
@@ -292,9 +297,9 @@
      wait: yes
      wait_timeout: 600
      tags:
-         Name: pg1_test_name_tag 
+         Name: pg1_test_name_tag
   register: rds
-
```
ryansb, line 297: Generally we try to avoid making whitespace changes to areas you aren't editing. If possible, can you set your editor to not automatically do this?

.gitignore
```diff
@@ -51,3 +51,5 @@ venv
 Vagrantfile
 .vagrant
 ansible.egg-info/
+# Visual Studio Code
+.vscode
```
ryansb, line 55: Please do this in a separate PR, since it's not related to the RDS module.

ryansb (changes requested):
Thanks for this PR. We have a decorator called AWSRetry to help with this sort of logic, can you use it instead?

cloud/amazon/rds.py
cooniur: OK,  will revert the whitespace changes.

cloud/amazon/rds.py
cooniur: @ryansb I'd love to dig more about the AWSRetry. But for the simplicity & quickness of fixing the api throttling issue, can we accept this for now? (and of course, do the AWSRetry conversion in a separate PR since the RDS module hasn't started using it yet anyway).

.gitignore
cooniur: OK, will do it in a separate PR.

cooniur:
Hi @ryansb , hopefully my last comments make sense to you. Also I just wanted to let you know that I still remember this PR and will just need to find some time to make the requested changes.

Thanks!

PR #5886, opened 2017-01-17
Yum module with restricted sudo (from stable 2.1.0)

sbernier:
Hi,
 Ansible supports sudo at the framework level, but it assumes unrestricted sudo permissions.
 We have to operate environments where sudo is restricted to certain commands only, hence Ansible at the framework level doesn’t help us.
 We have introduced sudo capabilities directly into a new yum module, in order to support restricted sudo and manage to deploy our solution.

What do you think of the solution, is there any other alternatives?

Thanks,
 Best Regards


sbernier:
Leave a comment

PR #5904, opened 2017-04-07
Fix fanout issue for OS10 10.2.2 release

skg-net:
Fix fanout issue for OS10 10.2.2 release

In OS10.2.2 release the fanout attribute was not passed as part of the show interface command.
Updated the code, not use the fanout attribute, instead parse based on subport.


skg-net:
@gundalow Can you please take a look at it. This fix is already in the latest development.

gundalow:
Merged into `stable-2.2`

